PR #193, opened 2014-12-15
Configs from Andrew H

dnash86:
Configs received from Andrew Hart


dnash86:
I am not sure if the issue regarding ctau values has been fully resolved here, but I've updated the configs to include the SLHA tables according to the example given here:/afs/cern.ch/user/b/bendavid/work/cmspublic/slhatest.py

If everything is okay, we can move forward with the request.  Let me know if there are further issues.  

Thanks!

bendavid:
I'll wait to merge this until the needed 71x release supporting the inlined SLHA tables is available.  (Haven't had a chance to look at the ctau issue yet either)

dnash86:
Hi Josh,

Oh, okay - sounds good.  Thanks a lot.

dnash86:
Now that RunIIWinter15GS is open, can this PR be approved?  Is 7_1_13 (the current release for RunIIWinter15GS) the needed 71x release for inlined SLHA tables?  

Thanks,
David

bendavid:
I still haven't had a chance to check the ctau issue.  Will move that up on the priority list...

dnash86:
Oh, alright.  Thanks!

python/ThirteenTeV/DisplacedSUSY_stopToBottom_M_300_100mm_TuneCUETP8M1_13TeV_pythia8_cff.py
```diff
@@ -0,0 +1,857 @@
...
+        parameterSets = cms.vstring(
+            'pythia8CommonSettings',
+            'pythia8CUEP8M1Settings',
+            'processParameters'
+        )
+    )
+)
+
+dirhadrongenfilter = cms.EDFilter("MCParticlePairFilter",
+    #Status = cms.untracked.vint32(1, 1),
```
bendavid: Independent of the ctau/decay question (looking at that now)
There is a  potential problem here.

For consistency all of the vectors should have the same length.  So if you want 10 particle id's then you have to repeat the MinPt, MinEta, etc 10 times... (or at least the code issues a warning if this is not the case, not sure what it does in that case)

bendavid: Ok, sorry never mind, looking at the code now.  Understand that the warning is just because you left out the "status" vector from the config.  (Which means that it is filled with 0, which means that the status is not used).

So ok the code is ugly (and also takes untracked parameters for the filter settings!) but fine.

bendavid:
Ok, I looked into the ctau issue.

It is indeed expected that the R-hadrons should decay despite having nominal proper decay length > 10mm, because R-hadron decays are handled by a dedicated mechanism which is not affected by the normal ParticleDecay settings.

dnash86:
Thanks a lot for the in-depth look.  Sorry about the code that looked wrong at first, I am often leery of making significant alterations myself to what analyzers give me without checking with both them and you (gen/prep-ops).

PR #201, opened 2015-01-29
aMC@NLO Patches for sysCalc and reweight card from Olivier

syuvivida:
multiple_pdf.patch fixes the output of PDF weighted cross section in sysCalc

reweight.patch fixes the erasing of scale/PDF weights if we run sysCalc first and run bin/madevent reweight second.


bendavid:
Hi eiko.
Master branch has been updated to 2.2.2. 
Do the patches still work there?  (And are you sure they don't conflict with the existing patches?)

syuvivida:
Hi Josh,
  I've tested this on the 2.2.2 scripts and source code.
The patches does not conflict with existing patches.
But note I had to use "patch < xxx.patch" to include the fixes from these patches.

Eiko

PR #210, opened 2015-02-18
Update of bin/Powheg: patch Roberto's fixes of PDF weights limitation on the fly

syuvivida:
1. Add one line in create_powheg_tarball.sh to patch the fixes for the two files
  a) POWHEG-BOX/include/pwhg_lhrwgt.h  (max number of weights)
  b) POWHEG-BOX/lhefwrite.f     (output of pdfinfo in the LHE file)

2. add one directory patches that includes the file to patch: pdfweights.patch


covarell:
Patch should be just lines 7-8 of pdfweights.patch in the first file and 60-75 in the second file.
I do not understand what are all the other lines removed by the patch, these should not have changed in principle

syuvivida:
The patche has been fixed and tested OK.

PR #214, opened 2015-02-24
Patches for sysCalc and reweight card

syuvivida:
1) add two lines in gridpack_generation.sh to patch sysCalc and lhe_parser.py 
2) multiple_pdf.patch patches the bug of sysCalc that gives incorrect PDF-weighted cross section
3) reweight.patch patches lhe_parser.py so that if sysCalc is run first and reweight_card is run second, 
the scale/PDF weights will not be erased by reweight_card.


bendavid:
Are the scripts currently capable of running syscalc first and reweighting second?

syuvivida:
Yes, exactly! These scripts allow us to run sysCalc first, reweighing second (but not the other way around).

bendavid:
Closing this since it is not really solving the problem until we can figure out how to run the reweighting in gridpack mode.

PR #216, opened 2015-02-24
macro for automated DJR studies

stosi:


bendavid:
At minimum we should extend this to also include the 4 parton events.

(Then we should probably also make it easier to switch between the three definitions of multiplicity)

covarell:
I agree for 4 partons. For the multiplicity we should maybe come to an agreement on a default before distributing to MC contacts.

As we discussed privately it should accept TChains to have enough stats for meaningful plots

bendavid:
I would make the default nMEPartons()
(but make it easy to switch between the three options)

PR #217, opened 2015-02-25
new version of a macro for DRJ studies

stosi:


covarell:
Could you replace the name of the main routine from "plotdjrorig" to "plotdjr" (so the macro can ve directly executed with .x)?

stosi:
change done

PR #220, opened 2015-02-26
Single top t-channel gridpacks 

maierbenedikt:
For 13 TeV. aMCatNLO 4FS setup. Default Pythia8, systematic Herwig++, 5FS as cross-check.

 Requires custom setscales.f & reweight_xsec.f. Added lines in gridpack_generation.sh


bendavid:
Can you explain a bit what is being customized in the setscales and reweight_xsec files?  (Perhaps even add some comment lines at the top of those files for future reference)

maierbenedikt:
I linked references to the modifications. 
The scale has to be modified because it relies on the kinematics of one specific outgoing particle (i.e. neither a fixed nor the default dynamical scale).
This in turn requires to remove some stop statements during the reweighting in the reweight_xsec.f. The reference is provided there in the header, too.

stosi:
Hi Benedikt
I gave a quick look and the proc_, run_ and madspin_ cards look OK to me. In gridpack_generation.sh, sorry if it's a stupid question, but why do you copy reweight_xsec.f into the directory Cards instead of SubProcesses ?

maierbenedikt:
That is a very good question ... It was a typo from my side. I fixed it!

bin/MadGraph5_aMCatNLO/cards/production/13TeV/st_tch_5f_ckm_NLO/st_tch_5f_ckm_NLO_proc_card.dat
```diff
@@ -0,0 +1,9 @@
+import model loop_sm-ckm
```
bendavid: this is not right for 5flavour generation.
You need:
define p = p b b~
define j = j b b~

bin/MadGraph5_aMCatNLO/cards/production/13TeV/st_tch_5f_ckm_NLO/st_tch_5f_ckm_NLO_run_card.dat
```diff
@@ -0,0 +1,146 @@
...
+# shower (HERWIG6 | HERWIGPP | PYTHIA6Q | PYTHIA6PT | PYTHIA8)         *
+# WARNING: PYTHIA6PT works only for processes without FSR!!!!          *
+#***********************************************************************
+  PYTHIA8   = parton_shower
+#***********************************************************************
+# Renormalization and factorization scales                             *
+# (Default functional form for the non-fixed scales is the sum of      *
+# the transverse masses of all final state particles and partons. This *
+# can be changed in SubProcesses/set_scales.f)                         *
+#***********************************************************************
```
bendavid: 5 flavour generation should be done with fixed scales?

bendavid:
(And sorry you'll have to rebase to avoid conflicts with my latest changes to gridpack_generation.sh, which added another optional file to copy.)

bin/MadGraph5_aMCatNLO/cards/production/13TeV/st_tch_5f_ckm_NLO/st_tch_5f_ckm_NLO_run_card.dat
maierbenedikt: Hi Josh, yes, the 5f generation is done with fixed scales, and of course, the process definition in the proc_card is wrong. I will create a new pull request.

PR #222, opened 2015-02-27
adding single top cards + modification to gridpack_generation.sh to allo...

maierbenedikt:
...w customized setscales.f + reweight_xsec.f


bendavid:
(for reference, this pull request is a revised version of https://github.com/cms-sw/genproductions/pull/220)

PR #227, opened 2015-03-04
Susy 13 TeV MC

alkaloge:
WJets, DYJets, ZJets HT samples

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/DYJets_LO_MLM/DY_HT-100to200_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/DYJets_LO_MLM/DY_HT-100to200_proc_card.dat
@@ -1,0 +1,43 @@
+#************************************************************
+#*                        MadGraph 5                        *
+#*                                                          *
+#*                *                       *                 *
+#*                  *        * *        *                   *
+#*                    * * * * 5 * * * *                     *
+#*                  *        * *        *                   *
+#*                *                       *                 *
+#*                                                          *
+#*                                                          *
+#*         VERSION 1.5.11                2013-06-21         *
+#*                                                          *
+#*    The MadGraph Development Team - Please visit us at    *
+#*    https://server06.fynu.ucl.ac.be/projects/madgraph     *
+#*                                                          *
+#************************************************************
+#*                                                          *
+#*               Command File for MadGraph 5                *
+#*                                                          *
+#*     run as ./bin/mg5  filename                           *
+#*                                                          *
+#************************************************************
+
+import model sm-no_masses
+define p = g u c d s u~ c~ d~ s~
+define j = g u c d s u~ c~ d~ s~
+define l+ = e+ mu+
+define l- = e- mu-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+# Define multiparticle labels
+define p u c s d b u~ c~ s~ d~ b~ g
+define l+ = e+ mu+ ta+
+define l- = e- mu- ta-
+define j  = p
+# Specify process(es) to run
+generate p p > l+ l- / h @0
+add process p p > l+ l- j / h @1
+add process p p > l+ l- j j / h @2
+add process p p > l+ l- j j j / h @3
+add process p p > l+ l- j j j j / h @4
+# Output processes to MadEvent directory
+output DY_HT_100-200 -nojpeg
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/DYJets_LO_MLM/DY_HT-100to200_run_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/DYJets_LO_MLM/DY_HT-100to200_run_card.dat
@@ -1,0 +1,247 @@
+#*********************************************************************
+#                       MadGraph/MadEvent                            *
+#                  http://madgraph.hep.uiuc.edu                      *
+#                                                                    *
+#                        run_card.dat                                *
+#                                                                    *
+#  This file is used to set the parameters of the run.               *
+#                                                                    *
+#  Some notation/conventions:                                        *
+#                                                                    *
+#   Lines starting with a '# ' are info or comments                  *
+#                                                                    *
+#   mind the format:   value    = variable     ! comment             *
+#*********************************************************************
+#
+#*******************                                                 
+# Running parameters
+#*******************                                                 
+#                                                                    
+#*********************************************************************
+# Tag name for the run (one word)                                    *
+#*********************************************************************
+  DY     = run_tag ! name of the run 
+#*********************************************************************
+# Run to generate the grid pack                                      *
+#*********************************************************************
+  .true.     = gridpack  !True = setting up the grid pack
+#*********************************************************************
+# Number of events and rnd seed                                      *
+# Warning: Do not generate more than 1M events in a single run       *
+# If you want to run Pythia, avoid more than 50k events in a run.    *
+#*********************************************************************
+    1500       = nevents ! Number of unweighted events requested
+      0       = iseed   ! rnd seed (0=assigned automatically=default))
+#*********************************************************************
+# Collider type and energy                                           *
+# lpp: 0=No PDF, 1=proton, -1=antiproton, 2=photon from proton,      *
+#                                         3=photon from electron     *
+#*********************************************************************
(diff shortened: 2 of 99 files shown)


bin/MadGraph5_aMCatNLO/cards/production/13TeV/DYJets_LO_MLM/DY_HT-100to200_proc_card.dat
```diff
@@ -0,0 +1,43 @@
...
+#*                                                          *
+#************************************************************
+#*                                                          *
+#*               Command File for MadGraph 5                *
+#*                                                          *
+#*     run as ./bin/mg5  filename                           *
+#*                                                          *
+#************************************************************
+
+import model sm-no_masses
```
bendavid: Default for 5-flavour should be sm-ckm_no_b_mass

clearly the ckm part makes no difference for drell-yan, but the implications of passing massless tau's to pythia have not been checked, so prefer not to do it.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/DYJets_LO_MLM/DY_HT-100to200_run_card.dat
```diff
@@ -0,0 +1,247 @@
...
+ 1d2  = etal    ! max rap for the charged leptons 
+ 0.6  = etaonium ! max rap for the quarkonium states
+   0  = etajmin ! min rap for the jets
+   0  = etabmin ! min rap for the b
+   0  = etaamin ! min rap for the photons
+   0  = etalmin ! main rap for the charged leptons
+#*********************************************************************
+# Minimum and maximum DeltaR distance                                *
+#*********************************************************************
+ 0.001 = drjj    ! min distance between jets
```
bendavid: Can be 0 in principle

bin/MadGraph5_aMCatNLO/cards/production/13TeV/DYJets_LO_MLM/DY_HT-100to200_run_card.dat
```diff
@@ -0,0 +1,247 @@
...
+#*************************************************************
+   0  = nhel          ! Number of helicities used per event
+#*******************                                                 
+# Standard Cuts
+#*******************                                                 
+#                                                                    
+#*********************************************************************
+# Minimum and maximum pt's (for max, -1 means no cut)                *
+#*********************************************************************
+ 10  = ptj       ! minimum pt for the jets
```
bendavid: Could be 0.01 to be really inclusive for the "non-qcd" partons.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/WJets_HT_LO_MLM/WJetsToLNu_HT-100to200_proc_card.dat
```diff
@@ -0,0 +1,51 @@
...
+# Define multiparticle labels
+define p = u c s d b u~ c~ s~ d~ b~ g
+define j = p
+define l+ = e+ mu+ ta+
+define l- = e- mu- ta-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+define all = j l+ l- vl vl~
+# Specify process(es) to run
+generate p p > w+ > l+ vl @0
```
bendavid: Decay chain syntax is faster (and physics is equivalent if bwcutoff=15000)

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZJets_LO_MLM/ZJetsToNuNu_HT100to200_proc_card.dat
```diff
@@ -0,0 +1,41 @@
...
+define j = g u c d s u~ c~ d~ s~
+define l+ = e+ mu+
+define l- = e- mu-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+# Define multiparticle labels
+define p u c s d b u~ c~ s~ d~ b~ g
+define j  = p
+# Specify process(es) to run
+generate p p > ve ve~ / h @0
```
bendavid: could use decay chain syntax plus large bwcutoff to speed things up, but ok

bendavid:
Made a few comments inline.

General comments:
1) since dy and w+jets MLM matching were already provided in the example cards, why not just change the HT cuts and use the cards verbatim?

2) For keeping track of things later, will be better to have seperate folders (can be subfolders) for each gridpack

alkaloge:
> On 04 Mar 2015, at 19:47, Josh Bendavid <notifications@github.com> wrote:
> 
> Made a few comments inline.
> 
> General comments:
> 1) since dy and w+jets MLM matching were already provided in the example cards, why not just change the HT cuts and use the cards verbatim?
> 

I started from the Phys14 gridpack cards - thought that the examples should not bear any differences 
> 2) For keeping track of things later, will be better to have seperate folders (can be subfolders) for each gridpack
> 
> 
ok - will do so

thanks

-a

bendavid:
Well a number of small things have changed since phys14 (in additional to examples here being based on default cards from 2.2.x)

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZJets_LO_MLM/ZJetsToNuNu_HT100to200_proc_card.dat
alkaloge: Should be equivalent ;-)
http://madgraph.hep.uiuc.edu/MadGraphData/Alexis.Kalogeropoulos@cern.ch/PROC149/HTML/info.html
http://madgraph.hep.uiuc.edu/MadGraphData/Alexis.Kalogeropoulos@cern.ch/PROC145/HTML/info.html

bendavid: Yes agreed, decay chain syntax should be a bit faster though (and combined with large bwcutoff also guarantees the z always appears in the event record, in case you care about that sort of thing)

bendavid: Anyway, for this particular case, don't care too much.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/DYJets_LO_MLM/DY_HT-100to200_proc_card.dat
alkaloge: So, use sm-ckm_no_b_mass in all of the WJets , DYJets, ZJets requests ?

bendavid: Yes.  Makes no difference for dy/z (but also no disadvantage, so might as well stay consistent)

For W it actually has some effect on the flavour combinations, and speed-wise we can afford it, so might as well.  (But here was tested only with the decay chain syntax + large bwcutoff, which helps keep the speed acceptable)

bendavid:
On the update: ok, but if some of the files added here are now superceded by the additional commit, please remove them (with git rm).

For the directory structure, I really meant that each gridpack (including each ht bin) should have it's own subdirectory at the very end.  (This is because mcm requests will have a link to a directory which is supposed to contain all the cards for that request and only for that request)

For the low mass drell yan, in the past this started at 10GeV right, are there analyses really going down to 5?

bendavid:
And why do we need seperate cards for W+jets with and without ckm?  Why not just one set of cards with sm-ckm_no_b_mass?

alkaloge:
Fixed directory structure - Different sample for WJets_toLNu WJets_to_QQ - The later needs some clarification from the requestor, so we want start the gridpack until then

bendavid:
Ok, but each commit is simply adding new files on top of the previous ones.  Please remove the old ones (or just open a new pull request with only the needed files).

(If you want to move files around or remove them in an existing pull request then it has to be done with "git rm" and "git mv")

alkaloge:
Ok, I will close this PR and create a new branch and PR to have a clean commit.

-a

PR #230, opened 2015-03-05
Run2 MG5_aMC data cards for ttH production

mharrend:
First version of the planned data cards for Run2 MG5_aMC production


bendavid:
Higgs should stay undecayed at lhe level.  Then only one set of cards/grid pack is needed for all Higgs decays.  (Which are configured later at the level of the pythia config)

bendavid:
It may be worth at least attempting ttH+0,1,2j (but it might be too slow, and maybe you tried it already even)

mharrend:
Hello Josh, I updated the cards according to your recommendations. At the moment I am trying both ttH+0/1jet and ttH+0/1/2jets.

Bests, Marco

bendavid:
Hi Marco,
The pull request now contains all the old files + the new files.  Can you remove the previous files?  (or just open a new pull request with only the needed ones?)

mharrend:
Hello Josh,

I changed the git history to remove the obsolete files. Sorry, that I forget this in the first step.

PR #233, opened 2015-03-05
SUSY PR

alkaloge:


bendavid:
Ah, I still see what look like duplicate cards:
 A bin/MadGraph5_aMCatNLO/cards/production/13TeV/WJets_HT_LO_MLM/WJetsToLNu_HT-100to200/WJetsToLNu_HT-100to200_proc_card.dat (24)
A bin/MadGraph5_aMCatNLO/cards/production/13TeV/WJets_HT_LO_MLM/WJetsToLNu_HT-100to200/WJetsToLNu_HT-100to200_run_card.dat (275)
A bin/MadGraph5_aMCatNLO/cards/production/13TeV/WJets_HT_LO_MLM/WJetsToLNu_HT-100to200_proc_card.dat (51)
A bin/MadGraph5_aMCatNLO/cards/production/13TeV/WJets_HT_LO_MLM/WJetsToLNu_HT-100to200_run_card.dat (275)

stosi:
Hi, in your run_cards sometimes you use
0.01  = ptj       ! minimum pt for the jets 
sometimes
10  = ptj       ! minimum pt for the jets 
and always with F  = auto_ptj_mjj
Is this intentional? Is it meant to be different for W and DY ?

alkaloge:
Hi


> On 05 Mar 2015, at 12:05, stosi <notifications@github.com> wrote:
> 
> Hi, in your run_cards sometimes you use
> 0.01 = ptj ! minimum pt for the jets 
> sometimes
> 10 = ptj ! minimum pt for the jets 
> and always with F = auto_ptj_mjj
> Is this intentional? Is it meant to be different for W and DY ?
> 

yes, this is what Josh suggested to do 

ptj = 0 .01 for the DY to be more qcd-inclusive

Josh - do you confirm ?

-a

alkaloge:
fixed

bendavid:
Yes, confirm that ptj = 0.01 and F = auto_ptj_mjj makes sense for processes with W's and Z's.

Silvano was saying that some of the cards had ptj = 10 though?

stosi:
Yes, all the run_cards for W have ptj = 10

alkaloge:
I think setting ptj = 0.01 is not making any sense both in terms of speed and also because practically all analyses have a ptj > 15 - 20 GeV or so. Are you sure you want the Wjets to have such a low ptJ  ? In the past I remember it was terribly slow to go that low (and even had problem with integration around certain low contributing channels which could give divergencies). I think also author recommend to have a minimum value of 5 or 10 GeV in these processes. 

Anyway, if you have already tested the 0.01 for the WJets I can modify if you find this critical, but honestly I don’t see the physics motivation here.

thanks

-a

bendavid:
Yes I've tested it and it does not slow things down much/at all.

What happens is that with both ptj and xqcut set, but F = auto_ptj_mjj,  then xqcut applies the cuts to all jets emitted from a qcd vertex.

The ptj = 0.01 setting practically only applies to the rare/special case when the Z/W is radiated off the quark.

We are not carefully treating this in the matching for now, but may want to improve that later (so would prefer that the gridpacks are already inclusive with respect to this point)

alkaloge:
Ok - ptj has been addressed as well in latest commit 
https://github.com/alkaloge/genproductions/commit/1f1d0fdff029a8395b45f1eb050888cc90a5a467

If no other pending issues, please approve

bin/MadGraph5_aMCatNLO/cards/production/13TeV/WJets_HT_LO_MLM/WJetsToQQ_HT-600toInf/WJetsToLNu_HT-600toInf_proc_card.dat
```diff
@@ -0,0 +1,25 @@
+import model sm-ckm_no_b_mass
+#switch to diagonal ckm matrix if needed for speed
+#import model sm-ckm_no_b_mass
+
+define p = p b b~
+define j = j b b~
+define q = j b b~
```
bendavid: This is confusing, since the b's are being added twice.  I assume madgraph is smart enough to remove duplicates from the definition?

bendavid:
One inlined comment on the WJetsToQQ cards.
Then I see some cards are still using sm-no_b_mass
Why not just use sm-ckm_no_b_mass everywhere for consistency?

alkaloge:
That was asked from the requestors that want the WToQQ sample - this is a special sample for the Razor analysis - Anyway, we won’t do the gridpack in the first round

bin/MadGraph5_aMCatNLO/cards/production/13TeV/WJets_HT_LO_MLM/WJetsToQQ_HT-600toInf/WJetsToLNu_HT-600toInf_proc_card.dat
alkaloge: yes, it does not hurt - will remove it for consistency though

bendavid:
Well also the Znunu and some of the Wtolnu cards have sm-no_b_mass still.

alkaloge:
fixed

bendavid:
There is still 10 = ptj and 0.001 = drjj in a few places (shouldn't make a difference, but let's try to make everything consistent!)

Also it looks like there are still duplicate cards for the ztonunu
/cards/production/13TeV/ZJets_LO_MLM/ZJetsToNuNu_HT100to200_proc_card.dat
/cards/production/13TeV/ZJets_LO_MLM/ZJetsToNuNu_HT-100to200/ZJetsToNuNu_HT100to200_proc_card.dat

PR #236, opened 2015-03-06
SUSY 13TeV MC

alkaloge:


bendavid:
For the QCD cards, can go with 0 = ptj and T = auto_ptj_mjj, since all partons are produced from qcd vertices.  But for gamma+jets I would keep it as for the W and Z cards.

For draj in gamma+jets I think we should make this much smaller to be more inclusive.  I would say 0.1 or even 0.05 if it's not causing any problems.

PR #237, opened 2015-03-06
SUSY 13TeV MC

alkaloge:
Adding GJets and QCD HT binned cards


bendavid:
Ok actually thinking a bit more, since the "photon radiated from quark" case is much more common than for the Z/W, and since we have not tested gamma+jets without the cuts, I would just do ptj=0, T = auto_ptj_mjj for both qcd and gamma+jets

alkaloge:
Fixed

PR #246, opened 2015-03-12
[Madgraph5_aMC@NLO] Input cards for gridpack production of narrow width dibson signals

syuvivida:
Input cards for the exotica diboson signals of the following 3 models: 
heft_radion (Spin-0), Vector_Triplet_free_Width_UFO (Spin-1), RS_bulk_ktilda (Spin-2). 

There are 35 channels in total.  All mass points and decay modes have the same run cards, but different 
process cards and customization  cards. We use customization cuts to set the mass value and the width of 
the new particle. 

There are 6 VBF channels that require only two mass points: 1800, 2500 
(I still uploaded cards for the other mass points for future production). Each mass point needs 50 K events.

There are 10 di-Higgs channels and we need the following events for 24 mass points:
mx = 260, 270, 300, 350, 400 GeV with 300k events
mx = 400, 450, 500, 550, 600, 700, 800, 900, 1000, 1200, 1400, 1600, 1800, 2000, 2500, 3000, 3500, 4000, 4500  GeV with 100k events

The rest 19 channels with 13 mass points need 50 K events for each mass point.
600, 800, 1000, 1200, 1400, 1600, 1800, 2000, 2500, 3000, 3500, 4000, 4500

There are 4.695 M events. But note the dihiggs signals occupies the budget of Higgs group as well.


bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson/Spin-0/Radion_VBF_ZZ_inclu/Radion_VBF_ZZ_inclu_narrow_M2500/Radion_VBF_ZZ_inclu_narrow_M2500_proc_card.dat
```diff
@@ -0,0 +1,14 @@
+set group_subprocesses Auto
+set ignore_six_quark_processes False
+output Radion_VBF_ZZ_inclu_narrow_M2500 -nojpeg
```
bendavid: There is a problem in this card: there are two output statements.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson/Spin-0/Radion_VBF_ZZ_inclu/Radion_VBF_ZZ_inclu_narrow_M2500/Radion_VBF_ZZ_inclu_narrow_M2500_proc_card.dat
```diff
@@ -0,0 +1,14 @@
...
+
+set complex_mass_scheme False
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+import model heft_radion
+define p = u c d s u~ c~ d~ s~
```
bendavid: there's an inconsistency here, p and j should normally have the same definition.  (The b should be included in both if and only if it is massless in the model)

bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson/Spin-0/Radion_VBF_ZZ_inclu/Radion_VBF_ZZ_inclu_narrow_M2500/Radion_VBF_ZZ_inclu_narrow_M2500_run_card.dat
```diff
@@ -0,0 +1,270 @@
...
+#*********************************************************************
+# KT DURHAM CUT                                                      *
+#*********************************************************************
+ -1    =  ktdurham        
+ 0.4  =  dparameter 
+#*********************************************************************
+# maximal pdg code for quark to be considered as a light jet         *
+# (otherwise b cuts are applied)                                     *
+#*********************************************************************
+ 4 = maxjetflavor    ! Maximum jet pdg code
```
bendavid: This should be set to 5 if the b is in the end massless and included in the proton and j definition.   (or stay as 4 otherwise)

bendavid:
Hi Eiko,
I made some inline comments for one of the cards (but please make sure the others aren't similarly affected).   I guess/hope you have some scripts for dealing with these given how many there are...

syuvivida:
Hi Josh,
  The output directory bug has been fixed. It happened only for the VBF channels. 
I have also made the jet definition consistent with that of proton. Because the diboson group decided to use consistently 4F definitions for proton, I remove b quark from the jet definition in VBF modes.
For b quarks that come from decays of particles, they are still kept.
Eiko

bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson/Spin-0/Radion_VBF_ZZ_inclu/Radion_VBF_ZZ_inclu_narrow_M1400/Radion_VBF_ZZ_inclu_narrow_M1400_proc_card.dat
```diff
@@ -0,0 +1,13 @@
+set group_subprocesses Auto
+set ignore_six_quark_processes False
+set loop_optimized_output True
+set complex_mass_scheme False
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+import model heft_radion
+define p = u c d s u~ c~ d~ s~
+define j = p
+define ferm = u c d s b u~ c~ d~ s~ b~ e- mu- ta- e+ mu+ ta+ vl vl~
```
bendavid: Hi Eiko,
When you say "For b quarks that come from decays of particles, they are still kept."
You mean using syntax like this?

bendavid: Are you sure the gamma* contribution is not important in this case?

syuvivida:
Hi Josh,
  Yes, the syntax for "define ferm" is exactly what I meant by including b quarks in the decay products of particles. About the gamma* contribution, I will let the original author of this card Alexandra reply to this question.

acarvalh:
Hi Josh

no, the gamma* contribution is not particularly important.
Its presence however increases the number of diagrams to be calculated and therefore the time to the ME converging,
that in VBF with this size of decay chain is not negligible.

This is why I remove from the card: it is not important and faster the simulation.

bendavid:
Ok yes, VBF topology, and the Z's from from the decay.  Ok, understood and agreed.

PR #248, opened 2015-03-13
Adding more single top cards

maierbenedikt:
for 13 TeV production. 

s-channel

t-channel mass up, down

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/st_sch_4f_ckm_NLO/st_sch_4f_ckm_NLO_run_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/st_sch_4f_ckm_NLO/st_sch_4f_ckm_NLO_run_card.dat
@@ -1,0 +1,146 @@
+#***********************************************************************
+#                        MadGraph5_aMC@NLO                             *
+#                                                                      *
+#                      run_card.dat aMC@NLO                            *
+#                                                                      *
+#  This file is used to set the parameters of the run.                 *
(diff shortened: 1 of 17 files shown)


bin/MadGraph5_aMCatNLO/cards/production/13TeV/st_sch_4f_ckm_NLO/st_sch_4f_ckm_NLO_run_card.dat
```diff
@@ -0,0 +1,146 @@
...
+#***********************************************************************
+# Cuts on the jets                                                     *
+# Jet clustering is performed by FastJet.
+# When matching to a parton shower, these generation cuts should be    *
+# considerably softer than the analysis cuts.                          *
+# (more specific cuts can be specified in SubProcesses/cuts.f)         *
+#***********************************************************************
+   1  = jetalgo   ! FastJet jet algorithm (1=kT, 0=C/A, -1=anti-kT)
+ 0.7  = jetradius ! The radius parameter for the jet algorithm
+   0  = ptj       ! Min jet transverse momentum
```
bendavid: We've seen problems with ptj=0.  Safer to set it to eg. 0.01

bendavid: but ok if it wasn't giving problems for t channel, maybe it's fine.  Just keep this in mind in case you hit strange errors.

PR #253, opened 2015-03-16
Adding single top Powheg cards

maierbenedikt:
tW and t channel

### DIFF
--- a/bin/Powheg/production/st_tW_5f_dr_ckm_13TeV/ST_tW_DR_5f_ckm_antitop_13TeV.input
+++ b/bin/Powheg/production/st_tW_5f_dr_ckm_13TeV/ST_tW_DR_5f_ckm_antitop_13TeV.input
@@ -1,0 +1,89 @@
+! ST-wtchannel production parameters
+
+! GENERAL POWHEG PARAMETERS
+
+numevts NEVENTS    ! number of events to be generated
+ih1   1           ! hadron 1 (1 for protons, -1 for antiprotons)
+ih2   1           ! hadron 2 (1 for protons, -1 for antiprotons)
+ebeam1 6500d0     ! energy of beam 1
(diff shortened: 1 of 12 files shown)


syuvivida:
Hi,
 Some of the cards have commented out the following lines or some cards even don't have 
the following lines:
  renscfact  1d0   ! (default 1d0) ren scale factor: muren  = muref * renscfact 
  facscfact  1d0   ! (default 1d0) fac scale factor: mufact = muref * facscfact 
   pdfreweight 1       ! PDF reweighting
   storeinfo_rwgt 1    ! store weight information
   withnegweights 0 ! default 0, 


I assume you want to generate weights with scale and PDF variations? If yes, you need to add 
the lines above.

Eiko

maierbenedikt:
Hi Eiko,


since the single top will be done with Powheg V1, the scales cannot be put into the LHE file :/. That's why only the dedicated scalp/scaledown scenarios have a value !=1 in the .input cards.

syuvivida:
OK, I see. it's because only ST_4F is added in V2?
Maybe we have to push authors again. 

Eiko

syuvivida:
Just to add, I don't see other problems in these cards.

maierbenedikt:
Right, ST_tch_4f is the only process that Powheg authors are currently busy with porting to V2. Final validation checks are being done by them (Emanuele Re). After that we should indeed push for tW as well!

maierbenedikt:
Hi Eiko, can the pull request than be merged? This would allow me to upload the gridpacks, set up the requests and report them in Thursday's MCCM.

Thanks a lot,
Benedikt

syuvivida:
Hi Benedikt,

  I don't have the permission to merge the pull request. 

Josh, could you merge this pull request? 

Thanks!!

Eiko

PR #254, opened 2015-03-16
Reduce temporary disk space usage for mg_amc@NLO

bendavid:


bendavid:
Yes indeed the lhe file is not part of the tarball.  It needs to be moved out of the processtmp area before that directory is deleted (as with the first tar.gz file)

Will fix it once I'm on a stable connection.

bendavid:
This should be fixed by
https://github.com/cms-sw/genproductions/pull/265

PR #260, opened 2015-03-17
Powheg input cards for DY mass binned processes

dnash86:
These are Powheg input cards for DY mass binned processes (Z->ee, Z->mumu) that is needed for EXO backgrounds.  I am new to Powheg, let me know if something is wrong here.  Thanks!

### DIFF
--- a/python/ThirteenTeV/LQToCMu_M_1000_TuneCUETP8M1_13TeV_pythia8_cff.py
+++ b/python/ThirteenTeV/LQToCMu_M_1000_TuneCUETP8M1_13TeV_pythia8_cff.py
@@ -1,0 +1,31 @@
+import FWCore.ParameterSet.Config as cms
+
+from Configuration.Generator.Pythia8CommonSettings_cfi import *
+from Configuration.Generator.Pythia8CUEP8M1Settings_cfi import *
+
+generator = cms.EDFilter("Pythia8GeneratorFilter",
+    maxeventsToPrint = cms.untracked.int32(1),
+    pythiaPylistVerbosity = cms.untracked.int32(1),
+    filterEfficiency = cms.untracked.double(1.0),
+    pythiaHepMCVerbosity = cms.untracked.bool(False),
+    comEnergy = cms.double(13000.),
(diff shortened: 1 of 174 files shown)


syuvivida:
Hi David,
  I don't see anything wrong in these cards. 

Eiko

dnash86:
Okay, thanks a lot Eiko.

It looks like I've accidentally attached another PR to this one as well.  They are a number of pythia cards for LQ1/2 production.  Can this PR be merged now?

bendavid:
@dnash86, I now see ONLY the leptoquark pythia cards in this pull request.  It looks like you completely overwrote the branch with the powheg cards.

Can you fix this one and/or make a new one with the powheg cards.  (and/or rename this one to reflect the fact that it now contains only the leptoquark pythia cards?)

The LQ cards anyways need to have the extra lines 
" 'Main:timesAllowErrors = 10000',
 'Tune:ee 3',
 'Tune:pp 5',
"
removed (since these are already covered by the common fragments/tune)

covarell:
Yes, it's better to close this PR and start over, since the committed stuff does not correspond to the title anymore (POWHEG things are removed by the second commit)

dnash86:
Yes, I will open the new PRs shortly, and do them one-by-one.

dnash86:
I have opened a new PR, named and described correctly.  It includes all LQ/1/2/3 pair production configs.  It is here:https://github.com/cms-sw/genproductions/pull/270

covarell:
Ok, we will have a look at these, please open a new one as well for powheg binned Z

dnash86:
Okay, how do I ensure I don't overwrite the new LQ one?  I'm not 100% sure what I did incorrectly last time, so I just wanted to be sure to avoid re-committing the same error.  If I add them to the same genproductions checkout area (in bin/Powheg/production of course) and then commit them, will that be okay?

covarell:
If you start over with a fresh genproductions area (which you may delete afterwards), copying inside the POWHEG stuff and committing there, the commit won't be overridden

bendavid:
You also have to make sure to use unique names for the github branches.

PR #263, opened 2015-03-18
[Madgraph5_aMC@NLO] exo-diboson signal gridpack cards update

syuvivida:
Update of cards for the Spin-0/2 -> hh -> tau tau bb and aa bb channels:
1) add 4 extra mass points upon the request of analyzers
2) remove definition of protons since they are already 4F by default


syuvivida:
Will commit more card changes in another pull request.

PR #266, opened 2015-03-22
[Madgraph5_aMC@NLO] Update of exo-diboson VBF and X->hh cards

syuvivida:
1) For the X->hh channels, add extra lower mass points upon the request of analyzers
Radion->hh -> aa bb
Radion->hh -> tau tau bb
Bulk Graviton->hh-> aa bb
Bulk Graviton->hh -> tau tau bb

Now, there are 28 mass points in these 4 channels. 
250, 260, 270, 280,300,320,340,350,400,450,500,550,600,650,
700, 800, 900, 1000,1200,1400,1600,1800,2000,2500,3000,
3500, 4000, 4500

2) For the W'->WZ and Z'->WW modes with VBF productions, change model name 
and also the process cards to remove contributions of DY-like processes.
Remove couplings of W' and Z' with quarks .

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson/Spin-0/Radion_hh_haahbb/Radion_hh_haahbb_narrow_M1000/Radion_hh_haahbb_narrow_M1000_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson/Spin-0/Radion_hh_haahbb/Radion_hh_haahbb_narrow_M1000/Radion_hh_haahbb_narrow_M1000_proc_card.dat
@@ -3,7 +3,6 @@ set ignore_six_quark_processes False
 set gauge unitary
 set complex_mass_scheme False
 import model heft_radion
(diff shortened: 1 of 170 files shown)


bendavid:
Few comments already.

1) You probably want F = cut_decays to avoid applying pt cuts to the partons from the decaying heavy particles.

2) Can you clarify which models have massive vs massless b's? (since of course this must lead to the consistent choice of b in/out of the proton/jet definition)

3) I see that all of these samples are without jet matching.  Are you sure no additional jets are needed?  (Like this one can have very large uncertainties on the pt distribution of the heavy system)

syuvivida:
Hi Josh,
1)  We are trying to be consistent with the 8 TeV cards, in which "cut_decays" was set to true.
But also note that our cut values are very loose. Since we already started production of gridpacks 
for most of the Spin-1 and Spin-0 modes, we would like to keep them the same.

2) All models have massive b

3) We have checked our kinematic distributions by adding one extra jet. Most of the daughter kinematics are driven by the mass of heavy mother, rather than the pt. Therefore, we do not need extra jets.

Eiko

PR #268, opened 2015-03-22
Add ZPrimeToTTJets samples cards

blinkseb:
This PR adds the cards used to the generation of Z' to ttbar Madgraph gridpacks.

A README as well as some helpers scripts are used to help the (re)generation of these cards if needed.

The main script `gridpack_generation.sh` has been modified to download the madgraph model used for the generation with all the others. For the moment, the model is downloaded from my dropbox. Feel free to grab it and upload it to some more reliable / official mirror.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZPrimeToTTJets_0123j_LO_MLM/README.md
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZPrimeToTTJets_0123j_LO_MLM/README.md
@@ -1,0 +1,10 @@
+# CMS Madgraph cards for Z' to top-antitop samples
+Author: Sébastien Brochet <sebastien.brochet@cern.ch>
+
+This folder contains all the Madgraph cards used for the generation of Z' to top-antitop CMS samples, with up to 3 extra partons in the matrix element. An helper script is provided in order to easily generate all the cards from a reference point, which is `ZPrimeToTTJets_M500GeV_W5GeV`.
+
+## How to generate the cards?
+
+First, edit the cards for the reference point, `ZPrimeToTTJets_M500GeV_W5GeV`. You can change the PDF, the center-of-mass energy, etc. in the `run_card`. You can also edit the process generated in the `proc_card`. The `madspin` card should not require any change. The content of the `customizecards` file is generated automatically, but **it has to be correct for the reference point**.
+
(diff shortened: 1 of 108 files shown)


bendavid:
I've added the file to the generators afs area.
Please modify the pull request to take the file from
https://cms-project-generators.web.cern.ch/cms-project-generators/topBSM_UFO.zip

Thanks.

bendavid:
There's an inconsistency in these cards.  The run cards (maxjetflavor and asrwgtflavor) are set to 5 (for 5 flavour matching).  The selected pdf is also a 5 flavour pdf.

The proton and jet definition do not include b b~ though.

Since the model would appear to have massive b-quark, the minimal change would be to go to 4 flavour matching (such that the additional partons do NOT include b's).

In that case one should change maxjetflavor and asrwgtflavor to 4, and then also change the pdf to 263400 (nnpdf30lo as=0.130, 4flavour)

blinkseb:
Thanks for the check. I'm wondering what is the best thing to do here: change the maxjetflavor and the PDF, or simply add the b's to the proton and jet? Both are easily done.

blinkseb:
I've updated the PR based on your suggestions. Thanks!

bendavid:
So indeed you have to decide whether you want/need to have the b's included in the extra jets.  If you want to add the b's to the proton and jet you would also have to modify the model slightly (adding a "restrict" file) to add and use the option of massless b's, since massive b's cannot/should not be included in the proton definition.

bendavid:
What is here now looks consistent for 4-flavour scheme matching at least.

blinkseb:
Ok thanks. I'm still not sure if we need b-jets in extra jets, so I'll ask in the B2G groups. Can you hold the merge until I have an answer?

bendavid:
Sure.

blinkseb:
Ok so we go for the 5-flavor scheme. I've updated the model to include a restriction file to set the b-quark mass to 0 (inspired from the SM restrictions). Can you please upload the new model to replace the old one?

https://www.dropbox.com/s/b27wbftflih2flz/topBSM_UFO.zip?dl=0

I'll update this PR to set maxjetflavors and co to 5.

Thanks!

bendavid:
Ok I've updated the zip file.

blinkseb:
Thanks a lot. I've updated the PR, everything should be correct.

PR #269, opened 2015-03-23
Pythia8 fragments for ttH+0/1jet, H->bb and ttH+0/1jet, H->Nonbb samples

mharrend:
Hello, 

I have added the pythia8 fragments for the ttH+0/1jet, H->bb and ttH+0/1jet, H->nonbb samples.

One question in this regard: In former times I had to add some TimeShower and other options for FxFx merging, e.g. see https://github.com/cms-sw/genproductions/blob/master/python/ThirteenTeV/Hadronizer_Tune4C_13TeV_aMCatNLO_FXFX_4f_1j_max1j_LHE_pythia8_cff.py

I have now used the following gen fragment as described in the SWGuideSubgroupMC https://github.com/cms-sw/cmssw/blob/CMSSW_7_5_X/Configuration/Generator/python/Hadronizer_TuneCUETP8M1_13TeV_aMCatNLO_FXFX_5f_max1j_max1p_LHE_pythia8_cff.py

Here the TimeShower options are missing, is this correct?

### DIFF
--- a/python/ThirteenTeV/ttHJetToNonbb_M120_13TeV_amcatnloFXFX_madspin_pythia8_cff.py
+++ b/python/ThirteenTeV/ttHJetToNonbb_M120_13TeV_amcatnloFXFX_madspin_pythia8_cff.py
@@ -29,8 +29,6 @@ pythia8aMCatNLOSettingsBlock,
   '25:m0 = 120.0',
   '25:onMode = on', 	# Allow all higgs decays 
   '25:offIfAny = 5 5',    # Switch decays of b quarks off
-  'Higgs:clipWings = off',  # Do not clip higgs width
-  'HiggsSM:NLOWidths = on' # On by default, mentioned for documentation
   ),
 parameterSets = cms.vstring('pythia8CommonSettings',
 'pythia8CUEP8M1Settings',
--- a/python/ThirteenTeV/ttHJetToNonbb_M120_13TeV_amcatnloFXFX_madspin_pythia8_cff.py
+++ b/python/ThirteenTeV/ttHJetToNonbb_M120_13TeV_amcatnloFXFX_madspin_pythia8_cff.py
@@ -1,0 +1,41 @@
+import FWCore.ParameterSet.Config as cms
+from Configuration.Generator.Pythia8CommonSettings_cfi import *
+from Configuration.Generator.Pythia8CUEP8M1Settings_cfi import *
+from Configuration.Generator.Pythia8aMCatNLOSettings_cfi import *
+generator = cms.EDFilter("Pythia8HadronizerFilter",
+maxEventsToPrint = cms.untracked.int32(1),
+pythiaPylistVerbosity = cms.untracked.int32(1),
+filterEfficiency = cms.untracked.double(1.0),
+pythiaHepMCVerbosity = cms.untracked.bool(False),
+comEnergy = cms.double(13000.),
+PythiaParameters = cms.PSet(
+pythia8CommonSettingsBlock,
+pythia8CUEP8M1SettingsBlock,
+pythia8aMCatNLOSettingsBlock,
+  processParameters = cms.vstring(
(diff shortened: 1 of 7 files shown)


mharrend:
I have added also a fragment for an exclusive tt+bb sample since I did not found more information. I just the standard FxFx gen fragment and set the nMaxJet parameter to 0. Or should I use another kind of Pythia8 fragment?

covarell:
Notice that in the configuration you are calling in this whole block of parameters:
https://github.com/cms-sw/cmssw/blob/CMSSW_7_5_X/Configuration/Generator/python/Pythia8aMCatNLOSettings_cfi.py
which may contain the TimeShower settings you wanted.

bendavid:
Indeed, the needed timeshower settings are all set in the common fragment.  The previously-used setting TimeShower:nPartonsInBorn is no longer needed when doing FXFX merging, since it is set automatically.  It is instead needed for non-FXFX samples.

For the tt+bb sample you should NOT use an fxfx fragment, the correct configuration should be based on the "NLO with no jet matching/merging" section of the twiki
https://twiki.cern.ch/twiki/bin/viewauth/CMS/QuickGuideMadGraph5aMCatNLO

python/ThirteenTeV/ttHJetToNonbb_M120_13TeV_amcatnloFXFX_madspin_pythia8_cff.py
```diff
@@ -0,0 +1,41 @@
...
+  'JetMatching:qCut = 40.', #this is the actual merging scale
+  'JetMatching:doFxFx = on',
+  'JetMatching:qCutME = 20.',#this must match the ptj cut in the lhe generation step
+  'JetMatching:nQmatch = 5', #4 corresponds to 4-flavour scheme (no matching of b-quarks), 5 for 5-flavour scheme
+  'JetMatching:nJetMax = 1', #number of partons in born matrix element for highest multiplicity
+  'SLHA:useDecayTable = off',  # Use pythia8s own decay mode instead of decays defined in LH accord
+  '25:m0 = 120.0',
+  '25:onMode = on', 	# Allow all higgs decays 
+  '25:offIfAny = 5 5',    # Switch decays of b quarks off
+  'Higgs:clipWings = off',  # Do not clip higgs width
```
bendavid: are you sure you want to turn off clipWings?  This should be standardized for all the higgs signal samples at least.

mharrend:
Okay, I see, so the Pythia8 fragments for the ttH+0/1jet sample should be fine besides the option clipWings.
I will ask Pietro and Olivier about this option.

Furthermore, I will upload a new version of the tt+bb fragment.

covarell:
Does clipWings have any effect of the lineshape of the Higgs is actually derived from another generator? I would be surprised that Pythia adds H->bb events with a mass outside the settings of bwcutoff...

bendavid:
Yes, though also probably better to remove HiggsSM:NLOWidths = on just to keep all the fragments consistent and avoid confusing people looking/comparing them later.

bendavid:
@covarell, I believe madgraph/aMC@NLO actually produces the undecayed higgs strictly on-shell in the LHE file, then it's up to pythia to do the momentum shuffling for the breit-wigner randomization.

bwcutoff in madgraph/aMC@NLO only has an effect with madgraph or madspin are doing the decays.

mharrend:
Okay, thank you, I checked now the Higgs mass in the LHE files and it is always 125 GeV.

mharrend:
I have updated the ttH+0/1jet , H->bb and H->nonbb samples and removed the two options.

I have used a new configuration for the tt+bb 4FS sample. Is setting nPartonsInBorn = 4 correct or should it be 0? Do I have to insert an additional option due to using a 4FS?

bendavid:
The tt+bb card looks correct (no special setting is needed for 4fs because there is no jet matching/merging for this sample)

PR #273, opened 2015-03-24
Fragments for EXO-RunIIWinter15GS for ADDmonoPhoton

davidsheffield:


bendavid:
I think all three of 
+ 'PhaseSpace:pTHatMin = 130.',
+ 'PartonLevel:ISR = on',
+ 'PartonLevel:FSR = on'

Can be removed.  ptHat at this level should be irrelevant, since there are heavy objects being produced.  Then the ISR and FSR flags are not needed, since they are part of the default settings.

davidsheffield:
The requester has asked that `'PhaseSpace:pTHatMin = 130.'` be kept just in case. I've removed the unnecessary lines for ISR and FSR.

bendavid:
to be kept just in case what?

Unnecessary and mostly-no-effect phase space cuts only make the acceptance of the resulting sample more difficult to understand.

Would suggest comparing the cross section with and without the cut.  If it really makes little or no difference (as it must), then much better to simply remove it...

PR #275, opened 2015-03-26
rename the ggH folder and add HWJ and HZJ folders in powheg higgs area

dsperka:
Hi, I am adding powheg cards for HWJ and HZJ, as well as JHUGen cards for their ZZ->4L decay and ZZ->2L+X decay, respectively. I also changed the name of the ggH folder

bin/Powheg/production/V2/13TeV/Higgs/gg_H_quark-mass-effects_JHUGen_HZZ4L_NNPDF30_13TeV/

to be

bin/Powheg/production/V2/13TeV/Higgs/gg_H_quark-mass-effects_NNPDF30_13TeV/

Since the same Powheg card can be used without JHUgen, it doens't make sense to have two folders with different names, one containing the JHUGen card the other not.


bendavid:
One of the reasons for having the cards in git is to have easy to follow links from the mcm requests to the cards.  Keeping the jhugen and non-jhugen version of the cards in separate folders would make things easier to understand I think, even if the powheg cards are the same.

dsperka:
Well, there are not "jhugen" and "non-jhugen" versions. With JHUGen, it will use the JHUGen input card and without JHUGen it will not. But, ok. How to proceed, should I make new pull request?

bendavid:
How does one choose to enable jhugen or not?  Based on the script used to create the tarball?  (Or a parameter given to the script at tarball creation time?)

dsperka:
It is a parameter given to create_powheg_tarball.sh when you create the tarball. You have to put either "none" or "JHUGen.input". The example from the twiki:

   ./create_powheg_tarball.sh slc6_amd64_gcc481/powheg/V2.0/src powhegboxV2_Feb2015 DMGG slc6_amd64_gcc481/powheg/V2.0/13TeV/examples/DMGG_NNPDF30_13TeV/DMGG_NNPDF30_13TeV.input eiko_DMGG none 100 12312
 
 Definition of the input parameters:
  (1) location or repository of powheg source tarball
  (2) name of source tarball 
  (3) process name
  (4) location of powheg input card 
  (either specify the location in the AFS generator area <scram_arch_version>/powheg/<powhegbox_version>/<beam_energy>/powheg.input 
  of the full path of your local input card)
  (5) name of output precompiled tarball
  (6) special additional cards (e.g. for JHUGen), in normal uses it should be just set to "none"
  (7) number of generated events
  (8) random seed

bendavid:
Ok thanks.  Since the script parameters are not easy to see from the mcm request I think it is clearer to keep separate folders at the cost of duplicating the powheg cards.  You can either update this pull request to copy the ggh instead of moving it, or create a new pull request.

PR #276, opened 2015-03-26
add HZJ and HWJ, and non-JHUGen folders

dsperka:
Sorry that there are 2 commits, second one undoing the first...


syuvivida:
I checked the cards and confirmed that the cards have all necessary settings for storing PDF/scale weights and the nominal PDF is correct.

covarell:
Question: why the JHUGen.input card is different in each process? Is it not HZZ4l SM in every case?
Please wait before starting gridpack production, I am uploading JHU5.2.5 and updating the script now

dsperka:
In WH it is HZZ4L, same as ggH. For ZH, to get an unbiased sample, we do 2L+X and then filter 4L events later. This is the only option right now for Powheg.

bendavid:
@dsperka do you know what the time per event is for ZHJ? (and what the filter efficiency is?)

This may not be a viable option if the event generation is too slow.  (One can get around this with Pythia8 user hooks potentially to guarantee that every event survives with an acceptable final state)

dsperka:
Hi Josh, what is the best way to accurately measure the time per even? The filter efficiency is about 15%.

bendavid:
If you have the tarball, you can just generate some events interactively and use "time"

"time ./runcmsgrid.sh 1000 101 1"

covarell:
15% does not seem critical however...

bendavid:
No, indeed probably it's ok.  (And there are madgraph samples which have matching efficiencies that low).  I was worried you about filtering at 2 or 3% efficiency or such.

dsperka:
With 50 events I get 55 seconds per event. Is it too high?

bendavid:
That is very high.  Combined with a filter efficiency of 15%, that means
more than 5 minutes of lhe generation per GEN-SIM event (much higher than
the SIM step).  Can you check if the time per event goes down when
generating a larger number of events at a time?

(The current slowest lhe request we have is ttbar+0,1,2jets aMC@NLO, which
is 23s/event with a matching efficiency of ~0.3.)

dsperka:
I will test with more events. Keep in mind though that these are small smaples unlike ttbar. We have 500k events requested for the sample and probably can reduce that a bit if needed.

dsperka:
Hi Josh,

With 500 events I still get 49s/event.

I also tried without producing all the weights. In this case, I get 2.9s/event. Is it expected that producing the various weights dominates the time/event?

bendavid:
OK thanks.  For 500k events after filtering I think this is fine.  It would probably be useful to work on an appropriate userhook for pythia8 moving forward, since I think this can gain a lot of flexibility for the future.

dsperka:
Hi Josh, ok yes it would be useful. Just need to convince the rest of the ZZ group that using pythia8 to decay everything is equivalent to JHUgen. I think JHUgen authors were also working on a similar thing within JHUgen itself.

Roberto, what is the status of the new JHUgen version? I produced gridpacks already with these cards using 5.2.3, and 5.2.5 is only minor updates. But I understand that 5.2.5 is "official".

bendavid:
Well I think a detailed comparison of JHUGEN vs pythia8 decays would be very useful in the gen only campaign.

covarell:
As far as I can understand, 5.2.5 differs from 5.2.3 only if you want to use ttH+JHUGen

dsperka:
That's my understanding, so then it is ok if I do not remake the gridpacks?

And yes, we will make a pythia8 only sample in the gen only campaign.

PR #277, opened 2015-03-26
Change the variable name for mass cut

KyeongPil-Lee:


bendavid:
uh, actually I'm not sure either of these variables exist for NLO running.  Have you checked?  You may need to implement the upper mass cut "by hand" using the cuts.f mechanism.

KyeongPil-Lee:
Oh, I am not familar with the madgraph yet, so maybe there are some mistakes...After your comment, I checked the madgraph files, and I saw that 'mmllmax' is defined in LO directory, not NLO directory. I will define the cut by hand and use them when I generate the gridpack. Thank you!

bendavid:
(You'll probably still need to remove the existing parameter from the card.  Anyways, suggest to do tests first with only the 0 jet process, since this should be much faster)

KyeongPil-Lee:
Hmm, Actually, I want to generate exactly same samples with 'dyellell012j_5f_NLO_FXFX' case except for the mass range, because I will combine them later. even in this case, are there some parameters that should be changed? existing parameters are same with the 'dyellell012j_5f_NLO_FXFX' card's parameter.

bendavid:
I mean you will have to remove the "mllmax_sf" parameter which doesn't exist for NLO.  (This parameter was not needed for the 50_infinity sample of course)

KyeongPil-Lee:
I'm not sure I understand properly. If I define 'mllmax_sf' variable in cut.f, can I use 'mllmax_sf' in the run card? then 'mllmax_sf' doesn't have to be removed. Is this right?

bendavid:
No, I don't think that parameter exists, therefore it has to be removed, and the upper mass cut has to be implemented using cuts.f.

If instead you have tested it and the parameter works, then maybe it can be used (but I doubt it)

KyeongPil-Lee:
Do you mean that I should give M<50 cut directly in the cut.f, instead of define 'mmllmax_sf' variable?
I'm sorry for bothering you.

bendavid:
Yes, I think this is the only way to do it.

KyeongPil-Lee:
Ok. then I will modify cut.f and test it. If it works well, I will give you the pull request to remove 'mmllmax_sf'. Thank you for your kind reply!

PR #278, opened 2015-03-26
param card and run card from H. Lee

dnash86:
This param card and run card were sent by the analyzers that had produced their own LHE files in Madgraph for a request.  It turns out only these two cards were used, and the mass parameter was varied in the param card for each run.  

If we want to proceed as normal, with param cards and run cards, I guess that we will need a separate param card for each mass point (named properly in each case), and all the gridpacks as well.


bendavid:
param cards are not allowed (and not used) by the gridpack creation scripts.
One should use the customize_cards mechanism to modify specific parameters with respect to the model defaults.

If model default values need to be set to zero for masses/couplings, then it may be needed to add a "restrict.dat" file to the model as it is for the SM case.

bendavid:
In addition I can say that the lhe's produced with these cards are not acceptable for run 2 production, among other reasons because they use cteq6l1 for the pdf, which is inconsistent with the rest of the production.  So I suggest to indeed set up proper cards for gridpack-based production.

bendavid:
One clearly needs also the process card (and without it I cannot judge whether all of the other settings are consistent/correct)

dnash86:
Okay, thanks, I will refer the analyzers to this information and we'll proceed with gridpack-based production.

dnash86:
Also, just to confirm, which pdf other than cteq6l1 should be used to be consistent with the rest of production?

bendavid:
This depends on LO/NLO/4f/5f
All cases are covered in the example cards which should always be the
starting point for production cards...
https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/examples

dnash86:
I will close this pull request as action is needed on these items from H. Lee anyways.

PR #279, opened 2015-03-26
remove the upper mass cut in the run card

KyeongPil-Lee:


bendavid:
The cuts.f will also have to be added as part of the cards, so I leave this one open until that detail is sorted out.

KyeongPil-Lee:
I included cuts.f in the card directory and rename it to match the format. please let me know if there are things should be done.

bendavid:
This looks good.  Can you just add a brief comment at the top of the modified cuts.f explaining what was changed and where?  (I found it and it looks ok, but took me a minute or two, so good to document it at the top)

KyeongPil-Lee:
OK. I added some comments on cut.f.

PR #281, opened 2015-03-27
Update Z' MadGraph cards

blinkseb:
This PR updates the cards used for the Z' to tt samples. Previous ones were faulty. Indeed, the decay of the Z' was not always tt when two extra jets are added ; final states like
p p > (Z' > u ubar) t tbar via gluon splitting for example

This issue is addressed by forcing the decay of the Z' to t tbar. I tried to let the tops to decay via MadSpin, but it fails miserably ( see tickets here: https://bugs.launchpad.net/mg5amcnlo/+bug/1435527 ). In the mean time, the tops decays chain is hardcoded in the proc card.

Cheers,
Sebastien


bendavid:
Did you actually try to run this?  I would expect the complexity of the decay chain causes trouble (at least we had problems with ttbar in the past)

If this works, then definitely fine, and definitely the correct way to do it.
(Otherwise if Olivier comments positively on your patch we can consider to apply it in the gridpack scripts so that the full chain can be done with madspin)

blinkseb:
Yes I tried it and was already able to generate a few gridpacks and validated them, so it's fine,

Even with my patch applied, I still have issue with MadSpin (something related to too many events with a wight greater than max weight). I didn't investigate too much because the production of the gridpacks is rather urgent, so I went for the easy path.

bendavid:
Ok, sure, like this is perfectly fine (even better than madspin technically)

PR #282, opened 2015-03-27
Fragments from Renjie Wang

dnash86:
Fragments from Renjie Wang for unparticles.

gridpack creation script had lines removed that I had added previously for a request that still needs to be reviewed.  Not sure about the missing space that was introduced...


bendavid:
Lines:
Main:timesAllowErrors
ParticleDecays:limitTau0
ParticleDecays:tau0Max = 10.

should be removed as they are already covered by the common fragments.

dnash86:
Sorry about that, it is done now.  Those lines are removed.

bendavid:
One more question, are you sure you want only e and mu decays?  Default is normally to include tau decays in "LL" type samples since there are also the leptonic tau decays which contribute.

dnash86:
The analyzers do actually want only e and mu decays, they do not use tau decays in their analysis.  I suppose then the config names are not 100% perfect, "ToLL" should be something like "ToEEAndMumu".  Is it okay to keep them as is though?

bendavid:
Whether they "use" tau decays or not, there is a contribution from leptonic tau decays which may or may not be relevant depending on kinematic cuts.

dnash86:
I believe the tau decays are irrelevant due to selection requirements, but I am checking with the analyzers currently.

dnash86:
I've checked with the analyzers, they have studied this, their Z mass window and Zpt cuts make the tau contribution negligible.  It seems in the past they have included the tau channel, but in effect this simply gives them 2/3 of the statistics in the sample without any benefit, since the selection requirements result in effectively vetoing all tau decays.  Thus they have chosen to actually remove them.

bendavid:
Ok thanks.  Suggest to rename the fragments then as you proposed.

dnash86:
Okay, they are renamed.

PR #285, opened 2015-03-31
add VBF_H powheg 

OlivierBondu:
This gives consistent distributions when comparing to powheg V1: 
https://test-obondu3.web.cern.ch/test-obondu3/2015-03-31-POWHEG_VBF_H_V1_VS_V2_test5_v2/


syuvivida:
The card have the necessary settings for producing PDF/scale weights. It looks fine to me.

covarell:
Does it differ in any way from the one in the example directory? If not, we may even skip this pull...

syuvivida:
Hi Roberto, actually for this process, I forgot to put input card. I will put Oliver's card into the example directory.

covarell:
OK, I will close this

PR #287, opened 2015-04-01
[Powheg] add powheg example input cards for VBF_H

syuvivida:
Put Olivier's VBF_H input card in the example/V2 directory


OlivierBondu:
Should I submit/PR example cards for the other mass hypotheses ?

syuvivida:
yes, I think so, in the production directory.

PR #289, opened 2015-04-02
add additional mass points for powheg ggH

dsperka:


syuvivida:
These cards have the proper parameters for getting scale/PDF weights.
Do we know if the following parameters work for ttH process too?
 bottommass 4.75d0    ! bottom quark mass - if defined it enables the
 charmmass 1.5d0     ! char quark mass - if defined it enables the charm

dsperka:
These cards are a copy of the 125 GeV point, which is already integrated, so yes I think it has the weights. An example of an LHE file created with the 125 GeV gridpack is here for you to check if youd like:

/afs/cern.ch/user/d/dsperka/public/Run2MC/cmsgrid_final_1_ggH_125.lhe

The only difference is the hmass and hwidth. The charmmass parameter is actually commented. It should work to have 4.75 for the b mass in ttH, but I think if it does not then it is a problem with ttH and not with ggH.

I just realized though, the hfact parameter I left the same for all mass points. I wonder if it should be changed to m(H)/1.2 for each point, instead of 125/1.2. What do you think?

dsperka:
Sorry, somehow I read your first sentence as a question.

syuvivida:
sorry I am not familiar with the hfact parameter, maybe Roberto could comment on this.

covarell:
Indeed the hfact parameter should be proportional to the mass

covarell:
Any news here? Could you just update hfact to mH/1.2?

dsperka:
Hi Roberto, we now have HRes predictions at 13 TeV and expect to be able to have some comparison over the weekend.

covarell:
Excellent!

syuvivida:
Could you remove the lines related to b quark mass and c mass since we know now that these parameters are not doing anything?

dsperka:
Hi, I need to make other pull requests besides this one. How can I do this easily? Can I just close this one, or can it be integrated? Sorry for my lack of expertise in git.

bendavid:
You can use a different branch for the new pull request (ie not "master")

Is this one still needed though?

dsperka:
This one is still needed, but also still don't know the best parameter to use for hfact.

covarell:
Sorry but here we do not have a final prescription for hfact do we?
Is it not better to wait until we understand Pythia differences between us and the POWHEG authors?

dsperka:
I agree Roberto, I just want to make another unrelated pull request. I figured out how now to do that now by creating a new branch, thanks Josh.

covarell:
Superseded.

PR #291, opened 2015-04-07
Update ttbb_4FS_ckm_amcatnlo_madspin_pythia8_cff.py

mharrend:
Changed number of nPartonsInBorn to 0 instead of 4


bendavid:
No no, 4 is the correct setting here.

PR #292, opened 2015-04-07
Adding W(->qq) + jets in HT bins for all-hadronic VV/VH/HH analyses

rappoccio:
Here are a couple of additional W+Jets MG cards for the W->qq channel (necessary for the all-hadronic VV/VH/HH searches, early EXO analyses). 

I copied the proc files from the WJetsToQQ_HT600toInf directory, and changed the names as appropriate, and copied the run files from the same HT bin as the corresponding WJetsToLNu directory.


bendavid:
Hi Sal,
Are these additional bins already on the google doc?  How many events are needed for the "first 1B" round of production? 

Currently I see there only the HT>600GeV bin, for  1M events.

https://docs.google.com/spreadsheets/d/16Z3rgM7XZzUMBxZr3DUXMldvGipeHwjDolUW5GSKn_s/edit#gid=0

rappoccio:
Hi, Josh,

Ahh, I see Andreas already put in the HT > 600 category. I think we only
need that one, I was putting this in just in case we wanted to have them.
But perhaps we don't even need it since the HT > 600 bin is already there?

Cheers,
Sal

bendavid:
Ok, I don't think we need the cards unless it was actually planned to produce the samples.

rappoccio:
Hi, Josh,

OK, I think we can close this then. Sorry for the noise.

Cheers,
Sal

PR #293, opened 2015-04-08
[Powheg] update ttH example cards to avoid confusion

syuvivida:
The original example card was for gridpack production. Revert it back to the card for official LHE production (to avoid confusion of ttH users).


syuvivida:
the change is fakevert and ncalls2

syuvivida:
ttH cards: change fakevert back to 0 and ncalls2=500000 

ggHZ card: this is LO process, replace the default PDF with NNPDF 30 LO (263000)

syuvivida:
Moved to pull requst 300

PR #296, opened 2015-04-09
Adding decay-hadronization fragment for VBF Hgg powheg

OlivierBondu:


bendavid:
Do the powheg LHE files have an SLHA table as in the madgraph case?  If so, may want to also include SLHA:useDecayTable = off to avoid similar problems.

OlivierBondu:
I see nothing of the sort in the gridpack output

OlivierBondu:
@bendavid : actually re-reading an old email thread of ours: should'nt we add '25:m0 = 125.0' ?
You were mentionning the reading of the Higgs mass from the lhe header was disabled

bendavid:
Ah, yes, forgot about that, sorry.  Please make another pull request updating it.

PR #301, opened 2015-04-12
Zbb adapted from Pierluigi

perrozzi:
Zbb adapted from Pierluigi Bortignon


bendavid:
Is this sample intended to be used together with an inclusive dy+jets sample?  There is no corresponding LO inclusive dy+jets sample requested so far for run 2.  (The corresponding sample from Run 1 was nominally replaced with the NLO 0,1,2j merged sample)

nstrobbe:
Is this Zbb sample the same as the one assigned to SUSY in line 312 of the spreadsheet?
I checked with our group, and we would prefer LO so that we can have an extra jet at matrix element level compared to NLO. Our DY HT-binned samples are LO, so it should be straight-forward to combine. 
We also agree with doing it in 5-flavour scheme, to make this combination possible.

covarell:
What is the status of this? Still needed or absorbed by SUS request?

perrozzi:
Ciao Roberto, actually I don't know, since Josh's and Nadia's comments disagree.
For what concerns Hbb, I would say that the NLO would be preferable, also according to Josh's comment.
In the past weeks I had tried to generate the NLO process without additional jets but had some jobs that didn't finish in the 2nw. This is quite strange and I would like to test it again.
(in the meanwhile I switched to the production of diboson gridpacks, that was urgent, and this is why I didn't update this commit yet)

bendavid:
Yes I think we basically agreed to do this both at LO and NLO.  So this is still relevant, and SUS will take care of the LO case.

PR #302, opened 2015-04-13
gridpack cards for T'+b, T'->b+W

peiffer:


bendavid:
The way the cards are currently named I don't think they're directly usable by the gridpack scripts, which expect cards with a name like <prefix>_proc_card.dat, <prefix>_run_card.dat.  Additionally, param_card.dat is not supported (and is ignored!) by the gridpack scripts, so one should change the parameter values through the customize_cards.dat mechanism.

Has the custom model being used been added to the gridpack_generation.sh script?  (The goal here is that someone could check out your cards + scripts from github and easily reproduce the gridpacks if needed)

peiffer:
I did not produce the gridpacks myself, I just extracted the cards from the the gridpacks. I have forwarded your objection to the B2G hypernews. The pull request should be done by the people who created the gridpacks originally because they should have the cards at hand.

bendavid:
Yes, fully agreed.

PR #304, opened 2015-04-13
adding single top tW DS cards

maierbenedikt:


covarell:
The fact that there is no storerwgtinfo etc. is because these are still for powheg V1?

maierbenedikt:
Hi Roberto, Yes, exactly!

covarell:
Excellent, I may be wrong but is there a reason why we do not use pdf 263000 here? (Also a question to @bendavid )

maierbenedikt:
263000 is the LO version if I am not mistaken:

NNPDF30_lo_as_0130


I was under the impression that 292200 (NNPDF30_nlo_nf_5_pdfas) was the "state-of-the-art" for 5FS NLO PDFs.

bendavid:
263000 is indeed LO.

292200 and 260000 have identical central values (nnpdf30 nlo alpha_s=0.118 5f).  The only difference is that 292200 includes alphas variations as additional members of the central set, rather than as completely separate sets.  This is needed for aMC@NLO due to the current limitations on including multiple pdf sets.  For powheg 260000 or 292200 can be considered equivalent as long as the alpha_s variations are anyways included separately.

PR #305, opened 2015-04-14
B2G single VLQ cards

dmajumder:
Modified gridpack_generation.sh to include VLQ model file.
Added cards for process pp->Ttj->tHtj for M=700-1800 GeV (in steps of 100 GeV) and LH and RH chiralities of T quark.


bendavid:
A few issues to start with.
1) param_card.dat is completely ignored by the gridpack scripts.  Parameter changes are intended to occur through the customize_cards mechanism only.  Please remove the param_card.dat from the pull request to avoid confusion.

2) It is dangerous to change parameters between 0 and non-zero in the customize_cards or the param_card, since madgraph may generate diagrams/apply extra optimizations which are later invalidated.  Looking at the model, it seems the defualt value for the "bsm 6" parameter is 1, and it is being set to 0 in the customize_cards.  It might be better to modify the model and add a "restrict.dat" file to set it to 0 from the beginning.

3) There are some git conflicts accidentally part of the pull request (will mark them inline)

dmajumder:
I've tried to address (1) and (3) with the modifed PR. I believe (2) will be common for all other proceses using this model and so should be updated inpendently

bin/MadGraph5_aMCatNLO/gridpack_generation.sh
```diff
@@ -364,6 +370,9 @@ cd processtmp
 echo "copying run_card.dat file"
 cp $CARDSDIR/${name}_run_card.dat ./Cards/run_card.dat
 
+#echo "copying param_card.dat file"
```
bendavid: Was this modification actually used when producing the gridpacks?

bendavid:
Since I see some commented lines in the gridpack script relating to the param_card, can you explain what was actually done when the gridpacks were produced?  The param_card was provided and copied in by the script?  The customize_cards were also used or have only been added after?

bin/MadGraph5_aMCatNLO/cards/production/13TeV/tH_t_topdecay_LH_narrow_M1100/tH_t_topdecay_LH_narrow_M1100_proc_card.dat
```diff
@@ -0,0 +1,41 @@
...
+define l- = e- mu-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+import model STP_UFO_freeWidth
+define Topo = t23 t23~
+define bb = b b~
+define WW = w+ w-
+define tt = t t~
+define ferm =  ve vm vt ve~ vm~ vt~ u c d s b u~ c~ d~ s~ b~ e- mu- ta- e+ mu+ ta+
+generate p p > j tt Topo  / h ,  tt > bb WW , ( Topo > h tt / z w+ w- ,  tt > bb WW )
```
bendavid: If I understand it correctly, this is doing the decay of the bsm particle, and decaying the tops to b W, but NOT decaying the W.  Any reason not to decay the W as well?  Doing it like this will leave the W decaying in pythia, losing polarization/spin correlation effects (and possibly also not handling correctly the width of the W)

dmajumder:
Both param_cards and customizecards were provided in respective "cards" folders which were inputed to the script. Was not entirely clear if the script generates param_card from model or modifies existing param_card according to the customizecard.

The only needed script modifications should be
[Line 107](https://github.com/cms-sw/genproductions/pull/305/files#diff-56bfba1423a749bf2d1829498400b7b4R107)
and 
[Line 108](https://github.com/cms-sw/genproductions/pull/305/files#diff-56bfba1423a749bf2d1829498400b7b4R108)

bin/MadGraph5_aMCatNLO/cards/production/13TeV/tH_t_topdecay_LH_narrow_M1100/tH_t_topdecay_LH_narrow_M1100_proc_card.dat
dmajumder: Yes, we decided to take the route, as the convergence time was way too large for tops decaying all the way t->bW-> b ferm ferm. For our model the polarization effect is small for most kinematic parameters. For those that matter, decaying up the W is considered sufficient.

bendavid:
The script indeed was intended to generate the param_card automatically from the model, then apply the modifications to that from the customize_cards.  So no param_card is needed in that case.

Are you sure the param_card that was provided matches exactly the one which is generated automatically by the model?

bin/MadGraph5_aMCatNLO/cards/production/13TeV/tH_t_topdecay_LH_narrow_M1100/tH_t_topdecay_LH_narrow_M1100_proc_card.dat
bendavid: Ok, but did you check that at least the finite width of the W is correctly taken into account?

dmajumder:
Apparently yes, as it was copied from the authors.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/tH_t_topdecay_LH_narrow_M1100/tH_t_topdecay_LH_narrow_M1100_proc_card.dat
dmajumder: The W has zero width in the LHE files as it is not decayed. For a decaying W, the width is set via the parameter "WW", so if decayed should be 
DECAY  24 2.085000e+00 # WW 
by default. Is that what you are asking?

bendavid: No,
I am asking if the width is actually reflected in the kinematics of the events after running them through pythia.  What I am afraid happens is that madgpraph produces the W's exactly on shell, and then pythia leaves them exactly on shell when doing the decays.  This should be checked, and if you want to use pythia for the decays then there may be a setting which needs to be changed in madgraph to sample properly the breit wigner of the W.  First step is to check what actually happens by running 100 or 1000 events through pythia and checking the distribution of the status 1 lv mass.

dmajumder: In Run1, samples were genertd only up to the T. All subsequent decays were done in Pythia6. If the width treatment between Pythia6 and Pythia8 are the same, we will get the same result. 
In any case this will be negliglible for MT in case of lv decay and M in case of jj decay. So a narrow-width approximation does not compromise the results.

bendavid: The effect of the W width is not negligible when looking at Mt tails, but ok maybe that is not relevant for your selection...

bendavid: Can you at least check what this configuration is doing with 1000 events
a) so that we know what we are producing
b) so that we have a comparison point in case we try to fix it at the madgraph configuration level (which will probably be useful also for other samples in the future)

dmajumder: We can do a fulll-fledged Pythia simulation if needed, but here are the mass distributions for the T quark, top and W, straight from the LHEFs. 

 * T quark mass showing width
hnpart11_m_tp

 * top quark mass showing width
hnpart11_m_t

 * W mass showing 0 width (undecayed)
hnpart11_m_w

bendavid: Indeed I would like to see for the W's the mass of the lepton-neutrino system formed from the status 1 lepton and neutrino.  It's the interaction between madgraph and pythia that matters here.  I think it should be fast/strightforward to run 1000 events through pythia.

dmajumder: Here's the GEN level M(lnu) where both l, nu comes from the W which comes from the top.
mlnugen

The mass is "predictably" at M(W) as Pythia does not insert a width. We did some research and apparenly there is no straightforward way to restore the width without a hack.

bendavid:
Indeed have discussed with the authors and there is not really a better solution for the widths at the moment.  So we can move forward, but just be aware of this point when analyzing the samples.

Unfortunately this pull request cannot be merged on account of other conflicting changes in the meantime. (In particular the model was already added in another pull request, so that part can simply be removed from this pull request, then it should be mergable again)

dmajumder:
@bendavid The conflict should be resolved now.

bendavid:
Sorry can you please remove completely the modifications of gridpack_generation.sh

bendavid:
No, sorry, now the pull request deletes the gridpack_generation.sh file, which is clearly not desired.  Please update the pull request so it does not make any changes to this file (either by restoring the version from the current master, or just starting from clean branch with the other files and opening a new pull request)

bendavid:
Any news on this?  Production cannot start until this is resolved.

If having trouble with git, suggest to start from a clean area and copy files in by hand.

dmajumder:
Apologies, was away. Does this work now?

bendavid:
Sorry, the pull request is still making changes to gridpack_generation.sh (you can see it in the list of "Files changed" at the top)

dmajumder:
Ok I close this now to re-initiate a new one

PR #309, opened 2015-04-14
LHE producers for EXO-RunIIWinter15GS for diboson

davidsheffield:
from Shin-Shan Eiko Yu

### DIFF
--- a/python/ThirteenTeV/BulkGraviton_VBF_WW_inclu_narrow_M1000_13TeV-madgraph_cff.py
+++ b/python/ThirteenTeV/BulkGraviton_VBF_WW_inclu_narrow_M1000_13TeV-madgraph_cff.py
@@ -1,0 +1,13 @@
+import FWCore.ParameterSet.Config as cms
+
+# link to cards:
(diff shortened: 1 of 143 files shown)


bendavid:
links to github should be to a specific git revision rather than "master" (since "master" can change underneath if things are updated later)

davidsheffield:
Do you mean that when I am creating a pull request, instead of trying to merge with cms-sw/genproductions master I should use the most recent commit? In this case cms-sw/genproductions 91ab3ea30e3c2280e4c31fdd7072a47eb2e5bdaa

bendavid:
No no,
I mean that for the links inside the cards
"# link to cards:
+# https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson/Spin-2/BulkGraviton_VBF_WW_inclu/BulkGraviton_VBF_WW_inclu_narrow_M1000
+"

These should point to the specific github revision of the madgraph cards used to generate the gridpacks.

davidsheffield:
I forgot that there were links to github in the fragments, hence my confusion. I've changed them to go to most recent commit of this repository at the moment, 91ab3ea.

PR #311, opened 2015-04-15
Requesting Z(->qqbar) + Jets

rappoccio:
For EXO VV all-hadronic analysis, we need Z+Jets, where Z decays hadronically. We only request HT > 600 GeV because the trigger cannot go any lower.


bendavid:
Are you sure you don't want the gamma* contribution here?

rappoccio:
Hi Josh,

I've been going back and forth on this myself. The chances are slim that we'll be able to see this at all (it's hidden under multijets). @ahinzmann may disagree, but I think we just need the on-shell Z production. 

Cheers,
Sal

ahinzmann:
I think we need no gamma*, only Z. Anything that is not peaky is for sure not visible at all under multijets. ALTAS showed they see W(qq)+jets, that's why we have this Z(qq)+jets as well.

bendavid:
Ok, given that the modification to the shape of the peak from interference will be ~completely washed out by the jet resolution, agreed this makes sense.  (Also practically it's probably not possible to do the full 6 jet process with Z/gamma* interference)

rappoccio:
Thanks Josh!

PR #312, opened 2015-04-16
Wgg enriched in FSR and ISR

wvieri:


bendavid:
param cards are ignored by the gridpack scripts.  Are these actually needed, or do they just correspond to the default values?

bendavid:
These cards correspond to 4 flavour scheme production.  Any particular reason why that is preferred here?  Any reason not to use off-diagonal ckm elements (is the analysis selecting or vetoing additional jets, do you care about the flavour composition of those?)

bin/MadGraph5_aMCatNLO/cards/production/13TeV/Wgg/FSR/LNuAA_FSR_NLO_PtG10_run_card.dat
```diff
@@ -0,0 +1,155 @@
...
+#***********************************************************************
+    1   = lpp1    ! beam 1 type (0 = no PDF)
+    1   = lpp2    ! beam 2 type (0 = no PDF)
+ 4000   = ebeam1  ! beam 1 energy in GeV
+ 4000   = ebeam2  ! beam 2 energy in GeV
+#***********************************************************************
+# PDF choice: this automatically fixes also alpha_s(MZ) and its evol.  *
+#***********************************************************************
+ 'lhapdf'  = pdlabel   ! PDF set
+ 236400    = lhaid     ! if pdlabel=lhapdf, this is the lhapdf number
```
bendavid: This is definitely not the correct pdf.
(this corresponds to "NNPDF23_nnlo_collider_as_0116")

bendavid:
Sorry this is now pulling in other changes and would need to be rebased.

Please just make a new pull request from a clean area.

PR #314, opened 2015-04-16
PhaseSpace:mHatMin = 350

perrozzi:
added cut PhaseSpace:mHatMin = 350


bendavid:
Can you explain a bit why the extra cut is needed?

perrozzi:
The cut was included in the HtauTau requests from which I cloned the cfg,
Quoting a mail from Clemens Lange, who required the Hbb samples:
"I think it’s a good idea to have this cut. In the past, we didn’t have it, but basically what happens is that you get a very long tail towards low masses because of certain parton configurations. However, we calculate our acceptance within a ±30% window of the generator mass, so this minimum 0.7*generator_mass cut would be fine, maybe set it to 0.6*generator_mass to be safe. The cut is documented here: http://home.thep.lu.se/~torbjorn/pythia82html/PhaseSpaceCuts.html"
Let me know if you want more info.

PR #315, opened 2015-04-16
gridpack cards for T'+b, T'->b+W including model in gridpack_generation....

peiffer:
This is a new pull request to replace #307. Now, the VLQ model is included and not taken from Devdatta's request (#305).


bendavid:
Sorry I still don't see the modification to gridpack_generation.sh here.

peiffer:
Here is the missing commit.

PR #317, opened 2015-04-16
Cards for tH_b LH couplings

roberts42:
MG Cards for tH_b LH couplings. Mass 800-1800 GeV.


stosi:
The cards look OK. Are you sure you want an unmatched sample?

PR #320, opened 2015-04-17
Cards for TTW and TTZ

nstrobbe:
Cards for TTWJets and TTZJets, NLO with FXFX merging, decays using MadSpin.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/TTWJets_5f_NLO_FXFX/TTWJets_5f_NLO_FXFX_customizecards.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/TTWJets_5f_NLO_FXFX/TTWJets_5f_NLO_FXFX_customizecards.dat
@@ -1,0 +1,4 @@
+#put card customizations here (change top and higgs mass for example)
+set param_card mass 6 172.5
+set param_card yukawa 6 172.5
+set param_card mass 25 125.0
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/TTWJets_5f_NLO_FXFX/TTWJets_5f_NLO_FXFX_madspin_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/TTWJets_5f_NLO_FXFX/TTWJets_5f_NLO_FXFX_madspin_card.dat
@@ -1,0 +1,32 @@
+#************************************************************
+#*                        MadSpin                           *
+#*                                                          *
+#*    P. Artoisenet, R. Frederix, R. Rietkerk, O. Mattelaer * 
+#*                                                          *
+#*    Part of the MadGraph5_aMC@NLO Framework:              *
+#*    The MadGraph5_aMC@NLO Development Team - Find us at   *
+#*    https://server06.fynu.ucl.ac.be/projects/madgraph     *
+#*                                                          *
+#************************************************************
+#Some options (uncomment to apply)
+#
+#directory for gridpack mode
+set ms_dir ./madspingrid
+
+#initialization parameters
+ set Nevents_for_max_weigth 250 # number of events for the estimate of the max. weight
+ #set BW_cut 15                # cut on how far the particle can be off-shell
+ set max_weight_ps_point 400  # number of PS to estimate the maximum for each event
+# 
+
+set max_running_process 1
+
+# specify the decay for the final state particles
+#(inclusive top decays in this case)
+decay t > w+ b, w+ > all all
+decay t~ > w- b~, w- > all all
+decay w+ > all all
+decay w- > all all
+
+# running the actual code
+launch
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/TTWJets_5f_NLO_FXFX/TTWJets_5f_NLO_FXFX_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/TTWJets_5f_NLO_FXFX/TTWJets_5f_NLO_FXFX_proc_card.dat
@@ -1,0 +1,13 @@
+import model loop_sm-ckm_no_b_mass
+#switch to diagonal ckm matrix if needed for speed
+#import model loop_sm-no_b_mass
+
+define p = p b b~
+define j = j b b~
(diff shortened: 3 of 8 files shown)


bendavid:
For ttZ I think you want to generate directly t t l+ l- (also including taus) to include the gamma* contribution.

nstrobbe:
Ok, that would make sense. 
We also need the other decays, so do you then suggest to do something like:
generate p p > t t~ l+ l-
add process p p > t t~ nu nu~
add process p p > t t~ q qbar
(with appropriate multiparticle labels, and adding the 1j counterparts)

bendavid:
Hi Nadja,
The qqbar case is really tricky to handle properly.  Are you sure you care
about this?  This should be negligible compared to inclusive ttbar + 2jet
production no?

simply doing add process p p > t t~ q qbar will not give you what you want,
since you will get instead pure qcd production rather than gamma*/Z
mediated production of the additional jets.

nstrobbe:
Hmm.. We for sure need the invisible decay of the Z, since that results in ttbar + MET. 
I was thinking to include also the hadronic decays for completeness, and for analyses that ask for boosted objects where the mass would matter.

bendavid:
Ok, if you need the hadronic case, the most feasible is to do p p > t t~ z and do the decay to qqbar in madspin.  This means you miss the gamma* contribution for the qqbar, but that should be fine.

This implies that you need to split into two samples at least.  (If you also want the off-shell breit-wigner tail for the neutrinos, then best is to produce the ll and nu nu together without madspin, and the qqbar in a separate sample with madspin if really needed)

alkaloge:
Another syntax is like
generate p p > t t~ l l / h @1
add process p p > t t~ vl vl / h @2
add process p p > t t~ z / h, z > q q  @3

so you get the q q~ via on-shell Z

bendavid:
That won't work at NLO though (since decay chain syntax is not supported)

nstrobbe:
Can we then do 
generate p p > t t~ l l / h @1
add process p p > t t~ vl vl / h @2
add process p p > t t~ z / h @3

and then decay the top and z in madspin, with only including z > q q

bendavid:
You can try, but this might cause problems (I'm not sure if madspin will
correctly skip the events with nothing to decay)

If not then it's sufficient to just split into two samples as said.

nstrobbe:
Madspin will in any case still have the top quarks to decay right? 

Would it also make sense to do
generate p p > t t~ l+ l- $ h @1
instead of 
generate p p > t t~ l+ l- / h @1
?

That way we are keeping the off-shell higgs and interference effects for the tau+ tau- final state. Or do you think this will be completely negligible?
The tth sample is being generated as p p > t t~ h, with the higgs decayed by Pythia.

bendavid:
The $ syntax is not supported at NLO (only $$ h, > h >, and /h), so no way to do it like that (anyway, don't think that's too important)

The possible problem with
generate p p > t t~ l l / h @1
add process p p > t t~ vl vl / h @2
add process p p > t t~ z / h @3

has to do with the manipulation of the cross sections and relative weighting between processes which madspin will attempt when it sees that the included Z decays account only ~70% of the total branching ratio.  As said, you could try it, but it might have problems in this direction.

nstrobbe:
We decided that we will try to generate it in one sample. We will of course check whether things make sense. 
I have updated the cards accordingly.

bendavid:
Alright, suggest to try with the ttZ first.  If that works you could also consider to do something similar for ttW (generate directly p p > t t l v for the leptonic case and do the hadronic decays with madspin) since there could at least in principle be some small polarization effects there which would not be captured by madspin. (These effects are likely pretty small, on the other hand, if you run into problems with combinatorics in madpsin in the ttW case it might even be faster to do it that way)

Will merge the cards as is for now and you can update the ttW ones if you want later.

bendavid:
(for the ttZ the thing to check is that you get the expected proportion of hadronic/leptonic/invisible decays)

bendavid:
Ah, one more thing.

When the inclusive ttbar requests were prepared, we had the TOP group set the maximum mll from the parton shower to 1GeV.  Looking at the google doc, this was matched with the HIG requests on lines 247-248

If you want a strictly continuous phase space when combining the inclusive ttZ with the ttbar sample, then you could also set mll>1GeV.  If you don't care and will mainly use this to cover phase space where the two leptons are resolvable, then also fine as it is.

nstrobbe:
What is the difference between the Higgs requests and this one in fact?

bendavid:
The Higgs request is only ttll (without the hadronic or invisible).  But indeed I think this is somewhere we simply missed in the deduplication/rationalization.  (And no one from Higgs has prepared either of those samples so far)

nstrobbe:
Since they have not prepared them yet, we can still harmonize the requests now. 
I like the idea of having a continuous phase space. How about we have two requests, with adjusted total number of events (keeping the overall budget the same, at 750k)
1) this one, with mll>5, 600k events
2) a second one, only the ttll, with 1 < mll < 5, 150k events
Does it make sense?

bendavid:
The planned Higgs requests were split into 1<mll<10 and mll>10, which probably makes more sense.  On the other hand, if you prefer to have one sample with mll>5 (and some analyses may use only that one) then also fine.

So yes, agreed to prepare two requests as you suggest.  Up to you whether to split at 5 or 10GeV.  (Then once the gridpacks are ready we may still discuss about the number of events to make sure everyone is happy.)

nstrobbe:
Presumably, if people care about the 5-10 range, they will also want to use the 1-5 range. So I will make the split at 10 as you suggest. 

Shall I already make a new pull request now with this change? Or should I wait till we know that the gridpack went fine, and there is no issue with the relative proportions of l/nu/q?

bendavid:
Yah you can already make a new pull request.

nstrobbe:
actually, looking at the cards, how does one implement a maximum mll cut at NLO?

bendavid:
You need to do it "by hand" using cuts.f
Take a look at
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/dyellell012j_5f_NLO_FXFX_M10to50/dyellell012j_5f_NLO_FXFX_M10to50_cuts.f

PR #321, opened 2015-04-17
Wgg at 8TeV

wvieri:


bendavid:
param card is ignored by the scripts!  I hope it's not needed.  If not, please simply remove it.

Then for the proc cards, no need to copy and paste the whole output here.  Also the model is still not correct for 5f.

I think it is sufficient to just have:

define p = p b b~
define j = j b b~
generate p p > l+ vl a a [QCD] @0
add process p p > l- vl~ a a [QCD] @0
output LNuAA_FSR_NLO_PtG10


Then finally, what is mlnpmlnacut?  Is this recognized by madgraph out of the box or requires some customization?

What is actually the difference between the ISR and FSR cards here?

bendavid:
Sorry I still see param_cards here.

wvieri:
Hi, yes sorry, I have deleted them now (as far as I understand, in the case of the aQGC the param_cards are actually used).

bin/MadGraph5_aMCatNLO/cards/production/8TeV/QGC/QGC_LM/LNuAA_QGC_LM_LO_PtG10_run_card.dat
```diff
@@ -0,0 +1,276 @@
...
+     4000     = ebeam2  ! beam 2 total energy in GeV
+#*********************************************************************
+# Beam polarization from -100 (left-handed) to 100 (right-handed)    *
+#*********************************************************************
+        0     = polbeam1 ! beam polarization for beam 1
+        0     = polbeam2 ! beam polarization for beam 2
+#*********************************************************************
+# PDF CHOICE: this automatically fixes also alpha_s and its evol.    *
+#*********************************************************************
+ 'cteq6l1'    = pdlabel     ! PDF set
```
bendavid: These should use the appropriate nnpdf30lo sets as production samples.

bendavid:
Yes ok, paramcards for the qgc stuff is fine, since those don't go through the scripts.  Pdfs for those should be fixed though.

bendavid:
The qgc cards look like for 4-flavour production (can you confirm the b is massive in this model?)

In that case the correct pdf would actually be 263400

bendavid:
can I suggest to decouple the aqgc cards from the wgg NLO and split into two pull requests?  the wgg NLO looks fine and could probably already be merged.  The qgc will take us a bit longer to converge it looks like.

PR #322, opened 2015-04-18
cards for single VLQ, T'+t, T'->bW

peiffer:
New gridpack cards for single VLQ production.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/TptTobW_M1000_LO/TptTobW_M1000_LO_customizecards.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/TptTobW_M1000_LO/TptTobW_M1000_LO_customizecards.dat
@@ -1,0 +1,7 @@
+set param_card mass 8000001 1000
+set param_card bsm 1 0.100000e+00
+set param_card bsm 2 0.000000e+00
+set param_card bsm 3 0.100000e+00
+set param_card bsm 4 0.000000e+00
+set param_card bsm 5 0.100000e+00
+set param_card bsm 6 0.000000e+00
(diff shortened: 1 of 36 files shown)


peiffer:
Comment from Nikoloz Skhirtladze, who prepared these cards:

I allowed Madgraph to do a full decay chain of W coming from Tp; but I didn't allow Madgraph to do a further decay of the associated top, the decay will happen using PYTHIA. I think this should be OK because associated top kinematics are pretty similar for the Tp mass range we consider for run II (you can see the pt and eta distributions below: Histogram 12 and 13). 

I have also made some relevant kinematic distributions for several mass points, and all looks ok to me:
http://tier2.ihepa.ufl.edu/~skhirtladze/TptTobW/HTML/

bendavid:
Ok, but again I think this has the problem of the narrow widths.  I believe it's possible to fix this at the madgraph configuration level, but I'm not sure off the top of my head what the needed setting is.

peiffer:
In these requests, the W decay is fully included in Madgraph (or in Pythia for the associated top), only the associated top is always on-shell. A top quark with a narrow width should not be a huge problem (compared to W or Z bosons with narrow width).

bendavid:
Ok, indeed we don't have a better solution for the narrow width issue, so we could go ahead like this.  Was it at least attempted to do the top decay in madgraph as well?

peiffer:
For this sample, it makes only a marginal difference whether the associated top is decayed via Pythia or Madgraph. We did some tests with other single VLQ modes and found that a precise description of top decays with Madgraph is only relevant in the case of tops from VLQ decays where the spin correlation is important. For this sample it is not worth the effort to include the top decay, it would just blow up the CPU usage.

PR #329, opened 2015-04-21
added dark matter madgraph cards

violatingcp:
Madgraph cards for dark matter models. Moels are available here:
/afs/cern.ch/work/p/pharris/public/bacon/prod/CMSSW_7_3_3/src/git/genproductions/afs/

and hopefully soon here:

Additional templates, intallation, and grid scripts are added in the directories for a reference


covarell:
About this pull request plus #330 and #332 .
These include changes in the POWHEG scripts (inlcuding a new variable number 9 which is not documented) plus a new generator interface to JHUGen
Needs to be discussed at GEN meeting first.

covarell:
More on MG cards: it seems like all these processes are run at LO. 
If so please change the PDF to LO: 260000 -> 263000

covarell:
AFAICT https://github.com/cms-sw/genproductions/pull/366/ supersedes this. closing.

PR #330, opened 2015-04-21
updated powheg branch with powheg change

violatingcp:
Adding dark matter models for Powheg. Also updated powheg scripts so that cross sections are stored in the LHE file and so that installation script can use custom scripts on top.

This powheg requires a brand new release avaible here 
 /afs/cern.ch/work/p/pharris/public/bacon/prod/CMSSW_7_3_3/src/git/genproductions/afs/

Soon to be availble here 

### DIFF
--- a/bin/Powheg/create_powheg_tarball.sh
+++ b/bin/Powheg/create_powheg_tarball.sh
@@ -40,6 +40,9 @@ echo "%MSG-POWHEG number of events requested = $nevt"
 rnum=${8}
 echo "%MSG-POWHEG random seed used for the run = $rnum"
 
+skipgen=${9}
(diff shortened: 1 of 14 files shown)


covarell:
Please put back line 151 in create_powheg_tarball.sh

Also we should make sure that your personal version of powheg is compatible with the ones currently in /afs/cern.ch/cms/generators/www (i.e. all other processes than the new ones are identical to the code already there)

syuvivida:
Why are lines 224-227 removed from create_powheg_tarball.sh?

violatingcp:
Hi,

The line is added lines 224-227 were moved up to 206 so that the generation
step can be skipped. This allows for the possibility of parameter scans in
a new script added into the directory

Phil

PR #331, opened 2015-04-21
added MCFM dark matter setup

violatingcp:
Setup to run MCFM gridpack production target design was to build a slew of gridpacks for dark matter simplified models. However, scripts can be easily adapted for you own needs. Some work could be be done to make them nicer, but this is a good start.


bin/MCFM/runcmsgrid_template.sh
```diff
@@ -0,0 +1,28 @@
...
+
+ncpu=${3}
+echo "%MSG-MG5 number of cpus = $ncpu"
+
+LHEWORKDIR=`pwd`
+basename=BASENAME
+
+cd $LHEWORKDIR/$basename
+export LD_LIBRARY_PATH=${LD_LIBRARY_PATH}:/cvmfs/cms.cern.ch/slc6_amd64_gcc481/external/lhapdf/6.1.5/lib/
+export LHAPATH=/cvmfs/cms.cern.ch/slc6_amd64_gcc481/external/lhapdf/6.1.5/share/LHAPDF/
```
bendavid: This is surely not safe for production.  You'll need to follow the procedure to determine this path automatically from the release as it's done eg in the madgraph run scripts

bin/MCFM/runcmsgrid_template.sh
```diff
@@ -0,0 +1,28 @@
...
+sed "s@EVTS@$nevt@g" input_gen.DAT > input_g0.DAT
+sed "s@XRAN@$rnum@g" input_g0.DAT  > input.DAT
+./mcfm
+#Replace the negative so pythia will work
+sed "s@-1000022@1000022@g" FILENAME > $LHEWORKDIR/cmsgrid_final.lhe
```
bendavid: This can be taken out, as we have a fix at pythia config level.

PR #337, opened 2015-04-23
cards for TTGJets

nstrobbe:
I put ptj to 20 to be consistent with the other ttbar requests. 
I also put 10  = ptgmin, and 2.5  = etagamma. I did not change the photon isolation compared to the example cards.


bendavid:
I would suggest to raise etagamma to 2.6 to leave some margin, and to reduce R0gamma to 0.1 or even 0.05, since 0.4 may actually remove quite a lot of phase space.

(The other point here is that the isolation is not really with respect to the final state jets, since this calculation is applied before the top decays.)

nstrobbe:
Thanks for the comments. I wasn't sure what to do with the photon isolation. 
I'll commit the changes.

nstrobbe:
Changes are committed.

PR #339, opened 2015-04-24
Cards for EXO-RunIIWinter15pLHE for RPVresonantToEMu

davidsheffield:
from Soeren Erdweg

These are for a set of pLHE requests that use CalcHEP. I've received the model files from the requester that produced these and included the extracted header from the LHE files. I'm unfamiliar with CalcHEP, so hopefully this is enough for you to check that the LHE files were generated correctly.


bendavid:
Can calchep be interfaced to lhapdf easily?  It would be better to use the same pdfs as for LO madpgrah samples (nnpdf30lo alpha_s=0.130) rather than cteq6l as is being done here.

serdweg:
So in principal it should be possible to interface lhapdf. How easy this will be is unclear. But of course it should have no effect on the signal if we choose a differnt pdf set. Therefore we would prefer to stay with cteq6l, as we are a little bit short on time to redo the complete LHE production.

bendavid:
Why do you say it should have no effect on the signal?

serdweg:
No effect in the sence, that the cross section and shape should not change as the only initial state for the signal is ddbar annihilation.

bendavid:
Well the pdfs can still affect the kinematics (even if just the rapidity distribution of the heavy particle)

One of the advantages of moving to NNPDF is also that one can have consistent uncertainty sets corresponding to the central value.
On the other hand given that calchep is not providing the pdf variation weights, you can also do the reweighting of the central value if needed when computing pdf uncertainties. (Which I guess you'll do the old fashioned way with lhapdf reweighting)

So this is ok for now, but please consider updating for any future calchep production.

PR #342, opened 2015-04-29
Cards for the additional mass points for VBFH, based on the existing one...

chayanit:
...s


syuvivida:
These cards have all necessary settings for getting scale/PDF weights.
How are the hwidth determined?

covarell:
These values are available from the LHC XS WG page

covarell:
For the moment the agreement with the Higgs group is to generate maximum 3 mass points per final state (125 GeV and two close to ir). Why are these needed?

chayanit:
Dear Roberto (cc MC Higgs conveners),

I might have missed the agreement but these mass points are preparing for Higg-Exo requests.

Thanks and regards,
Chayanit
Higgs-Exo MC contact


Sent from my iPhone

On Apr 29, 2015, at 8:18 PM, Roberto Covarelli <notifications@github.com<mailto:notifications@github.com>> wrote:


For the moment the agreement with the Higgs group is to generate maximum 3 mass points per final state (125 GeV and two close to ir). Why are these needed?

PR #343, opened 2015-04-29
added MadGraph5_aMCatNLO cards for H->llGamma (dalitz), for ggH, vbf and...

mmachet:
... vh


bendavid:
First a basic question:  Is there interest in doing these samples instead at NLO?  Should be possible with the right cards.

If staying at LO are you sure you want to do this without jet matching?

Then some issues with the current cards
1) param cards are not supported by the scripts.  parameter changes should be handled through the customizecards.dat mechanism
2) what is the ANO_HEFT_JGF_v2 model and what does it do?  Is this just to provide the effective h-a-l-l vertex?  Are we sure this is not available in any of the built-in models?
4) This looks like 4-flavour scheme production (with massive b's, b's not included in the proton definition).  Most of the SM higgs samples are being done with 5-flavour production.
5) Pdf configuration is not consistent with the rest of the production.  Depending on 4flavour or 5flavour, please see
https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/examples/dyellell01234j_4f_LO_MLM
6) use_syst  needs to be T
7) Suggest to loosen a bit the lepton eta cuts (at least to 2.6)

covarell:
And "gridpack" must also be true...

bendavid:
that one gets forced to true automatically by the scripts at least.

andreypz:
Hi all,

Let me answer some of the questions for H -> llg production.

 - Yes, we would be interested in a proper  NLO production for this process, with jet-matching and all included. If you could point to us an appropriate datacard for this, that would be great. Hopefully it would work with our decay/model.

If staying at LO, I'm not sure what would jet-matching mean in that case. I thought it is for NLO, which needs to be matched with pythia. If the process is at LO, there is no jets to match. (?)


 - The ANO_HEFT_JGF_v2 model that we use is  h ->llgamma vertex, which includes some anomalous coupling, but we only use SM Higgs part of it. I think h->llg process was not available in MG5 a few years ago, when we started. Now, I just checked with a recent mg5amcnlo version and, indeed, this process is available in the standard HEFT model. However it only includes h -> g*g -> llg process and does not include h -> Zg -> llg part.
In principle it is not a big issue for us, since the contribution from h -> Zg is quite small after m(ll)<50 GeV requirement. I checked some basic distributions from the LHE files produced with both models and they look similar, so, we may use it.
However, the other thing we want is  non-zero mass of a muon (i.e. real muon mass). I don't think this feature is available in heft model, is it? One would need to modify it. And in that case we may as well just stick with the ANO_HEFT model, where I already did it.

 - We would certainly like to produce the Higgs via the latest recommended  way. You mentioned 5-flavour production. Is that enough to  define: "p = g u c d s b u~ c~ d~ s~ b~" for that?
Same for PDF: sure we can use different PDFs, but I assume one can change PDFs for production, in the datacard? 

- Currently, I modified the cards to address some issues (changing the lepton cuts, setting use_syst, using lhapdf, etc). 
I put them together with the model at lxplus:
/afs/cern.ch/user/a/andrey/public/HiggsDalitz/MCat13TeV

Could someone, please, have a look and tell if these are Okay to be used?

bendavid:
Hi Andrey,
Looking briefly, neither the heft model (LO) nor the HC_NLO_X0_UFO (NLO) implement directly an effective h-l-l-gamma vertex.  Instead they both implement the effective h-gamma-gamma vertex, and the NLO model implements as well the effective h-z-gamma vertex.  (Then madgraph is able to use those vertices together with off-shell photons)

Is there a reason the h-l-l-gamma effective vertex would be expected to give a better description than the combination of the h-gamma-gamma and h-z-gamma effective vertices?

stosi:
> We would certainly like to produce the Higgs via the latest recommended way. You mentioned  5-flavour production. Is that enough to define: "p = g u c d s b u~ c~ d~ s~ b~" for that?
Same for PDF: sure we can use different PDFs, but I assume one can change PDFs for production, in the datacard? 

- To use a 5 flavour scheme, in addition to adding the b and b~ quarks in the proc_card in the way you mention above, you should also set maxjetflavor to 5 in the run_card :
5 = maxjetflavor

Concerning the PDFs, you can set the PDFs in the run_card in the following block (example for 5f, NLO):
# PDF choice: this automatically fixes also alpha_s(MZ) and its evol.  *
lhapdf    = pdlabel   ! PDF set                                     
292200    = lhaid     ! if pdlabel=lhapdf, this is the lhapdf number

And then there is a block for the PDF variations:
# Reweight flags to get scale dependence and PDF uncertainty           *
# For scale dependence: factor rw_scale_up/down around central scale   *
# For PDF uncertainty: use LHAPDF with supported set                   *
 .true.   = reweight_scale   ! reweight to get scale dependence
  0.5     = rw_Rscale_down   ! lower bound for ren scale variations
  2.0     = rw_Rscale_up     ! upper bound for ren scale variations
  0.5     = rw_Fscale_down   ! lower bound for fact scale variations
  2.0     = rw_Fscale_up     ! upper bound for fact scale variations
 .true.  = reweight_PDF     ! reweight to get PDF uncertainty
 292201   = PDF_set_min      ! First of the error PDF sets 
 292302   = PDF_set_max      ! Last of the error PDF sets

PR #344, opened 2015-04-29
Adding cards for TTTT

nstrobbe:


bendavid:
The cards look fine.  Slightly worried about the combinatorics in madspin for 4 tops, but maybe it will be ok.

PR #345, opened 2015-04-29
Cards for TTWW, 5f, inclusive decays

nstrobbe:


bendavid:
Assume you're happy to have only the on-shell W decays here.

nstrobbe:
Yes, on-shell decays only should be fine here.

nstrobbe:
This gridpack has been running for more than a week now, and many jobs are actually failing because they are hitting the 2nw cpu limit. So I fear that the automatic resubmission of those jobs won't help to make the gridpack actually finish. Is there a solution to get this gridpack produced anyways? Is the only way to try it at LO, or are there other things one can try?

bendavid:
At which step are the long running jobs appearing?

nstrobbe:
at the "setting up grid" step

bendavid:
OK.  The only available knob would be to increase req_acc to let's say 0.003.

Since this is not a huge request, that should be ok.  (Wouldn't want to generate 100M events with such a gridpack though)

nstrobbe:
I've been trying to create this gridpack using those settings, but now I run into memory issues. 
I tried requesting 80GB for both memory and disk, and I still get this error: 

```
INFO: Compiling code 
ESC[1;31mCommand "launch -n pilotrun" interrupted with error:
MadGraph5Error : Impossible to compile /pool/lsf/nstrobbe/679816811/TTWW_4f_NLO/TTWW_4f_NLO_gridpack/work/processtmp/madspingrid/full_me/SubProcesses/P0_gux_ttxwpwmux_t_bwp_wp_veep_tx_bxwm_wm_vtxtam_wp_csx_wm_vtxtam directory
        Trying to launch make command returns:
            [Errno 12] Cannot allocate memory
        In general this means that your computer is not able to compile.
Please report this bug on https://bugs.launchpad.net/madgraph5
More information is found in '/pool/lsf/nstrobbe/679816811/TTWW_4f_NLO/TTWW_4f_NLO_gridpack/work/processtmp/pilotrun_tag_1_debug.log'.
Please attach this file to your report.ESC[0m
quit
```

Any ideas? Is that update from the authors on its way? (Does it make sense to try the beta version?)

nstrobbe:
I should also comment that I switched to 4fs for this one as well (as for the WW, WWZ, WWW), and it is using diagonal ckm decays.

bendavid:
Ok, so the error is at the madspin step now.

You could try with the new version (no longer beta, but not yet cms default) in
https://github.com/cms-sw/genproductions/tree/mg230
though there are no significant changes I know of there that would be expected to help here.

Switching to LO may or may not help (the combinatorics is not reduced, but at least madspin would not have to deal with the extra emission for the real corrections)

What would help (but might be painful) is splitting up the sample in a few chunks according to different decays.

PR #348, opened 2015-04-30
Add cards for very large width Z' to ttbar

blinkseb:
Very large width (30% of the mass) Z' to ttbar samples have been requested. Here're the corresponding MadGraph cards.

Cheers,


stosi:
Everything looks OK to me

PR #350, opened 2015-04-30
Cards triboson

nstrobbe:
Cards for WWW, WWZ, WZZ, ZZZ, WWG, all in 5 flavour scheme, NLO with FXFX merging.


bendavid:
Few comments.
1) cards look ok in general
2) as far as I know no one has ever tried triboson + 1 jet, but ok, it might work.  Be prepared to fall back to NLO inclusive in case of problems (but other handles can be tried first as well)
3) most of the Z contributions could also have a gamma* contribution in principle.  Probably l+l-l+l-l+l- is impossible for the triple Z case, but if you cared about this contribution, w w l+ l- is likely possible, and w l+ l- l+ l- is maybe possible (but probably in both cases at the expense of not including the extra jet).  This of course introduces the same potential issue of mixed madspin/non-madpsin decays and might require further splitting the samples.

nstrobbe:
On 2) I had a look in the MG paper, and Table 3 lists triboson + jet, so I figured it should be doable. Of course, if there are issues, we are happy to switch back to NLO inclusive. 

On 3) As far as I know, we did not have the gamma* contribution in the past and nobody complained about it. So to avoid further complications, I would suggest to keep things without the gamma*.

PR #353, opened 2015-05-04
HZJ ZToEE and ZToMM for ZH HToInvisible analysis

chayanit:


covarell:
All PDFs should be changed to 260000 (NNPDF3.0nlo, and the fragment names updated) unless there is a reason to use CT10.
Again, for the first submission of samples, you should not have that many mass points. Even if this is for Higgs-EXO, you should choose 2-3 benchmark masses (cards however can be approved of course).

PR #355, opened 2015-05-04
Wgg at NLO

wvieri:


stosi:
Just a thought: in the run_card, is the photon isolation cut not too wide ?
0.3 = R0gamma   ! Radius of isolation code

bin/MadGraph5_aMCatNLO/cards/production/8TeV/Wgg/FSR/LNuAA_FSR_NLO_PtG10_run_card.dat
```diff
@@ -0,0 +1,154 @@
...
+#***********************************************************************
+# Tag name for the run (one word)                                      *
+#***********************************************************************
+  tag_1     = run_tag ! name of the run 
+#***********************************************************************
+# Number of LHE events (and their normalization) and the required      *
+# (relative) accuracy on the Xsec.                                     *
+# These values are ignored for fixed order runs                        *
+#***********************************************************************
+ 10000 = nevents ! Number of unweighted events requested
```
bendavid: can set this to 1000

bin/MadGraph5_aMCatNLO/cards/production/8TeV/Wgg/FSR/LNuAA_FSR_NLO_PtG10_run_card.dat
```diff
@@ -0,0 +1,154 @@
...
+#***********************************************************************
+  tag_1     = run_tag ! name of the run 
+#***********************************************************************
+# Number of LHE events (and their normalization) and the required      *
+# (relative) accuracy on the Xsec.                                     *
+# These values are ignored for fixed order runs                        *
+#***********************************************************************
+ 10000 = nevents ! Number of unweighted events requested 
+  0.001 = req_acc ! Required accuracy (-1=auto determined from nevents)
+    -1 = nevt_job! Max number of events per job in event generation.
```
bendavid: And this to 20

bin/MadGraph5_aMCatNLO/cards/production/8TeV/Wgg/FSR/LNuAA_FSR_NLO_PtG10_run_card.dat
```diff
@@ -0,0 +1,154 @@
...
+   0  = drll_sf ! Min distance between opp. sign same-flavor lepton pairs
+   0  = mll     ! Min inv. mass of all opposite sign lepton pairs
+  30  = mll_sf  ! Min inv. mass of all opp. sign same-flavor lepton pairs
+#***********************************************************************
+# Photon-isolation cuts, according to hep-ph/9801442                   *
+# When ptgmin=0, all the other parameters are ignored                  *
+#***********************************************************************
+  10  = ptgmin    ! Min photon transverse momentum
+  -1  = etagamma  ! Max photon abs(pseudo-rap)
+ 0.3  = R0gamma   ! Radius of isolation code
```
bendavid: Agreed with Silvano that this should be reduced.  Would suggest 0.05

bendavid:
Aside the inlined comments, I completely forgot that madspin is needed in this case in order to assign the correct helicity flags to the taus.

This can be done with a "dummy" decay.  You can take the madspin cards directly from
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/examples/wellnu012j_5f_NLO_FXFX/wellnu012j_5f_NLO_FXFX_madspin_card.dat

bendavid:
Finally, for the cuts.f files, please add a brief comment at the top explaining what is implemented and where for the custom cuts.

stosi:
Perhaps I'd also add a cut on the eta of the photons (2.6 ?)
-1  = etagamma  ! Max photon abs(pseudo-rap)

bendavid:
cutting on the photon eta could also be fine.  Depends what kind of acceptance studies analyzers might want to do.

bin/MadGraph5_aMCatNLO/cards/production/8TeV/Wgg/FSR/LNuAA_FSR_NLO_PtG10_run_card.dat
jkunkle: I don't particularly care what the isolation is, but it should be looser than our analysis cut of 0.4.  However I found that the success of the integration is dependent on where this is set, so I can check if 0.05 works.  Note that the Madgraph default is 0.4

jkunkle:
For the madspin card:  I had a lot of issues with using a madspin card -- it caused a crash when testing the created gridpack.  Also I could care less if spin correlations are kept in the tau decay.  I can try to use the given madspin card, but in the case that it does not work I would prefer to not spend more time debugging.

bendavid:
The difference in lepton pt spectra from leptonic tau decays can be large.  The card pointed to should hopefully work without issue...

bin/MadGraph5_aMCatNLO/cards/production/8TeV/Wgg/FSR/LNuAA_FSR_NLO_PtG10_run_card.dat
jkunkle: Actually the value of this cut is important.  This cut defines our fiducial volume which is at 0.3, so we keep it there

bendavid: But your fiducial volume is presumably defined at particle level no?  This cut is at parton level before the shower, so it's better if it's looser than your fiducial cut.  (But maybe doesn't need to be as small as 0.05)

jkunkle: Ah, you're right.  I was confusing myself.  So yes I think its fine as long as its 0.3 or less.  But I don't understand why we want 0.05.

bendavid: Well the idea is to try to be very safe that this cut is looser than any generator-level or reconstruction level isolation.

jkunkle: Right 0.3 already gives us some buffer and a smaller value would give more buffer.  As long as the cut is <= 0.3 and the integration converges I can't argue for one value or another.  So if you think 0.05 should be used, then I'll try that

bendavid: 0.05 is what was being used also for some run 2 samples.

In general, we try to make samples as generally usable as possible, hence the preference for looser cuts.
I'm not aware of any detailed study on how much headroom is really needed on parton level vs generator level isolation definitions, so better to be safe in my opinion.

jkunkle:
I just realized that I used a photon eta cut of 3.0 instead of 2.6 in the cards that I gave Vieri, but I assume that 3.0 is OK

bendavid:
sure, that's fine of course.

PR #356, opened 2015-05-04
Add very large width Z' to ttbar Madgraph cards for M=3TeV

blinkseb:
Hi,

A point at 3 TeV has also been requested after the generation of the others. You'll find the cards for Madgraph in this PR

Thanks

Cheers,


stosi:
Everything looks OK to me

PR #357, opened 2015-05-05
Cards for DYbb, consistent with HT binned DY samples

nstrobbe:


stosi:
The cards look OK to me

bin/MadGraph5_aMCatNLO/cards/production/13TeV/DYBBJetsToLL_M-50_5f_LO_MLM/DYBBJetsToLL_M-50_5f_LO_MLM_proc_card.dat
```diff
@@ -0,0 +1,15 @@
+import model sm-ckm_no_b_mass
+
+define p = p b b~
+define j = p
+define bb = b b~
+define l+ = e+ mu+ ta+
+define l- = e- mu- ta-
+
+# Specify process(es) to run
+add process p p > l+ l- bb bb / h @0
```
bendavid: usually preferred to have 
"generate" rather than "add process" for the first one
(I have seen this cause problems in the past at least)

nstrobbe:
Of course, sorry (copy paste issue)
Fixed now.

bendavid:
Ok, looks good now.

One thing to be aware of is of course even when properly combining this with the inclusive HT-binned sample, one will still have a contribution of b-jets from the parton shower (in addition to mistags) from the inclusive sample.

Most robust way to do the overlap removal is to explicitly veto events from the inclusive sample with two b quarks in the LHEEventInfo (this is less error prone than trying to do it at the genparticle level)

nstrobbe:
Ok, thanks. Is there a twiki or example documenting how to access/read the LHEEventInfo?

bendavid:
Not that I'm aware of, but it should be straightforward from the dataformats

https://github.com/cms-sw/cmssw/blob/CMSSW_7_1_X/SimDataFormats/GeneratorProducts/interface/LHEEventProduct.h

https://github.com/cms-sw/cmssw/blob/CMSSW_7_1_X/SimDataFormats/GeneratorProducts/interface/LesHouches.h

What you're interested in is the IDUP and ISTUP vectors containing the pdgid and status.  (You want to count ISTUP=1 b-quarks)

This data format is basically a 1:1 representation of what's in the <event> tag in the lhe file.

PR #358, opened 2015-05-05
Combined P.R.

dnash86:
I have accidentally combined these two issues by pushing both to my repository:

From Shih-Chuan, for the 8 TeV request

From Haneol Lee, for a wmLHE request

Is there an easy way to split them?  The 8 TeV configs should be an easy commit, the madgraph cards probably need to be reviewed first.


bendavid:
when in doubt just start from a clean area.
But please use unique branch names for each pull request.
eg.
git checkout -b majoranacards
git push myrepo majoranacards

dnash86:
I think the issue is that I am following these instructions: https://twiki.cern.ch/twiki/bin/viewauth/CMS/GitRepositoryForGenProduction
and I have already pushed both sets of changes to that repository.  So now when I try to make a separate branch, it includes both sets of changes.  Unless I am missing a key step?

bendavid:
Indeed those instructions will need to be updated.
Now I understand the problem: starting from a clean area doesn't help, because you checkout your repository rather than the base one.

You should be able to do something like
git pull upstream master:somenewbranch
git checkout -b somenewbranch
<make needed changes and commit>
git push origin somenewbranch

(basically to pull the master branch from the current upstream genproductions to a new branch in your local repository which can then be modified and pushed)

dnash86:
That is what I thought was happening.  Thanks for the help, I think I've gotten it figured out.  Here's a PR with the MG cards:

https://github.com/cms-sw/genproductions/pull/359

Will be setting up the one with just the configs momentarily.  This one can be closed, in that case.

PR #359, opened 2015-05-05
Cards from H Lee

dnash86:
Madgraph cards from Haneol Lee

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_majorananeutrino/MajoranaNeutrinoToEE_Tchannel_M-100_5f_LO/MajoranaNeutrinoToEE_Tchannel_M-100_5f_LO_customizecards.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_majorananeutrino/MajoranaNeutrinoToEE_Tchannel_M-100_5f_LO/MajoranaNeutrinoToEE_Tchannel_M-100_5f_LO_customizecards.dat
@@ -1,0 +1,2 @@
+set param_card mass 80000002 100.00
+set param_card yukawa 8000002 100.00
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_majorananeutrino/MajoranaNeutrinoToEE_Tchannel_M-100_5f_LO/MajoranaNeutrinoToEE_Tchannel_M-100_5f_LO_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_majorananeutrino/MajoranaNeutrinoToEE_Tchannel_M-100_5f_LO/MajoranaNeutrinoToEE_Tchannel_M-100_5f_LO_proc_card.dat
@@ -1,0 +1,40 @@
+#************************************************************
+#*                     MadGraph5_aMC@NLO                    *
+#*                                                          *
+#*                *                       *                 *
+#*                  *        * *        *                   *
+#*                    * * * * 5 * * * *                     *
+#*                  *        * *        *                   *
+#*                *                       *                 *
+#*                                                          *
+#*                                                          *
+#*         VERSION 2.1.0                 2014-02-21         *
+#*                                                          *
+#*    The MadGraph5_aMC@NLO Development Team - Find us at   *
+#*    https://server06.fynu.ucl.ac.be/projects/madgraph     *
+#*                                                          *
+#************************************************************
+#*                                                          *
+#*               Command File for MadGraph5_aMC@NLO         *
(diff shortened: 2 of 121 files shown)


bendavid:
Is there a particular reason to prefer 4 flavour scheme here?  (And does the model have massive b quarks by default?)

dnash86:
I am asking the analyzers this now.

dnash86:
I have a response from the analyzers:
----------------------
Is there a particular reason to prefer 4 flavour scheme here?
>> The signal does not include b jets so I didn’t change this part in the run card.
     (4 flavour scheme is default in MG5 run card)
     Diagrams for this channel can be found in arXiv:1308.2209v2.
     But I’ll check it and reply later.

And does the model have massive b quarks by default?
>> Yes. I attach the param card.
--------------------

So I now have the param card from them as well - I can add it to this PR if you'd like.

As for the 4 flavor scheme - will this be okay or something we'll need to change in the run cards?

dnash86:
Sorry about the large text size in the response, I guess I used a few symbols that github recognized as formatting without knowing it.

bendavid:
So 4-flavour scheme means in particular that b jets will not be included among the extra jets in the final state.

If the analysis is not sensitive to additional bjets, then I would say this is probably fine.  Otherwise 5-flavour scheme would be preferred.
(5-flavour is actually the default for cms production, despite not being default in madgraph)

bendavid:
(can we start cc'ing the analyzers when making these pull requests btw "by using their @githubusername tag in the description"

Would probably make the communication more efficient.

dnash86:
4 vs. 5 flavor scheme:  It seems okay then, but I will let the analyzer address that as well.

cc'ing the analyzers:  Agreed, I'm asking about his github account name now.

dnash86:
@helee can you comment on Josh's comment above?  Is the analysis in fact not sensitive to additional bjets?

helee:
There is no reason to exclude b jets.

I'll use 5-flavour scheme. Thanks.

dnash86:
Updated cards with 5-flavor scheme from @helee.

helee:
I have a question.

I don't understand well what 'massive b quarks' means.
In this model, c quarks and b quarks have masses. Is there any problem with using 5f scheme?

bendavid:
Yes there is.  5-flavour scheme requires a model with massless b-quarks.  (Many models have multiple variants provided through the "restrict.dat" files in the model directory, eg "sm" and "sm-no_b_mass")

If this is not the case for your model, you would have to create a modified version with an additional restrict-no_b_mass.dat file to set the masses as desired.

Otherwise you'll have to stay with 4 flavour scheme, and not have the b's included in the additional jets.

helee:
I have two questions about particle masses.

1.If I set b and c quarks masses to zero using customized card for 5f scheme, is it OK?

2.In the param_card.dat, there are pole masses and yukawa masses.
   These two mass values are equal for b quarks, c quarks, electrons and muons, but different for top quarks and heavy neutrinos. Is it better to set them equal?

bendavid:
Better for pole masses and yukawa's to be equal.

Masses can be modified in the customizecards but NOT set to 0.  Doing this properly requires modifying the model with a restrict-no_b_mass.dat (like in the SM case), because madgraph needs to know what is zero and what is non-zero already at process generation time.

helee:
OK. I'll add restrict card soon.
And one more question.
In the SM case, not only b and c quarks but also electrons and muons are massless in the restrict_no_b_mass.dat file. Is it better to set masses of e and mu to zero in my case?

bendavid:
Yes, it would be preferred to set u, d, s, c, b, e, mu (and corresponding yukawa couplings) to zero, since this greatly speeds things up.

Massive electrons and muons are only relevant in case one wants to do pair production down to very low masses. (ie below 2*m_mu)

helee:
Questions about the alpha_s value.
1.The alpha_s value is 0.118 in this model. Is this value used for parton showering using PYTHIA?
2.I've used NNPDF30_lo_as_0130 for the PDF so far, but I think I have to use NNPDF30_lo_as_0118. Is it right?

bendavid:
The alpha_s value is not taken from the model, but rather from the pdf.

For LO generation 0.130 is preferred, and a corresponding LO pdf should be used.  (4f or 5f consistent with the process and model)

dnash86:
Added updated cards from @helee and also changed generation script to reference model:
https://github.com/cms-sw/genproductions/pull/359/files#diff-56bfba1423a749bf2d1829498400b7b4R127
https://github.com/cms-sw/genproductions/pull/359/files#diff-56bfba1423a749bf2d1829498400b7b4R252

PR #361, opened 2015-05-06
Descope W+4jets to W+3jets for W->qq

rappoccio:
W+4jets with W->qq is taking way too long for practical purposes. We will not use much 4-jet phase space so descoping to W+3jets.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/WJets_HT_LO_MLM/WJetsToQQ_HT-600toInf/WJetsToQQ_HT-600toInf_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/WJets_HT_LO_MLM/WJetsToQQ_HT-600toInf/WJetsToQQ_HT-600toInf_proc_card.dat
@@ -13,13 +13,11 @@ generate p p > w+, w+ > q q $$ t t~ h @0
 add process p p > w+ j, w+ > q q $$ t t~ h @1
 add process p p > w+ j j, w+ > q q $$ t t~ h @2
 add process p p > w+ j j j, w+ > q q $$ t t~ h @3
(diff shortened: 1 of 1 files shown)


rappoccio:
Thanks Josh!

nstrobbe:
I was under the impression that the gridpack was produced already? Is there an issue with it?

/cvmfs/cms.cern.ch/phys_generator/gridpacks/slc6_amd64_gcc481/13TeV/madgraph/V5_2.2.2/WJets_HT_LO_MLM/WJetsToQQ_HT-600toInf/v1/WJetsToQQ_HT-600toInf_tarball.tar.xz

rappoccio:
OH! I had no idea. I thought I was producing them! If someone managed to get the W+4jets working then that would be fantastic. 

Should I also be doing the Z+Jets with Z->qq or are those also going to be done centrally? 

https://github.com/cms-sw/genproductions/pull/311

nstrobbe:
From line 390 in the rationalized spreadsheet, you can see that SUS was responsible for the W->qq. So we produced it along with the leptonic decays. 

We don't have the Z->qq in our list, so it would be great if you could handle that one.

bendavid:
aha,
Sal, can you open another pull request reverting the changes from this one (to avoid confusion in the future)

nstrobbe:
There's a button for that I think.

rappoccio:
Thanks everyone. I will do Z+Jets. @nstrobbe can you point me at the config you used for CMSSW and/or the instructions you followed? I have the gridpack already done for Z+jets. 

I don't actually know if I can do the revert from a "push" of mine (I can always modify it by hand but it's kind of messy). I think to do this cleanly you may have to do this from the server side. Any ideas? 

Cheers,
Sal

PR #363, opened 2015-05-06
make cleaning of temp output more robust during gridpack production

bendavid:


bendavid:
Fixes a moderately rare error reported by SUS group during ht-binned gridpack production.

PR #364, opened 2015-05-06
Added Templated Madgraph so Josh can read easily

violatingcp:
This is series of templated scripts to generate dark matter from templates. Template files are in Proc/template.

To run 
./install.py --carddir DMAxial_W/template --dmrange X --medrange X --name DMAxial_W_Gen

Files contain the LO lhaid (263000) as per the request of Roberto.


bendavid:
What was the purpose for copying/reimplementing runcmsgrid, gridpack_generation, and patches?

Also param_card.dat files should be removed, assuming that they are automatically generated from model + customize_cards

stosi:
The gridpack parameter in the run_card should be .true.

bendavid:
(that anyways gets overwritten by the scripts at least)

violatingcp:
>What was the purpose for copying/reimplementing runcmsgrid,
gridpack_generation, and patches?

A. grid_pack generation is split into two parts (install and integrate) so
that it can be run on the batch without making a a million madgraph
releases  Also, thought it was nice to make a standalone mg setup script
for my own leisure,
B.  patches is the same, I just copied it to this directory.
C. runcmsgrid is the same minus the HERAPDF sets, which were crashing.


>Also param_card.dat files should be removed, assuming that they are
automatically generated from model + | >customize_cards

I kept them in as a backup. They are not used at any point.

bendavid:
Ok, can you remove the param cards from the pull request?
(having them here implies they are used by the scripts which tends to confuse people)

PR #365, opened 2015-05-06
WZTo2L2Q 012j 5f NLO FXFX datacards

perrozzi:
Please cross-check


bin/MadGraph5_aMCatNLO/cards/production/13TeV/WZTo2L2Q012j_5f_NLO_FXFX/WZTo2L2Q012j_5f_NLO_FXFX_run_card.dat
```diff
@@ -0,0 +1,146 @@
...
+#***********************************************************************
+# Cuts on the charged leptons (e+, e-, mu+, mu-, tau+ and tau-)        *
+# (more specific gen cuts can be specified in SubProcesses/cuts.f)     *
+#***********************************************************************
+   0  = ptl     ! Min lepton transverse momentum
+  -1  = etal    ! Max lepton abs(pseudo-rap) (a value .lt.0 means no cut)
+   0  = drll    ! Min distance between opposite sign lepton pairs
+   0  = drll_sf ! Min distance between opp. sign same-flavor lepton pairs
+   0  = mll     ! Min inv. mass of all opposite sign lepton pairs
+  30  = mll_sf  ! Min inv. mass of all opp. sign same-flavor lepton pairs
```
bendavid: This might not be good for everyone.  Have asked in the thread.

bendavid:
Might be challenging with two extra jets, but can try.
Have asked in the mail thread about the minimum mll cut.

Cards look fine otherwise.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/WZTo2L2Q012j_5f_NLO_FXFX/WZTo2L2Q012j_5f_NLO_FXFX_run_card.dat
bendavid: Ok, let's go with 4GeV here.

I don't think the cross section becomes too outrageous, but after you have the gridpack, can check how much stats you lose in the Z-peak region from this and boost the size of the request to compensate if you like.

perrozzi: just to be paranoid: you mean
30  = mll_sf  ! Min inv. mass of all opp. sign same-flavor lepton pairs
to become
4  = mll_sf  ! Min inv. mass of all opp. sign same-flavor lepton pairs
right?

bendavid: yes, exactly

perrozzi:
concerning the number of jets, it two extra can be challenging I would prefer to start safe because could take up to 2 weeks to verify that is too demanding.
Shall I try one or even 0 extra jets?
What are others doing?

bendavid:
1 jet is for sure possible (Roberto did it for the 4 lepton case which is more complex than this one)

bendavid:
WZ cards look fine now.  Is it intentional that you have also the Zbb cards here?

perrozzi:
thanks, it was not intentional (the files didn't even appear in my local repository)
removed now. the gridpack creation went smooth and it is now ready

PR #366, opened 2015-05-07
Madraph darkmatter full

violatingcp:
The full suite of Madgraph Dark Matter Data cards in another pull request, as per request from Josh, to make validation made easier.


covarell:
@bendavid , need your input here

PR #367, opened 2015-05-07
Script for parallel POWHEG gridpack production.

yuanchao:
Initial commit from private repo.

### DIFF
--- a/bin/Powheg/run_pwg.py
+++ b/bin/Powheg/run_pwg.py
@@ -1,0 +1,606 @@
+#!/usr/bin/python
+
+'''
+Script for POWHEG generator production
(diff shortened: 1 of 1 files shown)


covarell:
It creates some branching from create_powheg_tarball.sh, meaning that if one is updated, the other must be updated as well.

Is there no other way to read in the .sh file than copying it by hand in the long f.write( statement, so that if this is updated, the other gets automatically updated as well?

yuanchao:
If some ankers could be place in create_powheg_tarball.sh, it would be possible. Though some extra code needed to parse the script in instead of the long f.write() statement, which would be still not pretty. So far I'll merge the changes by hand. Applying the diff patches is rather simple.

PR #368, opened 2015-05-07
HZJ ZToEE and ZToMM NNPDF30 for ZH HToInvisible analysis

chayanit:
Changed to NNPDF30 and hopefully, 4 mass points per channel as the first pass are accepted. Thanks!


syuvivida:
These files look OK to me. BTW, would your gridpack production benefit from the parallel-run mode? 
(as reported by Yuan Chao in the generator meeting on May 5, 2015).

chayanit:
I have a look at Yuan Chao slides and his twiki page but I don't quite understand what this parallel-run mode does for? Could you clarify?

syuvivida:
If it takes a long time to produce HZJ gridpacks, this script written by Yuan Chao allows the possibility to produce gridpacks using multiple CPUs.

chayanit:
OK. I already created gridpack using single-run mode and it took only 20 minutes for each mass point, so I haven't tried the parallel-run mode. Is that ok?

syuvivida:
OK, then there's no need to use parallel runs.

chayanit:
OK. Then, can you complete this pull request? Thanks in advance.

syuvivida:
Hi, Chayanit, only Josh and Roberto have the permission to merge the pull request.

PR #369, opened 2015-05-07
Last WW powheg for HWW, Fully hadronic

rebecacern:
Final WW sample of three, HWW background. Fully hadronic.


stosi:
Everything looks OK

PR #371, opened 2015-05-08
cards for aMCatNLO ttbar sample with m(tt)>1 TeV filter

peiffer:
Cards are a copy of tt012j_5f_ckm_NLO_FXFX cards, only the cuts.f file has been added with some modifications:

https://github.com/peiffer/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/tt012j_5f_ckm_NLO_FXFX_mtt1000toInf/tt012j_5f_ckm_NLO_FXFX_mtt1000toInf_cuts.f#L384-L399


stosi:
everything looks ok

PR #372, opened 2015-05-08
Madgraph cards from W Wulsin

dnash86:
From analyzers:

"These cards are for direct electroweak chargino-chargino and chargino-neutralino production.  We would like to use samples generated with these processes for the disappearing tracks search.  The chargino should be stable in the Madgraph simulation because its decay will be simulated by GEANT.  This can be done by setting the chargino decay width to 0.0 in param_card.dat; let me know if I should also send that card"

They will wait to produce gridpacks until these cards are reviewed/approved.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZZTo4L01j_5f_NLO_FXFX/ZZTo4L01j_5f_NLO_FXFX_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZZTo4L01j_5f_NLO_FXFX/ZZTo4L01j_5f_NLO_FXFX_proc_card.dat
@@ -1,0 +1,15 @@
+#import model loop_sm-ckm_no_b_mass
+#switch to diagonal ckm matrix if needed for speed
+import model loop_sm-no_b_mass
+
+define p = p b b~
+define j = j b b~
+
+define ell+ = e+ mu+ ta+
+define ell- = e- mu- ta-
+
+generate p p > ell+ ell- ell+ ell- [QCD] @0
+add process p p > ell+ ell- ell+ ell- j [QCD] @1
+
+output zz4ell01j_5f_NLO_FXFX -nojpeg
+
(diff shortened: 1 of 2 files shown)


bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_DisappTrks/C1C1_C1N1_2j_proc_card.dat
```diff
@@ -0,0 +1,43 @@
...
+#************************************************************
+#*                                                          *
+#*               Command File for MadGraph 5                *
+#*                                                          *
+#*     run as ./bin/mg5  filename                           *
+#*                                                          *
+#************************************************************
+import model mssm
+define p = p b b~
+define j = j b b~
```
bendavid: The rest of the cards look like they are set up for 4 flavour scheme.  The mssm model has massive b-quarks by default, so this would be 4 flavour scheme.  But in that case the proton and jet definition should not include the b quarks.

If you want the proton and jet definition to include the b quarks (also so that the additional partons include also b jets) then you should switch to "import model mssm-no_b_mass" and change also the pdfs and maxjetflavour for consistent 5 flavour scheme.

(5 flavour scheme is the default in CMS for jet matching with SM processes at least)

bendavid:
The gridpack scripts do not support param cards as input (intentionally).

Please provide a customizecards.dat which changes this parameter instead.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_DisappTrks/C1C1_C1N1_2j_proc_card.dat
wulsin: We do not have any preference for the 4 flavor scheme, so if the 5 flavor scheme is the default in CMS, we should use that.  I can switch to "import model mssm-no_b_mass" but I'm not sure which pdfs and maxjetflavour to use.  Are there any examples using this model I can follow?  I couldn't find any in our repository using this command:
genproductions/bin/MadGraph5_aMCatNLO/cards/production/13TeV]$ grep -r "mssm-no_b_mass" * 
Thanks for your help with this!

bendavid: The pdfs and other settings would be identical to the sm case.  The only special feature here is that you have to use
"import model mssm-no_b_mass"
instead of the sm model used there

see for example:

dnash86:
Accidental closure.  Conversation continues on #386

PR #373, opened 2015-05-08
cards for T'T' -> everything, including model (diagCKM) in gridpack_g…

dnowatsc:
…eneration.sh

I created cards for T'T' -> tHtH, tHtZ, tHbW, tZtZ, tZbW or bWbW, each for two chiralities of the T' (LH, RH) and 7 different mass points as it was decided in the B2G PAG.
T', t and W are decayed by Madgraph which is necessary to keep spin correlations which can cause a significant difference in the event topology between LH and RH chiralities (see e.g. https://indico.cern.ch/event/391467/contribution/5/material/slides/0.pdf, slide 5). Z and H are not decayed since this does not lead to such a big difference.
Please don't hesitate to ask me if you have any further question.


stosi:
From a quick look, the cards seem OK

bendavid:
Remember that the undecayed particles will have a zero width.  For Higgs this is fine, for Z decays to leptons it is probably not ok, since you would have an unphysical dilepton mass distribution.

Suggest trying to do at least the Z decays in madgraph as well.  (Or if it's too slow/not converging with the Z hadronic decays, to split off the leptonic Z's into a separate sample and do at least those in madgraph)

dnowatsc:
Thanks for your comment, the Z width is indeed an issue. I brought this up for discussion in the B2G group so it would be nice if you could hold off on merging until this is clarified.
Also I noticed a small bug in the customize_cards (apparently I modified the wrong couplings there) that I will resolve quickly

dnowatsc:
Ok, after discussing these matters internally in the B2G group we decided to change our approach completely and go back to letting MadGraph only do the production and let the decays of the T' quarks be handled by Pythia.
I will revise the code and update the PR accordingly tomorrow.

dnowatsc:
I modified the cards for the T'T' production (also changing the naming convention) so that now MG handles only the production of the Tprimes and all the decays are done in Pythia, as I already explained in my previous commit.
Please check the cards and let me know if there's anything wrong with them.

dnowatsc:
I also added the cards for B'B' production in a separate commit, they are essentially written in the same manner as the T'T' cards

bendavid:
Sorry I see that the pull request is now not making any changes to gridpack_generation.sh,  is the needed model already included in the scripts from some previous pull request that was already merged?

dnowatsc:
Yes, it is, sorry I forgot to mention that. As a matter of fact, the model that I included in one of my previous commits (STP_UFO_freeWidth_diagCKM) is not even needed anymore for the TpTp (or BpBp) production as it is right now as this only affected the CKM matrix elements for decaying the W in MG. Since all the decays are now handled by Pythia, the standard model for VLQs (STP_UFO_freeWidth) is now used.
Thanks for pointing that out

PR #375, opened 2015-05-09
added MadGraph5_aMCatNLO cards for TGJets sample

mmachet:
I started from the cards for TTGJets of pull request #337. I just removed the t~


syuvivida:
Don't you need the equivalent processes for the t~ or will they be generated separately? 
p p > t~ a [QCD]
p p > t~ a j [QCD]

mmachet:
Yes you are right. Is it sufficient to add:
generate p p > t~ a [QCD]
add process p p > t~ a j [QCD]

in the run card and:
decay t~ > w- b~, w- > all all
in the madspin card?

mmachet:
Yes you are right. Is it sufficient to add:
generate p p > t~ a [QCD]
add process p p > t~ a j [QCD]

in the run card and:


decay t~ > w- b~, w- > all all


in the madspin card?

2015-05-10 13:52 GMT+02:00 syuvivida <notifications@github.com>:

syuvivida:
In the process card, you could have the following lines

generate p p > t a [QCD] @0
add process p p > t a j [QCD] @1
add process p p > t~ a [QCD] @2
add process p p > t~ a j  [QCD] @3
In the madspin card,
Yes, I think adding this line would be enough
decay t~ > w- b~, w- > all all

mmachet:
ok thank you,
I'll change the cards and commit.

Martina

2015-05-11 7:49 GMT+02:00 syuvivida <notifications@github.com>:

mmachet:
cards changed.

2015-05-11 9:54 GMT+02:00 Martina Machet <martina.machet@cern.ch>:

syuvivida:
The cards look nice to me now. But please wait for the review of Josh/Roberto/Silvano.

mmachet:
ok!

2015-05-11 11:15 GMT+02:00 syuvivida <notifications@github.com>:

bendavid:
Hi Martina,
I'm not sure about the physics content of this.

Single top is always produced in association with an extra jet (and usually it's preferred to do the generation in the 4 flavour scheme where the additional bjet is explicit as well).  I would suggest to discuss with the top MC contacts/experts (mainly @maierbenedikt who has been handling the single top requests, and start from the default single top cards and add a photon)

Have you actually tried p p > t~ a [QCD]?  Are there any diagrams found?

maierbenedikt:
Hi Martina, 

p p > t(~) a  will indeed give you no diagrams for the Standard Model, nor at LO or NLO.

I assume you want to generate the single-top t-channel (i.e. the dominant production mode) with a photon?

We have had an 8 TeV tGamma sample produced at LO:

https://cms-pdmv.cern.ch/mcm/requests?page=0&dataset_name=tGamma_8TeV_madgraph

We used basically the 5fs (b's in proton) to generate

p p > a t j

However, in the meanwhile, the trend goes to NLO in the 4fs, so this is why I would advise to do 


generate p p > t b~ j a $$ w+ w- [QCD] 
add process p p > t~ b j a $$ w+ w- [QCD]

and decay the top with MadSpin.
This should be straightforward and is in line with what we do for the normal t-channel and also for tZ.

mmachet:
Hi Benedikt,
thanks for the prompt reply and the useful suggestions.
Josh, if you agree with Benedikt's proposal I'll change the cards and
commit them again.

Thanks

Martina

2015-05-12 16:29 GMT+02:00 maierbenedikt <notifications@github.com>:

bendavid:
Yes, I agree that Benedikt's proposal makes sense.

mmachet:
I changed the cards, hope to have put correctly the 4fs (I also changed 4 = maxjetflavor in run card).

Martina

mmachet:
sorry, changing the name of the directory I did reset in order to eliminate the previous directory.
Now I did a new pull request (#382)

PR #377, opened 2015-05-10
cards for bprime+b to tW, soon : bZ / bH

dgonzal:
all cards for B -> tW for right and left chirality for masses between 700 & 1800 in 100 GeV steps. In the first production only 2-3 RH mass points willl be used rest for completness


stosi:
The cards seem OK; the syntax p p >  b13 > jet bb tt WW identical for b13~ is a bit misleading, but it should work correctly.

dgonzal:
yes it is, but unlucky it seems to be needed. Otherwise MadSpin is complaining about too many decays or other things.

bendavid:
Any particular reason to use madspin at all here?  (Did you try doing all the decays as madgraph decay chains?)

dgonzal:
First thing was to try decay chains, but you end up with about ~44000 diagrams. It took seraval hours to get few hundred calculated and the prediction to finish was in the order of week(s) for one mass point and more then 200 gb diskspace. Anyway it crashed, and also for similar particles with this final state (X53) it never worked. 

The reason we have to go this way is to preseve spin correlation, since it changes e.g. the distributions of leptons in a non negligible way (specially the pT distributions). 

A different solution that I could just check today is to use a model that only has a diagonal CKM, which is what madspin does internally anyway. But it takes about 3 times more time for the computation and generation of the events and gives very similar results.

dgonzal:
The B -> Zb or Hb are eaier in the sense that they can be done fully in madgraph and follow the usual syntax

bendavid:
Ok, if it's working like this with madspin then it should be fine.
But why do you say madspin is using a diagonal ckm model internally?  Madspin by default uses the same model used to generate the events (but it is possible to override that in the madspin card in special cases)

dgonzal:
Sorry if I was a bit to unspecific or wrong. As far a I know madspin does not consider subdominant W/top decays as e.g. W -> u b or c b

bendavid:
By default it does if the model used to generate the events includes the off-diagonal ckm elements.  (And if not then it doesn't of course)

dgonzal:
Ok, thanks remembered it wrongly

bendavid:
Ah, one more question.
Is the model already available in the gridpack_generation.sh?

dgonzal:
I didn't add it, because I just tested it over the weekend. It is available in https://cms-project-generators.web.cern.ch/cms-project-generators/ but not yet implemented in the the gridpack_generation.sh. I think someone else is planning to use/using it.

bendavid:
Ok, sorry, please make an additional pull request which adds the model used for these cards to the script.

The goal for central production is that it should be possible to reproduce the gridpacks based on what is in the repository out of the box.

PR #379, opened 2015-05-11

MichaelButtignol:


stosi:
Param_cards will be ignored by the scripts. Usualy parameters should be changed with respect to the default using the customizecards.dat 
You use a 4 flavor scheme: are b massive in the model?
Also, is this model already part of the gridpack_generation.sh script in git? The model should be added to
such that it should be available from

MichaelButtignol:
You are right, sorry! I will repair this mistake as soon as possible.

Thank you,
Michael

PR #381, opened 2015-05-11
WpWpJJ SMP same-sign WW

albertobelloni:
MadGraph cards for three WW same-sign SMP samples: EWK, QCD, EWK+QCD


stosi:
Concerning the PDF, for LO 5f scheme, one typically should use
'lhapdf'    = pdlabel     ! PDF set                                  
263000    = lhaid   

In the cuts, you have
100  = mmjj    ! min invariant mass of a jet pair 
is this intentional?

In the proc_card, the "generate" line typically looks like, for example,
generate p p > l+ l- bb bb j j @0
without the QCD/QED flags

albertobelloni:
Hi Silvano,

let me add the gridpack producers to the conversation. Thank you very much for your quick response. Best,

Alberto

AndrewLevin:
OK, we will move to the recommended pdf set. 

Yes, the mjj cut is intentional -- we have an mjj cut of 500 GeV in our event selection.

The QCD and QED orders are needed to define which process we are generating in this case. I am not sure what the purpose of the @0 is.

bendavid:
Coupling orders make sense of course.

The @0 and @1
ie
generate p p > something @0
add process p p > somethingelse @1

is just for book-keeping, so that events appear with different process ids in the event header.  You may not care here.

One other possible issue, with cut_decays=T, these cards are applying |eta|<2.5 on the leptons.  Are you sure you want this cut at lhe level?  (If so, at least consider loosening that to 2.6 to keep some headroom)

Otherwise these look fine modulo the pdf change.

Were you planning to try this at NLO as well?

AndrewLevin:
Hi,

Thanks, I have changed the pdfset, the etal, and added @0 and @1:

/afs/cern.ch/work/a/anlevin/public/madgraph_cards/

I will reproduce the gridpacks now.

We cannot use NLO for reweighting model parameters in madgraph, so we will do the entire analysis in LO.

Andrew

bendavid:
Thanks,
(Alberto please update the pull request)

albertobelloni:
Hello,

I updated the card files in the repository; the commit message is "updated cards after pull request: PDF; eta cuts - fixed conflicts 2". Best,

Alberto

bendavid:
Sorry, something went wrong with the git updates. the pull request is now touching hundreds of files.

bendavid:
Aside from the issue with the git issue,
Andrew, ok this argument to work at LO makes sense, but wouldn't it at least be useful to produce the SM samples at NLO for comparison/systematic uncertainties for the missing higher orders? (given that scale variations at LO are not super-well-defined)

AndrewLevin:
I am not an expert on this, but since scale variations were used in run 1, I assumed they were also OK in run 2.

Can we request both LO and NLO versions of these samples?

bendavid:
Well it's always good to improve things.  And yes, can you remind how large these samples are?  I think they are small enough to request both versions.

AndrewLevin:
150000 events each. Maybe we will request the NLO samples later if the QCD scale systematic, measured with scale variations, turns out to be one of the largest systematic uncertainties.

albertobelloni:
Hi Josh,

I added a comment to the pull request, I hope that that works; please let me know if you preferred a brand new pull request to be created. Thanks! Best,

Alberto

bendavid:
Hi Alberto,
Normally that should be fine, but in this case for some reason it has resulted in a large conflict.

Probably easier to just open a new pull request at this point than trying to undo it.

PR #382, opened 2015-05-12
committed new cards for TGJets sample also changing the name of the d…

mmachet:
…irectory


bendavid:
Actually in this case the sample is not an FXFX merged sample, so there are changes also needed in the run card.  Suggest here also to start from the single top cards.

mmachet:
Hi Josh,
I just have to change the line for the FXFX merging or also some other
parts?
For example, looking at the single top run card (
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/st_sch_4f_ckm_NLO/st_sch_4f_ckm_NLO_run_card.dat)
I see that some values (lines 58,71,72,73,74,75,79,80,96,97) are slightly
different from the ones in my card.

Thanks

Martina

2015-05-12 20:09 GMT+02:00 Josh Bendavid <notifications@github.com>:

PR #385, opened 2015-05-12
amcnlo dibosons 01j nlo

perrozzi:
WWTo4Q, WZTo1L1Nu, ZZTo2L2Q, ZZTo2Q2Nu

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/WWTo4Q01j_5f_NLO_FXFX/WWTo4Q01j_5f_NLO_FXFX_madspin_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/WWTo4Q01j_5f_NLO_FXFX/WWTo4Q01j_5f_NLO_FXFX_madspin_card.dat
@@ -1,0 +1,34 @@
+#************************************************************
+#*                        MadSpin                           *
+#*                                                          *
+#*    P. Artoisenet, R. Frederix, R. Rietkerk, O. Mattelaer * 
+#*                                                          *
+#*    Part of the MadGraph5_aMC@NLO Framework:              *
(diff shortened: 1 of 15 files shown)


bendavid:
Since the SUSY group generally cares about the off-shell phase space,
would be preferred for the WZ->lvqq case to generate directly ell+ vl z rather than doing the W decay with madspin.

Otherwise looks good.

perrozzi:
didn't have an example to decay both W in proc_card
is it sufficient to move there the definitions below, from madspin?

define ell+ = e+ mu+ ta+
define ell- = e- mu- ta-
decay w+ > ell+ vl
decay w- > ell- vl~

bendavid:
No, there is no "decay" in NLO mode.

I meant something like
define ell+ = e+ mu+ ta+
define ell- = e- mu- ta-

generate p p > ell+ vl z [QCD] @0
add process p p > ell+ vl z j [QCD] @1
add process p p > ell - vl~ z [QCD] @2
add process p p > ell - vl~ z j [QCD] @3

ie just generate directly ell+ vl z
(this would also include non-resonant diagrams if there were any, but that should be fine)

perrozzi:
hi, what about this pull request? it has been closed, not merged

bendavid:
Sorry, missed the update (actually github doesn't automatically notify us when the pull requests are updated, so please add a comment after doing so)

For the cards.  If you haven't made the gridpacks yet, consider lowering ptj to 15 for consistency with some of the others which have been done (I don't think this should matter much, so if you've already made the gridpacks, don't bother)

perrozzi:
I already created WZTo2L2Q, WZTo1L1Nu2Q, ZZTo2L2Q and I am waiting for ZZTo2Q2Nu... 
After that, only ZZTo4Q and WWTo4Q are missing, I might change ptj for them if requested (but as far as I remember they are relevant mostly for Hbb)

PR #386, opened 2015-05-13
Update from Wells

dnash86:
Unintentionally closed this PR:  #372 

This is just an update from that conversation.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_DisappTrks/AMSBchargino100GeV/C1C1_C1N1_2j_run_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_DisappTrks/AMSBchargino100GeV/C1C1_C1N1_2j_run_card.dat
@@ -1,0 +1,270 @@
+#*********************************************************************
+#                       MadGraph5_aMC@NLO                            *
+#                                                                    *
+#                     run_card.dat MadEvent                          *
+#                                                                    *
+#  This file is used to set the parameters of the run.               *
+#                                                                    *
+#  Some notation/conventions:                                        *
+#                                                                    *
+#   Lines starting with a '# ' are info or comments                  *
+#                                                                    *
+#   mind the format:   value    = variable     ! comment             *
+#*********************************************************************
+#
+#*******************                                                 
+# Running parameters
+#*******************                                                 
+#                                                                    
+#*********************************************************************
+# Tag name for the run (one word)                                    *
+#*********************************************************************
+  tag_1     = run_tag ! name of the run 
+#*********************************************************************
+# Run to generate the grid pack                                      *
+#*********************************************************************
+  .false.     = gridpack  !True = setting up the grid pack
+#*********************************************************************
+# Number of events and rnd seed                                      *
+# Warning: Do not generate more than 1M events in a single run       *
+# If you want to run Pythia, avoid more than 50k events in a run.    *
+#*********************************************************************
+  1000 = nevents ! Number of unweighted events requested 
+      0       = iseed   ! rnd seed (0=assigned automatically=default))
+#*********************************************************************
+# Collider type and energy                                           *
+# lpp: 0=No PDF, 1=proton, -1=antiproton, 2=photon from proton,      *
+#                                         3=photon from electron     *
+#*********************************************************************
+        1     = lpp1    ! beam 1 type 
+        1     = lpp2    ! beam 2 type
+     6500     = ebeam1  ! beam 1 total energy in GeV
+     6500     = ebeam2  ! beam 2 total energy in GeV
+#*********************************************************************
+# Beam polarization from -100 (left-handed) to 100 (right-handed)    *
+#*********************************************************************
+        0     = polbeam1 ! beam polarization for beam 1
+        0     = polbeam2 ! beam polarization for beam 2
+#*********************************************************************
+# PDF CHOICE: this automatically fixes also alpha_s and its evol.    *
+#*********************************************************************
+ 'lhapdf'    = pdlabel     ! PDF set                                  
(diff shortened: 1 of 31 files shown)


bendavid:
you should be able to reopen the pull request if that's what would be preferred.

dnash86:
Interesting, I looked for an "open" button near the "closed" message.  I suppose that it is elsewhere?  The problem was when I accidentally pushed changes that removed the updated files before committing my changes to them.

dnash86:
Question re: multiple mass points.  Should each mass point have its own subdirectory or can we just put each customizecards.dat file in the "exo_DisappTrks" folder?

bendavid:
Each mass point should have its own subdirectory preferably (so that it it is easy to see from a single request exactly which cards were used for that request)

dnash86:
Okay, thanks.  

Re: github use - I suppose we will continue on this thread now, not the one I accidentally closed?

bendavid:
yes that's fine.

dnash86:
Individual masses are now separated by directory, via the python script "makeDirs.py".  The analyzers were interested in including that script in this commit, if that is okay, for possible future use/alterations that may be necessary.

bendavid:
No problem to have that kind of script in the commit.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_DisappTrks/AMSBchargino100GeV/C1C1_C1N1_2j_run_card.dat
```diff
@@ -0,0 +1,135 @@
...
+#*********************************************************************
+# KT DURHAM CUT                                                      *
+#*********************************************************************
+ -1    =  ktdurham        
+ 0.4  =  dparameter 
+#*********************************************************************
+# maximal pdg code for quark to be considered as a light jet         *
+# (otherwise b cuts are applied)                                     *
+#*********************************************************************
+ 4 = maxjetflavor    ! Maximum jet pdg code
```
bendavid: maxjetflavour should be set to 5, since this is 5flavour scheme

bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_DisappTrks/AMSBchargino100GeV/C1C1_C1N1_2j_run_card.dat
```diff
@@ -0,0 +1,135 @@
...
+#*********************************************************************
+# Beam polarization from -100 (left-handed) to 100 (right-handed)    *
+#*********************************************************************
+        0     = polbeam1 ! beam polarization for beam 1
+        0     = polbeam2 ! beam polarization for beam 2
+#*********************************************************************
+# PDF CHOICE: this automatically fixes also alpha_s and its evol.    *
+#*********************************************************************
+ 'lhapdf'    = pdlabel     ! PDF set                                  
+  263400    = lhaid     ! if pdlabel=lhapdf, this is the lhapdf number
```
bendavid: pdf should be 263000 (5f scheme again)

bendavid:
There is actually another major problem here.  It looks like a whole section of the run card is missing (the part with ickkw and related settings).  This is needed in order to correctly configure the jet matching.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_DisappTrks/AMSBchargino100GeV/C1C1_C1N1_2j_run_card.dat
wulsin: That is fine.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_DisappTrks/AMSBchargino100GeV/C1C1_C1N1_2j_run_card.dat
wulsin: That is fine.

wulsin:
Responding to comment from @bendavid about the missing section of the run card:  Can you recommend another run card in the production/13TeV directory that we could use as a model, or perhaps even copy identically?  I have very little Madgraph expertise so would prefer to use a run card that has been validated for another sample.

bendavid:
You can see an example card for 5f LO with jet matching here

Due to some special features of the drell-yan process which are not applicable here, with respect to the run card here, you'll want to change ptj to 0, and auto_ptj_mjj to T.

wulsin:
Thanks, @bendavid for the suggestion.  I took the drell-yan card you pointed me to, made the 2 modifications you recommended, and saved it here:  /afs/cern.ch/user/w/wulsin/workspace/public/disappTrk13TeV/forDavid_PullRequest/exo_DisappTrks/C1C1_C1N1_2j_run_card.dat

@dnash86, could you add this new version to the PR?  Thanks!

dnash86:
@wulsin: Shouldn't the run changes also be propagated to the run cards in the AMSBchargino*GeV folders?

wulsin:
@dnash86, yes the C1C1_C1N1_2j_run_card.dat card should be identical in all of the AMSBchargino*GeV folders.  The makeDirs.py script will propagate the version in the top-level directory to each folder.

bendavid:
OK but since the wmlhe requests will link to the specific folders in github, we need the final cards in github for each specific gridpack.

wulsin:
@bendavid, yes I agree that the final cards should be in github for each subfolder.  I was just suggesting a quick way for David to make that change.

dnash86:
The script is ran, updated cards should be correct now (with the modification in both the base cards and the individual mass cards).

wulsin:
Thanks @dnash86 !

wulsin:
@bendavid, is there anything else that you think should be changed before approving this PR?

PR #387, opened 2015-05-13
let us try again, hopefully I will manage to open a pull request now:…

albertobelloni:
… updated WpWpJJ cards - PDF, eta cuts


bendavid:
Ok, looks good from git perspective.

One last point that I missed before is that since this is 5 flavour scheme, it should use a model with massless b's

so either
"import model sm-ckm_no_b_mass"  (keeps off-diagonal ckm elements)
or
"import model sm-no_b_mass" (diagonal ckm matrix)

depending on what is preferred.

albertobelloni:
Hi Josh,

thanks! Andrew, could you please select your preferred model? I am ready to update the cards and copy the gridpacks (I suspect that they may need to be updated). Best,

Alberto

AndrewLevin:
Hi Alberto,

We will use sm-no_b_mass. Yes, I need to remake the gridpacks with this model. I will let you know when they are ready. 

Andrew

covarell:
Superseded by https://github.com/cms-sw/genproductions/pull/392

PR #389, opened 2015-05-13
aQGC cards

wvieri:


stosi:
param_cards are generally ignored by the scripts. if you need to change any parameters, you should use a customize_card

bendavid:
@wvieri, these cards are for pLHE requests right? (since they need reweighting, which is not supported by the gridpack scripts yet)
In that case having the param_cards is ok.

wvieri:
yes exactly @bendavid .

bendavid:
From the param cards/proc cards this looks like 4 flavour scheme (is this what's desired?)

In that case pdf's should be 263400 rather than 263000

bendavid:
(firstly please make a comment in the thread when updating the pull request, since otherwise we are not notified automatically)

The param cards are now inconsistent with the model (since they still have massive b's).  Are the param cards actually needed?  (Does anything need to be modified with respect to the defaults coming with sm-ckm_no_b_mass?)

jkunkle:
The param cards are required for the anomalous QGCs.  These are on lines 11-34.  The defaults in sm-ckm_no_b_mass can stay as they are.  What specifically do we need to change in the run card?

bendavid:
Ah, I was actually confused by a few things here.  The param cards are indeed fine, aside from the issue with the b masses.

First, if you want to use 5fs, to make sure the proton and jet definitions take effect, they should be moved AFTER the "import model SM_LM0123_UFO" statement.

Then if you want to use 5fs, the param cards need to be modified so that the b mass and yukawa couplings are zero.  (since massive b's cannot be included consistently in the proton definition)

cranelli:
I have sent a pull request with the latest version of the Cards.

I did not have the b masses set to 0 in the param card, so I have changed that, and will regenerate the lhe files.

bendavid:
Sorry which is the new pull request?  (For the future, things are easier to follow if the existing pull request is updated rather than opening a new one)

cranelli:
Ah, I cloned vieri's genproductions branch and then sent him a pull request:
https://github.com/wvieri/genproductions/pull/1

For the future, how do you modify someone else's pull request?  I agree that would be more efficient.

cranelli:
Thanks Vieri, the new cards are up.

cranelli:
The 0.5 M event, lhe files are located on my eos space, at:
/store/user/cranelli/lhe_files/LNuAA_LM0123_Reweight_LO_6_3 
/store/user/cranelli/lhe_files/LNuAA_LT012_Reweight_LO_6_3

We would like to run at least the GEN-SIM step on them.

bendavid:
Sorry @wvieri I got confused here whether this pull request is still relevant or was replaced by another?

bendavid:
@wvieri is this one needed or can we close it?

wvieri:
@bendavid yes,I think it can be closed now (sorry I missed the last post)

PR #390, opened 2015-05-13
committing cards for pair production of x53 in the same-sign dilepton final state

clint-richardson:
just as the title suggests, these are cards for the pair production of X53 particles with one of them decaying to same-sign dileptons. The other is allowed to decay inclusively


bendavid:
Is the model STP_UFO_freeWidth already included in the gridpack generation scripts?

clint-richardson:
Hello,

Yes it is: https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/gridpack_generation.sh#L107

bendavid:
ok, thanks.  Looks good then.

PR #391, opened 2015-05-13
WZto3L1Nu+01jets aMC@NLO Cards

senka:
Cards are ready to be reviewed


stosi:
Everything looks OK

covarell:
Please follow recent PR:
https://github.com/cms-sw/genproductions/pull/385

and relative conversation
In the end it seems that the syntax including offshell W is preferred:
p  p > l nu l l
with no madpsin. It should be consistent here.

trandbea:
To harmonize with other requests e.g.
#365 (WZTo2L2Q) and 
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZZTo4L01j_5f_NLO_FXFX/ZZTo4L01j_5f_NLO_FXFX_run_card.dat
Can we have:
mll_sf > 4 (instead of 12)
ptl > 0 (instead of 5, also might be interesting for soft lepton searches)

Also, for ZZTo4L and here WZTo3LNu ptj > 15, but for WZTo2L2Q ptj > 20. Is it possible to harmonize these as well?

senka:
Sure, we can go with mll_sf>4 and ptl>0, and do pp > l nu ll.
As for ptj cut, I decided to go with 15gev as in ZZTo4L case, since I know that for ZZTo4L this value was already tested together with several different qCut values in Pythia and shows good results. But I agree it would be nice to have the same value for all diboson samples. Should we have 15GeV in all these?

bendavid:
@senka what was the qcut value used in the ZZto4L case?

The ptj cut should make ~no difference to the physics as long as the qcut is sufficiently high.

senka:
For ZZto4l ptj>15gev was used

bendavid:
Yes but which qcut values were tested?

covarell:
I tested 30 GeV and it is already fine at the first shot in my opinion:
https://covarell.web.cern.ch/covarell/ZZMCchecks/testNLOqcut30.gif

bendavid:
Ok,
I would keep this at 15 if it's working alright in any case.

The requests can be synchronized at the level of qcut if needed (but even this should not be very important, since results should not depend much on qcut)

senka:
Sorry Josh, I missred your question. Ok, so we stay with ptj 15gev

bendavid:
Sorry there is a git conflict now/pull request is pulling in other changes.

senka:
Ups, sorry, my mistake. I moved to a clean repo and pull request with cards updated with suggested changes:
https://github.com/cms-sw/genproductions/pull/395

PR #393, opened 2015-05-14
Add WZ powheg production card

kdlong:
From the example card, changed m_ll to 12 GeV and set include negative weights to true.

### DIFF
--- a/bin/Powheg/production/WZTo3lNu_NNPDF30_13TeV/WZ_lllnu_NNPDF30_13TeV.input
+++ b/bin/Powheg/production/WZTo3lNu_NNPDF30_13TeV/WZ_lllnu_NNPDF30_13TeV.input
@@ -1,5 +1,5 @@
 ! WZ boson production parameters
-mllmin 12d0         ! default 0.1 GeV this is minimum invar mass for Z leptons
+mllmin 4d0         ! default 0.1 GeV this is minimum invar mass for Z leptons
 
 numevts NEVENTS   ! number of events to be generated
 ih1   1           ! hadron 1 (1 for protons, -1 for antiprotons)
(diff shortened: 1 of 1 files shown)


syuvivida:
Did you check that the fraction of negative weights is really not negligible and that "withnegweights" must be set to 1?
Here is instruction about how to check the fraction of negative weights.
https://twiki.cern.ch/twiki/bin/viewauth/CMS/PowhegBOXPrecompiled#How_to_check_the_fraction_of_eve

kdlong:
I saw that the fraction of negative weights was small, 2.5817156812542315E-003. I turned negative weights on because it is turned on in the default WZ input card distributed with POWHEG, and because storing vs. not storing the negative weights seemed like the conservative approach. Maybe this is not the correct way to think of it?

I will compare the distributions with and without negative weights as soon as my jobs finish. Any suggestions on what I should look for?

syuvivida:
Hi, if the fraction of negative weights is so tiny, I would suggest to keep "wighnegweights=0"
otherwise, you need to be careful when you fill histograms with weights and do any computation of efficiency. The uncertainty needs to be taken care of properly.

If you know which distributions are most important for your analysis (for example, pt of the WZ system, mass of the WZ system, pt of W, Z, ....), please make sure that the negative weights are not populated in only one or two bins.

kdlong:
Hi, I moved the m_ll cut to 4 GeV to be consistent with what we have in MadGraph5_aMC@NLO. With this change the fraction of negative weights is now 5.9521416054882977E-003

I checked the affects of negative weights as you can see here: http://www.hep.wisc.edu/~kdlong/POWHEG_wgts_comp/

My impression is that these can safely be ignored. Do you agree?

syuvivida:
Hi, The distributions look fine to me. Indeed, you could ignore the negative weights.

Josh and Roberto, I am OK with the cards. Could you merge them? Thanks!!

Eiko

PR #396, opened 2015-05-15
b*->tW param_card removed, 4f scheme changed to 5f scheme 

smshaheen:
Cards for b*->tW production have been uploaded with 2 chiratlities of 10 mass points.


bendavid:
Param cards are not supported by the gridpack scripts (and are ignored)

Parameter values should be set through the customizecards mechanism instead.

Is the model used here already available in the gridpack_generation.sh script?

bendavid:
This is not consistently 4flavour or 5flavour scheme.

Model, pdf and run card looks like 4 flavour scheme
but proc card has the b's included in the proton definition, which should not be the case for 4 flavour scheme.  (Which do you prefer, for 5f scheme, instead b would need to be massless in the model, pdf's updated, maxjetflavour set to 5, and b's included in both proton and jet definition)

smshaheen:
In the current case, the param card should be changed for each mass point since the decay width also varies with the mass of the b*. 
The model used here is not supported here. If its suggested then model can be uploaded too.

bendavid:
Yes, the goal is that gridpacks should be reproducible based on the cards and scripts in git!

For the param cards, yes this is understood, but you should use the customizecards mechanism instead of the param cards.  (Param cards are IGNORED by the gridpack scripts!).  See for example https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/tt012j_5f_ckm_NLO_FXFX/tt012j_5f_ckm_NLO_FXFX_customizecards.dat

bendavid:
Any update on this?

smshaheen:
I am almost finished with the cards but may need to ask that should the model be uploaded in each mass point directory ? Or I can upload only one copy of the model?

bendavid:
model should not be uploaded to github, but rather copied to /afs/cern.ch/cms/generators/www
(and then added to the gridpack_generation.sh script)

smshaheen:
Do I have permission to write at  /afs/cern.ch/cms/generators/www ? There is one more thing I did not understand properly about your comment for the consistency of the process scheme. Since this should be a 5f scheme so what particular settings do you expect to be made?

bendavid:
So first of all, the param cards need to be removed and replaced with customizecard.dat files. 
For 4f vs 5f, your model, run cards, and pdfs look like 4 flavour scheme.  In the 4flavour scheme, the b quarks should not be included as part of the proton definition.  (ie you should remove define p = p b b~)

Instead if you want 5flavour scheme, you need to switch to or modify the model to have massless b's, modify the pdf and run cards appropriately, and make sure you have both "define = = p b b~" AND "define j = j b b~" in your run cards.

For the model, if you point me to it I can copy it to the afs area.

smshaheen:
Hi. The model is available at /afs/cern.ch/user/s/sshaheen/public/Bprime.tar.gz . 
I had already committed the changes that includes the replacement of param_card with the customized_card. Since I am following the twiki page "Git Repository for GenProduction fragments" to push my changes to master branch. After updating the changes in my working directory I am trying "git push origin master " and I am ending up with the following msg. 
"To git@github.com:smshaheen/genproductions.git
 ! [rejected]        master -> master (non-fast-forward)
error: failed to push some refs to 'git@github.com:smshaheen/genproductions.git'
To prevent you from losing history, non-fast-forward updates were rejected
Merge the remote changes before pushing again.  See the 'Note about
fast-forwards' section of 'git push --help' for details. " Do you have any possible suggestion for the issue? I tried few things but did not work.

bendavid:
There are other files being touched by this pull request that look unrelated to b*->tW (higgs and top powheg cards).  You'll probably need to rebase and/or start again from a clean area.

bendavid:
Also proc cards are still present in the pull request, and should be removed completely to avoid confusion.

bendavid:
The model has now been copied to /afs/cern.ch/cms/generators/Bprime.tar.gz so you can also add the needed changed to gridpack_generation.sh to download and extract the model.

smshaheen:
Dear Josh , 
                      Ok for sure I will try to make it clean by starting from again. 
For the removing the process card. I should remove the previous cards and thus there should not be any comparison the one we are having but we will have still new process card for all mass points and chiralities. Am I right?

bendavid:
Ah, sorry, I mistyped.  I meant to remove the param cards, not the proc cards.  (The information in the param cards should be fully contained in the customizecards)

smshaheen:
Right. But I am sorry that I can  not see any param_card in the last commit. I see that all the param_card are replaced by customized card.

smshaheen:
For removing the other cards that are not associated to b*, can I delete the current forked repository and fork the repository again and implement my changes only? I tried to rebase multiple times but I am always stuck with some error of not merging some cards in some other directories (not the one related to b*).

bendavid:
That would be an extreme solution, but should work (just make sure you have a local copy of your cards!)

smshaheen:
OK. I will try at first to go with the re-basing again

PR #397, opened 2015-05-15
T+b, T->tH, t->Wb

dmajumder:
B2G Madgraph samples for the vector-like top partner generation: T+b, T->tH, t->Wb


stosi:
The cards look OK

PR #398, opened 2015-05-15
Adding model and cards for Z' > T' t samples

emanueleusai:
These four commits add the Z'>VLQ model in gridpack_generation.sh and the cards for Z' > T' t in the MadGraph card repository.
These MC samples refer to the B2G PAG.
The Z' is present in two width scenarios (1%, 10%) and three mass points
The T' is present in two width scenarios (1MeV, 10%), three mass points for each Z' mass point, two chiralities (LH, RH), and three final states (tZ, tH, bW).
Not all possible combinations of masses, widths, chiralities and final states are provided, but only the ones we're planning to generate.
The setup uses MadSpin to decay t, W, and Z particles, keeping spin correlation and resonance width intact. The H is left to decay to Pythia.
Up to 1 additional jet is simulated in the process.
Please don't hesitate to ask me if you have any question about the model or the cards.


emanueleusai:
Is there any problem with these cards? Could you have a look at them, please?

Thank you very much,
Emanuele

stosi:
From a quick look, everything looks ok to me. For the T widths you have
Twidths = [[0,'Nar'],[0.1,'Wid']]
the width set at exactly 0 does not give any trouble?

emanueleusai:
Hi,

Thank you for reviewing the cards.
Later in the script I have a chieck that changes 0 width to 1MeV:
https://github.com/emanueleusai/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZpToTpTop/generateCards.py#L58

This is because I wanted the narrow width for the T' to always be 1MeV regardless of the T' mass.
I know it doesn't make that much of a difference for such narrow widths, but I wanted it to be consistent with the single T' production cards.

Cheers,
Emanuele

emanueleusai:
Hi,

Could you merge this PR if there aren't other pending questions or issues, please?

Thank you very much,
Emanuele

emanueleusai:
@bendavid, @covarell can you merge the files?

bendavid:
are b-quarks massive in this model?

emanueleusai:
I've checked this now, yes, it's massive.

PR #399, opened 2015-05-15
Fixed use_syst

dnash86:
Fixed the systematics in run cards (use_syst bool, uncommented a few lines).


bendavid:
just fyi, everything below " Parameter of the systematics study" is anyways ignored by the gridpack scripts (the syscalc card is generated automatically by the script)

PR #400, opened 2015-05-16
added MadGraph5_aMCatNLO cards for TGJets process. Changed directory …

mmachet:
…name and changed also the run card

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/TGJets_4f_ckm_NLO/TGJets_4f_ckm_NLO_customizecards.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/TGJets_4f_ckm_NLO/TGJets_4f_ckm_NLO_customizecards.dat
@@ -1,0 +1,4 @@
+#put card customizations here (change top and higgs mass for example)
+set param_card mass 6 172.5
+set param_card yukawa 6 172.5
+set param_card mass 25 125.0
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/TGJets_4f_ckm_NLO/TGJets_4f_ckm_NLO_madspin_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/TGJets_4f_ckm_NLO/TGJets_4f_ckm_NLO_madspin_card.dat
@@ -1,0 +1,30 @@
+#************************************************************
+#*                        MadSpin                           *
+#*                                                          *
+#*    P. Artoisenet, R. Frederix, R. Rietkerk, O. Mattelaer * 
+#*                                                          *
+#*    Part of the MadGraph5_aMC@NLO Framework:              *
+#*    The MadGraph5_aMC@NLO Development Team - Find us at   *
+#*    https://server06.fynu.ucl.ac.be/projects/madgraph     *
+#*                                                          *
+#************************************************************
+#Some options (uncomment to apply)
+#
+#directory for gridpack mode
+set ms_dir ./madspingrid
+
+#initialization parameters
+ set Nevents_for_max_weigth 250 # number of events for the estimate of the max. weight
+ #set BW_cut 15                # cut on how far the particle can be off-shell
+ set max_weight_ps_point 400  # number of PS to estimate the maximum for each event
+# 
+
+set max_running_process 1
+
+# specify the decay for the final state particles
+#(inclusive top decays in this case)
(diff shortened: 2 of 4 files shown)


mmachet:
I have changed the run card (line 103), putting the FXFX merging to zero.
There are some other things that I have to change?
Looking at the run card of single top (genproductions/bin/MadGraph5_aMCatNLO/cards/production/13TeV/st_sch_4f_ckm_NLO/st_sch_4f_ckm_NLO_run_card.dat) they use slightly different values in the first part of the card.

bendavid:
Hi Martina,
I wonder whether the custom scales and associated files should also be applied here.  @maierbenedikt any thoughts as to whether the same custom scale makes sense here as for the single top without the extra photon?
(I would tend to say yes, since the spectator b quark is not fundamentally different in this case)

maierbenedikt:
Sorry for the late reply, I have been away the last days!

It is a good question. In general, the custom scale for the four-flavor scheme

Q = 4*sqrt(pt_b^2 + m_b+^2) 

is chosen because, with it's typical values of O(10^2 GeV), it is a sort of compromise between the two scales involved: the b mass and the top mass. 

If one adds a photon or a Z boson, this might be reconsidered. In http://arxiv.org/pdf/1504.00611.pdf for the single top + Higgs process another choice is made (right plot, page 5) after studying the scale dependence and picking the scale where 4 and 5FS are closest together. One can also achieve it with a fixed scale. tZq is surely comparable with tHq, and I will generate the tZq gridpacks with a fixed scale of ~75 GeV. For the taq, I am not sure myself which (fixed) scale is the best one.

In any case, taking the dynamical scale of the normal single top t-channel (https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/st_tch_4f_ckm_NLO) and using it without any modifications will probably result in wrong results, because it picks really the bottom quark. One first would need to check if its position has changed when adding a photon to the list of outgoing particles.

mmachet:
Hi,
so at the end I try with a fixed scale of ~75 GeV or something else?
I have to modify the run card or to add a setscales.f card like in https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/st_tch_4f_ckm_NLO?

Thanks

Martina

bendavid:
@maierbenedikt how large are the differences if one would blindly use the default dynamic scales for t-channel single top?

maierbenedikt:
Using the default functional form brings a cross section that is ~10% lower, while the distributions can also show some effect.

Having talked to Rikkert, it is the safest to use a similar setup as for tHq, where the dynamic scale is basically given by Ht/6.

This can easily be achieved by - since the default dyn. scale is Ht/2 - setting lines 81-84 of

https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/st_tch_4f_ckm_NLO/st_tch_4f_ckm_NLO_run_card.dat

to 1/3, i.e. 0.33.

I have already checked that then the systematic scale variations are calculated with respect to the correct Ht/6.

So chancing those four parameters and also raising the ptj to 0.1 to be safe against instabilities should be the way to go here. You don't need a separate setscales.f

mmachet:
Hi Benedikt,
thank you for your suggestions. I have modified the run card.

Martina

bendavid:
Hi Martina,
Sorry I don't see the change of the scale ratios to 0.33

Also since this is 4f the pdf's need to be changed for both the central pdf and the variations as here
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/examples/wplustest_4f_NLO/wplustest_4f_NLO_run_card.dat

mmachet:
Hi Josh,
I modified the run card. For the pdf's I modified lines 58,96 and 97 with the values from this card:
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/examples/wplustest_4f_NLO/wplustest_4f_NLO_run_card.dat

maierbenedikt:
Looks good to me.

For the QES scale, the actual factor is irrelevant, since it is an artificial helper scale which cancels out in the calculation. But according to Rikkert, setting it to the same value as muF and muR should make the calculations more stable.

mmachet:
So is it ok like this or I have to change something else?

2015-07-03 18:02 GMT+02:00 maierbenedikt <notifications@github.com>:

maierbenedikt:
For me it is okay!

PR #401, opened 2015-05-17
added mass points to VH inclusive, increased folding to reduce negative weights

perrozzi:
added mass points to VH inclusive, increased folding to reduce negative weights


syuvivida:
These cards look OK to me.

perrozzi:
reminder to merge. thanks

PR #402, opened 2015-05-18
Fragments from Josh Hardenbrook

dnash86:
Cards for XXto4J long lived production.

### DIFF
--- a/python/ThirteenTeV/XXTo4J_M-1000_CTau-1000mm_TuneCUETP8M1_13TeV_pythia8_cff.py
+++ b/python/ThirteenTeV/XXTo4J_M-1000_CTau-1000mm_TuneCUETP8M1_13TeV_pythia8_cff.py
@@ -1,0 +1,87 @@
+COM_ENERGY = 13000. # GeV
+MASS_POINT = 1000   # GeV
+CROSS_SECTION = 1 # pb
+CTAU_POINT = 1000 # mm
+
+import FWCore.ParameterSet.Config as cms
+
+from Configuration.Generator.Pythia8CommonSettings_cfi import *
+from Configuration.Generator.Pythia8CUEP8M1Settings_cfi import *
+
+generator = cms.EDFilter("Pythia8GeneratorFilter",
+    pythiaPylistVerbosity = cms.untracked.int32(0),
+    filterEfficiency = cms.untracked.double(1),
+    pythiaHepMCVerbosity = cms.untracked.bool(False),
+    comEnergy = cms.double(COM_ENERGY),
+    crossSection = cms.untracked.double(CROSS_SECTION),
+    maxEventsToPrint = cms.untracked.int32(0),
+    PythiaParameters = cms.PSet(
+        pythia8CommonSettingsBlock,
+        pythia8CUEP8M1SettingsBlock,
+        processParameters = cms.vstring(
+            "Higgs:useBSM = on",
+            "HiggsBSM:all = off",
+            "HiggsBSM:ffbar2A3H2 = on",
+            "35:m0 = %s" % MASS_POINT,
+            "36:m0 = %s" % MASS_POINT,
+            "35:tau0 = %s" % CTAU_POINT,
+            "36:tau0 = %s" % CTAU_POINT,
+            "35:0:bRatio = .2",
+            "35:1:bRatio = .2",
+            "35:2:bRatio = .2",
+            "35:3:bRatio = .2",
+            "35:4:bRatio = .2",
+            "35:5:bRatio =  0",
+            "35:9:bRatio =  0",
+            "35:10:bRatio=  0",
+            "36:0:bRatio = .2",
+            "36:1:bRatio = .2",
+            "36:2:bRatio = .2",
+            "36:3:bRatio = .2",
+            "36:4:bRatio = .2",
+            "36:5:bRatio =  0",
+            "36:9:bRatio =  0",
+            "36:10:bRatio =  0",
+            "35:0:meMode = 100",
+            "35:1:meMode = 100",
+            "35:2:meMode = 100",
+            "35:3:meMode = 100",
+            "35:4:meMode = 100",
+            "35:5:meMode = 100",
+            "35:9:meMode = 100",
+            "35:10:meMode = 100",
+            "36:0:meMode = 100",
+            "36:1:meMode = 100",
+            "36:2:meMode = 100",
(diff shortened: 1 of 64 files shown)


python/ThirteenTeV/XXTo4J_M-1000_CTau-1000mm_TuneCUETP8M1_13TeV_pythia8_cff.py
```diff
@@ -0,0 +1,92 @@
...
+            "HiggsH2:coup2l = 0",
+            "HiggsH2:coup2Z = 0",
+            "HiggsH2:coup2W = 0",
+            "HiggsH2:coup2H1H1 = 0",
+            "HiggsH2:coup2A3A3 = 0",
+            "35:onMode = off",
+            "35:onIfAny = 1 2 3 4 5",
+            "36:onMode = off",
+            "36:onIfAny = 1 2 3 4 5",
+            "ParticleDecays:limitTau0 = off"
```
bendavid: This has far-reaching consequences.  It affects ALL particles in the event, not just the ones from the X decays.

python/ThirteenTeV/XXTo4J_M-1000_CTau-1000mm_TuneCUETP8M1_13TeV_pythia8_cff.py
```diff
@@ -0,0 +1,92 @@
...
+    PythiaParameters = cms.PSet(
+        pythia8CommonSettingsBlock,
+        pythia8CUEP8M1SettingsBlock,
+        processParameters = cms.vstring(
+            "Higgs:useBSM = on",
+            "HiggsBSM:all = off",
+            "HiggsBSM:ffbar2A3H2 = on",
+            "35:m0 = %s" % MASS_POINT,
+            "36:m0 = %s" % MASS_POINT,
+            "35:mMin = 5.",
```
bendavid: The mMin and mMax settings look wrong
(mMin=5 and mMax=0)

Why are these needed at all?

dnash86:
Referring the analyzers to these questions.

hardenbr:
Is it the case that limitTau0 only effects hadrons? If so (I think it is), then I agree I can remove the limitTau0 line. The mMin was modified for the mX = 50 GeV scenario. Is  "mMin = 0" and  "mMax = 0"  preferable?

bendavid:
Sorry, the point is that having these settings with mMin>=mMax will result in null phase space->zero cross section ->crashes.

It would be preferable to remove these settings completely unless they are really needed.  Can you also explain the purpose of the lines modifying meMode?

Have you confirmed interactvely that these configs (or at least one of the longer lived phase space points) produces events and that that the long-lived particles are decayed by pythia with the expected lifetime?

hardenbr:
Using mMin = mMax = 0 the generation does not crash for me, but I'll concede to removing them. The meMode=100 lines are to force the branching ratios to exactly 0.2 per channel for (u,d,s,c,b). 

I have checked the proper lifetime distribution for 30mm and 300mm with O(1k) statistics and find the expected lifetime distributions (lifetimes are listed in [cm] for the mX=300 GeV points). ctau0 = lxyz / beta * gamma where lxyz is the distance from the quark GEN vertex to its mothers GEN vertex in 3D. 

http://hardenbr.web.cern.ch/hardenbr/XX4J_30mm.pdf
http://hardenbr.web.cern.ch/hardenbr/XX4J_300mm.pdf

Is 300mm far enough or do the 1000mm+ lifetimes pose some special concern?

bendavid:
300mm is probably far enough.

Why is changing the meMode needed to override the branching ratio?  My understanding is that meMode is rather related to the kinematics of the decay.

hardenbr:
I based this choice on this reference: http://home.thep.lu.se/~torbjorn/pythia82html/ParticleDataScheme.html

meMode: the mode of processing this channel, possibly with matrix elements; see the particle decays and resonance decays descriptions for the list of possibilities. Notably the default code 0 for a particle means pure phase space decays according to the given branching ratios, while for a resonance it means that code exists for the dynamic calculations of partial widths and thereby branching ratios as a function of the resonance mass (which is done e.g. at initialization based on the mass set by the user). Then codes 1 - 99 are reserved for various matrix-element-improved ordinary particle decays, and 100 - 103 for resonances where the partial width of a given channel is calculated from the total width and the stored branching ratio. **Thus, to enforce a new branching ratio for a resonance channel (with its own partial-width calculation code) it is not sufficient only to change the bRatio but also to set e.g. meMode = 100.**

bendavid:
@smrenna can you comment on this?

smrenna:
If Pythia8 has code to calculate the BR for a particle/resonance, and you try to set the BRs otherwise, we ignore your input.   However, if you tell Pythia8 to *forget* that is has code to calculate the BR, then it will use your input for the BRs.  That is what I think is going on here, and it isn't particularly different from the treatment in Pythia6.

bendavid:
Ok but does changing MEMode in this context also affect the kinematics of the decay?

smrenna:
These are just scalar, 2-body decays right?    The decay is just by phase space, so it should not matter.

hardenbr:
@bendavid Is this resolved? If so, I can send @dnash86 a new set of fragments with the mMin, mMax, and limitTau0 changes.

bendavid:
@hardenbr I thought those updates were already made to this pull request?

The open issue is still steve's question: "These are just scalar, 2-body decays right? The decay is just by phase space, so it should not matter."

If these are indeed scalar 2-body decays then the cards are fine as is.

hardenbr:
Yes, they are scalar 2-body decays. The updated fragments have the limitTau0 parameter removed, but still have mMin = 0 and mMax = 0. I will send @dnash86 new fragments.

bendavid:
Ok yes, please do.  (And yes, see that now)  Once the mMin/mMix are updated then I can merge them.

dnash86:
Just to clarify, what settings should be in the new fragments I should expect from you, @hardenbr?  You've said they will still have mMin, mMax=0, but @bendavid mentioned an update in mMin/mMax?

hardenbr:
@dnash86 The new fragments will not have mMin =0 and mMin=0 specified. These parameters will take the default common settings.

dnash86:
@hardenbr Okay, great, sounds good.  I'll commit them when I receive them from you.

dnash86:
Updated cards from @hardenbr.

PR #404, opened 2015-05-19
adding single top + Z cards

maierbenedikt:
tzq for z->ll and z->nunu final states


stosi:
Hi, everything looks OK to me. Just out of curiosity, why do you cut at 30 GeV on the mass of the lepton pairs?

maierbenedikt:
Hi Silvano, 

it was like this in Run-I, too. (Of course this is not an answer or justification.)
But the idea is that the analysts have applied / will apply a Z mass cut anyways, so that they could even have gone for 40 GeV, but we decided to leave it at 30; the chance is then higher the sample can also still be used as background sample in another three-lepton analysis.

maierbenedikt:
Is it possible to merge this pull request so that I can prepare the gridpacks?

thanks :)

maierbenedikt:
I received a "poles do not cancel" for the gridpack production of these cards. 

It happens in the subprocess

P0_ud_epemtbg

This process should however not be there, I suppose. There should only be " t b~ " or " t~ b ", but not the cases of both being particles or both being antiparticles at the same time.

I will update the proc_card, remove the multi particle labels and work with add process instead.

After I tried it, I'll do another pull request.

PR #406, opened 2015-05-19
Fragments for EXO-RunIIWinter15GS Heavy WToTauNu without tauola

davidsheffield:
From @bdelacruz

I closed #405 because these fragments should be separate from the tauola ones rather than replacing them.


bendavid:
Please remove the commented tauola lines, since this is just confusing.

Also the following lines need to be removed because they are already covered by the common/ue fragments

+            'Main:timesAllowErrors = 10000',
+            'Tune:ee 3',
+            'Tune:pp 5',

davidsheffield:
I've removed the lines.

bdelacruz:
Thank you for handling this,

Begoña

PR #409, opened 2015-05-20

MichaelButtignol:
This is going to replace the previous request #379


bendavid:
Are you sure you want cut_decays=T here?  (This means jet cuts will apply to the hadronic decays)

Also there is a merge conflict in gridpack_generation.sh which needs to be fixed.

MichaelButtignol:
Hi Josh,

You're right (again). We will keep the cut_decays to TRUE but change the 
value to 10 or 15 GeV for pt(jets).

Concerning the conflict, it will be fixed at the next iteration.

Thanks,
Michael

bendavid:
I would really suggest to just set cut_decays to False!
Then we are consistent with other samples, and completely safe, at the expense of a modest cost in statistics for what is probably anyways a small sample.  Feel free to increase the size of the request to compensate if you like.

MichaelButtignol:
OK, thank you fir the advice, we will do this then.

And OK I'll see this.

Thanks,
Michael

PR #411, opened 2015-05-21
Add qqZZ 4L Powheg cards (6 final states)

dsperka:


covarell:
Why are these needed? A gridpack is already there with the 6 (actually 9) final states merged

dsperka:
Hi Roberto,
These cards are similar to the one you mention. I just increased the foldings to reduce the fraction of negative weights (which are dropped, and checked to have no effect). I think it is better to have separate files in case we need to increase statistics later for a particular final state.

covarell:
That is OK but please add to the commit the card with fixed folding settings and all leptonic states together. The number of sample matters for official production, so the strategy will be to generate just one sample and in the future, if needed, use these cards to increase a specific f.s.

dsperka:
Ok, done.

PR #413, opened 2015-05-21
 Pythia cfg files for ZZ->2l2nu in VBF channel

amagitte:
Hi all,

I added 6 different cfg files to shower powheg lhe files using pythia8 in the case that the higgs boson is decayed by JHUGen or by Pythia8.
There are three different mass point: 400, 600 and 1000 GeV.
The cfg are the following:
1- Hadronizer_TuneCUETP8M1_13TeV_powhegEmissionVeto_3p_H_ZZ_2l2nu_VBF_M1000GeV_LepDECAY_LHE_pythia8_cff.py
2-Hadronizer_TuneCUETP8M1_13TeV_powhegEmissionVeto_3p_H_ZZ_2l2nu_VBF_M1000GeV_NoLepDECAY_LHE_pythia8_cff.py
3-Hadronizer_TuneCUETP8M1_13TeV_powhegEmissionVeto_3p_H_ZZ_2l2nu_VBF_M600GeV_LepDECAY_LHE_pythia8_cff.py
4-Hadronizer_TuneCUETP8M1_13TeV_powhegEmissionVeto_3p_H_ZZ_2l2nu_VBF_M600GeV_NoLepDECAY_LHE_pythia8_cff.py
5-Hadronizer_TuneCUETP8M1_13TeV_powhegEmissionVeto_3p_H_ZZ_2l2nu_VBF_M400GeV_LepDECAY_LHE_pythia8_cff.py
6-Hadronizer_TuneCUETP8M1_13TeV_powhegEmissionVeto_3p_H_ZZ_2l2nu_VBF_M400GeV_NoLepDECAY_LHE_pythia8_cff.py

Cheers,
Alessio

### DIFF
--- a/bin/Powheg/production/H_ZZ_2l2nu_NoJHUGen_NNPDF30_13TeV/VBF_H/VBF_NNPDF30_13TeV_M1000GeV.input
+++ b/bin/Powheg/production/H_ZZ_2l2nu_NoJHUGen_NNPDF30_13TeV/VBF_H/VBF_NNPDF30_13TeV_M1000GeV.input
@@ -1,0 +1,79 @@
+numevts NEVENTS    ! number of events to be generated
+ih1   1           ! hadron 1 (1 for protons, -1 for antiprotons)
+ih2   1           ! hadron 2 (1 for protons, -1 for antiprotons)
+! 131 cteq4m
+!  83 cteq4l
+!ndns1 131         ! pdf set for hadron 1 (mlm numbering)
+!ndns2 131         ! pdf set for hadron 2 (mlm numbering)
+ebeam1 6500d0     ! energy of beam 1
+ebeam2 6500d0     ! energy of beam 2
+
+! To be set only if using LHA pdfs
+! 19150 cteq4m
+! 19170 cteq4l
+! 10050 cteq6m
+lhans1  260000         ! pdf set for hadron 1 (LHA numbering)
+lhans2  260000         ! pdf set for hadron 2 (LHA numbering)	
+! To be set only if using different pdf sets for the two incoming hadrons
+! QCDLambda5  0.25 ! for not equal pdf sets 
+
+! Parameters to allow or not the use of stored data
+use-old-grid    1 ! if 1 use old grid if file pwggrids.dat is present (<> 1 regenerate)
+use-old-ubound  1 ! if 1 use norm of upper bounding function stored in pwgubound.dat, if present; <> 1 regenerate
+
+ncall1 125000  ! number of calls for initializing the integration grid
+itmx1    7     ! number of iterations for initializing the integration grid
+ncall2 75000   ! number of calls for computing the integral and finding upper bound
+itmx2    5    ! number of iterations for computing the integral and finding upper bound
+foldcsi   2    ! number of folds on csi integration
+foldy     5    ! number of folds on  y  integration
+foldphi   2    ! number of folds on phi integration
+nubound 175000 ! number of bbarra calls to setup norm of upper bounding function
+icsimax  1     ! <= 100, number of csi subdivision when computing the upper bounds
+iymax    1     ! <= 100, number of y subdivision when computing the upper bounds
+xupbound 2d0   ! increase upper bound for radiation generation
+
+! OPTIONAL PARAMETERS
+
+renscfact  1d0   ! (default 1d0) ren scale factor: muren  = muref * renscfact 
+facscfact  1d0   ! (default 1d0) fac scale factor: mufact = muref * facscfact 
+#ptsupp     0d0   ! (default 0d0)  mass param for Born suppression factor (generation cut) If < 0 suppfact = 1
+#bornonly   1      ! (default 0) if 1 do Born only
+#smartsig   0      ! (default 1) remember equal amplitudes (0 do not remember)
+#withsubtr  0      ! (default 1) subtract real counterterms (0 do not subtract)
+withdamp    1      ! (default 0, do not use) use Born-zero damping factor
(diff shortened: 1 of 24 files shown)


covarell:
They look OK, but the LepDECAY will also produce ZZ -> 4l and (many) 4nu events. You need to put a simple filter on the resulting HepMCProduct asking for exactly two charged and two neutrals and add it in the ProductionFilterSequence

amagitte:
Hi Roberto,

Many thanks!

This is what I am trying to do. 
I have some simple question to understand properly:
1- Why should I use HepMCProduct instead of reco::GenParticles?
2- is there any way to obtain informations of mother and daughter using HepMCProduct? Is GenVertex class useful for this purpose?

Alessio

covarell:
1- Filters based on reco:GenParticles cannot be in the fragment because they are produced later in the workflow. You can certainly put a temporary GenParticle producer with a different label which is then not kept in the output.

2 - For bare Pythia output, you have a broad choice of pre-existing filters here:
https://github.com/cms-sw/cmssw/tree/CMSSW_7_5_X/GeneratorInterface/GenFilters/src

amagitte:
Hi,

I added Powheg and JHUGen input files for H_ZZ_2l2nu using three mass points (1000GeV, 600 GeV and 400 GeV) 

Cheers,
Alessio

syuvivida:
Hi,
 Did you check if gg_H_quark-mass-effects produces non-negligible fraction of negative weights?

amagitte:
Hi,

I used David Sperka's recipe:

foldcsi 2         ! number of folds on csi integration
foldy   5         ! number of folds on  y  integration
foldphi 2         ! number of folds on phi integration

here you can see the numbers:

 btilde pos.   weights:  0.37914766552010731       +-
1.7412586029571060E-003
 btilde |neg.| weights:   3.4313059119969801E-004  +-
2.4951620569693830E-005
 btilde total (pos.-|neg.|):  0.37880453492890626       +-
1.7409819108418385E-003

Cheers,
Alessio

2015-05-27 10:37 GMT+02:00 syuvivida <notifications@github.com>:

> Hi,
> Did you check if gg_H_quark-mass-effects produces non-negligible fraction
> of negative weights?
>
> —
> <https://github.com/cms-sw/genproductions/pull/413#issuecomment-105824651>
> .
>



-- 
Alessio

syuvivida:
ok, then the powheg.input settings seem OK to me.
@bendavid, @covarell can you merge the files?

covarell:
Whoops, I did not notice this commit also contained the Pythia8 configuration (which still need filtering for 2l2nu). @amagitte can you open a new PR when you have done this?

amagitte:
I was writing a mail asking about the filter because I have some questions:

1- the fragment of the filter has to be put inside Pythia8 fragment or by
hand in cfg file obtained using cmsDriver.py and using the previous
fragment..
2- I have done it, and I have checked that works properly; I am wondering
if CMSSW_7_4_14 is the release that I have to use to make the pull request
or I have to use the 75X..

2015-05-27 10:45 GMT+02:00 Roberto Covarelli <notifications@github.com>:

> Whoops, I did not notice this commit also contained the Pythia8
> configuration (which still need filtering for 2l2nu). @amagitte
> <https://github.com/amagitte> can you open a new PR when you have done
> this?
>
> —
> <https://github.com/cms-sw/genproductions/pull/413#issuecomment-105826669>
> .
>



-- 
Alessio

covarell:
1. Must be in the pythia8 fragment, see an example here:
https://github.com/cms-sw/genproductions/blob/master/python/ThirteenTeV/LQToCMu_MuNuJJFilter_M_1000_TuneCUETP8M1_13TeV_pythia8_cff.py
2. genproductions repo has no release branches, only the master exists. So it is irrelevant

amagitte:
Ciao Roberto,

Thanks! About the second question I was talking about the EdFilter that I have written.

covarell:
Hmmm... You have *written* a filter?
This is a bit problematic because we would need a new 7.1.X release to have it working, just for you :-)
Is it really not possible to use a (combination of) existing filters? I believe you have simple diparticle filters in the directory, can you use two of them in chain (one for 2 charged leptons, one for 2 neutral)?

amagitte:
Yes, written and tested…
I will check the second option crossing the fingers..

Alessio

amagitte:
Hi Roberto,

I have checked inside 71X release..
https://github.com/cms-sw/cmssw/blob/CMSSW_7_1_X/GeneratorInterface/GenFilters/src/LHEDYdecayFilter.cc <https://github.com/cms-sw/cmssw/blob/CMSSW_7_1_X/GeneratorInterface/GenFilters/src/LHEDYdecayFilter.cc>

This filter should work apriori because it permits to select exactly 2 leptons, but considering we need ZZ_2l2nu contemporary we should implement a logic “or" and so generate more sequences inside the same fragment. 
Might it be possible?

Typically we would need to select
        (Z—>ee AND Z—>ve ve)
OR  (Z—>ee AND Z—>vmu vmu)
OR  (Z—>ee AND Z—>vtau vtau)
OR  (Z—>mumu AND Z—>ve ve)
OR  (Z—>mumu AND Z—>vmu vmu)
OR  (Z—>mumu AND Z—>vtau vtau)

so we need to call the filter twice per line..

Alessio

covarell:
OR is not possible.

Can I suggest something like:

poslepgenfilter = cms.EDFilter("MCSmartSingleParticleFilter",
                          MinPt = cms.untracked.vdouble(5.0,5.0,5.0),
                          ParticleID = cms.untracked.vint32(11,13,15),
)

and three more in cascade for the negative leptons and neutrinos?

PR #418, opened 2015-05-27
Adding tGamma FCNC model + cards, and fixing process definition in tZq

maierbenedikt:


stosi:
Hi, just one question: will you have a 4- or 5-flavor scheme? You set maxjetflavor to 4 in the run card, but in the proc card you add the b quarks to the j and p definition and you use sm rather than sm-ckm_no_b_mass. Was this intentional?

maierbenedikt:
This was not intentional I guess, I did not see it when I got the cards from the analyzers. However, since j and b cuts are the same throughout the run_card, I guess it does not matter here.

Thanks,
Benedikt

maierbenedikt:
PS: It is not the model sm that is used, but the custom model tqAandG_UFO.

This does have the b mass set to 4.7, but since there is neither a diagram in the process with a b in the initial state nor a diagram with a jet in the final state, including the b/b~ in the proton/jet would not have necessary, I suppose. 

What I however will back-check with the analyzers is whether they need the taus as leptons. Currently the W just decays into el/mu. Sorry I saw this only now.

But apart from that, if you are okay with the cards, maybe you can merge the pull request already and I will do another one adding l+ = e+ mu+ ta+ in the proc_card if it turns out they forgot it.

Thanks.

maierbenedikt:
Sorry, my fault. The tau IS included. Then everything okay from our side! Can you merge it?

Thanks!
Benedikt

stosi:
Can you please merge this request? Everything seems ok to me

bendavid:
1) There is a merge conflict now which needs to be fixed.

2) Are you sure you want cut_decays=T?  (then the jet cuts/etc will apply to the decay products, including drjj, etc)

maierbenedikt:
Hi Josh,

the analyzers just confirmed that they want cut_decays=T.

bendavid:
Ok, but almost all other samples with top have inclusive phase space for the top decays.  (and the cuts here are actually not that loose and may not even be guaranteed to be looser than analysis level cuts)

maierbenedikt:
Here is what the analyzers wrote:


"I prepared the card similar to the last round of analysis (8 TeV, in some cases even looser cuts are used).
For example, if we relax the cut on photon kinematics, in particular pT, 
the cross section will become divergent.

The samples are correct with the cuts."

bendavid:
Right, I'm not suggesting to loosen the cuts on the photon, only to set cut_decays = F so that the cuts do not apply to the top decay products.

(The top decay is never divergent, so no cuts are needed there.)

maierbenedikt:
Okay I will ask them once more and get back to you.

maierbenedikt:
Here is their reply:

"OK, at the end we have to apply several cuts on the top decay productions because
we do not have any electron or muon out of |eta| < 2.5 region and jets and leptons with pT > 20 GeV
will not considered. Why should we relax these cuts? If we relax these cuts we have to increase the number of events to have enough statistics in all regions of phase space."

They suggest to proceed with those cards/settings.

bendavid:
The correspondence between the matrix element level cuts and the reconstructed quantities is not 100%, especially for jets.  A cut of ptj 20 would require cutting on reconstructed level jets of at least 30 GeV to be safe.  There are also drjj cuts active here that are less easy to understand.  All other top samples produced in this campaign are without cuts on the top decay products.

maierbenedikt:
Hi Josh

the analyzers have updated the cards / gridpacks and removed (most of) the cuts, the cut_decays is set to false.

I made a new pull request, so this one is obsolete.

bendavid:
ok, to be clear it wasn't necessary to remove the cuts, just to set cut_decays=F to make sure they don't apply to the decay products.  Probably it's fine anyway, will look at the new pull request.

PR #419, opened 2015-05-27
[Madgraph5_aMC@NLO] ExoDiboson group:  W'/Z'->W/Zh->had/lep tautau input cards

syuvivida:
Upon the request of analyzers, we add the input cards for the following decay modes.
13 mass points are requested for each decay mode. Each mass point has 50K events.

W'->Wh->lnu tau tau
W'->Wh->qq tau tau
Z'->Zh->qq tau tau
Z'->Zh->ll tau tau


syuvivida:
@bendavid , @covarell 
  Could you merge this pull request? Many thanks!!
Eiko

bendavid:
Are you sure you want cut_decays=True here?  This means the jet pt cuts will apply to the hadronic decays.

syuvivida:
I will do the update for all the other modes too. Will close this pull request and open a new one.

PR #420, opened 2015-05-27
Pythia+Filter cfg files for H To ZZ_2l2nu in VBF channel

amagitte:
Hi,

I have modified the previous cfg in the case Pythia is used to decay the Higgs.
To filter the events according to our final state I have used MCParticlePairFilter code; the cfg was tested and works properly.

Alessio


amagitte:
Hi,

I added cfg files for H to ZZ_2l2nu, via ggH channel.
As in VBF, there are two typologies:
1- files used to hadronize the event because the decay is done by JHUGen
2- file used to decay the Higgs and for them a filter is applied.

Cheers,
Alessio

covarell:
I tested one of these files on H(125) (the only one for which I had a gridpack ready) and a find a filter efficiency of 2.600e-02 +- 7.117e-03. 

Does it look good? It is true that you also have pT cuts etc. but in principle this should be close to 2BR(z->ll)BR(z->vv)/BR^2(z->ll) + 2BR(z->ll)BR(z->vv) + BR^2(z->vv) = 37%

amagitte:
Hi Roberto,

Many thanks!
Umh..strange because yesterday I have generated 25000 events and the number of events with 2l2nu final state is 9289 that is ~37%

Alessio

covarell:
Ok, it may be that the 125 GeV mass has an effect on the lepton pT. I will approve now.

amagitte:
Hi Roberto,

To be sure, I have generated another lhe file with 250000 events: I have obtained the same percentage obtained before..

PR #427, opened 2015-06-01
tZq cards + inclusive tGamma FCNC

maierbenedikt:
new pull request with cuts for tGamma FCNC gridpacks removed


bendavid:
The cuts are irrelevant once cut_decays is set to false, but since they have no effect, ok like this.

One other point is that bwcutoff is set to 15000.  Do you particularly care about the offshell tops and W's here?  (probably it is ok for this process, so if that's what's desired then it's fine)

maierbenedikt:
The analyzers just confirmed 15000 is intended.

PR #430, opened 2015-06-02
changed folding parameters to decrease negative weight fraction for a…

chayanit:
…dditional mass points


syuvivida:
@covarell @bendavid the cards look fine to me. Could you merge them? Thanks!

PR #432, opened 2015-06-04
[Powheg] update of create_powheg_tarball.sh, fix of work directory name

syuvivida:
Previous version of create_powheg_tarball.sh has a silly feature that the work directory always has the same name (powhegbox_xxxx_xxx).
This is OK if users are running parallel jobs in different directory. But if users are running jobs 
in the same directory, the feature creates confusing output tar balls.

Fix this by giving the work directory a name that corresponds to the name of the output tar ball.


syuvivida:
@covarell , @bendavid could you merge this pull request that fixes the name of work directory? Thanks!!

PR #435, opened 2015-06-04
final settings for single top + higgs t-channel

maierbenedikt:
Raising the jet pt threshold to 1 GeV, removing some of the weights in the reweight_card.dat, because they were redundant (because of the matrix element being only sensitive to the relative (!) sign of the two parameters, i.e. M~(c_V - c_F)^2)


stosi:
The changes look OK to me.

PR #437, opened 2015-06-04
LO cards for ttbar dilepton

nstrobbe:
Adaptation of cards of the inclusive ttbar at LO.


stosi:
Everything seems OK to me

PR #438, opened 2015-06-04
Cards for HT binned ttbar request (LO)

nstrobbe:
Binning as agreed upon in email thread. Cards adapted from inclusive ttbar cards. I only changed the HT requirements and updated the proc card to contain the correct base card names.


stosi:
There is a redundant definition of the decays of the w bosons in the madspin card
decay w+ > all all
decay w- > all all
but it should not be a problem

PR #439, opened 2015-06-04
configs from Anirban

dnash86:
Configs for Zprime to tau tau production with pythia8

### DIFF
--- a/python/ThirteenTeV/ZprimeToTauTau_M_1000_TuneCUETP8M1_13TeV_pythia8_cfi.py
+++ b/python/ThirteenTeV/ZprimeToTauTau_M_1000_TuneCUETP8M1_13TeV_pythia8_cfi.py
@@ -1,0 +1,33 @@
+import FWCore.ParameterSet.Config as cms
+
+#from GeneratorInterface.ExternalDecays.TauolaSettings_cff import *
+from Configuration.Generator.Pythia8CommonSettings_cfi import *
+from Configuration.Generator.Pythia8CUEP8M1Settings_cfi import * 
+
+generator = cms.EDFilter("Pythia8GeneratorFilter",
+                         comEnergy = cms.double(13000.0),
+                         crossSection = cms.untracked.double(0.05115),
+                         filterEfficiency = cms.untracked.double(1),
+                         maxEventsToPrint = cms.untracked.int32(1),
+                         pythiaHepMCVerbosity = cms.untracked.bool(False),
+                         pythiaPylistVerbosity = cms.untracked.int32(1),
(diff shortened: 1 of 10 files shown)


bendavid:
Pythia 8.205 currently used in production does not handle tau decays well specifically for the Wprime/Zprime case.

This is fixed in 8.209.  If these samples are not urgent, suggest to wait.  Otherwise can produce them with tauola for the decays (which by default assumes SM-like helicity structure for the Zprime-fermion couplings)

dnash86:
Quick question: how long would the wait be?  (For pythia 8.209)

bendavid:
Could be several weeks given other pending backports and the need for careful validation.

dnash86:
Okay, I am checking with the analyzers.

dnash86:
So the analyzers actually want samples both with, and without taola.  So this set should wait for the fix, they have sent configs with taola, uploaded here: https://github.com/cms-sw/genproductions/pull/442

dnash86:
@bendavid It seems I've forgotten to follow up on this:  The analyzers would like these samples without the taola, with pythia 8.209 as you said, as well.  Do current CMSSW releases have pythia 8.209 yet?  (If so, we should merge this request and I should proceed with those samples.)

bendavid:
No, not yet backported to the production release, validation is still in progress.

dnash86:
Oh, okay, thanks.  Is there a sense of the timeline of that validation?  Or tough to tell?

bendavid:
There were some technical problems with the relvals which have slowed things down.  How urgent are these samples?

dnash86:
Checking on that now.

dnash86:
These samples are actually quite urgent.  Now that some of the other samples are finishing, the analyzers want these for the tau decays in the new Pythia8.

dnash86:
Just checking on the status of the validation for the pythia 8.209 backport.  If its ready, can we merge this PR?  (If not, I suppose there is nothing to do but wait.)  Thanks.

bendavid:
No, there are some issues (related specifically to tau decays) in 8.209.  We're waiting for a fix before moving forward with the pythia releases.

dnash86:
Okay, thanks.

dnash86:
@bendavid Any status update here?

bendavid:
Pythia release with the fix expected next week.

dnash86:
Okay, thanks

bendavid:
Fixed pythia release is out btw, in the process of being integrated in cmssw/tested/backported.

dnash86:
Great, thanks a lot for the update.

bendavid:
The validation and backport of the pythia release is now complete, so these can now go ahead.  (Be sure to use CMSSW_7_1_20 or later)

PR #440, opened 2015-06-05
[Powheg] additional update of the way we name work directory

syuvivida:
Small change of the way we name work directory.
Previous version take the name of the gridpack and the name of the powheg source tar ball to name the work directory.
To add one extra protection (to avoid users use the same tar ball name by accident), I also take into account the name of the powheg input card. Now the work directory name is 
${source_tarball_name}_${cardinput}_${tarball}.

@covarell @bendavid can you merge the request? Thanks!!


syuvivida:
@covarell @bendavid just a reminder, could you merge this pull request? Then I will make another announcement at hypernews after the merge. Thanks!!

PR #441, opened 2015-06-05
update hfact parameter for Powheg ggH to 60 GeV

dsperka:
Today at the HIG MC contacts meeting, we decided to choose 60 GeV for the hfact parameter. This is the recommendation of the Powheg authors to give the best fit of the whole distribution from 0-200 GeV before hadronization and MPI.


stosi:
The change is OK

PR #442, opened 2015-06-05
configs from Anirban, with taola

dnash86:
Configs for Zprime to tau tau, with taola included.  These are "sister" configs to those in this pull: https://github.com/cms-sw/genproductions/pull/439


stosi:
All seems OK to me

PR #445, opened 2015-06-05
Updated ttw cards

nstrobbe:
Split the originally inclusive sample into two parts: hadronic and leptonic. 
Also switch to diagonal ckm because madspin required too much memory. 
Organized all three cards directories into one new directory.


stosi:
Everything seems OK

nstrobbe:
Could this be merged?

PR #447, opened 2015-06-08
Powheg dy mll binned

dnash86:
Input cards for Drell Yan Mass binned processes in Powheg.


covarell:
Are all issues with low mass/high mass settings solved now? Do you actually see DY events in the mass bin as they should be?

dnash86:
Yes, the low mass/high mass settings work now.  In fact, the issue was that jobs shared a working directly, as Eiko explained on the thread.  When I ensured that jobs would not overlap in that way things were fixed.  So in fact the settings and input files were never wrong, rather the method I used to create gridpacks (which has been changed).

PR #448, opened 2015-06-08
produce w as lnu without mass cuts

perrozzi:
please check that I didn't introduce errors :-)


bendavid:
Madspin is now only needed to fill in spinup flags for the tau polarization (and only for W(->lv)H)
The appropriate madspin cards is the one containing a "dummy" decay as in
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/examples/wellnu012j_5f_NLO_FXFX/wellnu012j_5f_NLO_FXFX_madspin_card.dat

perrozzi:
thanks, please check again

bendavid:
for ZH madspin is not needed at all (so the madspin card should simply be removed)

perrozzi:
done

PR #451, opened 2015-06-09
CompHEP files for W'->t+b production, right handed

peiffer:


bendavid:
Default LO pdf for the run 2 production is nnpdf30lo alphas=0.130 5f

Do the requesters care if these samples are inconsistent?  Can CompHEP easily be interfaced to lhapdf?

peiffer:
Hi Josh,

I quote here the answer that I got from the creators of the CompHEP samples:

It is no problem to use cteq6 pdf set and I think we should keep the existing samples.
 In addition, it is not very important  - which PDF set is used at the generator stage because PDF4LHC recipe or the similar technique with reweighting will be used at the analysis stage.

bendavid:
Ok thanks.  Fine for these then.  I would still like an answer to the question whether comphep can be easily interfaced to lhapdf (so we can do this better in the future).

bendavid:
Since the changeset is too large to view on github, will have to wait for it to be merged until I have a chance to check on the command line.

peiffer:
As far as I could find out, lhapdf can be interfaced to comphep. At least, it is written in this update note http://xxx.tau.ac.il/pdf/0901.4757.pdf
But it seems that the requesters do not want to generate their LHE files again.

bendavid:
Right.  Please remind the requesters that in the future, even for pLHE production, cards should be pulled to genproduction BEFORE generating large numbers of lhe files.

If it's straightforward to interface comphep to lhapdf, then this should be done for any future production (and there's a possibility we may not accept future comphep requests with cteq6)

peiffer:
I have got new files for the W' samples from the requesters. So, I close this pull request and create a new one.

PR #454, opened 2015-06-10
Decay and hadronizer for ZH and VBF Invisible Higgs

chayanit:


covarell:
We do not need n files, but just one. m0 of the Higgs is fixed from the powheg LHE files

chayanit:
Hi Roberto,

I did ask Josh and he said m0 is needed as well. You can see his message to me below

By Josh "Yes, you should explicitly set the higgs mass in the pythia configuration (it is set at the powheg level, but it's also needed by pythia to initialize branching fractions, decay matrix elements, etc)"

Cheers,
Chayanit

covarell:
Hmmm OK, strange because the BRs are fixed by you, however ok I will approve as they are

PR #455, opened 2015-06-10
cards for WZ+JJ LO QCD+EWK sample; background sample for WpWpJJ analysis

albertobelloni:
This are the cards used to make the private LHE files used in
https://cms-pdmv.cern.ch/mcm/requests?prepid=SMP-RunIIWinter15pLHE-00001

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/WW_NLO/ww_nlo_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/WW_NLO/ww_nlo_proc_card.dat
@@ -1,0 +1,15 @@
+import model sm-lepton_masses_no_b_mass
+set group_subprocesses Auto
+set ignore_six_quark_processes False
+set loop_optimized_output True
+set complex_mass_scheme False
+define p = g u c d s b u~ c~ d~ s~ b~
+define j = g u c d s b u~ c~ d~ s~ b~
+define l+ = e+ mu+
+define l- = e- mu-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+define l+ = e+ mu+ ta+
+define l- = e- mu- ta-
+generate p p > w- w+ QED=2 QCD=99, w+ > l+ vl, w- > l- vl~
+output ww_nlo
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/LNuLNuJ_5f_NLO_FXFX/LNuLNuJ_5f_NLO_FXFX_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/LNuLNuJ_5f_NLO_FXFX/LNuLNuJ_5f_NLO_FXFX_proc_card.dat
@@ -1,0 +1,17 @@
+#import model loop_sm-ckm_no_b_mass
+#switch to diagonal ckm matrix if needed for speed
+import model loop_sm-no_b_mass
+
+define p = g u c d s b u~ c~ d~ s~ b~
+define j = g u c d s b u~ c~ d~ s~ b~
+
+define ell+ = e+ mu+ ta+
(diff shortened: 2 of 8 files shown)


stosi:
The cards are very similar to those for WpWpJJ and all seems OK. I wonder if, in the proc card, it could make sense to explicitely remove the Higgs diagrams

AndrewLevin:
Hi,

From this hypernews thread

https://hypernews.cern.ch/HyperNews/CMS/get/generators/1774/1/1/2/1/1/1.html

and this launchpad thread

https://answers.launchpad.net/mg5amcnlo/+question/200925

it seems pretty clear to me that we should have turned on jet matching for this sample, even though it has a single jet multiplicity and we are not planning to merge it with other jet multiplicities. I guess matching between QCD and EWK jets should work now? Please let me know if you agree and I will remake the lhe file with ickkw = 1.

Thanks.

Andrew

bendavid:
Hi,
No, matching of the EWK partons was unfortunately broken by one of the last fixes to the qcd matching.  This will hopefully be fixed in the near future.

In any case, as long as you don't intend to combine this with the inclusive sample, it is perfectly fine/correct to generate it with ickkw=0 and no matching.

If you want to combine it with the inclusive sample, then indeed you need jet matching, and then you'll have to wait for the corresponding fix.

senka:
Hi all,
my understanding from threads that Andrew sent (MG launchpad, CMS hypernews) was that correct way to generate exclusive X+2j sample, with exactly 2j from ME, is to use the ickkw=1 and jet matching. Even when the sample will not be used to merge with samples with different jet multiplicity. 
Can you please comment on this?

Senka

bendavid:
Hi,
But you are not trying to generate an exclusive  X+2j sample, but rather an inclusive X+>=2j sample (with additional jets produced by the parton shower).

In any case
a) arguments about sudakov's, etc, don't apply to the "ewk" partons
b) we don't have the technical means to consistently enable matching for the ewk production at this moment.

senka:
Hi,
I was assuming the problem with EWK partons for matching will be fixed soon. When can we expect this?

Senka

bendavid:
Unclear.  It was not a high priority so far.

AndrewLevin:
Hi,

OK, then please proceed with the current versions of the cards and lhe file. We want to have the Higgs diagrams.

Thanks.

Andrew

albertobelloni:
I am not sure why the old WZJJ pull request is still here, but I wanted to add the WW NLO and WW LO cards.

bendavid:
So the problem is that you are using the same branch.  If you use the same branch name, any new commits will be added to the existing open pull request.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/WW_NLO/ww_nlo_proc_card.dat
```diff
@@ -0,0 +1,15 @@
...
+set complex_mass_scheme False
+define p = g u c d s b u~ c~ d~ s~ b~
+define j = g u c d s b u~ c~ d~ s~ b~
+define l+ = e+ mu+
+define l- = e- mu-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+define l+ = e+ mu+ ta+
+define l- = e- mu- ta-
+generate p p > w- w+ QED=2 QCD=99, w+ > l+ vl, w- > l- vl~
```
bendavid: this is not an NLO process in any case

bendavid:
Suggest to remove the additional WW cards from this pull request, since the cards for the pLHE production were otherwise ready to merge.

bendavid:
Sorry @albertobelloni is this pull request still needed? (It looks like it is mixing two different things still in any case)

albertobelloni:
Hi Josh,

I think I ultimately gave up and tried to close the pull request.
Thanks! Best,

Alberto

PR #461, opened 2015-06-13
Add Vgamma and WVgamma datacards

jfaulkne:
The Wgamma + 012 jets, Zgamma + 012 jets, and WVgamma processes should include resolved jets as well as boosted/merged jets (dR -> 0.0 between jets).  The W boson is leptonic, while the V boson (W+, W-, or Z) decays to either 2 resolved jets or one merged jet.  Need to eliminate as much as possible any overlap between WVgamma and Wgamma + 012 jets production processes, given that the generated process includes p p > lep nu j j a in both.  Ideally, it would be great for the WVgamma proc_card/madpsin_card to explicitly produce p p > W V a, W > lep nu, V > j j where W is w+,w- and V is w+,w-,z, but I don't suppose MadSpin would be able to distinguish between the two possible W's in a single event. The photon needs to be isolated from the leptons/jets by dR > 0.5, and the jets need to be isolated from the leptons by dR > 0.3.


bendavid:
I think the way the processes are defined now is fine.
There is no overlap (aside from very small possible interference which is neglected) between the Wgamma+2 jets vs WVgamma, because 
p p > lep nu a j j [QCD] does NOT include diagrams where the jets come from a hadronic W/Z decay.  This is because by default madgraph generates only the diagrams which are of the lowest possible order in electroweak couplings.

One thing you could consider is a smaller value for the R0gamma parameter in the photon isolation, in order to be safe/sure that the generator cut is looser than any reco-level isolation cuts.  0.1 or even 0.05 should be ok.

jfaulkne:
Thank you for clarifying those issues.  I have changed the photon R0gamma values to 0.05.

PR #462, opened 2015-06-13
Polarized Sample Generation For Opposite sign WW.

ram1123:
One W decay in leptonic mode and another decays hadronically.


stosi:
Hi, 
you should not use the DECAY package to decay the particles, rather you should use MadSpin. 
You should not use a param_card: this will be ignored. If you need to change some parameters (mass of the top, or the Higgs for example), please use a customize_card.
Also, I do not see any proc_card: where do you state the process to be generated?
What do you need the file splitter and file merger for ?

ram1123:
Hello,

Our signal is longitudinally polarized WW. So, spin information becomes crucial for us. But, MadSpin does not preserve the spin information for the intermediate particles while the DECAY package is preserving this info. We discussed this problem with MadGraph authors and they suggested to use the DECAY package for this. The discussion link is  https://answers.launchpad.net/mg5amcnlo/+question/257782. 

We did not changed anything in the MadGraph cards.

We added a README file at link https://github.com/ram1123/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/WpWm_POL where we described our each steps in detail and reason. Please have a look at it and let us know if you need more details.

bendavid:
Ok, so this will be for plhe production clearly.  I think this is mostly fine (although we should switch back to madspin for such samples once the "bridge" method is available in future versions)

A few issues:
1) currently this is with 4-flavour scheme production.  Probably the difference is small, but 5 flavour scheme should also be fine/slightly more inclusive in case you're sensitive to additional b jets

2) You're currently using diagonal ckm matrix.  Again difference should be small, but if you're sensitive to the flavour composition of the jets, you might care.

3) Pdf's are not consistent with the rest of the production.  You should use 'lhapdf' with lhaid 263400 for 4flavour scheme, or 263000 for 5-flavour scheme.  (These are nnpdf30lo1 rather than nnpdf23)

bendavid:
Also, consider whether you want to turn on the systematic reweighting and run syscalc to populate per event weights for scale and pdf variations as for the central production.

ram1123:
Dear Josh,

We implemented your 1st and 3rd comment about b-quarks and pdf.

For CKM matrix as you said since its effect is very small so we are not using it. And about systematics we will do it at later stage.

Let us know if we need to do anything else.

arunhep:
Hello Josh,

When we changed the pdflabel and also the lhaid to lhapdf and 263000 then
it is crashing.
Could you please suggest where we may be doing wrong?

The settings we are using right now are:

'lhapdf'    = pdlabel     ! PDF set
263000            = lhaid     ! if pdlabel=lhapdf, this is the lhapdf number

Thanks and regards,
Arun.

bendavid:
Hi Arun,
Can you give more details/log about the error message you get?  Are you sure you have the path to lhapdf set up correctly? (easiest way to do this is to set up your working area using the "interactive" mode in the gridpack_generation.sh script, but it's also possible to do it by hand of course)

bendavid:
Also, for the systematics, why not just include the weights now, given that this is almost free?  For the pdf's the reweighting is at least possible (but more painful) to do later, but for the scale variations there is no way to do it later without generating new samples, at least up to generator level.

arunhep:
Hi Josh,
Problem regarding the lhapdf is solved as we had not interfaced lhapdf properly. But now its working. We also have added few generator level cuts after a discussion with Pietro Govoni to refine our signal events. We had a discussion with SMP MC Contacts as well and they suggested to remove some files from the request like .py and .cpp files. Ram has submitted the updated run card to the repository and also removed those files.
Regarding the systematics, since this is our iteration and we are trying to establish a cycle of generation of these samples, so to keep the things simple we are thinking that we will deal with the systematics at later stage of the analysis. As you suggested we can later generate small samples to evaluate them.
If you have any further queries, then please let us know or kindly approve the cards as soon as possible. 
Thanks a lot for your help.

PR #463, opened 2015-06-13
add bFDM cards

Xtian9:
bottom-flavored dark matter model for B2G


Xtian9:
I have the required model in zipped format for which I will need permission to add it to
/afs/cern.ch/cms/generators/www/

And then I can include it in the gridpack_generation.sh script.

bendavid:
Sorry can you point to the model and then we can copy it to the afs area?

bendavid:
Aside from that.  Firstly, please do not use the customizecards.dat to modify things which can just be set directly in the run card.

Second, I am not clear on exactly what you are trying to generate here
+generate p p > n1 n1~ b b~
+add process p p > n1 n1~ b
+add process p p > n1 n1~ b~
This in general is not a valid combination of processes, since there can be double counting between matrix element b's and b's from the parton shower.
Can you explain a bit better the production topology? (and maybe attach a feynmann diagram)

Is it possible to produce n1 n1~ with 0 b-quarks, or is it only possible in association with b-quarks?

bendavid:
Any news on this/is it still needed?

bendavid:
Closing this pull request since it is apparently not needed anymore.  Reopen if not the case.

PR #465, opened 2015-06-15
Added Zvv 012j NLO

violatingcp:
NLO Zvv+jets. We need it!


bendavid:
Pdf set should be 292200 to include also the alphas variations for pdf uncertainties (and then the pdf indexes for the systematics need to be changed as well)

Can just copy the numbers from 
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/examples/dyellell012j_5f_NLO_FXFX/dyellell012j_5f_NLO_FXFX_run_card.dat

PR #467, opened 2015-06-15
Hadronization fragments

mmachet:


bendavid:
Cards look fine, but this pull request is also pulling in the aMC@NLO diphoton+jets cards from the other pull request.

PR #473, opened 2015-06-18
powheg cards for WW and ttb processes

dnash86:


covarell:
@syuvivida was it verified for these two processes that the fraction of negative weights is negligible, thus "withnegweights 0" is safe?

syuvivida:
No, these two processes have not been checked yet. We checked inclusive W/Z/VBF_H, VBF_Z_Z, VBF_HJJJ processes so far. In the end, we recommend users to check this for their specific production.

dnash86:
@syuvivida do you have a suggestion for a quick method for checking this?

covarell:
Produce some o(10^4) events with withnegweights 1, check how many have negative weights (can be checked directly from the XWGTUP entry of the LHE file)

dnash86:
Okay, thanks.  I will do that and report back to this thread.

dnash86:
I sampled one for each process:
For ww->ee:  
negative weight fraction:   2.1048759642054281E-003

For ttbar->mumu:
negative weight fraction:   2.5414618957040609E-003

So both seem negligible.

PR #474, opened 2015-06-18
Filters for WW powheg mass bins

dnash86:
I actually have a few questions regarding these cards with filters, since I have measured oddly low efficiencies for the filters with these and the tarballs produced from the WW cards in https://github.com/cms-sw/genproductions/pull/473.

### DIFF
--- a/python/ThirteenTeV/Hadronizer_TuneCUETP8M1_13TeV_powhegEmissionVeto_WW_Mee_1200_2500.py
+++ b/python/ThirteenTeV/Hadronizer_TuneCUETP8M1_13TeV_powhegEmissionVeto_WW_Mee_1200_2500.py
@@ -1,0 +1,40 @@
+import FWCore.ParameterSet.Config as cms
+from Configuration.Generator.Pythia8CommonSettings_cfi import *
+from Configuration.Generator.Pythia8CUEP8M1Settings_cfi import *
+from Configuration.Generator.Pythia8PowhegEmissionVetoSettings_cfi import *
+
+generator = cms.EDFilter("Pythia8HadronizerFilter",
+                         maxEventsToPrint = cms.untracked.int32(1),
+                         pythiaPylistVerbosity = cms.untracked.int32(1),
+                         filterEfficiency = cms.untracked.double(1.0),
+                         pythiaHepMCVerbosity = cms.untracked.bool(False),
+                         comEnergy = cms.double(13000.),
+                         PythiaParameters = cms.PSet(
+        pythia8CommonSettingsBlock,
+        pythia8CUEP8M1SettingsBlock,
+        pythia8PowhegEmissionVetoSettingsBlock,
+        processParameters = cms.vstring(
+            'POWHEG:nFinal = 2',   ## Number of final state particles
+                                   ## (BEFORE THE DECAYS) in the LHE
+                                   ## other than emitted extra parton
+          ),
+        parameterSets = cms.vstring('pythia8CommonSettings',
+                                    'pythia8CUEP8M1Settings',
+                                    'pythia8PowhegEmissionVetoSettings',
+                                    'processParameters'
+                                    )
+        )
+                         )
+
+
+llgenfilter = cms.EDFilter("MCParticlePairFilter",
+                           Status = cms.untracked.vint32(23, 23),
+                           MinPt = cms.untracked.vdouble(0., 0.),
+                           #ParticleCharge = cms.untracked.int32(0),
+                           ParticleID1 = cms.untracked.vint32(11),
+                           ParticleID2 = cms.untracked.vint32(11),
+                           MinInvMass =  cms.untracked.double(1200.),
+                           MaxInvMass =  cms.untracked.double(2500.)
+                           )
+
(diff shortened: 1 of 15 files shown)


python/ThirteenTeV/Hadronizer_TuneCUETP8M1_13TeV_powhegEmissionVeto_WW_Mee_1200_2500.py
```diff
@@ -0,0 +1,40 @@
...
+                                    'processParameters'
+                                    )
+        )
+                         )
+
+
+llgenfilter = cms.EDFilter("MCParticlePairFilter",
+                           Status = cms.untracked.vint32(23, 23),
+                           MinPt = cms.untracked.vdouble(0., 0.),
+                           #ParticleCharge = cms.untracked.int32(0),
```
dnash86: I have tried these both with and without this line.  Both times I got extremely low efficiencies.  Is this filter configured correctly or have I done something wrong?

dnash86:
Filter efficiency for Mee_200_600: (119) / (5000).

This strikes me as quite low.  Is there perhaps a problem here?

covarell:
I would also leave out the Status line from the filter

bendavid:
@covarell, on this I'm not sure.  There can be many copies of the W's, so if the status line is left out, then there will effectively be a giant OR on the kinematics at all stages of the parton shower (and the samples will not be mutually exclusive)

It should be verified that there are exactly two status 22 W's in every event, then it should be correct to explicitly select on status 22.  (but indeed, status 23 is not correct, since this is only for outgoing matrix element particles, not decayed ones!)

covarell:
But the filter is on electrons, not W...
So for me it is fine to have no Status line, or even better Status = (1, 1)

bendavid:
Sorry, what exactly is the filter supposed to select?

dnash86:
Sorry, I thought I had mentioned what the filter should be on.  It is actually on M_ll (since this is a major background for Z').  Is status (1,1) okay for pythia8?  I thought that was a pythia6 status?

bendavid:
So in principle status 1 can also include non-prompt leptons.  In practice it's probably mostly ok for harder cuts.
The proper solution cannot be done solely with status codes, and requires the gen status flags, which are not available in this filter.  So that would require additional code changes.
If you want to be 100% sure there is no overlap between the samples, then this would be the way to go.  In practice the difference is probably small.

For tau's, if you want to do it only based on status codes, you would need to select on status 2 tau's (with the same caveats as above, that there can be non-prompt taus from hadron decays, etc in principle)

dnash86:
Overlap between samples is okay, as the main reason for these bins is to ensure high statistics are available at higher M_ll values.  The issue was that I was surprised by the low efficiency of this filter.

I can try with status(1,1) for ee and mumu and status (2,2) for tautau states to see what efficiency that yields.  

Would the possible negative weights issue brought up in https://github.com/cms-sw/genproductions/pull/473 have any effect here?  The efficiency taking weights into account was the same as without taking weights into account.

bendavid:
Negative weights should be negligible for this process.

And yes, the problem is that status 23 leptons may not always appear depending on the event history,
so status 1/2 would be the way to go.
(Those two status codes are universal)

dnash86:
I tried status(1,1) for ee and mumu states and still get some fairly low efficiencies:

(394) / (50000) for the 600 < M_ll < 1200 bin, for example.

I see this message in the output, as well:

"WARNING: MCParticlePairFilter : size of some vectors not matching with 2!!"

Is that indicative of an issue with the pair filter for this process and cut choice?  Or should very low efficiencies be expected?

bendavid:
I would suggest to compare this to what you get selecting by hand on the
leptons in the LHEEventProduct to be sure whether it is reasonable or not.

covarell:
In fact, I think m_ll in WW peaks at quite low values, so I am not totally surprised of < 1% for M_ll > 600. Why are you surprised?

dnash86:
I was surprised because the requested bins go quite high (~6000 GeV).  Thus, event counts at GEN level would have to be *quite* high to achieve reasonable statistics (~100k per bin) in the higher bins.  

But it seems that you might be right, and everything may be fine with the filter (although I guess then the GEN level requests will require very high event counts).  I saved the root file from one point (I wasn't before) and checked the LHEEventProduct.isPresent bool:

There were 238 entries in "isPresent", and the filter yielded this efficiency - (238) / (50000).  So it does seem to match.

bendavid:
What is the filter efficiency for the highest mass bin?

dnash86:
At the event counts I was running, it was zero.  Hence my concern something was wrong.

dnash86:
0/50000

bendavid:
OK, well without the ability to apply cuts at the powheg level I think the request is basically impossible then.

dnash86:
Okay, looking in to what to be done about this.

Will upload filters for the ttbar processes momentarily.

bendavid:
In general it is much easier to do this in madgraph (either at LO or NLO) since the cuts can be defined at the LHE level in that case.

dnash86:
Yes, madgraph may have to be the fall back.  Requesters originally wanted powheg.

PR #477, opened 2015-06-19
Semileptonic WVgamma, leptonic Wgamma+01jets and +23jets, leptonic Zgamma+012jets

jfaulkne:
I have removed previous pull requests for these processes in order to clean my personal repository, which got messed up.  I have also split the Wgamma+jets into two samples (Wgamma+01jets and Wgamma+23jets) because the WVgamma analysis will only need Wgamma+23jets, and the Vgamma analysis will only need Wgamma+01jets, and neither would need the extensive computationally intensive matching of 0123 jet multiplicities.  There is also a high photon pT Wgamma+23jets sample for photon pT > 500 GeV, which will be used to increase BKG statistics for AQGC limit studies above the SM expectation in the WVgamma analysis.  Overall, this is a semileptonic sample request, where the leading W or Z boson decays leptonically, the second W or Z boson decays to 2 jets.  There is a need to maintain both the resolved jets and boosted jets from the hadronic boson decay in the samples, i.e. there is no lower dR limit between the two jets.


bendavid:
Sorry, l v a j j j is definitely impossible at NLO from a computational complexity standpoint.

Probably lv a j j is impossible also.

So I think the best that can be done is l v a + 0 and 1 jets (then you also get the 2 jet events from the nlo real corrections to the 1 jet)

If you need more jets you'll have to go back to LO.
You can TRY 0,1,2 jets NLO, but there's a good chance it's not feasible.

jfaulkne:
I was afraid this was the case; however, I believe it will be no problem to switch the Wgamma+23 jets to LO, and leave the Wgamma+01jets at NLO.  What about the Zgamma+012 jets sample, can it be left as is or will it need to be split up as well?  Should I close this pull request or is it better to merge a new pull request?  I am not familiar with these commands, so it would be easier for me to just close it and open a new one.

bendavid:
You can just add commits to the existing pull request.
For Zgamma+0,1,2 jets NLO, you can try it, and then change in case of problems.

For Wgamma you could also try Wgamma+0,1,2 jets NLO (try with diagonal ckm at least).  Remember this gives you the equivalent accuracy for the 3rd jet as a 0,1,2,3 jet LO sample.

jfaulkne:
Would that mean using 'loop_sm-ckm_no_b_mass' model?

bendavid:
No, diagonal ckm would mean "loop_sm-no_b_mass"

the "ckm" version is with the off-diagonal ckm matrix (hence more diagrams/more compuationally intensive)

jfaulkne:
I have the Wgamma+012 jets and Zgamma+012 jets, both at NLO, still generating a gridpack at lxplus after 5 days, but the 'HighPhotonPT' Wgamma+012 jets NLO gridpack production has failed twice, both after about 45 hrs of running, with an error message: 

Command "import /afs/cern.ch/work/j/jfaulkne/genproductions/bin/MadGraph5_aMCatNLO/WAToLNuA012j_5f_NLO_FXFX_HighPhotonPT/WAToLNuA012j_5f_NLO_FXFX_HighPhotonPT_gridpack/work/WAToLNuA012j_5f_NLO_FXFX_HighPhotonPT_proc_card.dat" interrupted in sub-command:
"output WAToLNuA012j_5f_NLO_FXFX_HighPhotonPT -nojpeg" with error:
OSError : [Errno 12] Cannot allocate memory

I have submitted the jobs using the following command for the first attempt:

./submit_gridpack_generation_local.sh 30000 2nw WAToLNuA012j_5f_NLO_FXFX_HighPhotonPT cards/production/13TeV/WAToLNuA012j_5f_NLO_FXFX_HighPhotonPT 2nw

and I increased the memory request to 75000 (MB) for the second request, but this has nothing to do with this requested value, I suppose, since the job crashed in both cases.

Here is the resource usage stats from LSF (similar for both jobs):
Resource usage summary:
    CPU time   :  93198.91 sec.
    Max Memory :     35420 MB
    Max Swap   :     36483 MB
    Max Processes  :        14
    Max Threads    :        15

Is this symptoms of not being able to generate processes with two additional jets, or because I require the photon pT > 500 GeV, because the regular Wgamma+012 jets gridpack production is still ongoing (as already mentioned) with no errors yet?

bendavid:
So indeed I think that in both cases this is just an indication that l v a j j at NLO is just not computationally feasible with the current madgraph_aMC@NLO, so I would suggest to fall back to 0,1jets NLO (which means still the 2nd jet at LO accuracy from the real emissions as usual)

jfaulkne:
I've changed the Wgamma+jets and Zgamma+jets to NLO with only up to 1 additional jet; however, I have added in a LO Wgamma+2,3 jets that will be used for the WVgamma analysis, where the Vgamma analysis will use the NLO Wgamma+0,1 jets.  I used the datacards from the WJets process cards, but changed all the 1d5 (or 1d2) to '-1' in the run_card.dat, of course removed the H cuts, and left the rest unchanged.  The xqcut value has been determined previously with MG5 for this process.  I'm not familiar the lapdf value used, as previously I have used 'cteq6l1'.

bendavid:
Are you sure you only want 2,3 jets?  This sample cannot be easily combined with the other one in that case.

Maybe you would be better off doing 0,1,2,3 jets?

jfaulkne:
If it is not a concern for the additional computations and a larger stats sample produced, then I agree that 0123 jets is best for more applicability between analyses, but otherwise will the 0,1 jets stats drown out the 2,3 jets processes (given the WVgamma analysis will be interested in the backgrounds from mostly the 2,3 jets processes)?

bendavid:
How many events would you need in the 0,1,2,3 jet case?

jfaulkne:
The jet matching efficiency for MLM is about 30% for Wgamma+0123 jets in MG5, and for this same sample (produced privately in MG5), the WVgamma analysis acceptance x Eff. is 0.08%.  If I can get ~1500 events out of the pipeline, then that means we'd initially need to generate ~6.5M events.

bendavid:
What is the WVgamma selection roughly?  Any way to enhance things with smarter pt cuts, etc?

jfaulkne:
The list of cuts used in 8 TeV analysis is shown at https://twiki.cern.ch/twiki/bin/view/CMSPublic/PhysicsResultsSMP13009, and for the 13 TeV analysis we will use similar cuts in the resolved jets case, but we'll also be moving to boosted/merged jets.

bendavid:
Ok anyway, no problem to produce 6.5M events at LHE level (what actually matters is the number of GEN-SIM events, which is also fine)

jfaulkne:
I was able to successfully produce gridpacks for Wgamma+0123jets in LO with MLM, Wgamma+01jets in NLO with FXFX, and Zgamma+01jets in NLO with FXFX.  Still awaiting update to scripts for aMC@NLO 2.3.0 beta 1 before creating gridpack for WVgamma processes.

bendavid:
Ah sorry if you missed it, git branch for 2.3.0 beta is available here
https://hypernews.cern.ch/HyperNews/CMS/get/generators/2654.html

PR #478, opened 2015-06-19
new CompHEP cards for W'->t+b requests

peiffer:
Please note, that cteq6 is still used for these requests. I learned from the requesters, that some more work is required to get CompHEP running with LHAPDF.


covarell:
Is it normal to have > 4000 files uploaded?
What is the meaning of prt_<n> in each file?

peiffer:
For each sub-process there is an individual prt_ file. Since each process consist of ~50 sub-processes, we need more than 4000 cards. 

Note, that CompHEP is usually not steered with cards. You define the process and the parameters in an interactive mode and CompHEP creates all these files automatically.

bendavid:
Hi Thomas,
I assume that it is possible to define the parameters in "interactive" mode in a scriptable way.

If this is the case, can we just have instead the input necessary to generate the cards?

peiffer:
Hi Josh,

I could upload the headers of the LHE files as it has been done for CalcHEP requests. Would that be an option?

bendavid:
Do the headers contain the input needed to reproduce the cards/lhe's interactively?

peiffer:
I think so, at least all the sub-processes are listed and all relevant parameters like sqrt(S), particle masses, and other model parameters.

bendavid:
Ok, let's do that then.

peiffer:
Okay, here are the LHE header files. They are still quite huge but the number of files has been reduced to 132.

bendavid:
Ok, but I'm also not sure if this is the most useful form for this information.

Can you give an example of what kind of input is actually provided to CompHEP in interactive mode?

peiffer:
I have forwarded the question to Maxim who created the LHE files.

peiffer:
Hi Josh,
Maxim gave me some files with reduced size. Now, we just have the prt_1 file with all relevant parameters (which are the same for all sub-processes) and a list of all sub-processes in an additional file (cs.txt).

peiffer:
Hi Josh, do you think the documentation of the input variables is fine now?

PR #479, opened 2015-06-19
[Madgraph5_aMC@NLO] Exo-Diboson group: add  "Z'->Zh->nu nu bb"

syuvivida:
Upon the request of Tristan du Pree and Philip Harris, add this additional channel of Z'->Zh with Z-> nu nu and H->bb. The cards include process/run/customize card. There are 13 mass points. 100k events for each mass point.


syuvivida:
Just to add a note: the 39 files here includes the 3 cards for each mass points (13 *3 = 39)
3 cards include run card (identical for all mass points), process card, and customize card that sets the mass and width of new particles.

stosi:
Everything seems OK to me

PR #484, opened 2015-06-23
cards for single Tprime producion in tZ_b finale state

aspiezia:
Cards for single T' producion (pp->T'bq->tZbq process) for the B2G group:
- madspin is used
- both LH and RH
- masses in steps of 100 GeV in the range [700, 1800] GeV
- resonance width equal to 1 MeV


covarell:
They look ok to me

PR #485, opened 2015-06-23
added bbar cards

violatingcp:
Same ttbar cards with bbar


bendavid:
This needs to be migrated consistently to 4 flavour scheme.

Assuming the model already has massive b-quarks, should change maxjetflavor and asrwgtflavor to 4, and switch to 4fs pdf's.

PR #486, opened 2015-06-24
Higgs dalitz decay with HC model at LO

andreypz:
Hello,

This commit is supposed to address the  issues raised in pull #343 

Namely:
   * Switched to a more common HC model
   * Used 5-flavour scheme
   * Included pdf with uncertainties

Notice, however that:
   * It requires version 1.1 of the  HC_NLO_X0_UFO
   * Additional restrictions are used: masses of muons (and electrons) are set to non-zero. The corresponding files are attached.
   * This is yet only a LO production, due to problems with the NLO generation (to be addresses some other time).     
   * Only muon channel in gg production is added, other production modes and electron channel will be added once the cards are validated.
   * This cards are tested in a stand-alone madgraph, version 2.2.3

Please have a look and if it's okay, we shell produce some gridpacks and a  GEN sample.

Thanks,
Andrey


stosi:
Hi
the PDF in the run card is set to 292200, while it should be 263000 for LO, 5f scheme.
Also, you still use a param_card file:  param_card files are not supported by the scripts used for production. Parameter changes should be handled through the customizecards.dat

andreypz:
Okay,

I now modified the PDF ID and removed the param_card. The customize card has bee provided.

Andrey

bendavid:
@andreypz we'll have to think a bit about the right way to incorporate the extra restrict file into the repository.  I'll make a proposal later today.

andreypz:
Hi Josh,
Is there any update on this topic?

bendavid:
Sorry for the delay.  Here's what I would suggest.

1) remove the param card
2) move the extra restrict file to the "patches" area here https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/patches and add the necessary cp command to the gridpack_generation.sh script after the HC_NLO_X0_UFO model is unpacked.

andreypz:
Hi Josh,
I am having some problems with git, it did not do what I wanted it to do.
Although, I think it added the restrict files to this pull request, it did not commit the change to gridpack_generation.sh file

Here is what needs to be change. 
In this block:
```
  cd models
  unzip ../${HCNLO}
  tar -zxvf ../${SINGLEVLQ}
  unzip ../${SINGLEVLQ_diagCKM}

 # adding restrict files 
  cp $PRODHOME/patches/restrict_heft_muelmass.dat  ./HC_NLO_X0_UFO/
  cp $PRODHOME/patches/paramcard_heft_muelmass.dat  ./HC_NLO_X0_UFO/
  cd ..
```
Could you please do that change yourself.

Thanks,
Andrey

bendavid:
why do you need both the restrict and paramcard files?

andreypz:
Because otherwise it would not work. It took me a while to figure out ;)
You can see that in this model there are paramcards corresponding  restrict files. I don't know why.

bendavid:
Ok sure. 
Two related remaining points.
1) Generation would probably be faster using the decay chain syntax
generate p p > x0, x0 > mu+ mu- a

2) you could definitely add extra jets (at least 3, maybe even 4) and do LO jet matching.  You should strongly consider it if you care about the kinematics of extra jets and/or the pt of the higgs.

bendavid:
(if you don't add any extra jets then the speed difference from the decay chain syntax is of course totally irrelevant)

andreypz:
* I have tried adding jets, but then the jobs never went through for various reasons. At this point I think we should do it without jets and produce some samples. We can probably include jets later for an upgraded sample.
 * The generation was pretty fast for me, so I'd rather not change it.

bendavid:
These points are related, ie if you add jets you should/must switch to the decay chain syntax.

andreypz:
Hi Josh,
Could you please accept this pull request and edit the gridpack_generation.sh file as discussed above.
We would like to start the gridpack production at last.

bendavid:
The gridpack_generation.sh change has been merged here
https://github.com/cms-sw/genproductions/pull/555
(I also fixed the output name in the dalitz proc card, since the gridpack scripts require the output name to match the name of the cards)

andreypz:
Thanks a lot.

PR #487, opened 2015-06-24
Madgraph darkmatter missingpoints

violatingcp:
All the missing points in one fell swoop


stosi:
I have checked the first cards (my browser fails to open the whole set...) and they look OK to me. The cut on the invariant mass of lepton pairs at 50 GeV is intentional, right?

covarell:
The commit 2c919c56402e0b926596018c395fa296a6a73a67 is part of this PR, as well as #485 
I don't know if this can create a problem, since this commit needs to be updated according to Josh's comments

peiffer:
Hi Phil, Josh, Roberto,

could you please clarify the issues with this pull request so that I can submit the remaing ttbar+DM and bbbar+DM samples?

Thanks,
Thomas

bendavid:
@covarell I think this is ok as long as each pull request merges as expected.  (It means the diff for the second pull request to be merged has to be checked carefully AFTER the first one is merged)

Question about the cards, this looks like 4fs from the process definition, but not in the run card as far as pdfs and other settings are concerned.  Is this intended to be 4fs or 5fs?

peiffer:
Hi Phil,

could you please make a new pull requests with the missing ttbar+DM cards since this pull request has been rejected?

Cheers,

Thomas

PR #488, opened 2015-06-25
additional mass points for Higgs-Exo

chayanit:


covarell:
Please check values of masswindow... If it is 10d0 for high mass (i.e. allowing virtuality to go down to M_H - 10Gamma_H) it may go below 0 and crash the program

PR #490, opened 2015-06-25
ggzh cards with amcnlo

perrozzi:


bendavid:
Was there any further understanding on the speed difference between the ll and vv cases and/or do you intend to further optimize the choice of xqcut to keep the matching efficiency higher?

perrozzi:
"Was there any further understanding on the speed difference between the ll and vv cases"
- No
"do you intend to further optimize the choice of xqcut to keep the matching efficiency higher?"
- which values do you suggest to try?

bendavid:
well I think the point is to first generate events with xqcut=10, and use those to find a suitable value of qcut.  Then xqcut can be increased to let's say qcut/2 (and this will result in higher matching efficiency/more efficient use of the lhe events)

bendavid:
Since I would expect the same qcut to work for both the vv and ll cases, would be faster to generate (privately) the events needed for the qcut determination in the vv case I guess.

perrozzi:
Ok. never did this studies so I am not fully sure I know exactly what to do.
shall I:
- produce some lhe events with xqcut at 10 (how many)
- shower them with pythia8 using different qcut values (which set of values?)
- which figures of merit should I define? the usual differential jet rates dij?

bendavid:
Yes would say a few 10k events. (100k if you can manage it)

Then for a range of qcut values, would suggest 20-70 let's say in 10GeV steps.  Figure of merit is smoothness of the djr plots (only the 0->1 distribution is relevant in this case)

perrozzi:
I prepared the lhe events. Which pythia8 cfg should I use?

bendavid:
As any other LO process with jet matching.

perrozzi:
is this ok then?
https://github.com/mharrend/genproductions/blob/patch-1/python/ThirteenTeV/Hadronizer_TuneCUETP8M1_13TeV_MLM_5f_max1j_LHE_pythia8_cff.py

bendavid:
Yes

perrozzi:
I did some plots, I think that the situation with qcut 20 or 70 doesn't really change.
You might have a more trained eye to spot problems.
web through afs doesn't see to work so I post an image here. what changes is the number of jets. which might be covered in any case by scale variations (?)
image

perrozzi:
(najets = number of addifional jets other than the 2 b jets from the higgs, nbjets = number of be jets including the 2 b jets from the higgs. there is some genjet-parton matching efficiency effect)

bendavid:
Sorry do you have the d01 plot with the contributions of the two subprocesses individually plotted?  This is what's normally used to assess the smoothness (even if I agree things would seem to look ok)

perrozzi:
no. I guess I should create the 2 gridpacks separately and produce the events independently...

bendavid:
No, that's not necessary, the subprocess index is stored in the event.

bendavid:
And anyways for MLM matching you can access it from GenEventInfo nMEPartons() accessor.

bendavid:
(and there is a ~ready-made script to make these plots here: https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/macros/plotdjr.C)

perrozzi:
Here it is

qcut70

perrozzi:
(and thanks for pointing out the script, indeed it worked out of the box)

bendavid:
And for qcut=20?
(if qcut=20 also looks ok this probably gives reasonably high matching efficiency, and can be used directly with the xqcut=10 gridpack, avoiding the need for a new gridpack)

perrozzi:
here it is
qcut20

bendavid:
OK also looks fine.  What is the matching efficiency for qcut 20?  Would just go for qcut 20 with the existing xqcut 10 gridpacks in this case.

perrozzi:
24.7% for qcut 20, 26.5% for qcut 70
Ok I will go ahead with qcut 20 (unless these numbers don't sound to you)

bendavid:
Ok, I was actually expecting a bigger difference, but this just means things are really stable as a function of qcut.  You might want to go ahead rather with 30 as the nominal value in that case so that there is some room for downward variation in case someone wants to evaluate matching systematics later.

perrozzi:
could you please merge if ok?

PR #491, opened 2015-06-26
Filters for TTBar M_ll binned powheg

dnash86:
Likewise very small efficiencies, as in https://github.com/cms-sw/genproductions/pull/474.  Could these be checked to see if they're correct?

For use with the input cards here: https://github.com/cms-sw/genproductions/pull/473

### DIFF
--- a/python/ThirteenTeV/Hadronizer_TuneCUETP8M1_13TeV_powhegEmissionVeto_TT_Mee_1200_1800.py
+++ b/python/ThirteenTeV/Hadronizer_TuneCUETP8M1_13TeV_powhegEmissionVeto_TT_Mee_1200_1800.py
@@ -1,0 +1,40 @@
+import FWCore.ParameterSet.Config as cms
+from Configuration.Generator.Pythia8CommonSettings_cfi import *
+from Configuration.Generator.Pythia8CUEP8M1Settings_cfi import *
+from Configuration.Generator.Pythia8PowhegEmissionVetoSettings_cfi import *
+
+generator = cms.EDFilter("Pythia8HadronizerFilter",
+                         maxEventsToPrint = cms.untracked.int32(1),
+                         pythiaPylistVerbosity = cms.untracked.int32(1),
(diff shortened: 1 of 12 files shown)


bendavid:
How small?
If you're getting efficiencies less than about 1/10k then I don't think this is a very practical way of producing the events.

dnash86:
(31) / (5000) in the 500-800 bin.  So even smaller in the higher bins.

covarell:
Do you know who produced LHE files for 8 TeV searches (both for this and WW) and if they had to face this huge waste of events?

dnash86:
I think it may have been the TOP group back then (at least for this, ttbar) - although I am not 100% positive.

As for how to move forward for this and the WW samples, it seems that the analyzers will produce skimmed LHE files and provide all the input cards, scripts, etc for validation that they are okay.  Then we can move forward with the pLHE campaign, removing the need for GEN-SIM-level filters.

Is this okay with you, @covarell, @bendavid?  I know the wmLHE campaign is *strongly* encouraged for powheg, but in this case since the filter efficiency is so low it seems if we can validate the scripts used that the pLHE option may be the only way.

covarell:
This is OK for us, but a justification should also be provided of why this cannot move to MG5_aMCatNLO for all this samples (no skimming needed there)

dnash86:
As it turns out, after discussing further with the analyzers, they actually *are* willing to switch to MG5_aMCatNLO for the ttbar and WW production.

Starting with ttbar, now, I assume this is an okay starting point to edit in mass cuts:

https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/tt012j_5f_ckm_NLO_FXFX

Is that okay?

covarell:
It is OK, be aware that ttbar will have a significant amount of negative-weight events (to be determined), we do not know yet about WW. Sample sizes must be probably adjusted (increased) to compensate for these negative weight canceling positive weights

PR #497, opened 2015-07-01
Cards for single VLQ B in association with t

dgonzal:
The reason that the mg cards have the special way of writting the process is the same as in the Bpb_LO folder.


dgonzal:
any comments ? @bendavid @stosi

dgonzal:
If there are no comments could you pls accept the commit ? @bendavid @stosi @davidsheffield @dnash86 @syuvivida

bendavid:
If you wanted to be fully inclusive, you could probably even lower ptj to 0.01.  Otherwise looks fine.

bendavid:
Also in some of the processes the tops will be decayed by pythia.  Is that intentional and acceptable to you?  (It means you will lose spin correlation/polarization effects)

dgonzal:
I just took the recommended values for the ptj ... 
I had a closer look at the the top correlations, they have been thought to have negligible impact on the decay products, but actually they have some that could be just below the trigger threshold of the leptons so I will add the needed cards for completnes. Thanks for the comment

dgonzal:
The run card with the ptj ... values came from alexandra carvalho who is working on the theory side and I would stick to it, since also the Bpb_LO and other single VLQ samples where producd with it.

Any comments ?

dgonzal:
@bendavid

bendavid:
Does the last commit add also additional mass points?

dgonzal:
It adds different chiralities which means more samples but not more mass points.

PR #500, opened 2015-07-01
cards for the Non resonant hh group

sudhaahuja:
Could you please approve the standard model (*_sm) cards first.


covarell:
A few problems with these requests:
1) The model needs to be added to generator afs area, and to the gridpack_generation.sh script.
2) param cards are not supported by the gridpack scripts. Please use the customizecards mechanism instead
3) PDF's are not consistent with the rest of the production (please see the example cards, should use nnpdf30lo alphas=0.130 4flavour in this case)
4) Please reduce the number of events in the run card (should normally just be 1000 or 1500 for the gridpack creation step) and set gtridpack to True
5) the cards should not be named just run_card.dat, proc_card.dat but also have the name of the process in the filename e.g. HH4b_node_10_run_card.dat

sudhaahuja:
Hi Roberto,

Sorry for the delay but the changes requested have been implemented and the edited cards have been committed to the correct area this time. I could not commit to my last forked repository so I had to delete it and re-fork and create a new pull request. The pull request is this one - https://github.com/cms-sw/genproductions/pull/549 <https://github.com/cms-sw/genproductions/pull/549>

Please see comments inline -  
> A few problems with these requests:
> 1) The model needs to be added to generator afs area, and to the gridpack_generation.sh script.
> 
The model has already been uploaded, it is located at /afs/cern.ch/cms/generators/www/BSM_gg_hh.tar <http://cern.ch/cms/generators/www/BSM_gg_hh.tar> . The model has also been added in the gridpack_generation.sh (following the way other models are given in the script).
> 2) param cards are not supported by the gridpack scripts. Please use the customizecards mechanism instead
> 
Customise cards have been uploaded.
> 3) PDF's are not consistent with the rest of the production (please see the example cards, should use nnpdf30lo alphas=0.130 4flavour in this case)
> 
Corrected as given in the rest of the production cards.
> 4) Please reduce the number of events in the run card (should normally just be 1000 or 1500 for the gridpack creation step) and set gtridpack to True
> 
Done
> 5) the cards should not be named just run_card.dat, proc_card.dat but also have the name of the process in the filename e.g. HH4b_node_10_run_card.dat
> 
Done


Let me know if anything needs to changed.
Sudha.

PR #505, opened 2015-07-03
HToZATo2L2B

AlexandreMertens:
Adding an example of H->ZA->llbb set of cards for MG5 production.
The masses are MH = 500, MA = 200


bendavid:
A few things:
1) The model needs to be added to generator afs area, and to the gridpack_generation.sh script.
2) param cards are not supported by the gridpack scripts.  Please use the customizecards mechanism instead
3) cut_decays should be set to F normally
4) PDF's are not consistent with the rest of the production (please see the example cards, should use nnpdf30lo alphas=0.130 4flavour in this case)
5) Please reduce the number of events in the run card (should normally just be 1000 or 1500 for the gridpack creation step)

AlexandreMertens:
Thanks for the feedback,
Concerning the "2)", the customizecard should contain a command for each difference between the actual and the default param card of the model, is it correct?
Each masses for A and H induce different branching ratios and width. So it would be not optimal to do all those changes by hand. Is there already an automatic way to create the customizecard?

Thanks in advance,
Alexandre

bendavid:
Yes that's right, you'll need a different customizecard for each mass point.  There is no ready-made automated procedure, but it should be trivial to make a script to do it.

AlexandreMertens:
okay, thanks!

AlexandreMertens:
Hello,

Concerning 
1)  The model needs to be added to generator afs area, and to the gridpack_generation.sh script.

How do I have to proceed? 
related to that point, I have a set of cards and would like to test the gridpack generation with them before doing the pull request.
How can I add our 2HDM model to the genproduction?

Thanks in advance,
Cheers,

AlexandreMertens:
I have updated the cards, and I made a test and everything work fine.
Is there still modification needed or problem with this pool request?

Cheers,

bendavid:
Ok I see the model is now in afs and in the scripts from another request.
I see two remaining issues

1) the process looks like it has no jet matching.  In that case ickkw should be set to zero

2) the process is defined as "g g >".  In almost all cases we would use "p p >".  Are there some quark-initiated processes that you explicitly want to exclude here?

AlexandreMertens:
Thanks for the comments!
I updated the PR

PR #509, opened 2015-07-06
Updating thw configs 

maierbenedikt:
Moving to 5fs LO; this is the only setup that allows the reweighting in the cV-cF plane, since it doesnt require diagram removal or subtraction. Both wouldn't work with the reweighting.


covarell:
Output command with your own local directory does not look good. However, I understand this uses pLHE, so I will approve the corresponding McM request

maierbenedikt:
Fixed!

PR #511, opened 2015-07-06
higgs width bug fix

perrozzi:


syuvivida:
The cards look OK to me. @bendavid, @covarell could you merge them? Thanks!! Eiko

bendavid:
I guess this makes no difference in practical terms, but ok, might as well do it right.

PR #515, opened 2015-07-07
Added MadGraph5_aMCatNLO cards for WGGJets sample

mmachet:


bendavid:
This has ONLY the 1 jet process.  I would have thought this is supposed to include both 0 and 1 jets?

mmachet:
Hi Josh,
yes you are right. I modified the proc card, hope the syntax is correct.

2015-07-14 12:17 GMT+02:00 Josh Bendavid <notifications@github.com>:

PR #520, opened 2015-07-07
update charged Higgs cards and W+1jet cards

gitytakahas:


covarell:
- In H+ proc_card all statements in lines 46-57 should be removed I think, because they are set automatically. In general I think the statement "launch" should not be both in the proc and madspin cards
- W + 1Jets seems actually W -> tau nu + 1 jet. Is that intentional? I think in the spreadsheet inclusive W was intended

gitytakahas:
Regarding the H+ card, there is a specification of the mass in line 53. If I remove the line that you mentioned, where can I specify the mass value ? 

Concerning to W+1jet, I grab the card from, 

https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/examples/wellnu012j_5f_NLO_FXFX

Where they also have "dummy tau nu" decay in madspin card. Do you have any suggestion ? 

thanks,
Yuta

bendavid:
Yes, the dummy decay looks ok (@covarelli the sample is inclusive because the process is already generated through to ell+ vl at the proc card level.  madspin is just being used to fill in the spinup flags relevant for the tau decays in pythia)

For the W+1jets sample, it should be understood that this sample will be unphysical on its own.  This is intended to be combined with the existing inclusive W+jets sample for extra statistics right?

gitytakahas:
Yes, it is right, it is for increasing the stat.

Best,
Yuta

bendavid:
Ok, are you also planning to produce a W+2 jet sample along similar lines?

covarell:
Ah yes, OK for mhc, the rest can be suppressed

gitytakahas:
Hi Josh,

Yes, I'm planning to produce W+2,3,4 jets as well. 

https://docs.google.com/spreadsheets/d/16Z3rgM7XZzUMBxZr3DUXMldvGipeHwjDolUW5GSKn_s/edit#gid=0
(please see 251 - 254)

However, the gridpack production for these process takes forever ... it really takes time. Do you have any advice / suggestions ?

Thanks,
Yuta

bendavid:
No, sorry.  NLO goes up to 2 jets (3rd jet from real emission diagrams)

So you can only produce W+1j, W+2j in that case (and the W+2j sample has also W+3j events)

If you really care about W+4 jets then you need to instead switch to LO.  (but then switch to LO for all of them so that they can be combined properly later)

gitytakahas:
I now updated the cards for the charged Higgs. Please check.

Thank you,
Yuta

covarell:
Hi, it seems some more spurious line sneaked in:

set lhapdf /afs/cern.ch/user/e/epekkari/programs/LHAPDF-6.1.5/lhapdf-config

also to be removed

gitytakahas:
Hi Roberto,

Thank you, I just updated now. Please check again.

Thanks,
Yuta

gitytakahas:
Hi Roberto, Josh,

After I removed the lines starting from "launch" (except for "set mhc XXX"),
I got following errors.

Command "import /afs/cern.ch/work/y/ytakahas/gridpack/genproductions/bin/MadGraph5_aMCatNLO/ChargedHiggs_light_M160/ChargedHiggs_light_M160_gridpack/work/ChargedHiggs_light_M160_proc_card.dat" interrupted in sub-command:
"set mhc 160" with error:
InvalidCmd : Possible options for set are ['group_subprocesses', 'ignore_six_quark_processes', 'stdout_level', 'fortran_compiler', 'cpp_compiler', 'loop_optimized_output', 'complex_mass_scheme', 'gauge']
quit
 

If I again wrote the "launch", it succeeded. Probably, we cannot use "set mhc" without launch sentence.
Does this make sense ?

Thanks, Yuta

bendavid:
No you can't set parameters in the proc card.  These statements need to be moved to the customizecards.dat.  See for example 
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/examples/wplustest_4f_NLO/wplustest_4f_NLO_customizecards.dat

PR #524, opened 2015-07-08
hadronizer for LFV Higgs to mutau,etau,emu

chayanit:


covarell:
Pleasedo not redefine top mass to values different than 172.5 GeV (the value we use in CMS) unless there is a need for.
POWHEG emission veto settings should be called with the block, like you do for e.g. CUET tunes, since you call the python file containing that block
Why special tau decay settings are needed, rather than normal Pythia8 behavior?

chayanit:
Hi Roberto,

I received these fragment from analyzers and please find answer in lines.

Pleasedo not redefine top mass to values different than 172.5 GeV (the value we use in CMS) unless there is a need for.

OK. I will remove the line top mass.

POWHEG emission veto settings should be called with the block, like you do for e.g. CUET tunes, since you call the python file containing that block

Agree. I will remove them too.

Why special tau decay settings are needed, rather than normal Pythia8 behavior?

'TauDecays:mode = 2' forces the uncorrelated decay of taus coming from a "mother", which in this case is the Higgs ('TauDecays:tauMother = 25') to occur with 'TauDecays:tauPolarization = 0'. 0 is in fact the default value of tauPolarization, according to [1], but it probably wouldn't hurt to specify it just in case.

Also from [1]: The decays of tau leptons are categorized as correlated, where a tau pair is produced from a single process, or uncorrelated, where only one tau is produced.

So in this case the tau decay is uncorrelated because we are trying to model a BSM result where only one tau comes from the mother. So to summarize, we need the extra tau parameters to make sure that pythia is correctly modeling our exotic process.

Does it make sense?

Cheers,
Chayanit

[1]
http://home.thep.lu.se/~torbjorn/pythia82html/ParticleDecays.html

covarell:
Ok, please proceed and I will approve

PR #526, opened 2015-07-08
Drell Yan Zpt>150

tmulholland:


covarell:
Two small comments:
- in ZPt150 proc_card, the syntax seems confusing: proton and jets are first defined in 4f scheme, then there is a second define statement without equal sign... I am not sure 5f scheme is called correctly
- All tribosons are configured not to activate FxFx. Is that intentional? @bendavid

bendavid:
Are the triboson cards even intended to be part of this pull request?

tmulholland:
I'd like to pull all of them, if possible.

tmulholland:
I configured the Zpt150 proc card to conform exactly to the HT binned DY 
cards. See for instance:
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/DYJets_HT_LO_MLM/DYJets_HT_mll50/DYJets_HT-600toInf/DYJets_HT-600toInf_proc_card.dat

There are no additional jets in the new triboson sample, so there is no 
need to activate FxFx.

-Troy

covarell:
Thanks! @alkaloge can you have a look at the file mentioned by Troy in the github thread and confirm that the strange define statement (with no '=' sign) actually configures 5f scheme?

alkaloge:
Hi - "=" is also an acceptable argument for MG while defining multi particles ;-)

tmulholland:
Hi,

Does this mean we can accept the pull request or is there something else 
I need to change?

-Troy

PR #527, opened 2015-07-09
Adding VBS cards LO MG

jasperlauwers:
MG LO cards for: WWjj same and opposite sign, WZjj


covarell:
They look OK to me, however be aware that:
- this syntax will produce only events with on-shell W and Z
- the QCD-production samples are now configured to be without matching, so it will be impossible to merge this sample with inclusive WW and/or with other jet multiplicities

jasperlauwers:
Thank you for the comments.

- Although the decay chain formalism is used in the process card, the bwcutoff was set to 100 to take into account the off-shell intermediate states as well.
- The matching does not seem necessary to us. 

So we would keep the cards as they are. Do you agree it is fine like this?

cc: @amassiro

bendavid:
For other samples which have been done this way, the bwcutoff was set to 15000 to really have the full off-shell phase space (studies in the SUSY group have shown it can make a difference when looking in the far tails)

jasperlauwers:
Thanks! bwcutoff has been changed to 15000.
I did a small check myself and saw an increase of ~4% in the cross section of same sign WW ewk after increasing the bwcutoff to 15000.

jasperlauwers:
I have a question about the sysCalc part:
Is CT10nlo.LHgrid = sys_pdf correct that is in the cards now? Even though these are LO cards?
Or should it be NNPDF30_lo_as_0130 = sys_pdf ?

bendavid:
This part is ignored.  (The final syscalc card is overridden from here:
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/runcmsgrid_LO.sh#L59 )

jasperlauwers:
Thanks for the help.

I've made a few updates to the WZjj cards:
- specifying the z > l+ l- decay
- removing the rapidity cut on leptons so the sample can be used as a background for WWjj
- adding cards for studying Z/gamma interference

If everything looks ok, please merge the PR.

AndrewLevin:
Hi,

There are also same sign WWjj cards here

https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/WpWpJJ

although I think the ones here are more correct because of the breit-wigner settings. We should remove MadGraph5_aMCatNLO/cards/production/13TeV/WpWpJJ and just use the ones here in the future.

The only other possibly important difference I see is that ckm is turned on here. This can make the generation significantly slower, and it did not seem to affect the total cross section at least.

Andrew

AndrewLevin:
Hi,

Also, I am not sure I understand how we are planning to use some of the wzjj cards. There are the following cards:

WZjj ewk
WZjj ewk+qcd
WZjj qcd
Wajj ewk with mmll = 60
Wlljj ewk with mmll = 60

Is there a reason why we don't have just

Wlljj ewk+qcd
Wlljj ewk
Wlljj qcd

Also, for the wpwp analysis, where this sample will be used as a background, it would be useful to lower the mll cut from 60 to 12, since we want to try to include as many outside of acceptance leptons as possible.

Andrew

AndrewLevin:
Also, is there any reason to set etaj = 5.5 and not etaj=-1?

jasperlauwers:
Dear Josh and Roberto,

The gridpack production is going rather slow and sometimes freezes after a few days. Would it be ok to use sm-no_b_mass instead of sm-ckm_no_b_mass for these processes?

bendavid:
If you don't care about the flavour composition of the outgoing jets, then sure, although these shouldn't be particularily intensive gridpacks.  What do you mean by "freezes" exactly?  Can you point to a log?

AndrewLevin:
We are going to use (anti) b-tagging, so I guess this implies that we do care, right?

bendavid:
Yes.

jasperlauwers:
The WZ and WW opposite sign cards have been updated after discussion with Andrew. 

When trying to generating the gridpacks it became clear that some of the processes take too much resources for the gridpack production. For example Wlljj_lowMll_ewk_qcd submitted ten thousands of jobs and created a directory of more than 70GB.  
Is there a way to deal with this? Or should they be produced locally?

bendavid:
There is a way to deal with the disk usage at least.  Is the situation significantly better when trying to generate events directly without the gridpack step though?

jasperlauwers:
I have not tried generating events directly yet. 
Limiting the disk usage might be good enough. Could you expand on this?

bendavid:
If you use the submit_gridpack_generation.sh script as described in
then it will use the disk space on the scratch space of the worker node (and only the final tarball will be copied back to afs)

AndrewLevin:
Hi Josh,

Can you please merge this pull request? We are still having problems actually producing the Wlljj ewk+qcd sample, so once this is merged I will make a new pull request with Wlljj ewk+qcd without ckm, which is the best we can do right now.

Andrew

bendavid:
Why not just update this pull request then?

AndrewLevin:
Only the author of a pull request can update it, as far as I know. Jasper is on vacation now. If you want I can make a new pull request that includes this one plus the extra things that I want.

bendavid:
Yes, please do.  We want to avoid merging into github intermediate things if possible.

AndrewLevin:
OK, here is my new pull request:

https://github.com/cms-sw/genproductions/pull/576

PR #528, opened 2015-07-10
[Madgraph5_aMC@NLO] EXO Diboson group: cards for RS graviton -> ZZ->ZlepZhad

syuvivida:
This channel has the same run cards as all the other EXO Diboson channels.

There are 13 mass points. We would like to generate 50K events for each mass point.

Note the customization cards have more lines than the other channels because the default 
model files in Madgraph have inconsistent masses with the HVT model/Bulk Graviton/Radion models 
we used for the other channels. Therefore, we set masses of b/t/W/Z in the customize cards so to be consistent with the other channels.


syuvivida:
@bendavid @covarell are the customizecards ok?
If they are OK, could you merge them? Many thanks!!

PR #531, opened 2015-07-13
update charged Higgs cards

gitytakahas:
8 mass points for low mass, 13 mass point for heavy mass
each contains 4 files (madspin, custermized, proc and run cards), total 84 cards.


bin/MadGraph5_aMCatNLO/cards/production/13TeV/ChargedHiggs_light_M80/ChargedHiggs_light_M80_proc_card.dat
```diff
@@ -0,0 +1,45 @@
...
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+# Specify process(es) to run
+#generate p p > e- ve~ @1
+#add process p p > e- ve~ j @2
+#add process p p > t t~ @3
+generate p p > t t~ [QCD]
+
+# Output processes to MadEvent directory
+output
```
bendavid: this extra "output" line should also be commented/removed

bendavid:
A few comments.  For the "heavy" charged higgs, I think you also want to decay the tops in madspin (otherwise it will be done in pythia and you will lose spin correlation/polarization effects)

Also for the light charged higgs, it would be possible to add extra jets in a merged sample if desired (similar to the SM ttbar sample)

gitytakahas:
Hi Josh,

I updated the cards. Please check again. 

Thanks,
Yuta

PR #532, opened 2015-07-13
Zfullyhadronic + ggzh cards to the correct branch

perrozzi:


bendavid:
Sorry madspin is not supposed to be supported with loop-induced processes.  Does this actually work?

perrozzi:
sorry I just realized is a spurious commit, I had already closed it but I forgot to remove the files. it is done now...

perrozzi:
question related to ggzh01j_5f_LO_MLM_M125_ZToAll datacards: if is not possible to use madspin, is still possible to let the Z decay inclusively? if so, could please point me to an example?

bendavid:
No, I don't think this is possible for the moment.  You can at least combine ll and vv of course (but will be slower to generate the gridpack of course)

bendavid:
In fact for the hadronic decays, I think the only option for the moment is to leave the z undecayed and decay it in pythia (and loose the spin correlations/polarization effects)

PR #533, opened 2015-07-13
cuts.f file for dyellell012j_5f_NLO_FXFX with pt_ll>100GeV

wvieri:


bendavid:
Are the proc and run cards already there?

wvieri:
proc and run cards should be exactly the same as the default dyellell012j_5f_NLO_FXFX sample, I can copy them here if needed

bendavid:
yes, please do

wvieri:
done

bendavid:
You'll have to rename them for this to work with the gridpack scripts.

wvieri:
ok, done!

bendavid:
Now there are two copies of the files...

wvieri:
deleted

PR #536, opened 2015-07-14
[Madgraph5_aMC@NLO] Exo-diboson group: add other RS->ZZ/WW decays

syuvivida:
Add the following decay modes. Each decay mode has 13 mass points and each mass point has 3 cards (run card, customizecards, proc cards), which gives 195 cards in total. 
Each mass point requires 50K events.

The run cards are all identical to that of the other Diboson modes. The customize cards are also identical to that of the previous commit: RSGraviton_ZZ_ZlepZhad 
(in this pull request:https://github.com/cms-sw/genproductions/pull/528 )

RSGraviton_WW_inclu      
RSGraviton_WW_WlepWhad
RSGraviton_ZZ_ZhadZhad
RSGraviton_ZZ_ZincluZinv
RSGraviton_ZZ_ZhadZinv


syuvivida:
@bendavid , @covarell are these cards OK? If yes, could you merge them? Many thanks!!

syuvivida:
@covarell, @bendavid could you merge these cards? Thanks!! (They actually are similar to the graviton cards in https://github.com/cms-sw/genproductions/pull/543. Only the decay modes are different)

stosi:
The cards look OK to me

PR #540, opened 2015-07-15
Zgg NLO production cards for 8 TeV

maravin:
Dear all,

per the request from Alberto Belloni, I created a Zgg NLO production cards (we intend to submit a request for 0.5-1M events of this process).

Sincerely,
Yurii


bendavid:
suggest to reduce the number of events in the run card to 100 even.  (This is just for a small pilot production during the gridpack creation, so no sense wasting time here)

bendavid:
otherwise looks ok.

You could probably add an extra jet for a 0,1 jet merged sample if you care about the additional jet kinematics.

maravin:
Thanks Josh - we do not expect much stats at all for Zgg at 8 TeV, so a simple NLO sample would be sufficient. Should I modify the file to have 100 events and resubmit the pull request?

bendavid:
You can just add an additional commit to the branch to update the existing pull request.

maravin:
Done!

PR #541, opened 2015-07-16
s- and t-channel single top in LO MG

trandbea:
Here is a first try at the cards for single top at LO. Please let me know if these look ok, so I can start making the gridpacks. Thanks!

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/WJets_HT_LO_MLM/WJetsToLNu_HT-0toInf/WJetsToLNu_HT-0toInf_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/WJets_HT_LO_MLM/WJetsToLNu_HT-0toInf/WJetsToLNu_HT-0toInf_proc_card.dat
@@ -1,0 +1,24 @@
+import model sm-ckm_no_b_mass
+#switch to diagonal ckm matrix if needed for speed
+#import model sm-ckm_no_b_mass
+
+define p = p b b~
+define j = j b b~
+
+define ell+ = e+ mu+ ta+
+define ell- = e- mu- ta-
+
+generate p p > w+, w+ > ell+ vl $$ t t~ h @0
+add process p p > w+ j, w+ > ell+ vl $$ t t~ h @1
+add process p p > w+ j j, w+ > ell+ vl $$ t t~ h @2
+add process p p > w+ j j j, w+ > ell+ vl $$ t t~ h @3
+add process p p > w+ j j j j, w+ > ell+ vl $$ t t~ h @4
+
+add process p p > w-, w- > ell- vl~ $$ t t~ h @5
+add process p p > w- j, w- > ell- vl~ $$ t t~ h @6
+add process p p > w- j j, w- > ell- vl~ $$ t t~ h @7
+add process p p > w- j j j, w- > ell- vl~ $$ t t~ h @8
+add process p p > w- j j j j, w- > ell- vl~ $$ t t~ h @9
+
+output WJetsToLNu_HT-0toInf -nojpeg
+
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/WJets_HT_LO_MLM/WJetsToLNu_HT-0toInf/WJetsToLNu_HT-0toInf_run_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/WJets_HT_LO_MLM/WJetsToLNu_HT-0toInf/WJetsToLNu_HT-0toInf_run_card.dat
@@ -1,0 +1,275 @@
+#*********************************************************************
+#                       MadGraph5_aMC@NLO                            *
+#                                                                    *
+#                     run_card.dat MadEvent                          *
+#                                                                    *
+#  This file is used to set the parameters of the run.               *
+#                                                                    *
+#  Some notation/conventions:                                        *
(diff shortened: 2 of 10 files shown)


bendavid:
@maierbenedikt can you take a look.
What is the motivation to use 4fs for the s-channel production in this case rather than 5fs?  (naively I would think that 5fs has the advantage of including additional b's in the extra jets)

trandbea:
I thought we can't use the 5fs if the final state has b's, since the matching would get confused?

bendavid:
So actually, you are going to have a problem in this regard also for the t-channel case.  The proper solution (and this would allow also using 5fs for the s-channel) would be to have the proper exclusion of "electroweak partons" from the matching.  This will be fixed in the next pythia8 version, but is not working correctly right now.

As it stands, in the t-channel case, you will have all events with light jet pt<qcut removed.  Since it's not clear how high qcut will need to be, this might actually be a significant problem/showstopper for generating these samples until the proper exclusion of "electroweak partons" from the matching is available.

(Sorry for not bringing this up earlier.  Just thought of it now in fact in response to your comment about the b quark in the s-channel case)

bendavid:
@maierbenedikt was the 5fs single top with jet matching tested in CMSSW?  In which version exactly?  (this electroweak parton exclusion was partially working at some point, before it was completely broken in more recent releases)

trandbea:
hm, sorry, indeed i didn't connect the dots for the t-channel...

When is it expected to have the p8 update? Also, if it worked previously, is it a possibility to do it in that older CMSSW version?

Btw, to make use of b's in the proton to make an s-channel W, we have an extra suppression of |Vcb|^2 or |Vub|^2, so it should not contribute much. (I got to wondering if there are other reasons to stick to 4fs for the s-channel since I noticed that in the MG paper p.80 (http://arxiv.org/pdf/1405.0301.pdf), t-channel is quoted in both 4fs and 5fs, while s-channel is only in 4fs.)

maierbenedikt:
For the s-channel it really doesnt matter, it is no b-initiated process. In general, 4fs is the default of MG, so probably that is why they quote the s-channel in the 4fs only. However, the additional jets would not contain b jets in this case, but I'd expect this to be a minor effect as well. So I would also do the s-channel like this, with the caveat that I am not sure whether with this syntax > w+> you will ever get additional initial state radiation. Maybe you can check the diagrams?


I wasn't aware of the fact that the current P8 version doesnt exclude jets from an EW vertex anymore. That is bad and indeed a showstopper for this!

We used CMSSW_5_3_22_patch1,  and it worked correctly, the light jet was always excluded, you saw this from the nPartons vs. nPartonsMatched.

bendavid:
I think it will still be several weeks before this issue is resolved.

Indeed the main difference between 4fs and 5fs in the schannel case is not the initial state b's, but rather whether the extra jets include b's.  (in the 4fs case, extra b's would only come from the parton shower).  This may also be fine though, especially if you don't care much about extra b's in the analysis.

There is definitely not a production-ready cmssw release which contains even the old partially working setup unfortunately, so best would be to wait for the proper fix. (or else use powheg)

trandbea:
Ok, thanks. Should I close this pull request then, or do we keep these cards for when the update comes along?

bendavid:
let's leave this open for when the update is available.

trandbea:
Given Rikkert's concern for gauge invariance in LO with higher jet multiplicity, should we drop this? I guess I thought that single top + extra jets has been done at LO before so there must be a way but maybe not.

I want to do a pull for the WJets inclusive cards and I can't figure out how to keep this alive and do the new pull request simultaneously.

bendavid:
You can always make another pull request independently by using a uniquely named branch for each pull request.

maierbenedikt:
We did study it at LO, and saw that the differences in key distributions with respect to a well tested and understood configuration (t-channel 4fs at NLO) were small and most of the times (not always) covered by scale uncertainties. We did not study where those differences came from - from LO vs. NLO, from Pythia8 which runs with additional settings for NLO or indeed from presumably breaking gauge invariance.


But given that the exclusion of jets from EW vertices does not work right now, we could close this for the moment. 

The final test (a gridpack production with madspin_card.dat and proc_card.dat that are handed over) for Powheg+MadSpin is running for the t-channel. 
I expect to make a pull request tomorrow, then we can generate a rather large sample including Q2+PDF weights, which has

a) more statistical power than the aMCatNLO sample (btw, a 30M extension is in production)
b) only one jet less at ME level than for LO MLM in the 5F
("p p > t j j j j $$ w+ w-" for MG vs.  "p p > t j b~ $$ w+ w- [QCD]" plus the additional emission for Powheg).

trandbea:
Ah, sorry, the master's thesis LO to NLO comparison indeed slipped out of mind at this point. Though from Rikkert's email it was not clear to me how "problems with gauge invariance if you go beyond the lowest multiplicities" would show up in actual distributions, so then I can't tell if the comparison in the thesis is addressing his concerns... Maybe the effect is to show up in some particular region of phase space??

Anyway, it's great to hear the new Powheg request is on its way!

I'll go ahead and set up another branch for my new pr. Thanks, Josh

trandbea:
ok, I clearly messed this up. So can I close and remove my old fork and fork again? Sorry, still learning git...

bendavid:
Ok, at least back up the cards somewhere.
Closing this one for now.

trandbea:
For future reference, how do I make that second branch that is void of my prior commits?

PR #543, opened 2015-07-18
Graviton cards for ZlepZlep and ZlepZinv

dsperka:


syuvivida:
The RS graviton cards look OK to me.
@dsperka actually your customization cards of Bulk Graviton were the ones for RS graviton cards. (There are extra settings for RS model) In principle, they should also work for Bulk graviton. But I never tried. Have you tested them?

dsperka:
Sorry about that I can just redo them with the correct customizations.

syuvivida:
The updated cards look  OK to me.
@covarell @bendavid could you merge them?

PR #545, opened 2015-07-20
added MadGraph5_aMCatNLO cards for ZGGJets sample

mmachet:


bendavid:
Do you care about the gamma* contribution?

mmachet:
Do you mean that the second part $$ t t~ can be deleted?

2015-07-20 18:21 GMT+02:00 Josh Bendavid <notifications@github.com>:

bendavid:
no, I mean that you may want to generate directly l+ l- instead of z (at least for the leptonic decays)

mmachet:
Ah ok! So you suggest to generate l+ l- for the leptonic decay, and in a
similar way for the other decays? Or for the other decays I keep the z?

2015-07-20 18:57 GMT+02:00 Josh Bendavid <notifications@github.com>:

bendavid:
Yes, though to do it that way you will have to split into at least two samples.
Eg. one sample with l+ l- and the other sample with z (and then configure madspin to do only the neutrino and hadronic decays)

PR #548, opened 2015-07-22
bstar 5f scheme with b jet to be massless

smshaheen:
The following changes has been made.
1) param_cards are replaced by customized_cards.
2) b-quark mass is set to be zero.
3) Scheme has been changed from 4f to 5f with changing
    i) j = j b b~ has been added in the process.
   ii) The pdf used is 26300.
  iii)  maxjetflavor has been set to 5

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/bstar_tW_Left_M1200/bstar_tW_Left_M1200_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/bstar_tW_Left_M1200/bstar_tW_Left_M1200_proc_card.dat
@@ -33,7 +33,6 @@ define vl = ve vm vt
 define vl~ = ve~ vm~ vt~
 import model Bprime
 define p = p b b~
-define j = j b b~
 define l+ = l+ ta+
 define l- = l- ta-
 generate p p > t w- QAD=2 QCD=0 QED=0 , t > b j j , w- > l- vl~
@@ -41,8 +40,7 @@ add process p p > t w- QAD=2 QCD=0 QED=0 , t > b l+ vl , w- > j j
 add process p p > t~ w+ QAD=2 QED=0 QCD=0 , t~ > b~ j j , w+ > l+ vl
 add process p p > t~ w+ QAD=2 QCD=0 QED=0 , t~ > b~ l- vl~ , w+ > j j
 add process p p > t w- QAD=2 QED=0 QCD=0 , t > b l+ vl , w- > l- vl~
-add process p p > t~ w+ QAD=2 QED=0 QCD=0 , t~ > b~ l- vl~ , w+ > l+ v\
-l
+add process p p > t~ w+ QAD=2 QED=0 QCD=0 , t~ > b~ l- vl~ , w+ > l+ vl
 add process p p > t w- QAD=2 QED=0 QED=0 , t > b j j , w- > j j
 add process p p > t~ w+ QAD=2 QED=0 QCD=0 , t~ > b~ j j , w+ > j j
-output  bstar_tW_Left_M1200
+output bstar_tW_Left_M1200
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/bstar_tW_Left_M1200/bstar_tW_Left_M1200_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/bstar_tW_Left_M1200/bstar_tW_Left_M1200_proc_card.dat
@@ -33,6 +33,7 @@ define vl = ve vm vt
 define vl~ = ve~ vm~ vt~
 import model Bprime
 define p = p b b~
+define j = j b b~ 
 define l+ = l+ ta+
 define l- = l- ta-
 generate p p > t w- QAD=2 QCD=0 QED=0 , t > b j j , w- > l- vl~
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/bstar_tW_Left_M1200/bstar_tW_Left_M1200_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/bstar_tW_Left_M1200/bstar_tW_Left_M1200_proc_card.dat
@@ -1,0 +1,48 @@
+#************************************************************
+#*                     MadGraph5_aMC@NLO                    *
+#*                                                          *
+#*                *                       *                 *
+#*                  *        * *        *                   *
+#*                    * * * * 5 * * * *                     *
+#*                  *        * *        *                   *
+#*                *                       *                 *
+#*                                                          *
+#*                                                          *
+#*         VERSION 2.2.3                 2015-02-10         *
+#*                                                          *
+#*    The MadGraph5_aMC@NLO Development Team - Find us at   *
+#*    https://server06.fynu.ucl.ac.be/projects/madgraph     *
+#*                                                          *
+#************************************************************
+#*                                                          *
+#*               Command File for MadGraph5_aMC@NLO         *
+#*                                                          *
+#*     run as ./bin/mg5_aMC  filename                       *
+#*                                                          *
+#************************************************************
+set group_subprocesses Auto
+set ignore_six_quark_processes False
+set loop_optimized_output True
+set complex_mass_scheme False
+import model sm
+define p = g u c d s u~ c~ d~ s~
+define j = g u c d s u~ c~ d~ s~
+define l+ = e+ mu+
+define l- = e- mu-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+import model Bprime
+define p = p b b~
+define j = j b b~
+define l+ = l+ ta+
+define l- = l- ta-
+generate p p > t w- QAD=2 QCD=0 QED=0 , t > b j j , w- > l- vl~
+add process p p > t w- QAD=2 QCD=0 QED=0 , t > b l+ vl , w- > j j
+add process p p > t~ w+ QAD=2 QED=0 QCD=0 , t~ > b~ j j , w+ > l+ vl
+add process p p > t~ w+ QAD=2 QCD=0 QED=0 , t~ > b~ l- vl~ , w+ > j j
+add process p p > t w- QAD=2 QED=0 QCD=0 , t > b l+ vl , w- > l- vl~
+add process p p > t~ w+ QAD=2 QED=0 QCD=0 , t~ > b~ l- vl~ , w+ > l+ v\
+l
+add process p p > t w- QAD=2 QED=0 QED=0 , t > b j j , w- > j j
+add process p p > t~ w+ QAD=2 QED=0 QCD=0 , t~ > b~ j j , w+ > j j
+output  bstar_tW_Left_M1200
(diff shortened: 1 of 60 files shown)


bin/MadGraph5_aMCatNLO/cards/production/13TeV/bstar_tW_Left_M1200/bstar_tW_Left_M1200_proc_card.dat
```diff
@@ -0,0 +1,48 @@
...
+define j = j b b~
+define l+ = l+ ta+
+define l- = l- ta-
+generate p p > t w- QAD=2 QCD=0 QED=0 , t > b j j , w- > l- vl~
+add process p p > t w- QAD=2 QCD=0 QED=0 , t > b l+ vl , w- > j j
+add process p p > t~ w+ QAD=2 QED=0 QCD=0 , t~ > b~ j j , w+ > l+ vl
+add process p p > t~ w+ QAD=2 QCD=0 QED=0 , t~ > b~ l- vl~ , w+ > j j
+add process p p > t w- QAD=2 QED=0 QCD=0 , t > b l+ vl , w- > l- vl~
+add process p p > t~ w+ QAD=2 QED=0 QCD=0 , t~ > b~ l- vl~ , w+ > l+ v\
+l
```
bendavid: this looks like a spurious line which should be removed

bendavid: (also for the other cards)

bendavid:
So unfortunately the b-mass cannot be set to zero as part of the customize cards, but really needs to be zero in the model if you want to use 5fs.  (This is because setting it in the customize cards will not correctly propagate to the diagram generation step which occurs earlier)

If you point me to the location of the Bprime model on afs I can suggest the easiest way to fix it.

smshaheen:
Dear Josh Thanks for your comments. The model is available at 
/afs/cern.ch/cms/generators/Bprime.tar.gz
And for the comment of "spurious line" it could not follow you properly. Can you please mention the line number in the pro_card.dat ?

bendavid:
it's this line
https://github.com/cms-sw/genproductions/pull/548#discussion-diff-35524855R45
The line which contains just "l"

smshaheen:
OK sure I will change the mentioned line.

smshaheen:
And instead of setting the b mass to be zero in param_card.dat (which does not seem to be an elegant way), we  should go with the predefined setting for the mass of b quark?

bendavid:
No, setting it in the param card won't work either (it has the same problem as the customise cards).  It must be set in the model itself (or in the "restrict" file inside the model)

bendavid:
(or you could just change the cards to use 4fs, my original comment was only that things were not consistently 4fs or 5fs, so using 4fs settings consistently everywhere would also be fine, and this would not require changing the model)

bendavid:
This would mean removing the b's from both the p and j definitions (ie removing both
define p = p b b~
define j = j b b~
)

maxjetflavor -> 4, and pdf's changed to 4fs versions.

smshaheen:
Using 4f is not going to change any physical meaning of the process right? Is it so then I can retreat back to 4f scheme.

bendavid:
Right, it just means that b quarks are not included in the pdf definition.  (So you will miss contributions with b-quarks in the initial state, which are normally small)

bendavid:
But sorry, looking at your cards again, there is something which needs to be clarified:
"generate p p > t w- QAD=2 QCD=0 QED=0 , t > b j j , w- > l- vl~"

For SM production, this is actually a special case where the b-initial state is important (or else for 4fs you have to change the process definition so that the additional outgoing b is explicit)

What is the BSM production mechanism here?  (Can you point to a Feynman diagram?)

smshaheen:
Right. Attached is the Feynman diagram for the process 
bstar

bendavid:
Ok, and the flavour structure is completely/mostly diagonal?  Ie there is only a bb*g vertex and not eg a db*g vertex?

In that case you either need to use 5fs, or (probably better) use the 4fs, making the b-quarks from gluon splitting explicit

so instead of 
"generate p p > t w- QAD=2 QCD=0 QED=0 , t > b j j , w- > l- vl~"

you would have
"generate p p > t w- b~ QAD=2 QCD=0 QED=0 , t > b j j , w- > l- vl~"

(so the initial state becomes g g with one g splitting into b b~)

smshaheen:
As for as I remember, we only have bbg vertex.

smshaheen:
I think it would be better if we can go for 5f scheme. Hopefully we can find a way to fix the issue of b quark to be massless

bendavid:
Why do you say that?  4f scheme is normally preferred in eg the top group for these kinds of processes, because this tends to give a better description of the kinematics of the outgoing b-jet (which otherwise is still present, but comes from the parton shower)

smshaheen:
I am afraid I do not have any strong argument. So what would you suggest for the current case?

bendavid:
I would suggest to change to the 4f scheme in the run and proc cards and to change the process definitions to have the extra outgoing b quarks.
(then you can remove the line setting the b massless from the customize cards and no further modification of the model is necessary)

smshaheen:
OK. I will try to implement the changes as soon as I can. Thanks :-)

smshaheen:
Dear Josh,
I tried to switch to 4f scheme and tried to generate the process as suggested by you. But I am getting an error as follows.
MG5_aMC>generate p p > t w- b~ QAD=2 QCD=0 QED=0 , t > b j j , w- > l- vl~
INFO: Trying process: g g > t w- b~ QAD=2 QED=0 QCD=0  
INFO: Trying process: u u~ > t w- b~ QAD=2 QED=0 QCD=0  
INFO: Trying process: u c~ > t w- b~ QAD=2 QED=0 QCD=0  
INFO: Trying process: c u~ > t w- b~ QAD=2 QED=0 QCD=0  
INFO: Trying process: c c~ > t w- b~ QAD=2 QED=0 QCD=0  
INFO: Trying process: d d~ > t w- b~ QAD=2 QED=0 QCD=0  
INFO: Trying process: d s~ > t w- b~ QAD=2 QED=0 QCD=0  
INFO: Trying process: s d~ > t w- b~ QAD=2 QED=0 QCD=0  
INFO: Trying process: s s~ > t w- b~ QAD=2 QED=0 QCD=0  
Command "generate p p > t w- b~ QAD=2 QCD=0 QED=0 , t > b j j , w- > l- vl~" interrupted with error:
InvalidCmd : No amplitudes generated from process Process: g/u/c/d/s/u~/c~/d~/s~ g/u/c/d/s/u~/c~/d~/s~ > t w- b~ QAD=2 QCD=0 QED=0
	  Decay: t > b g/u/c/d/s/u~/c~/d~/s~ g/u/c/d/s/u~/c~/d~/s~
	  Decay: w- > e-/mu-/ta- ve~/vm~/vt~. Please enter a valid process

Am I missing something over here?

bendavid:
Yes, sorry the process is not QCD=0 in that case (because the gluon->b b~ splitting is a QCD vertex), so probably you need QCD=1

smshaheen:
Thanks for your kind reply one more thing 
p p > t ~w+ b QAD=2 QCD=1 QED=0 , t > b j j , w- > l- vl~ would be the case when we are having t~ w+ or b quark would stay fix during the whole generation ?

bendavid:
Yes the b quark also changes signs from b~ to b.

Btw, one other thing I noticed now.
", t > b j j" might be less efficient than "t > w+ b, w+ > j j"  (but should give very similar results)

smshaheen:
I apologize for my ignorance, in what context I should take this comment?
Btw, one other thing I noticed now.
", t > b j j" might be less efficient than "t > w+ b, w+ > j j" (but should give very similar results)

bendavid:
For the decay part of the process definition (the decay of the top)

smshaheen:
Dear Johs,
                               I tried to change to 4f scheme as you instructed but we are having a very low cross section aprroximatly (two times lower ). I tried to make sure by checking all the settings multiple times but still the cross sections is lower. However if we use 5f scheme (using b quark mass to be zero in parameter card) the cross section is quite reasonable. What is your suggestion?
Cheers,
Sarmad Masood Shaheen

smshaheen:
Dear Josh,
I tried to change to 4f scheme as you instructed but we are having a very low cross section approximately (two times lower ). I tried to make sure by checking all the settings multiple times but still the cross sections is lower. However if we use 5f scheme (using b quark mass to be zero in parameter card) the cross section is quite reasonable. What is your suggestion?
Cheers,
Sarmad Masood Shaheen

bendavid:
How do you know which cross section is more correct?

smshaheen:
In the following paper 
http://arxiv.org/pdf/1207.5179v1.pdf
at page 8 we have graph which shows the cross section for different mass point at different LHC CM energy. I am trying to compare the cross section we got from MadGraph with the one showed in the plot.

smshaheen:
Dear Josh,
                      What do you suggest how we should proceed forward? Should I upload some parton level distributions to compare the both schemes? Please do let me know accordingly.

smshaheen:
Dear Josh,
                               Should I upload cards to make sure that what I am using for 4f scheme is perfectly fine? Please do let me know accordingly.

smshaheen:
Dear Josh,
                   Hope you would be doing good. Is there any news or comment regarding to the this pull request?
Cheers,
Sarmad

bendavid:
So ok, most likely the difference in the cross section is within the uncertainty (which can be huge at LO)

If you prefer to reproduce the results from the paper, ok to do this in 5fs, but in that case you'll have to produce a modified version of the model tarball with an additional restrict file setting the b mass to zero.
Can you point me to the tarball for the existing model file so I can comment further?

smshaheen:
Dear Josh Thanks for your comments. The model is available at
/afs/cern.ch/cms/generators/Bprime.tar.gz

Thanks!

smshaheen:
Dear Josh,
                              Any suggestions regarding to issue of the massless b jets to be defined in the the model ?

Cheers,
Sarmad Masood Shaheen

bendavid:
I've put a modified tarball in
/afs/cern.ch/cms/generators/www/Bprime_no_b_mass.tar.gz

You'll need to add that to the gridpack_generation.sh script. And use the modified model "Bprime_no_b_mass"  (in that case you don't need to set the b mass to 0 in the customize cards)

bendavid:
(You may want to anyways produce one or two mass points in both 4fs and 5fs for comparison purposes, since it could make a large difference for this process)

smshaheen:
Dear Josh, 
                         Thanks for following up this pull request. I will surely let you know about the investigations related to comparison of the schemes. But one thing to make sure, for 4f scheme do I need to use the previous version of the model?

bendavid:
Yes, for 4f you still need to use the previous version of the model.

smshaheen:
great thanks I will surely update you with my outcomes. 
Thanks!

smshaheen:
Dear Josh,
                            I have checked and found that still found that the results with the 5f scheme are more convincing than 4f scheme. Since we have already implemented all the things with the 5f scheme, so I would clean the repository and will let you know so that the pull request can me merged if all the changes are consistent to what we had discussed.

smshaheen:
Dear Josh,
                         I have made certain changing in the cards and hope it would be able to merge now. 
The model after setting the b quark to be massless is available at 

/afs/cern.ch/user/s/sshaheen/public/Bprime.tar.gz .

Please do let me know if there is any other unnecessary things within the cards(Which I hope should not be :-) ) 
Cheers,
Sarmad Masood Shaheen.

smshaheen:
Dear Josh,
                          Any comments for the latest commits on the current pull?

Please do let me know. 

Cheers,

smshaheen:
Dear Josh,
                             I could not hear you from a long time. Please do let me know accordingly.


Cheers,

smshaheen:
Dear Josh,
                             Please make a comment on latest commits. 

Cheers,
Sarmad

bendavid:
(Looks ok now and has been merged)

smshaheen:
Dear Josh, 
                     That is good. I can proceed with the grid-pack production then. 
Thanks!

Cheers,
Sarmad Masood Shaheen.

PR #553, opened 2015-07-24
Cards for EXO-RunIIWinter15pLHE RPV Stop

davidsheffield:
from @alefisico

The analyzer, @alefisico, tried producing these with gridpacks. It worked without any extra jets, but failed when producing the two extra jets. I believe there was a problem with color connection in the two extra jets because of the RPV model. However, he can produce LHE files directly in madgraph without this issue for use in the pLHE campaign.


bendavid:
Sorry can we get some logs for the gridpack production failure?  I do not think falling back to plhe is necessarily the right thing to do (and may be masking other problems)

alefisico:
Hi @bendavid 

well no, the problem was not with the color connection, the problem is just that when I add two extra partons madgraph crashes. A good description of this problem I presented it in a multijets meeting, here (on slide 17):
https://indico.cern.ch/event/405433/contribution/2/attachments/1123910/1603869/RPVStopUpdate_Multijet_20150702.pdf
Unfortunately I think I erased already the log files but the main issue you can check it in the links that I included when I contacted the madgraph authors. (just in case I am trying to check again if I keep any of those log files)
cheers,

bin/MadGraph5_aMCatNLO/cards/production/13TeV/RPVStop/HadRPVStop1000_UDD312_13TeV/HadRPVStop1000_UDD312_13TeV_run_card.dat
```diff
@@ -0,0 +1,282 @@
...
+ 1        = scalefact        ! scale factor for event-by-event scales
+#*********************************************************************
+# Matching - Warning! ickkw > 1 is still beta
+#*********************************************************************
+ 1        = ickkw            ! 0 no matching, 1 MLM, 2 CKKW matching
+ 1        = highestmult      ! for ickkw=2, highest mult group
+ 1        = ktscheme         ! for ickkw=1, 1 Durham kT, 2 Pythia pTE
+ 1        = alpsfact         ! scale factor for QCD emission vx
+ F        = chcluster        ! cluster only according to channel diag
+ T        = pdfwgt           ! for ickkw=1, perform pdf reweighting
```
bendavid: pdfwgt should be F

bin/MadGraph5_aMCatNLO/cards/production/13TeV/RPVStop/HadRPVStop1000_UDD312_13TeV/HadRPVStop1000_UDD312_13TeV_run_card.dat
```diff
@@ -0,0 +1,282 @@
...
+     6500     = ebeam2  ! beam 2 total energy in GeV
+#*********************************************************************
+# Beam polarization from -100 (left-handed) to 100 (right-handed)    *
+#*********************************************************************
+        0     = polbeam1 ! beam polarization for beam 1
+        0     = polbeam2 ! beam polarization for beam 2
+#*********************************************************************
+# PDF CHOICE: this automatically fixes also alpha_s and its evol.    *
+#*********************************************************************
+ 'nn23lo1'    = pdlabel     ! PDF set
```
bendavid: pdf selection is not consistent with the rest of the production.  Please look at the examples.

bendavid: The current model is 4fs so if you stay like that you should use 4fs pdfs.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/RPVStop/HadRPVStop1000_UDD312_13TeV/HadRPVStop1000_UDD312_13TeV_run_card.dat
```diff
@@ -0,0 +1,282 @@
...
+#*********************************************************************
+# KT DURHAM CUT                                                      *
+#*********************************************************************
+ -1    =  ktdurham        
+ 0.4  =  dparameter 
+#*********************************************************************
+# maximal pdg code for quark to be considered as a light jet         *
+# (otherwise b cuts are applied)                                     *
+#*********************************************************************
+ 5 = maxjetflavor    ! Maximum jet pdg code
```
bendavid: if staying with 4fs this should be 4

bin/MadGraph5_aMCatNLO/cards/production/13TeV/RPVStop/HadRPVStop1000_UDD312_13TeV/HadRPVStop1000_UDD312_13TeV_run_card.dat
```diff
@@ -0,0 +1,282 @@
...
+#*********************************************************************
+# Jet measure cuts                                                   *
+#*********************************************************************
+ 30   = xqcut   ! minimum kt jet measure between partons
```
bendavid: 30 is probably much too high for xqcut.  You typically would use something like 10 so that you have more freedom in choosing qcut after.
(Have you already done the jet matching study and found a qcut value that works with xqcut=30?)

bin/MadGraph5_aMCatNLO/cards/production/13TeV/RPVStop/HadRPVStop1000_UDD312_13TeV/HadRPVStop1000_UDD312_13TeV_proc_card.dat
```diff
@@ -0,0 +1,37 @@
...
+set group_subprocesses Auto
+import model sm
+define p = g u c d s u~ c~ d~ s~ b b~
+define j = g u c d s u~ c~ d~ s~ b b~
+define l+ = e+ mu+
+define l- = e- mu-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+import model RPVMSSM_UFO -modelname
+define j = g u c d s u~ c~ d~ s~ b b~
```
bendavid: if staying with 4fs then neither the jet definition nor the proton definition should include b's.

bendavid:
So a general comment:
If there is a not understood "physics" error rather than a purely technical issue with the gridpack production, it is very likely that this issue is somehow present even when producing lhe's interactively.  (But the manifestation may be in the physics of the output rather than an outright error)

Then I have some specific comments on the cards as you can see.  A general issue is that 4fs vs 5fs is not consistently handled (the current model has massive b quarks so then you would be obliged to use 4fs consistently, which means additional partons will not include b's).  Alternatively you could modify/switch to a version of the model with massless b's and then move consistently to 5 flavour scheme. (Your current cards include the b's in the jet definition but not the proton, which is wrong in either case, and which in fact I suspect could contribute to some of the errors you have)

bin/MadGraph5_aMCatNLO/cards/production/13TeV/RPVStop/HadRPVStop1000_UDD312_13TeV/HadRPVStop1000_UDD312_13TeV_run_card.dat
alefisico: yes, I can provide some plots, but the qvalue that give us smooth distributions is xqcut=50. But I can generate with 10 the lhe files and then in pythia change this parameter to 50.

bendavid: You mean qcut=50 with xqcut=30?

alefisico:
Hi @bendavid thanks for your comments. 

I am confused about the last comment. You said that the jets definition include the proton but not the proton, but in line 25 of the proc_cards I am adding b's to the proton definition. 
In any case, I am not sure if 4fs or 5fs will be better for us because we have two "versions" of the analysis: when the RPV stops decay into a two light quarks and when decay into a light jet and a b quark. My naive idea will be to use 4fs for the first case while the 5fs for the second, am I right?

again thanks for your comments.

bendavid:
Why would you want to use different flavour schemes for the two cases?  The heavy resonance decays are decoupled from the extra jet production.

For the proton and jet definitions, the multiparticle definitions are typically reset when doing "import model" so the definition of p and j needs to happen AFTER the bsm model is imported.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/RPVStop/HadRPVStop1000_UDD312_13TeV/HadRPVStop1000_UDD312_13TeV_run_card.dat
alefisico: xqcut=10 (as you suggested) with qcut=50

alefisico:
So you think we should use only 4fs? Ok, I can change the cards according on what you suggested for 4fs.
And good to know about the import model. thanks.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/RPVStop/HadRPVStop1000_UDD312_13TeV/HadRPVStop1000_UDD312_13TeV_run_card.dat
bendavid: Ok, although if you already got smooth distributions with qcut=50 and xqcut=30 then that is probably ok as is.

bendavid:
Is the model already available in the generator afs area btw?

And for 4fs vs 5fs, in general 4fs is preferred only for processes which are dominated by subprocesses with b-quarks in the initial state.

Otherwise 5fs is preferred because this allows the additional jets to be inclusive in b multiplicity.

davidsheffield:
The model is in `/afs/cern.ch/cms/generators/www/RPVMSSM_af1_ufo.tgz`. It's from https://feynrules.irmp.ucl.ac.be/wiki/RPVMSSM.

bendavid:
Can you check with one point if gridpack production works after all the fixes?

alefisico:
It didn't work. The log file is here:
/afs/cern.ch/user/a/algomez/public/RPVSt100_UDD312_ISR2j_LO.log

bendavid:
And does it work producing an lhe file directly?

alefisico:
If I modified the current gridpack_generation.sh code to produce the lhe only I got this error:
/afs/cern.ch/user/a/algomez/public/RPVSt100_UDD312_ISR2j_LO_OnlyMG.log

bendavid:
Ok then I suggest to look at the logs inside one of the subprocess directories which failed
eg
P2_gg_su1su1xqq_su1_qq_su1x_qq/G1/

alefisico:
ok let me explain you the entire process because maybe I confused you a bit. 
I want to generate p p > stop stop + 0,1j,2j, stop > j j. I had several problems when I tried to create the gridpack. If I tried to generate the process with 0 or 1j, the gridpack generation works fine, but once I added the 2j I am having all the problems that I described in my slides.
Now, I spend several months trying to produce these samples and the only way that I can produce them, is with the SUSY prescription(https://github.com/CMS-SUS-XPAG/GenLHEfiles/tree/master/Run2Mechanism) 
where basically you produce an lhe file for the process p p > stop stop + 0,1,2j ALONE, modified the header of the lhe file to include the decay of the stops and then just make pythia8 handle the decay. The cards that I uploaded are for this case.

Now, after you gave me some suggestions about the cards, I though that you asked me to try to generate the entire process again (including the decay). Then, the 2 log files that I gave you are for these cases, but unfortunately as you see I still have issues. The error of the last log file (RPVSt100_UDD312_ISR2j_LO_OnlyMG) is exactly the error that I had before and that I asked the MG authors:
https://answers.launchpad.net/mg5amcnlo/+question/265473
https://answers.launchpad.net/mg5amcnlo/+question/265751
for that they created an special version of MG trying to fix this issue but it does not work. If you run this process P2_gg_su1su1xqq_su1_qq_su1x_qq  ALONE,  it works, but if you run the whole chain it does not.

Finally, I talked to @davidsheffield and I think a "version" of the SUSY prescription that may work is to produce the gridpacks for the p p  > stop stop + 0,1,2j and then just add these modifications as SLHA tables in the pythia8 cards. I think is doing the same that the SUSY people does, so I will test it asap.

Hopefully now it is more clear, or if not please let me know.

bendavid:
Ok thanks, it's clear now.

The solution proposed by @davidsheffield is by far preferred and should work.  So hopefully you can produce gridpacks with the updated versions of the cards after incorporating my comments.

(and I understand now that the real issues you encountered had nothing to do with gridpack vs not, but rather with including the decay vs not.  This makes much more sense)

In the future, to avoid wasting time, I suggest that you discuss these kinds of issues with MC contacts + generator group much earlier.
(and in particular the recommended procedure is to make pull requests for cards BEFORE producing large numbers of gridpacks or lhe files, so that suggestions/fixes can be incorporated without wasting time)

alefisico:
I did it :)
https://hypernews.cern.ch/HyperNews/CMS/get/generators/2559.html
and I can point you to a long chain of emails that I exchanged with Steve Mrenna. But ok, sure I will test all of these and let you know.

davidsheffield:
@alefisico provided updated param cards for the original set and cards for a second RPV coupling. I also added the RPV model to `gridpack_generation.sh`.

bendavid:
Sorry this one got a bit lost.
Is it planned to do these with gridpacks finally?
In that case the param cards should be replaced with customize cards as usual.

davidsheffield:
Gridpacks are planned. Ale sent me some new cards that I need to add (I've been busy with the BH and DM requests but I'll be able to add the updates to this PR tomorrow). I'm not sure if Ale had the same trouble but the MC for my own analysis, which uses the same RPV model file, I couldn't use a customize card to set the mass of the SUSY particles. It just would not recognize them. I could change the mass of SM particles fine with the customize card. What I ended up doing for my tests and what I believe Ale also did was modify the gridpack creation script to replace the param card with one we give.

bendavid:
Did you try to set the mass by name or by pdg id?

bendavid:
(both should work, but if one does not, then worth trying the other)

davidsheffield:
I used pdgid. I'm not sure if I tried the name. Ale, is that something you can test?

alefisico:
Hi @bendavid, 
yes, we will definitively generate this samples. I have been iterating with @davidsheffield about it, maybe I should add you in cc. 
Yes, I modify the gridpack generation script to add the param card not only for the mass of the stop but for the couplings and the decays of the stops. At the beginning I tried with the custom cards and I didnt have any problem. I just switch just to be sure that the parameters that I was using were the right onces. 

I can make the PR if David is busy, I have everything in place just to make the PR for these cards. 

Now, the only issue is about the number of events that we want to generate for the low mass. Since we are doing something special, I am not sure how to get the correct efficiency factor to know the number of events that we need to request. (I thought that you were in that email, but I will cc you now).

bendavid:
Hi,
We really really really prefer to have customize cards rather than param cards, because it is much easier to track what is changed from the defaults. (and as such we do not want to commit such a modification to the gridpack script)

When you generate a gridpack using the customize cards, the final param card is present in the cards directory inside the final tarball, and can be inspected/diff'd to be absolutely sure about what was used.

Will answer in the mail about the efficiencies.

alefisico:
Hi @bendavid 

at the end @davidsheffield update the cards with the generator level cuts. As we discussed, I am adding the param_cards instead of the customizecards, because this process needs to include the decay of the particles and this cannot be done with customizecards alone.

let me know if there is something else that I need to do.

bendavid:
Ok, but now it looks like there is some inconsistency about the flavour scheme.  In particular proc and run cards are set up for 5fs generation, but the b-quarks are massive in the param card.

alefisico:
Maybe it is because we have two models here: one it is when the stops decay into to light quarks (the files reflect this coupling with UDD312) and the case when the stops decay into a light quark and a b quark (UDD323). But still, I think I am a bit confused, in the definition of extra partons we are including b-quarks for both models, shouldn't the bquarks been massive?

bendavid:
No it's the opposite.  There are two possibilities

4 flavour scheme:  massive b-quark, b not included in proton or jet definition (extra jets do not include b quarks)

5 flavour scheme:  massless b-quark, b included in proton and jet definition (extra jets do include b's)

In addition to the p and j definition in the proc card, the pdfs plus a few other settings in the run card also need to be consistent with the chosen flavour scheme.  The reason why 5 flavour scheme requires/implies massless b-quarks is that in the 5 flavour scheme, b's are included in the proton parton distribution functions (and therefore need to be treated as massless to be on the same footing as the other quarks)

alefisico:
ok great, that makes sense. Thank you for the explanation.
So I guess that what we need is the 5 flavour scheme, so I have to make the b quarks in the param card massless? something else?

bendavid:
Changing masses or couplings between 0 and non-zero with the param card only is not safe (because to be fully properly taken into account it needs to be known at process generation time whether the b is massive or massless)

The right way is to add a restrict.dat file to the model (make a new tarball and change the name of the model also to avoid name conflicts)

More details in
https://answers.launchpad.net/mg5amcnlo/+faq/2312

bendavid:
(or if you have trouble with that can just produce a modified tarball for the model with a new name and the b mass changed directly in the parameters.py)

bendavid:
(superseded by https://github.com/cms-sw/genproductions/pull/663)

PR #557, opened 2015-07-27
Alpgen cards from Emrah

dnash86:
Cards from Emrah for Majorana Neutrino to EE, MuMu, EMu, MuE LHE files.

### DIFF
--- a/bin/Alpgen/cards/production/13TeV/MajoranaNeutrinoToLL/MuMu/input_MuMu_M_100
+++ b/bin/Alpgen/cards/production/13TeV/MajoranaNeutrinoToLL/MuMu/input_MuMu_M_100
@@ -1,0 +1,30 @@
+1           ! imode
+hvyn        ! label for files
+0 ! start with: 0=new grid, 1=previous warmup grid, 2=previous generation grid
+10000   4  ! Nevents/iteration,  N(warm-up iterations)
+10000000    ! Nevents generated after warm-up - (it depends on how many events wants to be generated for each mass)
+*** The above 5 lines provide mandatory inputs for all processes
+*** (Comment lines are introduced by the three asteriscs)
+*** The lines below modify existing defaults for the hard process under study
+*** For a complete list of accessible parameters and their values,
+*** input 'print 1' (to display on the screen) or 'print 2' to write to file
+iqopt 1
+qfac  1            
+ih2   1 
+ebeam 6500.
+ndns   7 ! ndns is the PDF choice code in ALPGEN documentation
+ptlmin 0.
+etalmax 10.
+drlmin  0.
+v21  0.000d0      !  v21  0.0054d0
+v22  1.000d0     !  v22  0.0096.d0
+v23  0.00d0       !  v23  0.016.d0
+ilnv 1                 
+il1  2             
+il2  2              
+ima  1               
+***
+***
+indec 1
+idf   1
+mn   100.d0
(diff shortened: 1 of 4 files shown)


bendavid:
Can I ask what pdf was used here?

dnash86:
The pdf was: CTEQ6M.

bendavid:
Ok, but that is an NLO pdf (and quite obsolete).  This also implies an inappropriate alpha_s value is used (~0.118 vs 0.130).  This is also inconsistent with most of the rest of the production.

Can alpgen be easily interfaced with lhapdf?  Do the analyzers care about doing this with pdfs consistent with the rest of the production?

dnash86:
I definitely do not know about how easily alpgen can be interfaced with lhapdf.  I will refer the analyzers to your comment and see what they say.

bendavid:
Even if it can't be interfaced to lhapdf, even CTEQ6L1 would make much more sense than CTEQ6M

dnash86:
Just so you know, this question is not being ignored, the main analyzer I was in contact with said the group is discussing it.

etiras:
We used NLO pdf for Run I analysis with good reason and now just to make sure again we asked Prof. Tao Han, who is a theorist at the University of Pittsburgh and an expert on Majorana Neutrinos, provided estimates for heavy neutrino production at the LHC. Here is our question and his answer:  

Q: 

For the new Run II analysis we have the choice of using 6TEQ6M or CTEQ6L1 pdf for our Leading Order heavy neutrino MC signal samples. Which one do you recommend? We used CTEQ6M before and we applied a kfactor of 1.34.
 
A: 
Although “theoretically”, the LO cross section should match the LO pdf, I strongly recommend you to still use the NLO pdf’s, namely CTEQ6M. I could tell you pros and cons, but you may not need to know these “theoretical arguments”.

So, we would like to continue with NLO pdf, CTEQ6M.

dnash86:
@bendavid Is this answer satisfactory, or is a different pdf set required?  If it is satisfactory, we should be ready to move forward with the request.

bendavid:
I would actually like to know what the arguments are.  We use LO pdfs for ~every single other LO sample in CMS.

dnash86:
@bendavid The analyzers are preparing a comparison study to understand the theoretical differences are.  

In the meantime, would it be possible to do a "partial approval" and proceed with generation of only a few (5-6) MuMu channel points with the NLO pdf for them to start work on the analysis with them?  (With the provision that they may change to the LO pdf for the full production.)

bendavid:
Can we produce samples with both pdf's for a few points rather?  Do we already have an answer as to whether alpgen can be easily interfaced with lhapdf?

etiras:
After discussions with theorists and comparing cross sections, we decided to go with LO PDF, which is CTEQ6L1. We also checked with ATLAS, they are using CTEQ6L1 as well. 
Alpgen is not easily interfaced with lhapdf. We searched for it but couldn't find too much information about it. 
We have generated LHE files with CTEQ6L1, we'll send them soon. Thanks-

dnash86:
@bendavid Just a note, there will be 4 mass points only for this cteq6m production:

MuMu, with M=50,100,500,1000.

I can put "cteq6m" in the dataset names to make it clear.  If that is okay, should I leave this PR for the entire set of cards for ease, or just the 4 that I will be entering corresponding entries in McM?

bendavid:
Let's put just the cards we're going to use in production.

dnash86:
Okay, I'll update remove all but those 4 files from this PR.

dnash86:
I've reduced this PR to just the 4 masses needed, is it okay now?

PR #558, opened 2015-07-27
gg high mass HWW with JHU, 450, 650, 1000GeV

rebecacern:


rebecacern:
I added the VBF as well

rebecacern:
Can you approve these cards? I would like to prepare the McM requests and this is the first step...

PR #559, opened 2015-07-27
[Madgraph5_aMC@NLO] Dihiggs group: RS->hh cards

syuvivida:
Add the following channels. Each mass point has 3 cards (run card, proc card, and customize cards)
RSGraviton_hh_hVVhbb_fullLep  --> 24 mass points
RSGraviton_hh_hbbhbb   --> 24 mass points           
RSGraviton_hh_hVVhbb_inclusive --> 24 mass points
RSGraviton_hh_htatahbb --> 28 mass points.

Above 1 TeV, the request will be done via EXO group. 
Each mass point requires 50K events.
M=(1000,1200,1400,1600,1800,2000,2500,3000,3500,4000,4500)

Below 1 TeV, the request will be done via the Higgs group.
260 GeV: 300K
270 GeV: 300K
300 GeV: 300K
350 GeV: 300K
400 GeV: 300K
450 GeV: 100K
500 GeV: 100K
550 GeV: 100K
600 GeV: 100K
650 GeV: 100K
700 GeV: 100K
800 GeV: 100K
900 GeV: 100K

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson/Spin-2/RSGraviton_hh_hVVhbb_fullLep/RSGraviton_hh_hVVhbb_fullLep_narrow_M1000/RSGraviton_hh_hVVhbb_fullLep_narrow_M1000_customizecards.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson/Spin-2/RSGraviton_hh_hVVhbb_fullLep/RSGraviton_hh_hVVhbb_fullLep_narrow_M1000/RSGraviton_hh_hVVhbb_fullLep_narrow_M1000_customizecards.dat
@@ -1,0 +1,13 @@
+set param_card mass 39 1000
+set param_card decay 39 0.001
+set param_card yukawa 4 0
+set param_card yukawa 5 4.8
+set param_card yukawa 6 172.5
+set param_card mass 4 0
+set param_card mass 5 4.8
+set param_card mass 6 172.5
+set param_card mass 23 91.1876
+set param_card mass 24 80.385
+set param_card decay 6 1.491500
+set param_card decay 23 2.4952
+set param_card decay 24 2.085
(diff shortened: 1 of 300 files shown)


stosi:
Hi, the cards look OK to me

syuvivida:
@covarell @bendavid could you merge the pull request? We would like to produce gridpacks this week. Many thanks!!

bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson/Spin-2/RSGraviton_hh_hVVhbb_fullLep/RSGraviton_hh_hVVhbb_fullLep_narrow_M1000/RSGraviton_hh_hVVhbb_fullLep_narrow_M1000_customizecards.dat
```diff
@@ -0,0 +1,13 @@
+set param_card mass 39 1000
+set param_card decay 39 0.001
+set param_card yukawa 4 0
+set param_card yukawa 5 4.8
+set param_card yukawa 6 172.5
+set param_card mass 4 0
```
bendavid: Is this something which has already been done for other requests?  In principle one should not toggle masses or couplings between 0 and non-zero using the customizecards, since this is not taken into account properly at the diagram generation level.

syuvivida:
Hi Josh,
  Yes, this is also done for the RS->VV gridpack production. The produce LHE files reflect the changes we made in customizecards.

Are you referring to the masses and yukawa couplings of charm, bottom and top quarks?
I thought this was legal as I saw the same settings in this card:
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/tt0123j_2l_5f_ckm_LO_MLM/tt0123j_2l_5f_ckm_LO_MLM_customizecards.dat

bendavid:
Yes it's perfectly fine to change masses or yukawas in the cutomizecards.  It's changing them between 0 and non-zero which could cause problems.

bendavid:
So in particular I'm worried about 
"set param_card mass 4 0"

bendavid:
Maybe it's ok changing it from nonzero -> zero, just technically less efficient.  Changing it from zero->nonzero is likely to cause real physics problems.

syuvivida:
So would it be better if we change the parameters.py (where the default c quark mass is set) in the model?

bendavid:
The proper way to do it is by including a "restrict" file in the model, so you can do
modelname-no_c_mass as for the various sm models.

If you've already produced samples like this and it looks ok, then it might be fine as is (because it's non-zero -> zero as said)

syuvivida:
Just to be safe, I will remove these lines that set c quark mass to zero.

bendavid:
Ok, but then you need to modify the model.  You definitely don't want to generate these with massive c quarks.  (Why does the model even have this?)

syuvivida:
oops, ok, No change then. I will keep the c quark mass lines. Is there a quick a way to check if the models were built with massive c quarks? Which files should I check? not particles.py or parameters.py?

bendavid:
No no, if there are no restrict_ files then you need to check the parameters.py
If the c quark is already massless in the model then the extra lines from the customizecards can of course be removed.

syuvivida:
Hi Josh, ok, I will add a new RS model then.

syuvivida:
Move to pull request https://github.com/cms-sw/genproductions/pull/565

PR #561, opened 2015-07-28
use diagonal ckm

perrozzi:


perrozzi:
could you please point me to the correct 4f pythia fragment?

bendavid:
You just need to change JetMatching:nQmatch to 4

PR #563, opened 2015-07-29
POWHEG Gridpack generation script: Two stage procedure for ttH

mharrend:
After talking with the authors of the ttH package in POWHEG Box V2 and with @eiko I have extended the up to now used POWHEG Gridpack generation script, so that automatically a two stage procedure is used.

@kskovpen

### DIFF
--- a/bin/Powheg/create_powheg_tarball.sh
+++ b/bin/Powheg/create_powheg_tarball.sh
@@ -248,9 +248,85 @@ fi
 cat ${card} | sed -e "s#SEED#${seed}#g" | sed -e "s#NEVENTS#${nevt}#g" > powheg.input
 cat powheg.input
 
-#make sure env variable for pdfsets points to the right place
-export LHAPDF_DATA_PATH=`${myDir}/lhapdf-config --datadir`
-../pwhg_main &> log_${process}_${seed}.txt
+ 
+# Use one stage procedure by default and two stage procedure if ttH process is used
+if [ "$process" != "ttH" ]; then
+	echo "Default one stage procedure is used"
+	#make sure env variable for pdfsets points to the right place
+	export LHAPDF_DATA_PATH=`${myDir}/lhapdf-config --datadir`
+	../pwhg_main &> log_${process}_${seed}.txt
+else
+	# Save powheg input before starting two stage
+	cp powheg.input powheg.input.original
+	echo "Two stage procedure is used for ttH process"
+
+	# Check if required options are set in input card, otherwise set them
+	## use-old-grid must be set to 1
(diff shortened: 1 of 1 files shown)


bendavid:
Are these parameters being changed from step 1 to step 2 unique to the ttH process?

mharrend:
Hello Josh,

in the first draft I had introduced an option to activate the two stage
procedure for all POWHEG processes. Since the two stage procedure is
available only for a small fraction of POWHEG processes Eiko asked me to
activate it only for the ttH process.

Marco A. Harrendorf
KIT - IEKP

syuvivida:
Hi Josh, 
  I was mainly referring to the option "fakevert". I have not seen this option in the example cards of other processes.

bendavid:
@syuvivida can you clarify something.  This pull request touches create_powheg_tarball.sh.

Is this the currently used version of the script, or would one also/instead have to update run_pwg.py or runcmsgrid_powheg_hdampgrid.sh (or any of the other scripts in the bin/powheg area)?

syuvivida:
if I understand correctly, Macro is producing the ttH gridpack in the serial mode (not using run_pwg.py) and the settings here are for optimizing the computing time when producing gridpacks in the serial mode only. But one could also make use of run_pwg.py to produce ttH gridpacks in the parallel mode and enable the fakevert option. the short answer is that this script could be checked in without modifying other scripts.

bendavid:
Ok but do we want to keep updating the two scripts in parallel? (as a more general question)

syuvivida:
No, we shouldn't. It's just that I haven't got a chance to try run_pwg.py on every process and my test framework were set up for create_powheg_tarball.sh. 

Hi Josh,

I am updating create_powheg_tarball.sh now for the next round of tar ball July2015 (for including the updates of POWHEG authors in July). This latest change broke 4 processes and we need additional patches.
Once it's tested and approved by conveners, I will pass the 
change to Yuan Chao. Then, we will test every process with run_pwg.py. If everything is fine, we will remove create_powheg_tarball.sh once and for all. Does that sound OK to you?

syuvivida:
@bendavid, @covarell Marco has tested his script for the ttH process and I also tested his uploaded script on other POWHEG processes we have included so far and the script is OK too (didn't break what we already had). Could you merge the update of Marco? Thanks!! Once you merge this update, I will make another pull request that add additional patches and fixes for July 2015 version of POWHEG-BOX-V2. Thanks!!

bendavid:
Ok will merge it, but am getting that we are diverging between different versions of the script.  Who will add the two step support to the parallel script?  (Same goes for the powheg+madspin support)

The sooner we van converge on one single script the better.

PR #566, opened 2015-07-29
aMC@NLO cards for WV aTGCs 

mmozercern:


bendavid:
The lhe's for these are going to be produced privately I guess rather than with gridpacks? (given the reweighting which is not presently supported in gridpack mode)

bin/MadGraph5_aMCatNLO/cards/production/13TeV/SMP/WW_aTGC_semilep/run_card.dat
```diff
@@ -0,0 +1,280 @@
...
+# Automatic ptj and mjj cuts if xqcut > 0
+# (turn off for VBF and single top processes)
+#**********************************************************
+   T  = auto_ptj_mjj  ! Automatic setting of ptj and mjj
+#**********************************************************
+#                                                                    
+#**********************************
+# BW cutoff (M+/-bwcutoff*Gamma)
+#**********************************
+  5  = bwcutoff      ! (M+/-bwcutoff*Gamma)
```
bendavid: this is a narrower cutoff compared to the default (15) or the ultra high value (15000) being used for eg LO W+jets production to get all the off-shell phase space.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/SMP/WW_aTGC_semilep/run_card.dat
```diff
@@ -0,0 +1,280 @@
...
+ 1        = scalefact        ! scale factor for event-by-event scales
+#*********************************************************************
+# Matching - Warning! ickkw > 1 is still beta
+#*********************************************************************
+ 1        = ickkw            ! 0 no matching, 1 MLM, 2 CKKW matching
+ 1        = highestmult      ! for ickkw=2, highest mult group
+ 1        = ktscheme         ! for ickkw=1, 1 Durham kT, 2 Pythia pTE
+ 1        = alpsfact         ! scale factor for QCD emission vx
+ F        = chcluster        ! cluster only according to channel diag
+ T        = pdfwgt           ! for ickkw=1, perform pdf reweighting
```
bendavid: pdfwgt should be False

bin/MadGraph5_aMCatNLO/cards/production/13TeV/SMP/WW_aTGC_semilep/run_card.dat
```diff
@@ -0,0 +1,280 @@
...
+#                                                                    
+#**********************************
+# BW cutoff (M+/-bwcutoff*Gamma)
+#**********************************
+  5  = bwcutoff      ! (M+/-bwcutoff*Gamma)
+#**********************************************************
+# Apply pt/E/eta/dr/mij cuts on decay products or not
+# (note that etmiss/ptll/ptheavy/ht/sorted cuts always apply)
+#**********************************************************
+   T  = cut_decays    ! Cut decay products
```
bendavid: cut_decays is usually False (since usually we are trying to produce inclusive samples as much as possible)

mmozercern:
On 07/29/2015 05:58 PM, Josh Bendavid wrote:
>
> The lhe's for these are going to be produced privately I guess rather 
> than with gridpacks? (given the reweighting which is not presently 
> supported in gridpack mode)
>
> —
> <https://github.com/cms-sw/genproductions/pull/566#issuecomment-125998334>.
>

Hi Josh

Collecting all your comments in one e-mail:

 > The lhe's for these are going to be produced privately I guess rather 
than with gridpacks? (given the reweighting which is not presently 
supported in gridpack mode)

Yes, but the SMP MC contacts asked me to upload them on github anyway

 > this is a narrower cutoff compared to the default (15) or the ultra 
high value (15000) being used for eg LO W+jets production to get all the 
off-shell phase space.

This is a signal sample for an aTGC search, so the there's not that much 
interest in the tail in this case. I can generate a test sample to see 
the difference, though.

 > pdfwgt should be False

This was suggested in order to get PDF uncertainties included with 
syscalc. Is that false?

 > cut_decays is usually False (since usually we are trying to produce 
inclusive samples as much as possible)

Will check the effect as well.

I'll let you know.

Matthias

bendavid:
Yes, we want cards in github also for privately produced lhe's (both to review them, and so they can be easily accessible later).

For the bwcutoff, I would suggest to at least use the madgraph default of 15.

For the pdfwgt setting, no this is completely orthogonal to the pdf uncertainties.  (The setting which is related to that is "use_syst" which is already correctly set to true).  Pdfwgt controls some interplay between the pdfs and the jet matching, and has so far produced not fully understood kinematic differences which experts will need to look into before we can/should use it.

mmozercern:
On 07/29/2015 06:14 PM, Josh Bendavid wrote:
>
> Yes, we want cards in github also for privately produced lhe's (both 
> to review them, and so they can be easily accessible later).
>
> For the bwcutoff, I would suggest to at least use the madgraph default 
> of 15.
>
> For the pdfwgt setting, no this is completely orthogonal to the pdf 
> uncertainties. (The setting which is related to that is "use_syst" 
> which is already correctly set to true). Pdfwgt controls some 
> interplay between the pdfs and the jet matching, and has so far 
> produced not fully understood kinematic differences which experts will 
> need to look into before we can/should use it.
>

urgh. That's definitely a problem. If I have to redo the lhe files 
because of that I can adjust the other issues as well. I was hoping I 
would be able to get away without redoing the lhe files, but no such luck.

Matthias

bendavid:
Indeed this is why (as with gridpacks) we're suggesting pull requests for cards before producing gridpacks/lhe's...

bendavid:
Actually, sorry to say, but there is a more fundamental problem here.

Pythia8 in CMSSW is not currently correctly supporting jet matching for EWK-type processes.  This is something currently being worked on together with the authors, and I would expect it to converge in the next several weeks, but if you want to already go ahead, then you'll have to do these without jet matching.

mmozercern:
On 07/29/2015 06:35 PM, Josh Bendavid wrote:
>
> Actually, sorry to say, but there is a more fundamental problem here.
>
> Pythia8 in CMSSW is not currently correctly supporting jet matching 
> for EWK-type processes. This is something currently being worked on 
> together with the authors, and I would expect it to converge in the 
> next several weeks, but if you want to already go ahead, then you'll 
> have to do these without jet matching.
>
> —
> <https://github.com/cms-sw/genproductions/pull/566#issuecomment-126009857>.
>

Fair enough. I hadn't heard about that, I roughly followed the old 8TeV 
production that still ran Pythia6.
How exactly do you propose to proceed?
Use the same processes an turn ickkw to 0?
use only the WW+0jet process?
What other parameters would be affected?

Matthias

bendavid:
Pythia6 didn't support this either.  Can you point to the run 1 samples you are referring to?

Yes, you need to have only the WW+0jet process and also set ickkw to 0.  That should be all.  (Then the pythia configuration will clearly be different without matching)

mmozercern:
On 07/29/2015 06:42 PM, Josh Bendavid wrote:
>
> Pythia6 didn't support this either. Can you point to the run 1 samples 
> you are referring to?
>
> Yes, you need to have only the WW+0jet process and also set ickkw to 
> 0. That should be all. (Then the pythia configuration will clearly be 
> different without matching)
>
> —
> <https://github.com/cms-sw/genproductions/pull/566#issuecomment-126013378>.
>

I got pointed to this twiki page for the production:
https://twiki.cern.ch/twiki/bin/viewauth/CMS/WWaTGCMadGraphFullSimulation
 From what I can tell, this did not result in an central production as 
production CMSSW didn't forward the event weights yet, but it seems like 
the samples may have been used as privately produced MC in some analysis.

Matthias

bendavid:
Ok, there's something missing from the instructions, since the jet matching configuration is not specified anywhere, but I don't see how this could have given sensible results.

mmozercern:
On 07/29/2015 07:02 PM, Josh Bendavid wrote:
>
> Ok, there's something missing from the instructions, since the jet 
> matching configuration is not specified anywhere, but I don't see how 
> this could have given sensible results.
>
> —
> <https://github.com/cms-sw/genproductions/pull/566#issuecomment-126019119>.
>

Thanks for spotting this, I'll bring it up in the working-group. There's 
definitely the potential for other people to fall into this trap as well.

Matthias

mmozercern:
On 07/29/2015 07:02 PM, Josh Bendavid wrote:
>
> Ok, there's something missing from the instructions, since the jet 
> matching configuration is not specified anywhere, but I don't see how 
> this could have given sensible results.
>
> —
> <https://github.com/cms-sw/genproductions/pull/566#issuecomment-126019119>.
>

Ok, I redid the configs with your comments up submitted a new pull 
request. Please have another look.
While you're at it, could you check that the PDF I use is the correct one?
New pull request is


  #567



Yours
  Matthias

PR #567, opened 2015-07-29
Update to WV aTGC request with Josh's comments

mmozercern:


bendavid:
So the process definition and pdf's look consistent for a 5fs scheme process, however the model has massive b's.  If you want to stay with 5fs, then the b needs to be set massless (and also yb=0) in the param card.

mmozercern:
Ok, should be consistent 5flavor scheme now.

Thinking more about the MLM machting I'm not so surprised any more that there's trouble: if I read the Madgrpah output correctly, only the initial quark lines are eligible for real emissions, but not the W decay quarks. However, the W decay quarks get showered in Pythia just as well.

bendavid:
No that part is handled properly at least (W decay products are correctly excluded from the matching)

PR #569, opened 2015-07-30
Enabling powheg + madspin

maierbenedikt:
Modified are:

1.) the create_powheg_tarball.sh: if a madspin_card.dat and a proc_card.dat are lying in the directory from which the script is executed, they are copied into the final tarball. A link to how a proc_card should look like is given in the script as well. (The folder does not exist yet, but is part of this pull request.)

2.) the runcmsgrid_powheg.sh: if it detects the madspin_card, it will at the end download MadGraph v2.2.2, splice the proc_card into the header of the cmsgrid_final.lhe and run MadSpin on it.

Open questions:
- The runcmsgrid_powheg.sh does not apply/download the patches for MG that are downloaded in the gridpack_generation.sh. Is this a problem? I understood that the patches were there for correct submission of jobs to the cluster (which might not be relevant here), and for editing some models. It could be that for instance for ttbar in the 5F, the appropriate model would be "loop_sm-ckm_no_b_mass", which is a customized model, isnt it? This would not be there then. Maybe it is easy to apply the patches here too?
- Do we want the madspin option also for runcmsgrid_powheg_hdampgrid.sh and runcmsgrid_powhegjhugen.sh?
- I tested it with 10 events, and the time needed for the creation of the final cmsgrid_final.lhe increases by ~50% compared to the case without the application of madspin, mostly because it takes some time to compile. This 50% should be reduced however once one goes to ~1000 events: The time for compiling stays the same, while decaying one event is quicker than generating+reweighting it. So I guess it is still justifiable.

If you want to see how it works, here is a tarball:

/afs/cern.ch/work/b/bmaier/public/ST_tch_antitop_4f_ckm_13TeV-powhegV2_tarball.tar.gz

Just do: 

./runcmsgrid.sh 10 123 1


bendavid:
Indeed I am worried here also about maintaining these changes in one/multiple versions of the scripts @syuvivida 

@maierbenedikt the patches must be applied, since in many cases those have included bugfixes for madspin.  As said, I would strongly prefer that the scripts reuse the madgraph scripts as much as possible (maybe even calling them directly if possible in "interactive" mode and copying the needed output directories around)

One other thing is that I think calling the files containing headers to be spliced "proc_card.dat" is a bit misleading since more is contained there than just the proc card.

maierbenedikt:
Letting the runcmsgrid_powheg.sh download MadGraph was the simplest and most straight forward modification of the scripts I could think of ... the two issues are that there are multiple versions of this script, and that the patches are not applied. But I don't know how I can call the gridpack_generation.sh or parts of it from the runcmsgrid_powheg.sh; if this was possible, I guess also the work for maintaining the various powheg scripts could be reduced.

bendavid:
If the powheg scripts are used in a consistent git checkout, then the relative path of gridpack_generation.sh should always be known...

maierbenedikt:
Sorry, you mean starting from the Powheg working directory something like

"../MadGraph5_aMCatNLO/gridpack_generation.sh"?

bendavid:
yes

maierbenedikt:
Okay but then downloading MadSpin will have to happen in the create_powheg_tarball.sh script and must be packed into the tarball somehow.

Because in the end, the runcmsgrid.sh itself is lying somewhere in /cvfms/.../slc.../*tarball.gz 
At this point, I guess it would not know about the "../MadGraph5_aMCatNLO/gridpack_generation.sh" anymore?

Or one did directly a "wget https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/gridpack_generation.sh"  in the runcmsgrid.sh ...

bendavid:
Yes,
And sorry I misunderstood what the current changes do.  You'll have to change it along these lines anyway.

The tarball must be self contained
runcmsgrid_powheg.sh
is NOT allowed to do wget, which is absolutely forbidden in production jobs.

maierbenedikt:
Okay, I did not know that. Then I will do it from the create_powheg_tarball.sh...

For my education, why is it not allowed to do wget in runcmsgrid.sh?

bendavid:
Because then reproducibility cannot be guaranteed if the underlying web server changes.
(Plus this has the danger that thousands of production jobs could hammer the webserver)

bendavid:
Plus, when running madspin "from scratch" this has code compilation and phase space integration.
Code compilation is also not allowed, because the worker nodes are not guaranteed to have consistent development environment.

So the madspin initialization (with pre-compiled directory) must be done as part of the tarball creation.

maierbenedikt:
Okay. Just à la gridpack_generation.sh. Understood.

Then this will also speed up the actual event generation/decay.

syuvivida:
@bendavid, the best person to maintain run_pwg.py is Yuan Chao. I think he could take over my powheg jobs. But I need to double check with him about how long he could do it.

Eiko

covarell:
Was there an updated commit here?

maierbenedikt:
Hi Roberto,

Not yet, I was on vacation the last three weeks.
I will catch up with the latest changes to the powheg scripts and try to bring this to an end.

Benedikt

PR #575, opened 2015-08-03
[Powheg] update create_powheg_tarball.sh and patches to fix the compiling error of 4 proceses

syuvivida:
Updates include:

1) create_powheg_tarball.sh :
a) call the patches for Wgamma/W_ew/ttb_nlo_dec and trijet so that the POWHEG-BOX-V2 source tar ball (dated July 30, 2015) could compile without problem.

b) If an old directory with the same name exits, instead of moving the directory to a different name, stop the program completely

c) if the user does not use July2015 tar ball for this script, stop the program completely.

2) patches for the processes listed above.

### DIFF
--- a/bin/Powheg/create_powheg_tarball.sh
+++ b/bin/Powheg/create_powheg_tarball.sh
@@ -3,12 +3,12 @@
 fail_exit() { echo "$@" 1>&2; exit 1; }
 
 #set -o verbose
-EXPECTED_ARGS=8
+EXPECTED_ARGS=5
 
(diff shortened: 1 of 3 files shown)


bendavid:
Can't we just move to a mode where the script downloads the tarball itself as we do for Madgraph?

syuvivida:
@bendavid, I have updated script so that now it downloads the tarball itself.
in addition, the output tar ball name is also given by the script (same as madgraph script)

PR #576, opened 2015-08-03
cards for WW scattering and WZ scattering analyses

AndrewLevin:
Here are the differences with respect to https://github.com/cms-sw/genproductions/pull/527:

--added Wlljj diagonal ckm cards
--added lepton_masses to model in Wlljj cards
--changed group_subproceses to auto in Wlljj cards


bendavid:
Are you sure you want cut_decays=T for some of these?

bendavid:
Also for the Twj processes, can you comment how this relates to the existing powheg samples with cards in

https://github.com/cms-sw/genproductions/tree/master/bin/Powheg/production/st_tch_4f_ckm_13TeV
https://github.com/cms-sw/genproductions/tree/master/bin/Powheg/production/st_tW_5f_dr_ckm_13TeV
https://github.com/cms-sw/genproductions/tree/master/bin/Powheg/production/st_tW_5f_ds_ckm_13TeV

AndrewLevin:
Regarding the cut_decays: for the Wlljj samples, I think we only want the mll cuts to apply to the l+l-, so actually I think we do not want them. I will update the pull request. For the other samples, I think cut_decays should have no effect.

Regarding the top samples: I don't know the answer. I will remove those from the pull request for now.

bendavid:
cut_decays also refers to the ptl and etal cuts, so I would think it affects all the samples.

bendavid:
some of the cards still have cut_decays=T, which imposes pt and eta cuts on the leptons, is this really desired?

AndrewLevin:
It is only the WWjj_OS  and WWjj_SS cards where there is an etal cut. This cut is set to 2.6, beyond where CMS has acceptance. We are planning to select both of the leptons in these samples, so it makes no difference for us whether this cut is applied or not. I don't expect any other analysis will use these samples without selecting the leptons, so I propose we don't change them.

AndrewLevin:
Actually, for W to tau decays, I am not sure if it is possible for a tau to be produced with etal > 2.6 and then decay to an electron or muon with etal < 2.6, do you think that is something we should worry about?

bendavid:
For example.  In any case the default/preferred for MC production is to produce samples as inclusive as possible, also so that you can measure the acceptance using the same MC samples for example.

AndrewLevin:
OK, I don't really know what to do about this. I will need to discuss with Jasper. For now, can you please approve just the Wlljj ewk+qcd diag ckm cards, where cut_decay is set to false and the only cut on leptons is the mll cut. These will be used as a background for the WpWpJJ analysis. The gridpacks will be ready tomorrow, and I hope to request them then.

bendavid:
But why not just set cut_decays to False everywhere?

AndrewLevin:
I see no reason to make cuts on the leptons, so I think we should just set etal = -1 and then cut_decays will not make any difference. The only disadvantage I see is that we will lose statistics in the interesting region. I think we probably don't care about measuring acceptance from a fully inclusive region to our signal region, so it may be reasonable to make these cuts.

Also, we should probably reconsider the etaj cut (since matrix element jets can radiate gluons that could have smaller eta) and maybe also the mjj cut and the ptj cut.

Do you have any opinion about these cuts?

bendavid:
all cuts should at least be looser than any jet-level cuts.

PR #577, opened 2015-08-04
missing mass points for scalar ttbar+DM

peiffer:
This pull reqeusts is going to replace #487. I could only find the scalar ttbar+DM cards in the directory of violatingcp, thus the cards for the pseudo scalar ttbar+DM samples are still missing.


bendavid:
https://github.com/cms-sw/genproductions/pull/487 still had the unresolved issue of inconsistent 4fs vs 5fs definition.  Has this been resolved and/or otherwise discussed how to proceed?

@violatingcp

peiffer:
Hi Josh,
 
as I understood, the 4f /5f issue concerns only the bbar+DM samples. We have already produced other mass points for ttbar+DM with the same settings.

bendavid:
Yes I know that some samples were already produced with these settings, but this issue DOES affect the ttbar+DM samples as well (including the already produced samples).  So we need to decide how to proceed.

Either reset the old requests and reproduce everything with fixed cards, or verify with one point that the difference is negligible and proceed as is.

peiffer:
Hi Josh,

I don't see why 4f or 5f scheme should give any difference for ttbar+DM because their is no Feynman diagram with initial state b quarks involved for this process.

bendavid:
So first of all, sure there is.  b b~ -> g -> t t~ + X.  But also this concerns the outgoing jets.

Independent of that, as usual, there should be customizecards instead of paramcards.  (but the other issue needs to be resolved first anyways)

peiffer:
Okay, then I will go back to the requesters and ask them to provide the missing input.

PR #578, opened 2015-08-05
changed the file cuts.f for the cards diphotonJets

mmachet:


bendavid:
Just for future reference.  The default cuts.f is responsible for actually implementing many of the cuts in the run card, so one needs to make only the minimal modification/addition to it starting from the version in the madgraph release.

mmachet:
ok!

2015-08-05 17:33 GMT+02:00 Josh Bendavid <notifications@github.com>:

PR #580, opened 2015-08-07
ggZH with inclusive Z decay and H undecayed, for HWW

rebecacern:
Title says it all, new powheg process feature (the inclusive Z decay)


syuvivida:
The cards look fine to me. @covarell @bendavid can you merge them?
Thanks!!
Eiko

PR #582, opened 2015-08-10
add VBS WWjj SS cards

AndrewLevin:
These cards have two important fixes compared to the cards in the cards/production/13TeV/WpWpJJ:

-- the breit-wigner width is set to 15000
-- the etal cut is removed


covarell:
Why do you fully rename the directory instead of doing an update?

AndrewLevin:
There are a lot of VBS cards now eg

https://github.com/cms-sw/genproductions/pull/583

and probably putting them all in one directory is more organized. I have updated the pull request so that it removes the wpwpjj directory now.

PR #589, opened 2015-08-19
Upload of cards for FOURTOP_NLO production

heilman:


nstrobbe:
Have you seen these cards:
https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/TTTT_5f_NLO 

The sample produced with these is already available:
https://cmsweb.cern.ch/das/request?input=/TTTT_TuneCUETP8M1_13TeV-amcatnlo-pythia8/RunIISpring15DR74-Asympt25ns_MCRUN2_74_V9-v1/MINIAODSIM

heilman:
These cards seem to not be generating just SM tttt.  The top mass seems to have been shifted down slightly as well as an enhancement in the b content of the proton defined in the generation process.  We are searching for SM tttt and need a clean sample.

nstrobbe:
The top mass was set to the same value as used by all other ttbar samples. 

If you need 4f in particular, then you will indeed need a new sample. 
But it would not go so far as to say it would not be SM four-top production. The only interactions present in this sample, are SM interactions.

bendavid:
Hi indeed the only relevant difference I can see between these cards and the ones @nstrobbe is referring to is 4f vs 5f.  (these cards would also need some fixes for the pdf definitions, but ok)

In any case there are no significant differences expected between 4f and 5f for this process.  Have you done some tests which indicate otherwise?  Can you point to some plots/slides illustrating the issue you are referring to in the existing sample?

heilman:
Hi Everyone,  Sorry for my ignorance but I'm not an MC expert.  I'm just following instructions provided to me by my MC contact who is on holiday to try to get the production chain going.  After consulting with some of my colleagues, I think that you are correct and we are fine with a 5f sample.  What we do need is about 4 times as many events as the sample currently has.  But I assume that I should take that up somewhere else.  Thanks for your comments and you can close this pull request if you like.

PR #590, opened 2015-08-19
Cards for EXO-RunIIWinter15pLHE RPVresonantToLL

davidsheffield:
from Soeren Erdweg


davidsheffield:
@bendavid regarding you question about the pdfs at today's MCCM, we already submitted the EMu version of these in #339 (this PR is for the ETau and MuTau channels). The pdfs were discussed in that PR.

bendavid:
Fine for these, but I strongly suggest analyzers to look into this/improve the choice of pdfs for the future.

PR #592, opened 2015-08-20
update AtoZh cards and modify gridpack_generation.sh accordingly

gitytakahas:


covarell:
@bendavid @govoni , looks fine to me. Since the number of exotic model is growing, I wonder if we should let the script download all of them in every run (this seems the case now)?

bendavid:
Indeed this is something which could be improved.

covarell:
If you would agree, I can modify the script to accept an extra argument with the model name. If that argument exists and makes sense, then the relative tarball is unpacked

bendavid:
Small issue is that this would break compatibility with existing wmlhe fragments.

covarell:
OK, but there are very few committed I think (mostly they are written directly in the fragment box, so they are bound to a specific release)

bendavid:
Alright, let's hold off on that change until we move to the 2.3.x branch in that case.

bendavid:
For this pull request, please replace param cards with customizecards as usual.  Also you probably want cut_decays = F in the run cards to avoid applying unintended cuts to the decay chains.

gitytakahas:
I now updated the cards. 

Regarding the cut_decays values, should I need to reproduce the gridpacks again after fixing cut_decays = F ?

Thanks,
Yuta

bendavid:
Yes you would, although looking at the cards, given that the decay chains are always to leptons, and there are no cuts on lepton pt or eta, it should actually be ok to leave it as is if gridpacks have already been produced.

bendavid:
How did you produce the gridpacks using the param cards though?  The gridpack scripts ignore those completely.

gitytakahas:
Hi Josh,

oh, it is just my mistake when I asked pull request.
I produced gridpacks using custermized cards. Sorry for the confusion.

Best,
Yuta

2015-08-22 16:05 GMT+02:00 Josh Bendavid <notifications@github.com>:

> How did you produce the gridpacks using the param cards though? The
> gridpack scripts ignore those completely.
>
> —
> <https://github.com/cms-sw/genproductions/pull/592#issuecomment-133706927>
> .
>



-- 
Yuta Takahashi
LHC - CMS experiment,
CERN (European Organization for Nuclear Research)

bendavid:
ok, If you're going to use the existing gridpacks with cut_decays=T I think it is ok, but then please update the pull request so that the cards match what was actually used.

gitytakahas:
Sorry, I, by mistake push "close" now it is reopened.

Yuta

PR #593, opened 2015-08-24
Updates on run_pwg.py for CMSSW 7.1.17 and bug fixes

yuanchao:
Updated version of run_pwg.py for the latest POWHEGBOX source and CMSSW 7.1.17. Updates from the original create_tarball.sh for specific processes are includes. Bug fixes for ttH, gg_H and Wgamma processes.

### DIFF
--- a/bin/Powheg/run_pwg.py
+++ b/bin/Powheg/run_pwg.py
@@ -573,13 +573,15 @@ def createTarBall(parstage, folderName, prcName, keepTop, seed, scriptName) :
 '''
 
 cd $WORKDIR/$folderName
+echo "Processing folder: "
 pwd
 
-rm -rf $folderName'_'$process
+rm -f $WORKDIR/$folderName'_'$process'.tgz'
+
 cp -p $WORKDIR/run_pwg.py $WORKDIR/$folderName
 
-cp -p $WORKDIR/$folderNamepwggrid-0001.dat $WORKDIR/$folderNamepwggrid.dat
-cp -p $WORKDIR/$folderNamepwg-001-stat.dat $WORKDIR/$folderNamepwg-stat.dat
+cp -p $WORKDIR/$folderName/pwggrid-0001.dat $WORKDIR/$folderName/pwggrid.dat
+cp -p $WORKDIR/$folderName/pwg-0001-stat.dat $WORKDIR/$folderName/pwg-stat.dat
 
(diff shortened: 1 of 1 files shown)


syuvivida:
Hi John,
   it seems that the current option -p01239 only allows local
 runs. could we make it possible to submit to the batch system?

 Some of the processes would take a long time to run through p01239 
and would need to submitted to the lxplus batch system.

Eiko

yuanchao:
Although one could submit the whole python script to LSF, it would be better that I write everything into the script. The most issue is LSF runs in a different directory. One need to change the directory back such that CMSSW env. can be properly setup.

syuvivida:
@yuanchao ,
python $PWD/run_pwg.py -p 01239 -i powheg.input -m DMGG -f my_DMGG -q 1nw -n 100
 is this the right command to run the one-shot option? 
I tried this oneshot option at lxplus, but the batch job complained that the grid output files are not found.
and I think it was looking in the wrong directory:
cp: cannot stat `/afs/cern.ch/work/s/syu/Big_Generators/MyPowHeg/v20/slc6_DMGG/debug/CMSSW_7_1_17/src//pwggrid-0001.dat

yuanchao:
I found some bug with this option and the fixes are on github. The error
you see is not relavent. That's for parallel grid production hack on
runcmsgrid.sh script.
2015/8/27 下午6:19於 "syuvivida" <notifications@github.com>寫道：

syuvivida:
@yuanchao 
I just tried on DMGG process, but in the output my_DMGG_DMGG.tgz, it seems that there is no files that store computed grid. The files that are kept in the tar ball are executable, runcmsgrid.sh, powheg.input, but no computed grid files, like pwg*.dat

syuvivida:
This latest update works fine now. @bendavid and @covarell can you merge the pull request?

Many thanks!!

Eiko

syuvivida:
Hi,
 this python file has been tested with single-run code in the following processes and they work OK.
@bendavid, @covarell, @govoni could you merge the pull request? Many thanks!!

dijet
DMGG
DMS
DMV
gg_H_2HDM
gg_H_MSSM
gg_H_quark-mass-effects
ggHZ
HJ
hvq
HW
HWJ
HZ
HZJ
trijet
ttb_dec
ttH
VBF_H
VBF_HJJJ
W
W_ew-BMNNP
Wgamma
Wj
WW
WZ
Z
Z_ew-BMNNPV
Zj
ZZ

PR #594, opened 2015-08-25
Excited Lepton configs from Chia-Ming

dnash86:
Excited lepton configs from Chia-Ming Kuo.

We were late to process these (missed the request from analyzer), so if possible, please review and merge soon.  Thanks a lot, sorry for being the cause of urgency.


dnash86:
Also, they will need pythia 8.205 or greater - is that in the CMSSW version used for the RunIIWinter15GS campaign, or not?

bendavid:
Yes, the campaign is currently using 8.205

dnash86:
Excellent, thanks.

PR #595, opened 2015-08-25
Excited lepton

dnash86:
Did not at first notice the extra parentheses here, removed now.


dnash86:
@bendavid Fixed a missing parenthesis here, from the previous PR: https://github.com/cms-sw/genproductions/pull/594

Sorry for spam.

PR #596, opened 2015-08-26
Added 4f WWG and WWZ at NLO

tmulholland:
Exactly the same as the current 5f NLO sample, but with 4f scheme instead of 5f.


bendavid:
was there a problem producing the 5fs versions? (same as for the WW samples?)

tmulholland:
Hi Josh,

Yes, with the 5fs versions we were getting "poles don't cancel" errors.

-Troy

PR #598, opened 2015-08-27
ExtendedWeakIsoSpin in CalcHEP, LHE headers and model files

dnash86:
All model files are in "model/".
All LHE Headers from <header> to </header> are in "LHEHeaders/".
Let me know if any more information is needed.

Note that the pdf used was NNPDF30_lo_as_0130 (the five flavor scheme).


dnash86:
I will wait to upload LHE files until this is reviewed and approved, thanks.

dnash86:
Also, clearly I cannot use XML formats in github comments, I meant to write that all LHE Headers from "header" to "/header" are included.

PR #599, opened 2015-08-28
Added lower cut on invariant mass

mweinberg2718:


covarell:
Ok, two remarks
- Please close this and upload instead to production/13TeV/ etc. (the "examples" directory should be static)
- Please run a few events to check that the 50 GeV cut is really applied... ( @dnash86  I do not remember the outcome of your investigations, but there were issues with this parameter)

dnash86:
I took a quick look, it looks to me like it is the input file/parameter that is okay.

The "Z" process has mass_low and mass_high as configurable parameters, if I remember correctly the other powheg processes for Z production (with electroweak corrections) only had mass_low as a parameter.  So it should be okay, but of course checking is good in any case.

mweinberg2718:
Thanks for taking a look. I wrote out a hundred or so events and 
confirmed that the Z mass was above 50 GeV in all cases (dipped as low 
as 58 GeV for one event). I've closed the original pull and opened a new 
one with the fragment in the production folder.

Cheers,
Marc

PR #603, opened 2015-09-02
Config file fragment and slha file for 100 GeV, 100 cm sample

wulsin:
This PR just has one mass, lifetime point.  Once it is approved, I will add the corresponding configs for the other mass and lifetime samples.  One question I have is whether a path needs to be specified for the slha file when it is defined in the cff.py file:  
particleFile = cms.untracked.string('geant4_AMSB_chargino_100GeV_ctau100cm.slha')


bendavid:
The production release is CMSSW_7_1_X.

I assume the file SimG4Core/CustomPhysics/data/RhadronProcessList.txt is already present in this release?  Otherwise a patch release is needed.

For the slha table, if the file here is not already in the release, please use instead the inline syntax.  (Let me find an example)

wulsin:
Yes, RhadronProcessList.txt is already in the release:
https://github.com/cms-sw/cmssw/blob/CMSSW_7_1_X/SimG4Core/CustomPhysics/data/RhadronProcessList.txt

bendavid:
The SLHA table can be inlined in the config as here
https://github.com/cms-sw/genproductions/blob/cecab601e4bd05ecead33cbab93cfd241514802c/python/ThirteenTeV/DisplacedSUSY_stopToBottom_M_1000_100mm_TuneCUETP8M1_13TeV_pythia8_cff.py

So then no patch release is needed.

wulsin:
Thanks for the example.  How should I proceed?  Cancel this PR and start a new one?  Or should I modify the existing PR?

bendavid:
You can modify this one. (please remove the separate slha file from the pull request in any case)

wulsin:
Ok, I'll do that.

wulsin:
I committed a change so that the slha file is defined inline.

bendavid:
Ok sorry maybe I'm confused, do you need the SLHA table to be read by pythia or by geant?

For pythia the correct syntax is 
SLHATableForPythia8 = cms.string('%s' % SLHA_TABLE),

wulsin:
The SLHA file needs to be read by SimG4Core/CustomPhysics/src/CustomParticleFactory.cc <http://customparticlefactory.cc/>, which then passes the information to GEANT.  Given this, I think that defining the slha file inline will not work.

bendavid:
Yes that's right, the inlining mechnanism only works for when the slha file needs to be read by pythia.

In that case indeed you'll have to add it to
https://github.com/cms-data/Configuration-Generator
and then we'll need a patch release

wulsin:
Ok.  So I add the slha file to https://github.com/cms-data/Configuration-Generator <https://github.com/cms-data/Configuration-Generator> but leave the cff.py where it is?

bendavid:
Yes that's right.

Then the path to the slha file in the cff should point to "Configuration/Generator/data/<filename>.slha"

wulsin:
Great, thanks.  I'll do that.

wulsin:
I made a PR:  https://github.com/cms-data/Configuration-Generator/pull/4.

degano:
@wulsin @bendavid I've merged this into 71X and 76X.

bendavid:
The pull request was merged for cmssw, but I think we're still waiting for a new 71x release to proceed with production.

In the meantime, for the gen fragment here, are these samples going to be produced with jet matching?  (can you point me to the madgraph cards used for the lhe part of the request?)

bendavid:
(These changes should be in CMSSW_7_1_19_patch1 already, but please check to confirm)

Would still want to look at the madgraph cards in order to make sure the pythia configuration is consistent.

wulsin:
Hi @bendavid, sorry, but I'm not sure how to check that the changes are already in CMSSW_7_1_19_patch1.  When I look at https://github.com/cms-sw/genproductions, there are no CMSSW_X_Y_Z tags.  

Also, which madgraph cards would you like to check?  The ones here?:  https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_DisappTrks/AMSBchargino100GeV

Please let me know if you have any comments on them.  The last thing I need to check are the jet-matching parameters (qcut and xqcut).  Besides that, they should be ready.  But I would appreciate your checking them, since you may notice a problem that I missed. 

Thanks!

bendavid:
If they were already merged into genproductions, then it means we already checked them.

For the release, just try running your gen fragment and see if it finds the extra customize functions which were added.

wulsin:
Great, thanks I'll try that!

bendavid:
Any news on this?

govoni:
Any news on this? If not, I will close it tomorrow.

wulsin:
Hi Pietro, Josh,

I apologize for the delay, but I have not yet had a chance to work on this.  I would request that you give me another week to put my attention to this.

Thanks,
Wells

perrozzi:
Any news on this?

wulsin:
Obviously I didn't have time to work on this lately.  Once I finish some other projects I will try to come back to this and pick it up again.  Fine to leave it closed for now.

PR #606, opened 2015-09-04
Branon 13 TeV

dnash86:
Madgraph cards from Ashim Roy

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_Branon/Branon_MB10000_fB100/Cards/Branon_MB10000_fB100_customizecards.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_Branon/Branon_MB10000_fB100/Cards/Branon_MB10000_fB100_customizecards.dat
@@ -1,0 +1,3 @@
+#put card customizations here 
+set param_card fBRANON 100
+set param_card MPI 10000
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_Branon/Branon_MB10000_fB100/Cards/Branon_MB10000_fB100_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_Branon/Branon_MB10000_fB100/Cards/Branon_MB10000_fB100_proc_card.dat
@@ -1,0 +1,36 @@
+#************************************************************
+#*                     MadGraph5_aMC@NLO                    *
+#*                                                          *
+#*                *                       *                 *
+#*                  *        * *        *                   *
+#*                    * * * * 5 * * * *                     *
+#*                  *        * *        *                   *
+#*                *                       *                 *
+#*                                                          *
+#*                                                          *
+#*         VERSION 2.2.2                 2014-11-06         *
+#*                                                          *
+#*    The MadGraph5_aMC@NLO Development Team - Find us at   *
+#*    https://server06.fynu.ucl.ac.be/projects/madgraph     *
+#*                                                          *
+#************************************************************
+#*                                                          *
+#*               Command File for MadGraph5_aMC@NLO         *
+#*                                                          *
+#*     run as ./bin/mg5_aMC  filename                       *
+#*                                                          *
+#************************************************************
+set group_subprocesses Auto
+set ignore_six_quark_processes False
+set loop_optimized_output True
+set complex_mass_scheme False
+import model sm
+define p = g u c d s u~ c~ d~ s~
+define j = g u c d s u~ c~ d~ s~
+define l+ = e+ mu+
+define l- = e- mu-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+import model Branon
+generate p p > a pi pi
+output Branon_MB10000_fB100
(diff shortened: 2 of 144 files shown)


bendavid:
Param cards are not supported, replace with customize cards as usual

dnash86:
@bendavid updated with customize cards now.

bendavid:
Is the model included in gridpack_generation.sh?
Is this 4fs or 5fs? (massive or massless b's in the model)
The pdf is using cteq6l1, but should be switched to appropriate nnpdf 3 set as for the rest of the production.
Process definition looks like without jet matching, but in that case ickkw needs to be set to 0

dnash86:
The model is not yet included in gridpack_generation.sh as far as I know.  I've forwarded the other questions to the analyzers, but I think I will need to update gridpack_generation.sh

gomber:
Hi Josh, 

Replying here directly only, so that we can get this done quickly.
So this is new model Branon model, and branons are massive.

For pdf, shouldn't we change it to following as done in other request
  'lhapdf'      =  pdlabel      ! PDF set
  263400        =  lhaid      ! if pdlabel=lhapdf, this is the lhapdf number

CKKW will be set to 0.
Could you please reply on the previous 2 comments, then we can provide the new cards.

bendavid:
The correct pdf's and flavor definition depend on whether the b-quarks are massive or massless in the model (not the Branons, sorry for the confusion)

If the b-quarks are massive then this should be 4fs, and then the proc cards are ok, and 263400 is the correct pdf set (and the pdflabel needs to be changed to 'lhapdf')

ickkw=0 is indeed the correct setting.

But maxjetflavor in the run card should also be set to 4 for 4fs.

If the b-quarks in the model are massive then instead you need 5fs
(b b~ as part of proton and jet definition, 263000 for the pdf, and maxjetflavor=5)

Aside from that, can you please explain why bwcutoff is set to 1?

gomber:
Hi Josh,

Thanks for the reply.
So for this particular model, it doesn;t matter if we use b-quarks massive or massless, so we can use
massive.

And Ashim that implies, you need to do following changes in the card :
) pdflabel needs to be changed to 'lhapdf'
)But maxjetflavor in the run card should also be set to  4
) ickkw = 0

For us even changing bwcutoff does;t affect the cross section. So that's why we set it to 1.
Also read madgraph documentation, regarding this https://cp3.irmp.ucl.ac.be/projects/madgraph/wiki/FAQ-Cards-2 and as it says "The value of bwcutoff does not affect the cross section of any processes, except those defined using the decay chain formalism, ab>(c>de)(f>gh)" and that is not the case for us.

So I think we are fine.

So if you agree then Ashim can prepare new cards with the following change, so that we can move forward.

 ) pdflabel needs to be changed to 'lhapdf'
)But maxjetflavor in the run card should also be set to  4
) ickkw = 0

Regards
Bhawna

bendavid:
Even if it doesn't affect the cross section, bwcutoff does affect the lhe event history and in turn how much the parton shower may modify the mass distribution.  So unless there is a good reason to change it, best to leave the default value of 15.

gomber:
Thanks Josh for the prompt reply.
We will fix it to the default value i.e 15
Ashim please do the following changes in the card and provide new cards

 ) pdflabel needs to be changed to 'lhapdf'
)But maxjetflavor in the run card should also be set to  4
) ickkw = 0
) bwcutoff = 15

Regards
Bhawna

dnash86:
@bendavid Updated cards from Ashim, please check now.

bendavid:
The pdf definition is still not right.
(The lhaid is commented out, but should be 263400 in this case)

dnash86:
The lhaids are now 263400

covarell:
And I think you need 
T  = use_syst   
everywhere (as we discussed for previous EXO requests)

dnash86:
Ah, good point @covarell.  Fixed now.

dnash86:
Are these okay now, with use_syst fixed?

covarell:
Good. This will be approved. We need:
1. The tarball of the exotic model uploaded to https://cms-project-generators.web.cern.ch/cms-project-generators/
2. An extramodel card for each fB and MB point

See new instructions at:

dnash86:
@covarell Question regarding the extramodel cards:  What is the role of these?  Should they each look like this:

https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/examples/ggh012j_5f_NLO_FXFX/ggh012j_5f_NLO_FXFX_extramodels.dat

with the .tgz model file in place of the zip file in the example?

dnash86:
@covarell Also, I was looking for instructions about how to upload to https://cms-project-generators.web.cern.ch/cms-project-generators/, but couldn't find them - where do I copy the model file?

bendavid:
The gridpack script has recently been modified such that it no longer downloads all models for every run.  So the extramodels card tells it which additional models need to be downloaded and unpacked into the models directory.

bendavid:
This area is pointing to /afs/cern.ch/cms/generators/www
So the files should simply be copied there

dnash86:
@bendavid Thanks for the explanation.  The model file is there now: /afs/cern.ch/cms/generators/www/branon_UFO.tar.gz

Regarding the extramodels card - I guess then it is just a list of file names?  So in this case it should just read:

branon_UFO.tar.gz

and they should be named Branon_MB10000_fB100_extramodels.dat, one for each point?

bendavid:
Yes that sounds right.

PR #607, opened 2015-09-04
Cards for EXO QBHToEMu for RunII

davidsheffield:
from @serdweg

This is for the QBH generator, where the parameters are set directly in the C++ code. Soeren has extracted the parts of the code used to set the parameters and provided them in text files as the "cards".

The reason these use CTEQ6L1 rather than NNPDF:

"the main reason is to be consistend between the QBH generator, where we can easily change the PDF set and CalcHEP where it is not so easy. We use CalcHEP to cross check our QBH results, and therefore it seems useful to use consistend PDF sets, so Cteq6l1. In the future this is definitly a point of improvement, but the effect on our sample should be small as the only contributing channel is q-q_bar annihilation with one valence quark."

Soeren also submitted the recent CalcHEP requests also using CTEQL1, although they were for a different process.

### DIFF
--- a/bin/QBH/cards/production/13TeV/exo_QBHToEMu/ADD/QBHToEMu_M-1000_n4_ADD_TuneCUETP8M1_13TeV-pythia8.txt
+++ b/bin/QBH/cards/production/13TeV/exo_QBHToEMu/ADD/QBHToEMu_M-1000_n4_ADD_TuneCUETP8M1_13TeV-pythia8.txt
@@ -1,0 +1,60 @@
+    /// QCD scale definition
+    qbh->setQscale(false); /// Definition of QCD scale for PDFs false(= QBH mass), true (= inverse gravitational radius) (Default = true)
+
+    /// Yoshino-Rychkov stuff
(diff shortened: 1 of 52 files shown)


bendavid:
Does QBH use pythia8 internally? Is the name TuneCUETP8M1_13TeV-pythia8 appropriate?

serdweg:
Hi,
yes it runs via Pythia8.
Best,
Soeren

bendavid:
Does it actually use the CUETP8M1 tune?

serdweg:
The tune is only relevant in the hadronization of our signal, as the 
final state is exclusivly electron and muon. Pythia8 is used in the 
production only as a framework for the QBH code. The requested 
hadronizer file of course uses the CUETP8M1 tune.

bendavid:
Ok, but then it's a bit misleading/imprecise to include TuneCUETP8M1 in the name of the txt files no?

serdweg:
The only reason for this naming convention is, that the name is the same 
as the correspondng sample name, should I rename all files?

bendavid:
Ok.  No I guess it's fine as is.

PR #615, opened 2015-09-10
DYJetsToNuNu_Zpt-80toInf_5f_LO_MLM and WJetsToLNu_Wpt-80toInf_5f_LO_MLM

perrozzi:
to be (currently) used for b enriched sample using PS filter


perrozzi:
gentle reminder to check the datacards

bendavid:
Cards look ok, but you could speed up the Z to nu nu case by using the decay chain syntax + bwcutoff=15000 as in the W case (since there is no gamma* contribution here)

perrozzi:
done, thanks

bendavid:
No no, I meant that the process definition should/could also be
p p > z, z > vl vl~

perrozzi:
should be ok now then, thanks

PR #617, opened 2015-09-15
proc and run card for W+jets at NLO, including the W decay in madgraph

mstoye:
Hi Josh,
This is supposed to be used to make a request for the W+jets sample with NLO precision also for the W decay products. We intent to bin in 0,1, and 2 jets. Ideally we want this to be as close as possible to the previous 8 TeV jets (inclusive LO) using Pythia6. We might use this to reweight the CMSSW_5_3_X samples and to first order GEN would be ok. The aimed stats are about 200M events and presumably mainly in 1&2j.
Can you comment if the files make sense to you for that purpose?
Cheers,
Markus

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/8TeV/WJToLNu_NLO/wellnu2j_5f_NLO_FXFX_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/8TeV/WJToLNu_NLO/wellnu2j_5f_NLO_FXFX_proc_card.dat
@@ -8,8 +8,8 @@ define j = j b b~
 define ell+ = e+ mu+ ta+
 define ell- = e- mu- ta-
 
-generate p p > ell+ vl jj $$ t t~ h [QCD] @0
-add process p p > ell- vl~ jj $$ t t~ h [QCD] @1
+generate p p > ell+ vl j j $$ t t~ h [QCD] @0
+add process p p > ell- vl~ j j $$ t t~ h [QCD] @1
 
-output wellnu012j_5f_NLO_FXFX -nojpeg
+output wellnu2j_5f_NLO_FXFX -nojpeg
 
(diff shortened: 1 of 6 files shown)


bendavid:
This can only be done using pythia 8, so you'll have to change parton_shower in the run card to PYTHIA8.

ickkw will also have to be changed to 3.

mstoye:
Hi Josh,
I changed it to Pythia 8. By "this" you mean that onlu with Pythia 8 we can do the NLO W decay?
Cheers,
Markus

bendavid:
the FXFX merging for combining the different jet multiplicities at NLO is only working with pythia 8 (and in general there is only limited support for aMC@NLO with pythia 6, and we have not produced any such samples in CMS or tested it at all)

bin/MadGraph5_aMCatNLO/cards/production/8TeV/WJToLNu_NLO/wellnu2j_5f_NLO_FXFX_proc_card.dat
```diff
@@ -0,0 +1,15 @@
...
+#switch to diagonal ckm matrix if needed for speed
+#import model loop_sm-no_b_mass
+
+define p = p b b~
+define j = j b b~
+
+define ell+ = e+ mu+ ta+
+define ell- = e- mu- ta-
+
+generate p p > ell+ vl jj $$ t t~ h [QCD] @0
```
bendavid: I think there needs to be a space between the "jj" for this to work.

bin/MadGraph5_aMCatNLO/cards/production/8TeV/WJToLNu_NLO/wellnu2j_5f_NLO_FXFX_proc_card.dat
```diff
@@ -0,0 +1,15 @@
...
+add process p p > ell- vl~ jj $$ t t~ h [QCD] @1
+
+output wellnu012j_5f_NLO_FXFX -nojpeg
```
bendavid: in general the name of the output needs to match the name of the cards.  eg this should be
"output wellnu2j_5f_NLO_FXFX -nojpeg"

mstoye:
ok, sorry changed the output names and jj->j j

PR #591, opened 2015-08-20
DYBJetsToLL, DYBJetsToNuNu and WBJetsToLNu datacards

perrozzi:
adapted from https://github.com/cms-sw/genproductions/tree/2a3a99875e5424505436ed7378e15dac5a6f9f27/bin/MadGraph5_aMCatNLO/cards/production/13TeV/DYBBJetsToLL_M-50_5f_LO_MLM

PR #620, opened 2015-09-15
WBJetsToLNu_Wpt-80toInf_5f_LO_MLM DYBJetsToNuNu_Zpt-80toInf_M-50_5f_L…

perrozzi:
same as https://github.com/cms-sw/genproductions/pull/591 with 
80  = ptllmin  ! Minimum pt for 4-momenta sum of leptons(l and vl)

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/WBJetsToLNu_Wpt-80toInf_5f_LO_MLM/WBJetsToLNu_Wpt-80toInf_5f_LO_MLM_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/WBJetsToLNu_Wpt-80toInf_5f_LO_MLM/WBJetsToLNu_Wpt-80toInf_5f_LO_MLM_proc_card.dat
@@ -1,0 +1,21 @@
+import model sm-ckm_no_b_mass
+
+define p = p b b~
+define j = p
+define bquark = b b~
+
+define ell+ = e+ mu+ ta+
+define ell- = e- mu- ta-
+
+# Specify process(es) to run
+generate p p > ell+ vl bquark / h @0
+add process p p > ell+ vl bquark j / h @1
+add process p p > ell+ vl bquark j j j / h @2
+add process p p > ell+ vl bquark j j / h @3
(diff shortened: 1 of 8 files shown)


bin/MadGraph5_aMCatNLO/cards/production/13TeV/WBJetsToLNu_Wpt-80toInf_5f_LO_MLM/WBJetsToLNu_Wpt-80toInf_5f_LO_MLM_proc_card.dat
```diff
@@ -0,0 +1,21 @@
...
+define j = p
+define bquark = b b~
+
+define ell+ = e+ mu+ ta+
+define ell- = e- mu- ta-
+
+# Specify process(es) to run
+generate p p > ell+ vl bquark / h @0
+add process p p > ell+ vl bquark j / h @1
+add process p p > ell+ vl bquark j j j / h @2
```
bendavid: these are "out of order" which might cause confusion downstream when looking at the matching efficiency tables

bendavid:
Could do the same with the decay syntax + bwcutoff for the znunu here in principle.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/WBJetsToLNu_Wpt-80toInf_5f_LO_MLM/WBJetsToLNu_Wpt-80toInf_5f_LO_MLM_proc_card.dat
perrozzi: fixed, thanks

perrozzi:
done

bendavid:
Ah, so in fact the decay chain syntax plus large bwcutoff should ALSO be used for the w samples.
(This is how we did the inclusive samples in this round also)
Otherwise the 4 jet case with off-diagonal ckm becomes quite intensive.

perrozzi:
no problem, done

PR #625, opened 2015-09-17
Cards for ZGamma, ZtoNuNu

nstrobbe:
Z+photon sample, with Z decaying to neutrinos.
Main analysis in need of this sample needs high pt photons, so they will use the pt > 80 sample.
Since no such samples exist at all, we also propose to complement this with a 10 < pt photon < 80 GeV sample.


nstrobbe:
Any comments on this pull request?

covarell:
Sorry, is there a reason why most VV samples use NLO + up to 1 jet and this is LO + 3 jets?

nstrobbe:
We decided on LO since we can have more jets at matrix element level that way. Of course if there are strong objections, we can go to NLO as well.

nstrobbe:
Is there any strong preference? Or can I go ahead and produce the gridpacks?

PR #627, opened 2015-09-18
[monoHiggs group] Pythia8 configuration file for decaying Higgs into 5 subchannels

syuvivida:
The gridpacks for the monoHiggs signal do not include decays of Higgs.
These 5 configuration files further decay Higgs into aa, bb, tau tau, WW(llnunu) and ZZ(4l).
The gridpacks were produced with LO Madgraph5_aMC@NLO and zero jet multiplicity. Therefore, 
there is no jet matching in the configuration file.


syuvivida:
@covarell @govoni @bendavid 
  We have produced step1 test samples using these pythia8 configuration files. The analyzers of each channel (aa, bb, tau tau, WW, ZZ) have all confirmed that the root files look OK. Could you merge the pull request? Many thanks!!

Eiko

PR #629, opened 2015-09-19
Patch VBF processes to have configurable H mass window bounds

yuanchao:
* Patches for VBF_H and VBF_Z_Z processes so that H mass window can be configurable with the code of gg_H_quark-mass-effects, following the suggestion from David Sperka.
* Patches for the new POWHEG source tarball of Sep. 2015.

### DIFF
--- a/bin/Powheg/create_powheg_tarball.sh
+++ b/bin/Powheg/create_powheg_tarball.sh
@@ -19,7 +19,7 @@ echo "   ______________________________________________________    "
 repo=slc6_amd64_gcc481/powheg/V2.0/src
 echo "%MSG-POWHEG source repository = $repo"
 
-name=powhegboxV2_July2015
+name=powhegboxV2_Sep2015
 echo "%MSG-POWHEG source tarball name = $name"
 
 process=${1}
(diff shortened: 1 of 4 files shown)


covarell:
@dsperka can you please confirm this is perfectly in sync with the change you did to the shell script?

dsperka:
there are only some tiny differences:
- here the default is 10 sigmas if the parameter is not set, in my commit it was 30
- here the absolute minimum is 0.5 GeV, in my commit it was 0.0 GeV

These differences should not matter. However, we should not apply this patch and still have the lines in the shell script, of course. Sorry I didn't know about this commit.

covarell:
Even if these differences do not matter, it would be better to make this patch absolutely identical to the one already approved...

dsperka:
We should only have one patch, either from the "patches" folder or in the shell script. If you apply the change from the "patches" folder and then also apply the fix from shell script, it could / probably will cause problems. I think probably 30 sigmas as default is better since that is what the VBF_H process actually was hard coding before. For the minimum, 0.0 GeV is again what the VBF_H process was doing before.

yuanchao:
Added updates for run_pwg.py and create_powheg_tarball.sh to fetch the latest source tarball.

PR #630, opened 2015-09-21
Cards for bbH NLO with MG5

OlivierBondu:
also with readme instructions since this is going to be pLHE


covarell:
Sorry, if this is going to be pLHE, why do we need this stuff in genproductions?

bendavid:
It's still good to have links from the pLHE requests.

OlivierBondu:
plus, I would prefer to be sure about the param_cards, I took the latest mH, mT, mb from discussing with Eiko, but I do not know where I should expect to find the latest-greatest default CMS settings ?

bendavid:
please please please start from the default param cards in aMC@NLO

bendavid:
(in particular the default b mass in aMC@NLO is 4.7GeV)

OlivierBondu:
I took the param card directly from the authors, then modified only the aforementioned masses, I'll revert the b mass to 4.7 GeV

bendavid:
Ok,
Please do a diff against the default card to make sure any other differences are understood.

OlivierBondu:
so the notable differences with the parameters from ``MG5_aMC_v2_3_2_2/models/sm_v4/param_card.dat`` are:
```
    5 4.700000e+00 # MB 
    6 1.725000e+02 # MT 
   25 1.250000e+02 # MH 
```
I did correct the Higgs width also: 
```
DECAY  25 4.070000e-03 # WH
```

However for the top width, I have no clue of what should be the correct one for mt = 172.5 (the default card having a mass of 173 GeV, I don't know where to find the info)
```
DECAY   6 1.491500e+00 # WT
```

the rest seems the exact same

bendavid:
You should compare to loop_sm actually (but should be similar)

Fine to leave the top width as is. (Also what we do for ttbar production)

OlivierBondu:
@bendavid : comparing with ``models/loop_sm/restrict_default.dat`` there is one additional modification:
```
Block yukawa
    5 4.187000e+00 # ymb: mb(mb) 
```
 instead of 
```
Block YUKAWA
    5 4.700000e+00 # ymb 
```

which is expected (latest LHCHXSWG recommendations on how to treat the inputs 
for mb(muR) inside the bottom Yukawa coupling, implemented by the authors)

PR #638, opened 2015-09-24
Z+jets divided per jet multiplicities 0j/1j/2j

wvieri:


bendavid:
Don't you also need the run cards?

wvieri:
@bendavid  I need just one run_card, right?

bendavid:
It can be the same, but to actually make the gridpacks you need one for each gridpack with the proper name/location.

wvieri:
added

bendavid:
only see 2 run cards

wvieri:
done

covarell:
They seem correct to me now

bendavid:
Yes, looks good.

Note that maxjets in the pythia configuration has to be set to "2" for all three of these samples.

PR #640, opened 2015-09-25
Adding the data cards for GJets with minimum distance between jet and photon = 0.4

parbol:


trandbea:
Could these be merged? Thanks!

bendavid:
etaa 2.5 is not really safe with respect to the detector acceptance.  Should be 2.6 at least.  (Probably this was also the case with the previous Gjets cards unfortunately)

More generally, can you explain a bit what these are for?  Are these intended to completely replace the samples with more inclusive cuts?  To add more statistics?  Are there studies which show that the 0 < dr < 0.4 region is not relevant after selection?

trandbea:
Yes, you are right that 2.5 is a bad choice and indeed legacy, but hopefully our resolution is good enough that this is a small effect?

As for the change in DR. These are intended to replace the current GJets sample. The problem is that analysis are interested in isolated photons and the sample with draj >0.05 has large fraction of the stats in draj (0.05, 0.4). This is exacerbated by the fact that the pT cut was also lowered so for those interested in high pT isolated photons it's a double hit in stats.

bendavid:
Ok, but the first question is whether the (0.05,0.04) phase space really has a negligible contribution to the selected events.  Otherwise that phase space still needs to be covered (then we can discuss the best way to do it while ensuring adequate statistics)

For the eta cut, the issue is not really the resolution, but rather the correspondence between physics eta and detector eta, which depends on the z position/spread of the primary vertex (and introduces an effective smearing on the order of at least 0.02 in eta)

trandbea:
ok, i am afraid i don't have more precise info on the DR-cut impact, so I should let Pablo/others comment...

gzevi:
The phase space (0.05, 0.4) is very small after photon isolation cuts, while it completely dominates the sample before isolation cuts. The actual DR distribution depends entirely on the HT required in the event, but since the pt cut in this sample is 25 GeV, the peak at low DR is very large.

As far as I know, this phase space is rarely included in GammaJets simulation (and it was not included in previous versions of these samples).
Instead, QCD samples are used to simulate the "fragmentation photons" inside of jets. 

The reason for this request is to undo a past change which, by introducing the (0.05, 0.4) region in these samples, resulted in a much lower statistical power when studying isolated photons in high HT events. 

Sorry, writing from a phone...

gzevi:
In the MT2 analysis , we estimate the fragmentation photon (DR<0.4) contribution after full analysis selection to be ~8%, based on QCD samples.

bendavid:
Ok, but small angle QED FSR from an outgoing quark is not the same as fragmentation photons.
QCD samples of course also include small angle QED FSR from the parton shower, so if this is consistently included from the QCD sample as well, then that can be a reasonable solution.  (But it implies a more complicated overlap removal scheme)

gzevi:
Our current practice is to take prompt photons with DR<0.4 from QCD, the rest from GammaJets. 
I believe this was also done in the past, but I might be wrong.

What is the alternative strategy you are suggesting?

bendavid:
Well if the fragmentation photons are relevant they anyways have to come from the QCD sample.
The problem is that any component which comes from the qcd sample will have a large relative weight because of the much larger cross section (ie likely the equivalent integrated luminosity is less.)

So in choosing the delta R cut there's a tradeoff.  Choosing small value for the draj cut in the gamma+jets sample wastes statistics in the gamma+jet sample as you have observed, but reduces the phase space/number of events which will come from the likely poorer statistical power QCD sample.  If this region is really ~completely removed by the isolation cuts, then indeed a larger draj cut makes sense.

gzevi:
Is it just a question of statistics?
Has it been demonstrated that the Matrix Element calculation can describe successfully this part of phase space (DR < 0.4, effectively photons inside of jets)?
In terms of statistics, I would prefer to have good statistics for the bulk of the sample, and a low stat sub-section, than overall low statistics.
In terms of practicality, I would also prefer to require DR > 0.4, rather than adding a factor of 10 statistics to the current samples to achieve comparable statistics to Phys14 in the DR>0.4 region.

gzevi:
On the other hand, it would be nice to have a better understanding of this region (DR<0.4), so if you have a suggested strategy, or if you think the Matrix Element calculation could really take care of the whole thing, perhaps we could invest in a large production specifically designed to study the DR < 0.4 region. Is this what you are suggesting?

bendavid:
Well, whatever cut is put on the matrix element, you are effectively relying on the parton shower for the complementary phase space.  Not sure if there is a detailed study of the crossover behaviour, especially for the case with several jets in the matrix element.

In any case, efficient use of the statistics would imply a tighter cut than 0.4 for the overlap removal at analysis level if generating the gamma+jet matrix element with draj=0.05.

On the other hand, if you want to use dR<>0.4 as the separation point, it may still be safer to use an intermediate draj in the matrix element generation (maybe 0.3) to make sure there are no phase space gaps.

Using dr<>0.4 to seperate gamma+jets and qcd with the current samples should be ok, but inefficient statistically.  I think the question is whether the statistical power of the QCD component is ok when using dR<0.4 to select it.  Ie do you have too many events with too large relative weight from the QCD sample as it stand?  In that case just changing the cuts for the gamma+jets sample will not solve that part of the problem clearly.

bendavid:
Since you already have now the samples with a small draj cut, you can in principle test what happens when you modify the analysis level dR cut used to separate the QCD from the gamma+jet sample (both in terms of the combined distributions and the statistical power of the combined sample)

You could think of this a little like xqcut and qcut by the way.  draj in the run card plays the role of xqcut (the matrix element cutoff in the dr(photon, quark) distribution) and your dR<>0.4 cut plays the role of qcut, and needs to be sufficiently looser than draj in order to avoid edge effects from additional QCD ISR/FSR.  (how exactly are you calculating the dR to cut on to separate the samples btw?)

gzevi:
I have looked at the new samples, and found the prediction from GammaJet at DR (0.05, 0.4) to be a factor of 2 or 3 smaller than the QCD ones in the same region, as expected since the Matrix Element should not model well the low-DR region. 

So again: we now have a sample that is dominated by photons at low DR that don't pass the isolation cuts, and even when they do pass it, they make useless predictions, since they don't include fragmentation effects (and  -as you say above- fragmentation photons have to come from the QCD samples anyway). Hopefully one of these statements is wrong, because otherwise I don't understand what we are discussing. 
Perhaps you could point me to the studies and plots motivating the DR change from 0.4 to 0.05, which might clarify some of these issues for me? And they might also contain some answers to the questions you asked above. 

>> I think the question is whether the statistical power of the QCD component is ok when using dR<0.4 to select it. Ie do you have too many events with too large relative weight from the QCD sample as it stand? <<
No, we are not talking about the QCD statistics here. I am trying to solve the problem with the GammaJets sample statistics for isolated photons.

bendavid:
So, I'm not sure it is expected that the gamma+jets would give a smaller preidction for DR(0.05,0.4) than the QCD sample.  I'm also not sure which prediction is more correct in that region.  (it's not clear what exactly constitutes "low DR" in this context.)

The point about the QCD statistics is that IF the qcd prediction gives a large contribution with poor statistical power in the dR(0.05,0.4) region, then this could be improved by simply making the transition point something smaller than dr=0.4 (and therefore requiring some draj<0.4 for the production of the gamma+jets sample).

If the prediction of the QCD sample in this region is not statistically problematic, then at least from the statistics standpoint I agree with you that splitting the samples at dR=0.4 is a good solution and that draj in the run cards for the photon+jet generation should be increased (but probably not all the way to 0.4, since this risks some small phase space gaps).

bendavid:
(probably we should discuss further in person)

gzevi:
I'm happy to meet any time (my skype is giovannizdp, to coordinate). But I would like to get this request moving forward in the meantime, since the DR>0.05 sample already exists, and effectively we are simply asking for more statistics for the region with DR>0.4. 
Would that be possible?

bendavid:
Ok, discussed with @gzevi and ok to go ahead with draj=0.4, but the etaa cut should definitely be updated.

bendavid:
As discussed with SUSY group, we go ahead as is to save time since the gridpacks with the current cards already exist.  (But analyzers to be aware of possible boundary effects)

PR #641, opened 2015-09-25
ttWJets and ttZJets LO sample config

mharrend:
Datacards for a ttW012jets_5FS_LO_MLM sample and a ttZ01jets_5FS_LO_MLM sample


govoni:
@mharrend are you ok with mll > 10 GeV cut?

mharrend:
@govoni Yes, I assume so.

I was setting this value according to Giovannis email from 15th of September:
"Hi Pietro,

I had generated 8 TeV madgraph events by hand, but that was with a
different madgraph, without using madspin for the top decays, and I don't
know the procedures for preparing cards and gridpacks for the CMS
production system, nor what is the setup (magdraph model restrictions,
parameters, pdfs, cuts, weights, ...)
To me it would be much simpler if anybody who knows the rest of the
infrastructure can make them.
Eventually, for ttW + 0/1/2j it should be just
      define wpm = w+ w-
      generate p p > t t~ wpm @0
      add process p p > t t~ wpm j @1
      add process p p > t t~ wpm j j @2
and for ttZ + 0/1j, in order to get also the gamma* for the dilepton final
state one needs
      generate p p > t t~ z, z > j j @0
      add process p p > t t~ z, z > vl vl~ @1
      add process p p > t t~ l+ l- @2
      add process p p > t t~ z j, j > j j @3
      add process p p > t t~ z j, z > vl vl~ @4
      add process p p > t t~ l+ l- j @5
and set the m(l+l-) cut to 10 GeV
and in the end one should tell madspin do decay the tops inclusively."

I even send the configuration later around and let it get checked. Since the request is according to Cristina rather urgent and I am already waiting to get the corresponding gridpacks finished I would keep this parameter since decreasing it could also lead to problems with generating the gridpack.

BTW: I was mistaken that Cristina is the SUSY MC contact, so of course I can also push the request through McM instead of you.

mharrend:
Dear all, 
I am getting problems due to using MadSpin for the top decay.:

##############################
INFO: decay channels for t : ( width = 1.476401 GeV )
INFO:        BR                 d1  d2
INFO:    1.000000e+00            b  w+
INFO:
INFO: generating the production square matrix element
INFO: generate p p > t t~ z, z > j j @0  --no_warning=duplicate;add process p p > t t~ z, z > vl vl~ @1  --no_warning=duplicate;add process p p > t t~ l+ l- @2  --no_warning=d$
INFO: Done 114.6
INFO: generating the full square matrix element (with decay)
INFO: generate p p > t t~ z, (t~ > b~ w- , w- > all all  QCD=99), (t > b w+ , w+ > all all  QCD=99), ( z > j j , (t~ > b~ w- , w- > all all  QCD=99), (t > b w+ , w+ > all all $
^[[1;31mCommand "launch" interrupted with error:
KeyError : 23
Please report this bug to developers

           More information is found in 'MS_debug'.

           Please attach this file to your report.^[[0m
MadSpin>

rm: cannot remove `/pool/lsf/mharrend/702450951/ttZ01j_5f/ttZ01j_5f_gridpack/work/unweighted_events.lhe.gz': No such file or directory
#############################################

Since I am running the job on the lsf batch system I cannot access the MS_debug file or at least I do not know how to access it.
Any ideas how to fix this?

mharrend:
Sorry, some lines of the error message were truncated. Here is the full error log:

INFO: decay channels for t : ( width = 1.476401 GeV ) 
INFO:        BR                 d1  d2 
INFO:    1.000000e+00            b  w+  
INFO:    
INFO: generating the production square matrix element 
INFO: generate p p > t t~ z, z > j j @0  --no_warning=duplicate;add process p p > t t~ z, z > vl vl~ @1  --no_warning=duplicate;add process p p > t t~ l+ l- @2  --no_warning=duplicate;add process p p > t t~ z j, z > j j @3  --no_warning=duplicate;add process p p > t t~ z j, z > vl vl~ @4  --no_warning=duplicate;add process p p > t t~ l+ l- j @5  --no_warning=duplicate; 
INFO: Done 114.6 
INFO: generating the full square matrix element (with decay) 
INFO: generate p p > t t~ z, (t~ > b~ w- , w- > all all  QCD=99), (t > b w+ , w+ > all all  QCD=99), ( z > j j , (t~ > b~ w- , w- > all all  QCD=99), (t > b w+ , w+ > all all  QCD=99)) @ 0  --no_warning=duplicate;add process p p > t t~ z, (t~ > b~ w- , w- > all all  QCD=99), (t > b w+ , w+ > all all  QCD=99), ( z > vl vl~ , (t~ > b~ w- , w- > all all  QCD=99), (t > b w+ , w+ > all all  QCD=99)) @ 1  --no_warning=duplicate;add process p p > t t~ l+ l- , (t~ > b~ w- , w- > all all  QCD=99), (t > b w+ , w+ > all all  QCD=99) @ 2  --no_warning=duplicate;add process p p > t t~ z j, (t~ > b~ w- , w- > all all  QCD=99), (t > b w+ , w+ > all all  QCD=99), ( z > j j , (t~ > b~ w- , w- > all all  QCD=99), (t > b w+ , w+ > all all  QCD=99)) @ 3  --no_warning=duplicate;add process p p > t t~ z j, (t~ > b~ w- , w- > all all  QCD=99), (t > b w+ , w+ > all all  QCD=99), ( z > vl vl~ , (t~ > b~ w- , w- > all all  QCD=99), (t > b w+ , w+ > all all  QCD=99)) @ 4  --no_warning=duplicate;add process p p > t t~ l+ l- j , (t~ > b~ w- , w- > all all  QCD=99), (t > b w+ , w+ > all all  QCD=99) @ 5  --no_warning=duplicate; 
Command "launch" interrupted with error:
KeyError : 23
Please report this bug to developers

           More information is found in 'MS_debug'.

           Please attach this file to your report.
MadSpin> 
 
rm: cannot remove `/pool/lsf/mharrend/702450951/ttZ01j_5f/ttZ01j_5f_gridpack/work/unweighted_events.lhe.gz': No such file or directory

bendavid:
There are two issues
1) proc cards are missing the model import (so they would be using "sm" by default with massive b's)
you'll need "import model sm-ckm_no_b_mass"  or " import model sm-no_b_mass"

2) madspin doesn't support processes which have already been defined with decay chains.  So the the processes with z > j j and z > vl vl~ won't work as is.  Probably this means at least the hadronic z decays will have to be split in a separate sample from the rest.

covarell:
@bendavid For 2), could it be a viable solution to remove the statements z > j j , z > vl vl~ from the process card, and write z > neutrorjet neutrorjet in the madspin card, where neutrorjet is a multiplarticle tag of neutrinos and jets?

bendavid:
That would result in the wrong relative fraction of events from the different processes (because "generate p p > t t~ z" will take the full cross section of ttZ without taking into account any branching ratio)

Madspin may or may not be able to correct reject events later to compensate.  (and even if it did, it would cause problems with the externalLHEProducer)

So I think we can't mix madspin and non-madspin decays of the z in the same sample.

mharrend:
I removed the MadSpin cards in both processes, so we should have the run 1 config again.

bendavid:
Proc cards are still missing the model import.

Also are you sure you don't want to use madspin?  It can be fixed just by skipping the decay chain syntax as said.  (But probably means you would have to split the ttZ into two samples)

mharrend:
Ah sorry, I updated my local cards already in this regard, but forgot to update the merge request in this regard.

I tried, e.g., to get a ttW gridpack using MadSpin many times and always got some problems during the MadSpin step. Since the depending samples are rather urgent, thus, I would produce the two samples in the old way for now.

bendavid:
Ok,
Maybe better to at least decay the W's in madgraph using the decay chain syntax then?

mharrend:
I am not familiar with the decay chain syntax. Would the right definition of the ttW proc card be then something like the following?

import model sm-ckm_no_b_mass

define p = p b b~
define j = j b b~

define wpm = w+ w-
generate p p > t t~ wpm, w+ > j j, w- > j j @0
generate p p > t t~ wpm, w+ > l+ vl, w- > l- vl~ @1
add process p p > t t~ wpm j, w+ > j j, w- > j j  @2 
add process p p > t t~ wpm j, w+ > l+ vl, w- > l- vl~  @3
add process p p > t t~ wpm j j, w+ > j j, w- > j j @4 
add process p p > t t~ wpm j j, w+ > l+ vl, w- > l- vl~ @5

output ttW012j_5f -nojpeg

mharrend:
@bendavid I have now added the w decay in ttW and further improved the ttZ proc card.
Can we merge the request now?

bin/MadGraph5_aMCatNLO/cards/production/13TeV/TTZJets/TTZJets_5f_LO_MLM/ttZ01j_5f_proc_card.dat
```diff
@@ -0,0 +1,15 @@
+import model sm-ckm_no_b_mass
+
+define p = p b b~
+define j = j b b~
+
+generate p p > t t~ z, z > j j @0
+add process p p > t t~ z, z > vl vl~ @1
+add process p p > t t~ l+ l- @2
+add process p p > t t~ vl vl~ @3
```
bendavid: vl vl~ decays are double counted here.  (because you have it defined both with and without the decay chain syntax)

Suggest keeping only the t t~ z, z > vl vl~ version.

bendavid:
Few remaining issues

1) the double counting of z > vl vl~ decays I mentioned inline

2) l+ and l- by default are only e and mu (don't include taus).  If you want them then I suggest to define
ell+ =  e+ mu+ ta+
ell- = e- mu- ta-
and use those instead

3) For the decay chain syntax, the bwcutoff parameter in the run card does matter, and does act as a real cut.  If you want the full off-shell phase space (I think you do) then suggest to set this to 15000

mharrend:
@bendavid Thanks, updated the cards, is everything okay now?
I updated the bwcutoff parameter in both samples, hopefully this is okay.

PR #642, opened 2015-09-25
Added Pythia Cfg files

amagitte:
Hi all,

I have added the Cfg files for Pythia8, both in VBF and ggH channel and for all the mass points that we are planning to generate.

Cheers,
Alessio


bendavid:
There are git conflicts here apparently.

amagitte:
Hi Josh,

What I have done is to remove the old and useless Pythia cfg file  used during 1B campaign by ZZ—>2l2nu.
I removed the cfg file where the Higgs is decayed  by Pythia in ZZ and then the boson into 2l2nu, then I changed the name for the cfg file where the decay is done by JHUGen.
In the case of the cfg files for ggH I had a line to specify the number of particle.

Cheers,
Alessio

bendavid:
Ok, but to start with there is a git merge conflict in a technical sense.

Suggest to start from a clean area using the instructions in 
https://twiki.cern.ch/twiki/bin/view/CMS/GitRepositoryForGenProduction

and open a new pull request.

PR #645, opened 2015-09-29
missing sample WZTo2Q2Nu01j_5f_NLO_FXFX

perrozzi:


bendavid:
to be safe this should probably be done in mg 2.3.2.2 once the branch is pushed. (since it could be affected by the same factorization scale issue as the other WZ sample).
More news on that shortly.

bendavid:
Master has been updated to 2.3.2.2 now, so gridpacks can be generated using that.

Can you double check the value of ptj in the run card?  I think some of the diboson samples were generated with lower values (10 or 15)

perrozzi:
same as here
https://github.com/cms-sw/genproductions/pull/385#issuecomment-102870841
(but I didn't make the gridpacks yet, let me know what you prefer)

covarell:
This is NLO, so it should not matter much, for me they can be approved as they are

PR #646, opened 2015-09-29
Filters for EE and MuMu Single LQ production

dnash86:
Hadronizers for Single LQ production to filter out decays to neutrinos (In final state topologies of llj, lnuj, and nunuj, select only llj).

### DIFF
--- a/bin/Powheg/create_powheg_tarball.sh
+++ b/bin/Powheg/create_powheg_tarball.sh
@@ -19,7 +19,7 @@ echo "   ______________________________________________________    "
 repo=slc6_amd64_gcc481/powheg/V2.0/src
 echo "%MSG-POWHEG source repository = $repo"
 
-name=powhegboxV2_July2015
+name=powhegboxV2_Sep2015
 echo "%MSG-POWHEG source tarball name = $name"
 
 process=${1}
--- a/bin/Powheg/run_pwg.py
+++ b/bin/Powheg/run_pwg.py
@@ -302,7 +302,7 @@ if [[ -s ./JHUGen.input ]]; then
 fi
 
 ### retrieve the powheg source tar ball
-export POWHEGSRC=powhegboxV2_July2015.tar.gz 
+export POWHEGSRC=powhegboxV2_Sep2015.tar.gz 
 
 echo 'D/L POWHEG source...'
 


covarell:
Have you checked that status 23 is fine for all events? (I do not know what is going to be the source of events)

dnash86:
Sorry, I should have put more details (in this case, I am actually "the analyzer"):

The source for the events will be CalcHEP (1st gen and 2nd gen LQ models).  

To check that these filters were okay, I did the following:
Over the full scan of mass points (around 100 for L-handed and R-handed each), I examined 100 events.  Looking by hand, status 23 looked good.  

I also measured the filter efficiency for 100 events for each mass point.  The nominal efficiencies to expect would be 1.0 for the R-handed case and 1/3 for the L-handed case.  Over the full scan I got 1.0 in each case for R-handed (so it seems all the products are status 23 indeed) and some fluctuations around 1/3 (as expected for only 100 events per test).  

So everything seems good, if that sounds good to you.

bendavid:
Are the particles being filtered on produced directly by calchep, or by decays in pythia?

dnash86:
They are produced directly by CalcHEP.

bendavid:
Can you point me to an example LHE files?

This may not be safe, because sometimes pythia overwrite status 23 with status 1.
The simplest thing to do would actually be to filter instead directly on the LHEEventProduct

dnash86:
Ah, interesting, I did not know that pythia sometimes did that.  It seemed to all read out as status 23 to me.  

The LHE headers are here: https://github.com/cms-sw/genproductions/pull/639

The LHE files have been uploaded to the repository in eos, so picking randomly here, here's an example file:

/store/lhe/14904/SLQ_Lhanded-MLQ2100g1l0.6.lhe.xz

bendavid:
How many events did you check?  Did you verify that all events have the expected particles present in status 23?

bendavid:
Anyways you could probably easily switch to the generic lhe filter
https://github.com/perrozzi/cmssw/blob/ddea33b48a5fc55a5af49da2c46bb59ce2af9267/GeneratorInterface/GenFilters/python/LHEGenericFilter_cfi.py

dnash86:
I looked through a couple LHE files by hand.  But I think the more thorough test was running the filter on the R-handed ones, which should always have charged leptons only (/store/lhe/14821/SLQ_Rhanded-MLQ2600g1r0.8.lhe.xz for example).  In that test, I ran on 100 events ~100 times and got 1.0 for each mass as expected.

dnash86:
If thats preferred I could switch to that, run a few tests, and then update this PR.

bendavid:
I would suggest to switch to be absolutely safe.
In the future we can/should extend the generic filters to allow using the genstatus flags to be robust with respect to this issue.

dnash86:
Okay, will do.

dnash86:
Woops, accidentally closed this one with forgetting to do a commit before a push.  For some reason "Reopen and comment" isn't available...

PR #652, opened 2015-10-01
Change to LHEGeneric filter

dnash86:


dnash86:
From https://github.com/cms-sw/genproductions/pull/646#issuecomment-144315309

Rather than reopening the old one, sorry about the git confusion...

dnash86:
These filters (based on LHEGenericFilter) work fine, thanks for the advice.

bendavid:
Are these intended to be used with pLHE or wmLHE requests?

For wmLHE the "src" will have to be changed to "externalLHEProducer" in order for it to work.

dnash86:
These are intended for pLHE requests.  I had tested them, and they seemed to work okay locally.

bendavid:
Ok yes, they will work as is either locally (on lhe files) or with plhe requests.  If you want to use them with wmLHE requests you'll need to commit an additional version with the alternate "src" in that case.

PR #654, opened 2015-10-02
Forced offshell generation to be on shell (happens 1/10000 times)

violatingcp:
Fix to the Dark matter decay to force onshell decays of the mediator. 1./10k the JHUGen generation mass would shift to be offshell. This fixes production and has no effects on the final state event kinematics.


covarell:
Was this causing the JHUGen issue in production?

violatingcp:
Hi,

Yes, so there are two issues. One which is small, which can be fixed with
an updated JHUGen. The other which is bigger, this is the bigger fix.

I have remade the gridpacks with the bigger fix already here :

 eos ls eos/cms/store/cmst3/user/pharris/gen3/

Thanks,

Phil

On Fri, Oct 2, 2015 at 12:52 PM, Roberto Covarelli <notifications@github.com

PR #659, opened 2015-10-05
Update to madgraph_aMC@NLO 2.3.2.2

bendavid:
Also includes a change to how extra models are handled.

Extra model files must still be put in /afs/cern.ch/cms/generators/www/ as before, but now instead of being added to gridpack_generation.sh and always retrieved/unpacked, they are specified per-process using the extramodels.dat file as in
cards/examples/ggh012j_5f_NLO_FXFX


bendavid:
This also includes consistent fixes for both issues observed with low mll in diboson samples.  (Additional adjustment of lepton masses in NLO event generation for pythia8 is always skipped, issue needs further discussion for Herwig++ though)

The patch from Rikkert related to the factorization scale in the WZ samples is also included.
(see https://answers.launchpad.net/mg5amcnlo/+question/270111)

PR #660, opened 2015-10-05
HWW mass scan

rebecacern:
Low mass points for HWW mass scan. ggH, 2l2nu final state.


govoni:
@rebecacern the phase space is not folded at all (foldcsi etc), is ot ok like that?

govoni:
@rebecacern the negative weights are excluded:
withnegweights 0
is it done on purpose?

covarell:
@govoni for ggH these two settings were checked before by David Sperka and we know they are OK

govoni:
@covarell @rebecacern can't we keep also the negative weight events in the generation, to be consistent with the rest of the production and to be able and cross-check numbers in the future?

covarell:
The fraction has been checked to be ~10-5, so I think it just increases the complexity of the sample with no gain...

govoni:
ok, it's not relevant then. My call would be to keep the negative weights as default, to avoid possible problems in the future of copy and paste, to let us always re-check the negative fraction, etc, so if it's not too much of a burden I'd suggest to change it, otherwise we can revise this in the future generations.

govoni:
@rebecacern do you have time to keep negative weights, or are you so in a hurry that you want the fragments to be merged as they are?

rebecacern:
So my choice would be not to keep them, they are small and people tend to forget taking MC weights into account. But if you rather have the negative weights included always, I can re-do the full set of gridpacks for this and VBF and push them again.

govoni:
Foe vbf the neg weights are in. I prefer people to think that they always need to use weights in fact and the main reason for letting you redo gridlock would be to avoid copy and paste into future. Then I merge as is now, but for the next generations we'll put neg weights in.

rebecacern:
Ok, then I will do the ggH again. When the gridpacks are finished I will push them again.

PR #661, opened 2015-10-05
mono-Higgs ZpBaryonic and Scalar model cards

dburns7:
The MadGraph UFO tarball is in the top directory of each model. These are the next highest priority models for mono-Higgs, with other simplified and EFT requests delayed until further studies of parameters are done.


bendavid:
The tarball should not go in git, but rather in /afs/cern.ch/cms/generators/www

As per the new changes discussed in the gen meeting today, you'll also need to add the corresponding _extramodels.dat cards to use it with the current version of the scripts.

dburns7:
I have removed the model tarballs and added the extramodels.dat cards. Please let me know any additional revisions.

bendavid:
I don't actually see the model in 
/afs/cern.ch/cms/generators/www/Higgs_scalar_UFO.tar.gz

If you let me know where I can find it, can copy it there.

dburns7:
By "model" I assume you mean the madgraph model implementation, or UFO
directory. Is that file empty? I have attached the UFO tar for this model
here in case this is what you need.



Dustin Burns
Physics Department
University of California, Davis

bendavid:
Yes I mean the tarball which is referenced in the extramodels.dat
(if you attached it to the email, then it got dropped, because the github comment system doesn't support that I think.  Just put it in your public afs directory or something)

dburns7:
/afs/
cern.ch/user/d/dburns/work/public/MonoHiggs_models/Higgs_scalar_UFO.tar.gz



Dustin Burns
Physics Department
University of California, Davis

bendavid:
Alright, cards look fine except for one thing.  The output directory in the proc card needs to match the name of the cards.

eg for "Scalar_MZp10000_MChi1_proc_card.dat" you need to have
"output Scalar_MZp10000_MChi1 -nojpeg" rather than "output scalar"

dburns7:
I have made this change in the Scalar and ZpBaryonic model cards. Should I
proceed to producing gridpacks now?



Dustin Burns
Physics Department
University of California, Davis

bendavid:
yes, good to go now that it's merged.

PR #663, opened 2015-10-08
New Cards for EXO-RunIIWinter15pLHE RPV Stop

alefisico:
Continuing the issue here: #553 open by @davidsheffield , here are the new cards with the massless bquark and the new extramodels cards. 
I think I will only need to include the new model in the /cms-project-generators/ website. You can copy it from here: /afs/cern.ch/user/a/algomez/public/forDavid/RPVMSSM_af1_ufo.tgz
I modified it accordingly to https://answers.launchpad.net/mg5amcnlo/+faq/2312 adding a restrict_no_b_mass.dat 
Please let me know if the cards are good now in order to start creating the gridpacks. thanks


bendavid:
@alefisico I already see a file with the same name
-bash-4.1$ ls -lh /afs/cern.ch/cms/generators/www/RPVMSSM_af1_ufo.tgz
-rw-r--r--. 1 dsheffie zh 589K Jul 30 22:02 /afs/cern.ch/cms/generators/www/RPVMSSM_af1_ufo.tgz


so I've copied your model to
/afs/cern.ch/cms/generators/www/RPVMSSM_af1_ufo_extra_restrict.tgz

please update your extramodels.dat in the pull request accordingly.

bendavid:
Then in order to actually use the new resrict card, you'll need rather in the proc card
"import model RPVMSSM_UFO-no_b_mass -modelname"

alefisico:
actually I am having an error with the restrict cards in the version of madgraph of the latest gridpack_generation.sh:
```
Error detected in "import model RPVMSSM_UFO-no_b_mass -modelname"
write debug file MG5_debug 
If you need help with this issue please contact us on https://answers.launchpad.net/madgraph5
MadGraph5Error : Invalid restriction card (not same block)
	    set(['mass', 'wolfenstein', 'yukawa', 'sminputs', 'decay']) != set(['umix', 'snumix', 'msoft', 'msu2', 'vckm', 'fralpha', 'upmns', 'msd2', 'rvlamlqd', 'msl2', 'decay', 'tu', 'selmix', 'rvtlle', 'td', 'te', 'usqmix', 'rvtlqd', 'dsqmix', 'ye', 'yd', 'sminputs', 'yu', 'mse2', 'rvtudd', 'rvlamudd', 'nmix', 'imvckm', 'vmix', 'msq2', 'mass', 'rvlamlle', 'hmix']).
	    Missing block: umix,snumix,msoft,msu2,vckm,fralpha,upmns,msd2,rvlamlqd,msl2,tu,selmix,rvtlle,td,te,usqmix,rvtlqd,dsqmix,ye,yd,yu,mse2,rvtudd,rvlamudd,nmix,imvckm,vmix,msq2,rvlamlle,hmix
	    Unknown block : wolfenstein,yukawa
```
Do you have any idea what is happening? I just copy the restrict card from the sm model.
I'll write the madgraph authors asap

bendavid:
Ah, no, you can't copy it from the sm, you'll have to make one based on the param cards from your own model.

The error is just that there are some additional and/or missing parameters between the restrict card and the model.

alefisico:
ok, I just create a param_card from the model where the bquark mass is zero. I renamed that card and run a quick test. I think is working because I got this message:
```
MG5_aMC>import model RPVMSSM_UFO-no_b_mass -modelname
INFO: Restrict model RPVMSSM_UFO with file models/RPVMSSM_UFO/restrict_no_b_mass.dat . 
INFO: Run "set stdout_level DEBUG" before import for more information. 
Pass the definition of 'j' and 'p' to 5 flavour scheme.
Kept definitions of multiparticles l- / vl / l+ / vl~ unchanged
Defined multiparticle all = g u c d s b u~ c~ d~ s~ b~ a ve vm vt e- mu- ve~ vm~ vt~ e+ mu+ go su1 su2 su3 su4 su5 su6 sd1 sd2 sd3 sd4 sd5 sd6 su1~ su2~ su3~ su4~ su5~ su6~ sd1~ sd2~ sd3~ sd4~ sd5~ sd6~ t t~ z w+ h01 h02 a0 h+ sv1 sv2 sv3 sl1- sl2- sl3- sl4- sl5- sl6- w- h- sv1~ sv2~ sv3~ sl1+ sl2+ sl3+ sl4+ sl5+ sl6+ n1 n2 n3 n4 x1+ x2+ tau- x1- x2- tau+
```
The new tgz file is here:
/afs/cern.ch/user/a/algomez/public/forDavid/RPVMSSM_af1_ufo_extra_restrict.tgz

bendavid:
Since the old one hasn't been used for any production, I've updated it in place.

bendavid:
(by the way, the "pass the definition of 'j' and 'p' to 5 flavour scheme" message is a new feature of madgraph 2.3x to help users get consistent 4f/5f configurations)

alefisico:
Hi @bendavid 
I make a silly mistake by applying the HT selection in this process. I was producing a small sample to validate this process before I submit the generation and I realized that the HT selection here in the cards, since the stops do not decay in madgraph, are just calculated based on the extra jets.
I think the correct approach should be to then filter the events on pythia8, and more correctly after the matching step. What do you think?
Also, I was searching around the pythia fragments and I couldn't find an example of this filter, do you have some example in mind?
In any case, I think I will have to update this PR with the new cards.

bendavid:
So it's perfectly fine to do the filtering at the madgraph stage, assuming that you adjust your cuts to take into account the fact that the stop decays are not included in the counting.  There is no conflict between this and the jet matching.

On the other hand, it's true that filtering after parton shower/matching/hadronization will give you a closer correspondence to the real detector level quantities.  So if that is important then you could go that way.  I don't think we have such a filter already existing, so you would have to write it yourself (and add it to https://github.com/cms-sw/cmssw/tree/CMSSW_8_0_X/GeneratorInterface/GenFilters plus backport to 71x.

Simply relaxing the cuts in madgraph (but still applying them there) is still simpler and faster if you can find acceptable cut values.

alefisico:
If I got it right, madgraph calculate HT as the sum of the pt of quarks in the process. Since the process in the cards only have quarks coming from extra partons and not the decay of the stops, the ht selection is only applied to those quarks. Therefore, I think we cannot apply an ht selection in madgraph if the stops do not decay there. 
Then I think I will have to create a filter to do this selection after the matching.
what do you think?

bendavid:
Well it depends what you are trying to do.

The stop decay products will give HT on the order of the stop mass presumably, so you could just lower the HT cut in the run card accordingly.

alefisico:
I think this is true for the case where the stop are created at rest, but that is exactly the point of the low mass analysis where the stops are boosted and then I dont think we can just lower the HT cut in the card. Am I wrong?

PR #664, opened 2015-10-08
Llgamma sherpa

dnash86:
Sherpa data cards from Otman, and the python configs for them.
I am not very familiar with sherpa requests, is this sufficient for them?


thuer:
The compiled phase space results (Sherpacks) are missing, Otman should have them since the checksum is written in the python fragments. These tarballs have to be put into the folder where the pyton file is executed (either put them there by hand or store them using the Frontier Client) To use the Frontier Client store them at /afs/cern.ch/cms/generators/www/$SCRAM_ARCH/sherpa/2.1.1/13TeV/ and put the part after www into the python card under SherpackLocation and set the switch FetchSherpack to True

dnash86:
@thuer Thanks for the explanation.  Just one clarifying question about the sherpacks to be sent.  I am not sure which should be loaded.  I have been provided *_crdE.tgz (cards, maybe?), *_crss.tgz (cross sections I assume), *_libs.tgz (probably a library of some kind), *_logL.tgz (I assume runtime logs), *_migr.tgz (not sure), and *_MASTER.tgz.  I assume I should put the *_MASTER.tgz files in the file location you sent. 

Also, I assume $SCRAM_ARCH must match the architecture in which they are created?

thuer:
Yes the MASTER one (the others contain only one part each, pretty much what you expected, migr stands for multiple interaction grid, e.g. underlying event)
And SCRAM_ARCH should be the one which is used for the production (and creation of the sherpack, but I guess Otman took care to use the same CMSSW version as for official production)

bendavid:
@dnash86 do you have everything which is needed in the pull request now?

Can I ask about the binning?  From the file names it looks like 10...13000, 130..400, 400..13000

Was it intentional or did you want exclusive samples instead?

dnash86:
@bendavid  Sorry, no, I haven't updated the PR yet.  Will do that momentarily.

And yes, the binning was intentional.  The requester specifically wanted one inclusive bin, one medium pT bin, and one high pT bin.

dnash86:
@thuer Question regarding sherpack storage: I tried making a directory for the sherpacks, here:

/afs/cern.ch/cms/generators/www/slc6_amd64_gcc481/sherpa/2.1.1/13TeV/

I got a permission denied message.  I guess the directory doesn't exist and I do not have the permissions to make one there.  Should I ask someone (you, someone else?) to do so?

I will update the PR with the correct sherpack locations once I have them in place.  Thanks.

vciulli:
@dnash86 There is already a directory named 
/afs/cern.ch/cms/generators/www/slc6_amd64_gcc481/13TeV/sherpa/2.1.1
can you try to upload the sherpack there?

dnash86:
I can list the contents of that directory, but not make a subdirectory inside it to put the sherpacks in.

covarell:
Sorry, @bendavid @vciulli is there a reason why we should use the old afs generator web repository for sherpacks and not the new EOS -> cvmfs storage procedure?

bendavid:
The sherpa interface in CMSSW has not been updated to take the sherpacks from cvmfs.  Indeed this is something which should be done.

dnash86:
Is such an update possible in a short time frame?  Or should I be given permission to create folders in /afs/cern.ch/cms/generators/www/slc6_amd64_gcc481/13TeV/sherpa/2.1.1 in the meantime as a stopgap measure?  (I tried making a folder called EXO_llgamma in that directory, but could not due to permissions.)

thuer:
@bendavid @covarell could you point me to some example where the EOS-> cvmfs storage proceudure is used?
I do not think that this will be implemented fast, so it would be easier to use the current working approach for now, can you copy the tar files somewhere on afs and I  (or someone else) copy them to the www directory)

bendavid:
@dnash86 yes yes this request should go ahead from afs.

bendavid:
@thuer, cvmfs is mounted as a regular filesystem so you can just cp the file or untar it directly as we do in the madgraph scripts
https://github.com/cms-sw/cmssw/blob/CMSSW_7_6_X/GeneratorInterface/LHEInterface/data/run_generic_tarball_cvmfs.sh#L33

(how it works is that the whole /eos/cms/store/group/phys_generator/cvmfs area is automatically mirrored to /cvmfs/cms.cern.ch/phys_generator/ )

vciulli:
@bendavid @covarell @govoni will you request afs access right for @dnash86?

vciulli:
digging a bit more I just found that dnash is already in the group zh:gen_writers, but the latter has no permissions on 
/afs/cern.ch/cms/generators/www/slc6_amd64_gcc481
and subdirectories
I will write to cms.computing (we have no rights to change acl)

dnash86:
The issue is solved now.  

I've moved the *_MASTER.tgz sherpacks to that location, and updated the configs to have the location and FetchSherpack is set to true.

bendavid:
sorry looks like there is a merge conflict now.

bendavid:
(pull request now touches 148 files)

dnash86:
Aha, yes, I see that now (refreshed).  Must have done something weird with commit, was trying something.  Will redo now.

dnash86:
Seems to only touch the 12 files intended now, check from your end?

Sorry about that!

bendavid:
Is the sherpack location really supposed to contain the full absolute path on afs?  (as opposed to just the relative path under /afs/cern.ch/cms/generators/www) ?

thuer:
should be just the part after www, so beginning with slc6_....

dnash86:
Ah, yes, I missed that comment you made earlier, sorry.  Is it okay now?

PR #665, opened 2015-10-09
Cards from Varun

dnash86:
Pythia config cards for Photon+Jet samples binned in pT from Varun Sharma

### DIFF
--- a/python/ThirteenTeV/G_Pt_0to15_TuneCUETP8M1_13TeV_pythia8_cfi.py
+++ b/python/ThirteenTeV/G_Pt_0to15_TuneCUETP8M1_13TeV_pythia8_cfi.py
@@ -1,0 +1,35 @@
+import FWCore.ParameterSet.Config as cms
+
+from Configuration.Generator.Pythia8CommonSettings_cfi import *
+from Configuration.Generator.Pythia8CUEP8M1Settings_cfi import *
+
+generator = cms.EDFilter("Pythia8GeneratorFilter",
+                  comEnergy = cms.double(13000.0),
+                  crossSection = cms.untracked.double(8.269e+07),
+                  filterEfficiency = cms.untracked.double(1.0),
+                  maxEventsToPrint = cms.untracked.int32(0),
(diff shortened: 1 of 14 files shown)


bendavid:
Why is this needed given that we already have the HT-binned madgraph samples?

dnash86:
Not sure actually, I had thought the requester was aware of the HT-binned ones but upon double checking an email thread that may not be the case.  Checking now.

dnash86:
Okay, they are aware of it.  They use the single photon dataset for their analysis, with single photon trigger thresholds of 165, 175, 250, and 300 GeV.  The HT binned samples have a highest bin of HT=600-inf, I believe, so they do not have sufficient statistics at higher pT.

The analyzers point out another thing I was not aware of: The GJets sample does not have all prompt photon production modes on (only numbers 201,202, and 203).  For these, PromptPhoton is set to "all", which apparently includes 201-205.

bendavid:
That is intentional, since 204 and 205 are double photon production, which are covered by other samples.  We do NOT want to include this in the gamma+jet samples, since this makes their use by other analysis groups much more complicated.

And ok, if the only reason the madgraph samples are not good is because they don't go high enough, why not just request additional madgraph HT bins?

dnash86:
I'll link the requesters to this thread so they can explain further.

In general it seems that these samples are fairly standard and used by a few groups inside EXO.  Weren't there pT binned samples back in 2012, or not?

bendavid:
Yes, there were pt binned pythia6 samples in Run1.  

In principle they don't add anything on top of the madgraph samples though aside from different kinematic binning, and the madgraph samples should be preferred.

So if it's just a matter of statistics it would be better to extend/add bins for the madgraph.

govoni:
Any news on this? If not, I will close it tomorrow.

dnash86:
@govoni Sorry I left this issue dangling.  Yes, it seems we will likely be going another route with this issue (not using pt-binned samples).  I'll close the request.

PR #666, opened 2015-10-10
reweights in gridpacks

mmozercern:
This is the prototype for getting madgraph reweighting running with gridpacks.

The changes to runcmsgrid_LO.sh and gridpack_generation.sh are straight forward: if a reweight card exists it's copied and used during the pilot.run as well as the proper generation.
examples/WW_aTGC_reweight contains an example config.

There's three things that are not 100% straight forward:
-> there is a bug that prevents the reweighting to work right with relative paths as reweight-directory.
  a patch is included. Alternatively, we could just not support the reweigth-directory functionality (forcing the use of the default directory)
-> madgraph recognizes the precompiled reweight code by running 'make', so that nothing will be recompiled if the binaries are up to date. However, cleaning the source files breaks this. I've added a line to the cleanup script that empties all makefiles, effectively suppressing the recompilation.
-> I've added the path for this bug: https://bugs.launchpad.net/mg5amcnlo/+bug/1504089 because it's necessary for my test-case.


bendavid:
Why do you need to add support for param_card.dat?  (This is intentionally not present to encourage the use of customizecards.dat instead)

bendavid:
For the issue in https://bugs.launchpad.net/mg5amcnlo/+bug/1504089 isn't the correct solution to just run with massless b's/consistent 5fs in the first place?

bendavid:
Have you discussed with the authors about the fix for the relative paths? (Would be good if this is eventually included upstream, but fine to keep it as a patch in the cms scripts for now)

Then finally for the make files.  For me the best solution would be that ./bin/madevent reweight supports an additional commandline option to skip compilation.  Then no hacking of the makefiles is needed.  Can you open a launchpad discussion on that one as well?

mmozercern:
So this is where that functionality has gone. I wasn't aware of that. 
I've tried to use customizecards.dat instead, but there appears to be a bit of a tricky issue:
parameters in param_card.dat are superseded by the items in the reweight file
parameters in customizecards.dat overwrite the items in the reweight file
So customizecards.dat cannot be applied when pre-compiling the reweight code.
It's not clear to me whether this is necessary, though. If there is anything in customizecards.dat that's relevant at compile time (opposed to runtime), this is going to be a problem.
At least for the EWDim6 example it appears to be Ok, but I'm not sure what else could show up in the customize snippet.

Concerning the relative path issue: I haven't raised that with the authors yet. (will do)

Concerning the 5f patch: In the current setup the param_card (or alterantively customize snippet) correctly sets the b-mass and yukawa couplings to zero and still requires the patch. Apparently it's necessary to rejigger the model as well? I will check back with Olivier on that.

Concerning the make-files:  I haven't raised that with the authors yet. (will do)

bendavid:
Yes, for either customize cards or param_card it is NOT in general safe to change masses or couplings between 0 or non-zero  (the reason is that this affects the behaviour of the model at process generation time)

Best way to handle this is to make sure your model has a "restrict-no_b_mass.dat" or similar file included.  (Can add a modified version of the model to afs which includes this if needed)

For customize cards are you saying this overwrites the parameter values in BOTH the param_card AND the reweight card?

mmozercern:
I've removed the 5f patch and set up a custom model for it instead. Where should the model file go?

For customize cards, I looked a bit more closely and things should probably be fine.
The stdin parameters overwrite the values in BOTH the param_card AND the reweight card. However, technically things are handled slightly diffferently, so that things are probably fine. In the reweight-step, the param_card is not read at all. Instead all the corresponding information is extracted from the lhe header, which contains a copy of the param card inclduing the customization. For the EWDim6 case I've been using as a test everything works fine by just not putting any customization in the reweighting step as all the relevant customization has been stored in the lhe header in the pilotrun. The big question is whether we allow anything in the customization that does not leave a trace in the lhe header. 

I'll see that I can get in contact with Olivier on the other remaining issues on Monday

bendavid:
No, by construction the lhe header contains the very final set of parameters used after all modifications
(model + restrict file -> param card -> additional set commands from customize card -> lhe header)

There is a corner case where madspin can be used with an alternate model from the generation (but this we can address separately, and in any case, usually it's possible to avoid madspin for LO generation)

The custom model should go in /afs/cern.ch/cms/generators/www/
(and then you can get the script to automatically load it by adding an "extramodels.dat" file as in https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/examples/ggh012j_5f_NLO_FXFX/ggh012j_5f_NLO_FXFX_extramodels.dat)

If you are just adding an extra restrict card to an existing model, then please prepare a tarball with just the extra restrict file.

mmozercern:
I just forgot to commit the extramodels.dat

However, I don't have permissions to put the model file into  /afs/cern.ch/cms/generators/www/

mmozercern:
Am 11.10.2015 um 18:38 schrieb Josh Bendavid:
>
> No, by construction the lhe header contains the very final set of 
> parameters used after all modifications
> (model + restrict file -> param card -> additional set commands from 
> customize card -> lhe header)
>
> There is a corner case where madspin can be used with an alternate 
> model from the generation (but this we can address separately, and in 
> any case, usually it's possible to avoid madspin for LO generation)
>
> The custom model should go in /afs/cern.ch/cms/generators/www/
> (and then you can get the script to automatically load it by adding an 
> "extramodels.dat" file as in 
> https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/examples/ggh012j_5f_NLO_FXFX/ggh012j_5f_NLO_FXFX_extramodels.dat)
>
> If you are just adding an extra restrict card to an existing model, 
> then please prepare a tarball with just the extra restrict file.
>
> —
> <https://github.com/cms-sw/genproductions/pull/666#issuecomment-147216854>.
>

Here's the model file in case that's easier than setting up the permissions.

Matthias

bendavid:
Email attachments don't work through github, sorry, put it on afs public area?

mmozercern:
On 10/11/2015 07:00 PM, Josh Bendavid wrote:
>
> Email attachments don't work through github, sorry, put it on afs 
> public area?
>
> —
> <https://github.com/cms-sw/genproductions/pull/666#issuecomment-147221207>.
>

Ok, It's on ~mmozer/public/EWdim65f.tgz

mmozercern:
Hi Josh,

I talked to Olivier here: https://answers.launchpad.net/mg5amcnlo/+question/270128
Solutions to both of the outstanding issues are forthcoming.
For the reweightpath, the solution is identical to the patch we have now and is going to end up in MG soon.
For the makefile issue Olivier suggested a new solution instead of a command-line parameter: automatic detection of the presence of the final object file. For my testcase this works fine. I have a  patch ready for this one too if we decide not to wait for the MG update.

Let me know how you want to proceed

bendavid:
Ok yes, fine to include this patch in the scripts for now then.

bendavid:
The tarball with the extra restrict file/card is available in /afs/cern.ch/cms/generators/www/EWdim65f.tgz, so you can add it to the extramodels.dat in your example

mmozercern:
That was already included

bendavid:
Ok thanks.
Two remaining "housekeeping" things then.

First, can you remove the commented lines relating to the param_cards (to discourage others from bringing them back ;)

Second, for the proc card in your example, can you remove everything before "import model EWdim6-EWdim65f"

everything before that is automatically generated by madgraph anyways and we want to discourage people copying at pasting it around.  (It's being done anyways in many cases for production cards, but let's not encourage it further by having it in the examples)

mmozercern:
On 10/12/2015 07:14 PM, Josh Bendavid wrote:
>
> Ok thanks.
> Two remaining "housekeeping" things then.
>
> First, can you remove the commented lines relating to the param_cards 
> (to discourage others from bringing them back ;)
>
> Second, for the proc card in your example, can you remove everything 
> before "import model EWdim6-EWdim65f"
>
> everything before that is automatically generated by madgraph anyways 
> and we want to discourage people copying at pasting it around. (It's 
> being done anyways in many cases for production cards, but let's not 
> encourage it further by having it in the examples)
>
> —
> <https://github.com/cms-sw/genproductions/pull/666#issuecomment-147465658>.
>

Sure

PR #670, opened 2015-10-12
CompHEP files for Wprime->t+b with mixed chirality

peiffer:


bendavid:
These look like they're using cteq6m (NLO) for the pdf.

Assuming it's still not convenient to interface comphep with LHAPDF, would still be much more consistent to use cteq6l1

perfilov:
Hello,
It is convenient to use LHADPF for CompHEP and we will use it for the ingoing anomalous samples but  we used CTEQ6M for the newest "mixed W' " samples because for previous samples for W' (left+SM; right) production (since Summer) we used CTEQ6M also.
We need mixed samples for the constraints on W' masses for the"mixed W' " case and also we need to include these "mixed W' " samples  in the combination  with the existed W' samples (which also use CTEQ6M) for the W' couplings constraints.  That is the reason why I didn't change the PDF set for the newest samples  - we can't combine, e.g., "left W' +SM" samples with CTEQ6M and "mixed W' " with another PDF set.
 Cheers,
        Max

bendavid:
Ok, why did we use CTEQ6M for those samples though?

perfilov:
We used the mormalization to the NLO CS for W' (with the help of some procedure), it was better to use CTEQ6M then. And, I used the same machinery as for old (7, 8 TeV) samples while preparing the new 13 TeV ones (with the changes in energy etc.)
It doesn't matter which PDF were used at generator level because one needs to reweight the samples with PDF4LHC recipe in the real analysis, am I right?

PR #673, opened 2015-10-14
[HIG-HH][bugfix] lambda_hhh cannot be exactly 0

OlivierBondu:


bendavid:
There should be no problem having couplings as zero BUT you cannot switch them between zero and non-zero in the customize card.

Proper solution is to add an additional restrict.dat file to the model.

OlivierBondu:
while I agree that this would be the clean solution to add this in the model, it means delaying further these 4 samples out of a bunch of 56...

can't we get this in while we check with the authors that the end-product is the same ?

OlivierBondu:
Looking in the default parameters of the model in `parameters.py`: `ctr` and `cy` have correct default values of 1.0 (SM values) but BSM parameters have non-zero default values: `(c2, a1, a2) = (-1., 1., 1.)`

details below:
```python
c2 = Parameter(name = 'c2',
                  nature = 'external',
                  type = 'real',
                  value = -1.0,
                  texname = 'c2',
                  lhablock = 'BSM',
                  lhacode = [ 30 ])

a1 = Parameter(name = 'a1',
                  nature = 'external',
                  type = 'real',
                  value = 1.0,
                  texname = 'a_1',
                  lhablock = 'BSM',
                  lhacode = [ 31 ])

a2 = Parameter(name = 'a2',
                  nature = 'external',
                  type = 'real',
                  value = 1.0,
                  texname = 'a_2',
                  lhablock = 'BSM',
                  lhacode = [ 32 ])

ctr = Parameter(name = 'ctr',
                  nature = 'external',
                  type = 'real',
                  value = 1.0,
                  texname = '\\text{ctr}',
                  lhablock = 'BSM',
                  lhacode = [ 188 ])

cy = Parameter(name = 'cy',
                  nature = 'external',
                  type = 'real',
                  value = 1.0,
                  texname = '\\text{cy}',
                  lhablock = 'BSM',
                  lhacode = [ 189 ])
```

bendavid:
Do you know what those parameters are?  To be fully safe it would be better to handle this with a restrict file in the model as discussed.

OlivierBondu:
yes, sorry for not writing the details, I agree the naming could be more explicit...:
  - `cy` is `yukawa_top / yukawa_top_SM`
  - `ctr` is `lambda_hhh / lambda_hhh_SM`
  - `c2` is anomalous quartic coupling top-top-h-h
  - `a1` is anomalous triple coupling gluon-h-h
  - `a2` is anomalous quartic coupling gluon-gluon-h-h

as for the restrict file, we are currently in touch with the authors, I'll update this thread.

OlivierBondu:
I have confirmation that the end result should be the same by switching between zero and non-zero in the customize card for all parameters but the one we are discussing here (`ctr`), which should never be zero to begin with.

It is true that we could speed up generation by adding these restrict.dat cards and not having all diagrams around, but it should not affect the results (and the process is simple, we do not expect significant changes in speed).

OlivierBondu:
up, any further modifications that would be needed ?

PR #674, opened 2015-10-15
Configs from Begona

dnash86:
These are the configs for the M_100 bins for WToLNu.

### DIFF
--- a/python/ThirteenTeV/WToENu_M_100_TuneCUETP8M1_13TeV_pythia8_cfi.py
+++ b/python/ThirteenTeV/WToENu_M_100_TuneCUETP8M1_13TeV_pythia8_cfi.py
@@ -1,0 +1,28 @@
+import FWCore.ParameterSet.Config as cms
+
+from Configuration.Generator.Pythia8CommonSettings_cfi import *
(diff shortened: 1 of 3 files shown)


dnash86:
Could these be merged if there are no problems so I can push requests to Defined?

bendavid:
Two issues
1) there seems to be a third empty file added by the pull request

2) These lines should be removed, since they conflict with the default/tune settings
+            'Main:timesAllowErrors = 10000',
+            'Tune:ee 3',
+            'Tune:pp 5',

dnash86:
Accidentally closed, reopened with edits now

bendavid:
Looking back at the higher mass bins for this request, I see that the WToTauNu samples were produced both with and without Tauola.
(Logic was that without tauola is preferred to some extent since that is the default for most run 2 production, but Tauola was also needed for closer correspondence to the Wprime samples which still needed Tauola).

Would suggest to do the same here (add an additional card for WtoTauNu without tauola)

Will merge this as is not to further delay the other requests, but please go ahead and add the additional card in a new pull request.

(Btw, the issue with WPrimeToTauNau in pythia is now solved, so those samples can also be produced without tauola)

PR #676, opened 2015-10-16
[Madgraph5_aMC@NLO] ExoDiboson: add higher HT bins for DYJetsToLL and ZJetsToNuNu

syuvivida:
In order to have enough statistics to model the background of high mass Z'/Graviton, we need to add higher HT bins for the following modes:
DY(->LL)+Jets and Z(->nunu)+Jets

The number of events we would like to request is as follows:

DY+Jets:
600-800 GeV : 4M
800-1200 GeV: 1.5M
1200-2500: 250K
2500-Inf: 250K

Z(->nunu)+Jets:
600-800 GeV : 8M
800-1200 GeV: 3M
1200-2500: 500K
2500-Inf: 500K


bendavid:
I assume these are exact copies of the existing cards aside from the min/max HT?

syuvivida:
Yes, exactly.

PR #678, opened 2015-10-18
Cards for Single top t-ch 4f 8TeV MadGraph5_aMCatNLO+Herwigpp

rgoldouz:


bendavid:
Does the setscales.f start from the default version in mg_aMC 2.3.2.2 now used in production?

reweight_PDF should be true (and range of lhapdf id's properly set) to have pdf uncertainty weights.

PDF set is ct10nlo 4fs, which is fine for this sample.  I assume you prefer to be consistent with existing 8TeV samples rather than using nnpdf30 as for run2 production?

rgoldouz:
Hi,
Thanks for the comments,
I will update the setscales.f to the newest version of MG5 with new runcard and pull request again in one hour.

As you mentioned we want to be consistent with 8 TeV samples and used  ct10nlo 4fs for PDF and  we set reweght PDF flag to false because we don’t have proper 4f error sets for CT10.

-Reza

bendavid:
Ah thanks, did not realize that the CT10 4f error sets were not available.  Indeed no choice but to leave pdf weight false in that case.

rgoldouz:
Hi,
new  committed  setscales.f and card file is committed. I do not know if I should pull request again or it can be merged in this request.

-Reza

bendavid:
If you push to the same branch as on open pull request, it is automatically updated.

PR #681, opened 2015-10-20
removed madspin card for WWWJets

tmulholland:
Unable to generate gridpack with madspin decays for WWW sample. Will let pythia handle decays instead.


covarell:
What do you mean exactly with "Unable to generate gridpack"? Which errors do you get? 

Notice that if you let pythia do the decays, all W bosons will have a mass *exactly* equal to mW (no resonance width), this is the only drawback. If you are fine with this, I will accept the PR

tmulholland:
Hi,

I continually receive an AssertionError. I've tried to debug but 
Analyzers would rather have the sample with pythia decays instead of 
waiting any longer.
I've been told that we are content with pythia decays.

-Troy

PR #689, opened 2015-10-22
Dark matter cards from Renjie W.

dnash86:
More dark matter madgraph cards, I put them in the same directory as a previous request from the same group.


covarell:
OK, all proc_cards look like they have a carriage return in "output... no -jpeg" marked with a backslash, is this syntax accepted in Madgraph (i.e. have you tried the cards as they are?)

dnash86:
I just fixed the syntax, rather than "risk" it.

PR #690, opened 2015-10-22
W jets fully hadronic

perrozzi:
same as ZJetsFullyHadronic_HT180_LO_MLM, just W instead of Z


covarell:
Cards OK, I just wonder if the McM sample name should reflect many other requirements in the run card (single-jet pT, dijet mass etc.)

perrozzi:
Ciao Roberto, indeed this is a VBF oriented sample and has some tailored cuts. For the Z case I used ZJetsFullyHadronic_HT180_LO_MLM and probably would be good to use the same for W, for consistency. I will keep it in mind for the next cases
(please let me know if you want to change the name instead)

PR #691, opened 2015-10-23
Hgg signals new masses

mmachet:


covarell:
Is there any other modification w.r.t. standard cards except the new Higgs mass?

mmachet:
I would say no

2015-10-26 11:33 GMT+01:00 Roberto Covarelli <notifications@github.com>:

PR #694, opened 2015-10-27
Cards for EXO-RunIIWinter15pLHE RPV Stop without HT selection

alefisico:
Same as #663 without HT selection. 
I am creating a filter for events after pythia and jet matching to filter the events in HT. I will add a fragment with this filter as soon as possible.


bendavid:
I thought these were going to be wmLHE in the end?

Is it intentional that this pull request is removing some files completely?

alefisico:
Hi  Josh

the files that I am removing are for very high mass that we are not sensitive at 13TeV, so no reason to generate those mass points.

And for the generation, I am removing the HT selection in the madgraph cards because we will create a filter after matching to keep events with HT larger than 500. I am testing this filter now but I am experiencing some crab problems. As soon as it is tested, I will send the request.

cheers,

PR #701, opened 2015-10-29
Removed WVA 5f combined process, added separate WWA/WZA 4f processes

jfaulkne:
We fixed problem with WVA Madspin - needed to separate WWA/WZA processes due to different BR for W > j j and Z > j j (Madspin reweighs events after decay using a single value, could not merge different BR, thus crashed).  Also modified some cuts to better suit WVA analysis.


bendavid:
Few things.

1) You seem to have switched from 5fs to 4fs.  Cards look consistent, but was there a reason/need to do that?

2) the loop_sm model does not have the off-diagonal ckm elements.  Again was there a particular reason for that?

(did you run into the same issues with the WW processes as concerns top-related diagrams appearing?  In that case 4fs + diagonal ckm does make sense as the solution)

jfaulkne:
Yes, in fact it was this process (WVgamma, V > j j) that initiated the
discussion on HN and the semileptonic WW group was able to find the
solution, as it pertained to them as well.  Following that discussion, we
switched to 4fs and loop-sm.

-James

bendavid:
Ok sounds good.  Last question, looks like bwcutoff has been changed in the run card from 15 to 30.  What's the reason for this?  Note that this will only have a marginal effect (will change for which events in the intermediate W/Z appears for the leptonic decays).  The full off-shell phase space will anyways be included for the leptonic W/Z either way, so probably better to leave it as default.  (The hadronically decaying W/Z will instead always be approximately on-shell)

jfaulkne:
Since we will be using the dijet mass (including checks using the sideband
region outside of the W/Z mass window) in the analysis, 30 is better than
15. For example, for Z>jj we will have Mjj in [90-30*2.44, 90+30*2.44], or
~[18,162] GeV for bwcutoff=30,  while [53, 127] for bwcutoff=15. We
anticipate that the latter might bias us when we perform some sideband
checks or techniques.

bendavid:
I think there is some misunderstanding about the meaning of the parameter.
bwcutoff is NOT a cut, but rather a technical parameter which divides the "on-shell" vs "off-shell" phase space.

For the process as you have defined it here, the FULL off-shell phase space is included for the leptonic decay irrespective of the value of bwcutoff.  This setting will have only a minor effect on the treatment of the l nu system in the parton shower, and also change a bit how the generator history is written.  For this reason it is suggested to just leave it as the default 15.

qliphy:
bwcutoff indeed can affect the V->jj offshell-ness.  You can check here
https://twiki.cern.ch/twiki/pub/CMS/PKUWVA15Spring/compare-bwcutoff.png
where we compared the mjj distributions for bwcutoff=15 or 100, for the process pp>Za, Z>jj.

bendavid:
That is a different case because the decay chain syntax was used.  In that case bwcutoff really does act as a cut.

jfaulkne:
We have decided to go to bwcutoff = 15 (default value), as we no longer believe it is necessary to increase it.  Since we are using MadSpin to produce the decay chain, we have verified that changing bwcutoff between 15 and 30 has no measurable effect on the V > j j distributions.  Thanks for helping point this out.  I have pushed the changes.

PR #702, opened 2015-10-30
POWHEG gridpack configs for ttH

mharrend:
* Updated MH 125 GeV config to set correct Higgs width and include negligable negative event weight fraction
* Moved MH 70 GeV data card and set some options like in MH 125 GeV
* Created new configs for some MH based on MH 125 GeV data card

Please check one config just to make sure that everything is up to date.


covarell:
@syuvivida @yuanchao could you check these cards, especially because in the script I see that ttH has a special treatment.

@mharrend, if the gridpacks were not done yet, it will be good if you could beta-test our brand new powheg scripts and instructions:
https://twiki.cern.ch/twiki/bin/viewauth/CMS/PowhegBOXPrecompiled

mharrend:
Hello Roberto,

the input cards should be in principle correct since I am reused the old
configurations. Nonetheless, it is good to get them checked.

I ran into some problems with the new Python script for the POWHEG
gridpack production and send already a mail to the GEN mailing list in
this regard.

Thanks,

Marco
-- 
Marco A. Harrendorf <marco.harrendorf@cern.ch>
KIT - IEKP 


Am Freitag, den 30.10.2015, 04:07 -0700 schrieb Roberto Covarelli:

syuvivida:
@mharrend 
  Just to double check, the PDF set 260000 is for 5F production. Does the ttH process allow 5F production?

Eiko

syuvivida:
Otherwise, all other parameters look OK to me.

PR #703, opened 2015-10-30
W gstar

perrozzi:


bendavid:
Suggest to remove the eta cut on the photons if statistics allows.  Otherwise at least increase it to 3.1 (or at LEAST to 2.6)

Also, suggest to use ckm_lepton_masses_no_b_mass
(this will make it a bit slower, but I think for 0,1,2 jets should be ok)

Should also add the $$ t t~ h
syntax to the process definition to be safe

bendavid:
Actually on second thought, the etaa cut is irrelevant, since there is no final state photon.  Suggest to just set it to -1 so as not to cause confusion

perrozzi:
done, please check

bendavid:
Sorry the model should still have sm at the front.
"ckm_lepton_masses_no_b_mass" -> "sm-ckm_lepton_masses_no_b_mass"

perrozzi:
done

perrozzi:
for the record ckm_lepton_masses_no_b_mass might slow down too much the process.
I submitted both WGstarToLL012j and WGstarToLL01j and the latter ended due to time limit reached on the 2nw queue, I don't expect the 012j case to be better. I submitted the 0j to check, but maybe would be good to use the diagonal model to allow for some extra jets, let me know what you think.

bendavid:
You submitted WGstarToLL01j  with diagonal or off-diagonal ckm?  When did you submit it?  It really shouldn't time out on the 2nw queue before 1 week at least.

perrozzi:
off-diagonal, yesterday (or 2 days ago). log attached, job report below


Job <export PRODHOME=/afs/cern.ch/work/p/perrozzi/private/git/Hbb/genproductions/bin/MadGraph5_aMCatNLO; gridpack_generation.sh WGstarToLL01j_5f_LO_MLM cards/production/13TeV/WGstarToLL01j_5f_LO_MLM > was submitted from host <lxplus0091.cern.ch> by user <perrozzi> in cluster <batch>.
Job was executed on host(s) <p05496706e37669.cern.ch>, in queue <2nw>, as user <perrozzi> in cluster <batch>.
</afs/cern.ch/user/p/perrozzi> was used as the home directory.
</afs/cern.ch/work/p/perrozzi/private/git/Hbb/genproductions/bin/MadGraph5_aMCatNLO> was used as the working directory.
Started at Mon Nov  2 12:06:54 2015
Results reported at Mon Nov  2 19:49:55 2015

Your job looked like:

------------------------------------------------------------
# LSBATCH: User input
export PRODHOME=/afs/cern.ch/work/p/perrozzi/private/git/Hbb/genproductions/bin/MadGraph5_aMCatNLO; gridpack_generation.sh WGstarToLL01j_5f_LO_MLM cards/production/13TeV/WGstarToLL01j_5f_LO_MLM
------------------------------------------------------------

TERM_CPULIMIT: job killed after reaching LSF CPU usage limit.
Exited with signal termination: CPU time limit exceeded.

Resource usage summary:

    CPU time   : 437336.00 sec.
    Max Memory :       422 MB
    Max Swap   :     11099 MB

    Max Processes  :       152
    Max Threads    :       217

The output (if any) follows:


---
According to the CERN-LSF setup the standard output of your job <*_LO_MLM >,
id <715599870>, has been moved to </afs/cern.ch/work/p/perrozzi/private/git/Hbb/genproductions/bin/MadGraph5_aMCatNLO/LSFJOB_715599870/STDOUT>.

perrozzi:
can't attach, log here http://perrozzi.web.cern.ch/perrozzi/LSFJOB_715599870.txt

bendavid:
Looks like you forgot to specify a queue for the worker jobs.  So the job ran in multicore mode, running on 32 cpu's in parallel on the one batch worker.

bendavid:
Probably would be fine if it would submit jobs.

perrozzi:
Ciao, I used the command
 ./submit_gridpack_generation.sh 30000 50000 2nw WGstarToLL0j_5f_LO_MLM cards/production/13TeV/WGstarToLL0j_5f_LO_MLM
so unless I remember wrongly the 2nw is specified.
I now remember that there is also the option to start the main job locally, with sub-job submission through lxbatch (though I might run out of space if the process is heavy)

bendavid:
you missed the final argument of the script. 
eg
./submit_gridpack_generation.sh 30000 50000 2nw WGstarToLL0j_5f_LO_MLM cards/production/13TeV/WGstarToLL0j_5f_LO_MLM 1nw

This means "submit the master job to the 2nw queue and have it submit the worker jobs to the 1nw queue"

As you did it before, you submitted the master job to the 2nw queue, but had it run any workers locally in that single batch job using all the cores of the worker node.

PR #706, opened 2015-10-30
External tarball + madspin

maierbenedikt:
Enabling generating events from an external tarball and decaying them with MadSpin.
For this, it just detects whether a $name_externaltarball.dat can be found among the input cards (see example here https://github.com/maierbenedikt/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/st_tch_4f_antitop_ckm_NLO_powheg_madspin), then generates the equivalent MG5_aMCatNLO process and saves all the necessary information from the pilotrun LHE's header in a txt file that is stored together with the external tarball, in the gridpack.

The runcmsgrid_NLO.sh then generates events from the external tarball, splices the header into the LHE and calls madspin.

Tested was:
	•	script works with a PowhegV1 tarball
	•	script works with a PowhegV2 tarball
	•	script still works when using it in the old style, without an externaltarball.dat, i.e. if one just wants to generate aMCatNLO events.

Not possible at the moment: decaying external LO events. (But modifications to extract the header from the pilotrun LHE in the LO section of the gridpack_generation.sh and to the runcmsgrid_LO.sh should be straightforward).

### DIFF
--- a/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
+++ b/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
@@ -424,6 +424,7 @@ if [ "$isnlo" -gt "0" ]; then
       gunzip ./Events/pilotrun_decayed_1/events.lhe.gz
       sed -n '/<MG5ProcCard>/,/<\/slha>/p' ./Events/pilotrun_decayed_1/events.lhe > header_for_madspin.txt
       mv header_for_madspin.txt $WORKDIR
+      gzip ./Events/pilotrun_decayed_1/events.lhe
   fi
   
   echo "mg5_path = ../mgbasedir" >> ./Cards/amcatnlo_configuration.txt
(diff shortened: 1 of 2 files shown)


maierbenedikt:
Did you have time to look at this yet?

I just received a request from some groups of analyzers who need inclusive t-channel samples. (Current ones have leptonic top decays). Since we wanna switch from aMCatNLO to Powheg as default, this would be the first use case of such a Powheg+MadSpin combination.

Thanks,
Benedikt

bendavid:
Ok, so looks mostly fine.

One suggestion I would have to save space and IO on the worker node, is to untar the external tarball into a dedicated subdirectory already in gridpack_generation.sh, instead of doing it every time on the worker node in runcmsgrid.sh.

This would also clean up the logic checking for tar.gz, tar.xz, .tgz, etc

maierbenedikt:
Thanks, done!

bendavid:
Why does the externaltarball.dat file need to be included inside the tarball?  Is it just for documentation/traceability purposes?

maierbenedikt:
Yes exactly. It could also be removed, but then I was thinking that maybe it would be good to have the information on the original location somewhere in the tarball.

PR #707, opened 2015-10-31
Wbb cards

trandbea:
Adding cards for Wbb @ LO + up to 2 extra partons.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/WBBJetsToLNu_5f_LO_MLM/WBBJetsToLNu_5f_LO_MLM_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/WBBJetsToLNu_5f_LO_MLM/WBBJetsToLNu_5f_LO_MLM_proc_card.dat
@@ -1,0 +1,19 @@
+import model sm-ckm_no_b_mass
+
+define p = p b b~
+define j = p
+define bb = b b~
+define l+ = e+ mu+ ta+
+define l- = e- mu- ta-
+
+# Specify process(es) to run
+generate p p > l+ vl bb bb / t t~ H @0
+add process p p > l+ vl bb bb j / t t~ H @1
+add process p p > l+ vl bb bb j j / t t~ H @2
+
+add process p p > l- vl~ bb bb / t t~ H @3
+add process p p > l- vl~ bb bb j / t t~ H @4
+add process p p > l- vl~ bb bb j j / t t~ H @5
+
+# Output processes to MadEvent directory
+output WBBJetsToLNu_5f_LO_MLM -nojpeg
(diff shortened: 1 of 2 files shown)


bendavid:
Is this intended to be used standalone, or combined with the existing HT-binned (and/or inclusive LO) samples?

trandbea:
they can't be combined, right?

bendavid:
Well it depends.  There are two options

1) 5fs with settings matched to inclusive/ht-binned samples.
In this case the Wbb sample along will be unphysical (because the jet matching will act on the b's and remove phase space below qcut)

2) 4fs, with kinematic cuts on the b's (pt and mbb) defined independently of the light jets/matching
In this case the sample can be used standalone, but cannot be easily combined with the inclusive/HT-binned

for 1) I think the sample may already exist/in progress from the Higgs group.

bendavid:
@perrozzi can remind the current status

trandbea:
So, right now I've defined it as in 1), but I thought in that case it can't be used with the inclusive since it would be double counting. What am I missing? Can you just do simple gen veto of events with two b's in the inclusive and then add them up? Isn't there still overlap with gluon splitting from parton shower?

bendavid:
Yes, you can do a simple veto to remove the double counting, but it has to be based on the LHEEventInfo directly.  Once that's done, there's no more double counting because the b's from the parton shower will have been removed by the jet matching itself.

There's another subtlety here, that if you want to cover as much of the b-content as possible, you shouldn't do
w bb bb
w bb bb j
w bb bb j j

but rather
w bb
w bb j
w bb j j
w bb j j j

The reason being that there's a strong correlation between the ME flavours and the parton shower.  (And in particular, if you have an event with only 1 b quark at matrix element level, then it means there is an initial state b, and you will get at least an additional soft b from the parton shower ~100% of the time)

(and For QCD this becomes
bb j
bb j j
bb j j j
)

Note that the above argument does not apply to the 4fs case, because in that case there is no initial state b (and in fact the single b case is impossible, except in the W case from off-diagonal ckm couplings)

bendavid:
Anyways, I would focus on the QCD request for now, because I'm pretty sure the Wbb case in 5fs is already taken care of by the higgs group as said.

bendavid:
There is a disadvantage to the 5fs approach though:
In this case you are bound to place the b cutoff/transition to the inclusive sample at the qcut value.

For the 4fs approach you can likely set much looser cuts on the b's (but you can't combine it)

For what the Higgs group is doing they have in the end three samples
1) inclusive W+jets
2) Wbb 5fs 
3) Special Wjets sample with ME b's vetoed, and parton shower b's enriched with a special filtering setup at generator level (to reuse the same lhe events multiple times to boost the efficiency)

In this way, the combinability is preserved, and also the statistics of the parton-shower b component is enhanced along with the ME b component.

perrozzi:
(sorry to come very late on this, I discussed earlier today with Ana.
for the record it will be interesting to produce the 4f as well)

govoni:
Any news on this?

perrozzi:
Ciao,
the samples with one single b at matrix element, complemented by the corresponding samples enriched at parton shower, have been succesfully produced in 74X.
In HBB we have started validating them but not finished.
We observed something we don't completely understand regarding the ps enriched samples, which are filtered by counting the number of b hadrons with status 2.
We will try to wrap up asap.

We didn't attempt yet to produce the 4F samples (in that case only a single sample is needed, starting from a ME containing at least an outgoing b quark)

I think that the SUSY group is using the samples double enriched at matrix element which they originally produced.

Let me know if you need more info

PR #708, opened 2015-10-31
BBJets cards

trandbea:
Request to add bb~ + up to 2 extra partons at LO.


bendavid:
The relationship between this sample and the HT-binned qcd is similar to Wbb vs W
(but here there is no existing effort from the Higgs group)

So again, consider 4fs standalone vs 5fs combinable (ie combination mandatory)

PR #710, opened 2015-11-02
QCD sample with VBF preselection for VBF H->invisible

chayanit:
These are the cards for QCD process with additional VBF pre-selection below

30.0 = dijet_pt  ! min pT of at least one jet pair
0.0 = dijet_eta ! min delta era of at least one jet pair
800.0 = dijet_mjj ! min invariant mass of at least one jet pair

Gridpack is also produced by adding these additional parameters in related files when proceeding 

Let me know if you have any comments

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/qcd_vbf/qcd_vbf_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/qcd_vbf/qcd_vbf_proc_card.dat
@@ -1,0 +1,43 @@
+#************************************************************
+#*                     MadGraph5_aMC@NLO                    *
+#*                                                          *
+#*                *                       *                 *
+#*                  *        * *        *                   *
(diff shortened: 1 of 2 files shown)


bendavid:
Can you remind what the cuts will be at genjet and/or reconstruction level?

You can speed up madgraph by setting
T = auto_ptj_mjj
together with 
0.0 = ptj

chayanit:
Hi Josh

there will be next step of generator filters to get two different samples
#1 at least one dijet with pT>40 Eta<4.8 DeltaEta>3.0 Mjj>1000 DeltaPhi<=2.15 
#2 at least one dijet with pT>40 Eta<4.8 DeltaEta>3.0 Mjj>1000 DeltaPhi>2.15

Run2 analysis haven't decided about the cuts at reco jets yet. In Run1, we cut dijet pT>45 GeV as a reference. 

OK. I will add this to speed up MG. Thanks for your comment.

PR #711, opened 2015-11-03
QCD sample with VBF preselection for VBF H->invisible

chayanit:
Re-commit the cards after fixed two parameters as suggested by Josh here #710 
I will also close that one when this one is approved. Thanks in advance.


bendavid:
Thanks,
Have you checked that these madgraph cuts are loose enough with respect to the generator filters so that you're not missing any phase space?

chayanit:
Hi Josh,

Yes. I confirm they are loose enough. 

Cheers,
Chayanit

PR #713, opened 2015-11-05
Adding CT14 NLO PDFSET as requested

yuanchao:


covarell:
I see CT10 is not removed (as @bendavid suggests)
However, please consider that many jobs are already close to the memory limit, I am afraid they will crash with other 56 weights

yuanchao:
As pull request #724, CT10 is replaced with CT14 as requested in all three scripts.

govoni:
Dear Yuan, could you please add one weight more to the events, corresponding to the central value of the CT10 pdfs? This is important to be able and perform proper comparisons to Run1.

covarell:
Pietro, see my comment above. Too many weights just crash the run.

bendavid:
OK but one more weight for the central value won't make much difference.

covarell:
Ah ok, sorry, I missed "central value". Yes, ok, I agree with having CT14 with all variations and just one for CT10.

yuanchao:
Done in #729

PR #714, opened 2015-11-09
turn off jet matching and fix the processes

chayanit:
could you also check if they look fine before approve? then I will submit batch jobs to generate gridpacks. thanks in advance.


chayanit:
Hi, could you merge these changes if there is no further comment? Thanks in advance.

bendavid:
Can you change the names so that these are no longer "MLM" (since this refers to the jet matching)

chayanit:
Ah yeah. I will pull new commit with names changed.

PR #716, opened 2015-11-10
split w+ and w-

perrozzi:
otherwise is not working...


perrozzi:
apparently I got another error while producing the gridpack. It is a compilation error before to produce the pilot run. May it be a site-related error or does it have something to do with something more fundamental? (for the record I am in sync with cms-sw/genproductions master)
Here is the log
/afs/cern.ch/user/p/perrozzi/public/WJetsFullyHadronic_HT180_LO_MLM_gridpack.log

covarell:
It seems that these two variables 'mqqvbf' and 'deltaeta2' are not recognized by madgraph:
https://github.com/perrozzi/genproductions/commit/5f833aac82a666b15a3454a520b23563ac333fb7#diff-8f0f3846069c7a26e3ce4b16377c71abR232
Are you sure that their name is correct (maybe they are available only in the NLO version)?

perrozzi:
Ciao Roberto,
I was able to use the very same settings to produce the Z fully hadronic at LO, 
https://github.com/perrozzi/genproductions/blob/48d71e08dc700e60f5ec5fa53a40e5cbd3965f4a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZJetsFullyHadronic_HT180_LO_MLM/ZJetsFullyHadronic_HT180_LO_MLM_run_card.dat#L232

I don't know if something changed with the new version....

covarell:
can you check the file where you get the compilation error if the naming has changed?

bendavid:
Are you sure these variables were actually used by the old version?

Maybe what has changed is more strict checking of the names...

covarell:
Yes indeed I was also wondering...

perrozzi:
so I asked the customer of the sample (@azzurrip) to have a look and indeed these 2 cuts are not needed nor they have an impact on the Z. I will simply remove them and try again to produce the gridpack

PR #717, opened 2015-11-11
Cards and fragment for EXO Seesaw Type III

davidsheffield:
From Kevin Feigelis.

The analyzers had a question about central production:

"We noticed an issue in our internal generation, and wanted to fix it before going for official production. The issue was that MadGraph and Pythia were using different points for the beam spot, roughly 46 cm part. This ended up drastically reducing our signal yield, since Pythia was using the wrong beam vertex from bosonic decay/hadronization. How do we know that this will not be an issue in official MC generation?"

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_SeesawTypeIII/SeesawTypeIII_M-140/SeesawTypeIII_SIGMA+SIGMA-HH_M-140_13TeV-madgraph/SeesawTypeIII_SIGMA+SIGMA-HH_M-140_13TeV-madgraph_customize_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_SeesawTypeIII/SeesawTypeIII_M-140/SeesawTypeIII_SIGMA+SIGMA-HH_M-140_13TeV-madgraph/SeesawTypeIII_SIGMA+SIGMA-HH_M-140_13TeV-madgraph_customize_card.dat
@@ -1,0 +1,4 @@
+set param_card mass 8000018 1.400000e+02 # Mtr0
+set param_card mass 8000020 1.400000e+02 # Mtrch
+set param_card decay 8000018 5.82234e-12 # Wtr0
+set param_card decay 8000020 5.82234e-12 # Wtrch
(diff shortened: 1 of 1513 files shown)


peterthomassen:
4.6cm (not 46cm)

bendavid:
The way things are set up for central production, both madgraph and pythia use (0,0,0) for the vertex, then the vertex smearing to produce a realistic beamspot is done after the fact by the CMS generator interfaces (and the 4 vectors + vertices of all particles are coherently adjusted at the level of the HepMCProduct).

Actually in principle madgraph does not have a concept of "vertex position" per se.  I have never tried to run it myself using a non-(0,0,0) vertex position in pythia.

peterthomassen:
Thanks for your input. What are the implications now for our generation request - can we be confident that beamspot inconsistency will not be an issue? Do you need further information from us?

bendavid:
There should be no implications for the request - everything should be consistent same as any other request. 

There are a few other issues/questions though.
There are some inconsistencies for what concerns 4fs/5fs (the process definition implies 5fs, but maxjetflavour=4 is rather for 4fs)
Either way the pdfs need to be updated to "lhapdf" and the appropriate lhapdf id depending on 4fs or 5fs (please see the example cards)

Does the model you are using have massive or massless b's?

peterthomassen:
We have fixed the 4fs vs 5fs inconsistencies (it is now all 5fs). We also changed the PDF set to lhapdf and put the desired ID. Our model has massive b's.

We have notified David already, who I think now needs to update the pull request. To my understanding, everything is ready for production then. Just in case - do you need anything else from us?

bendavid:
Hi,
Sorry, if the model has massive b's, then everything needs to be set consistently for 4fs.  (Since massive b's cannot be included in the proton definition consistently as implied by 5fs)

KTFeigelis:
Hi,
I have updated all run cards currently hosted on lxplus to reflect the inclusion of massive b's.

currently using 4fs, maxjetflavor=4, and lhapdf id # 263400

Cheers,
Kevin

peterthomassen:
Hi! Do you need anything else form us?

bendavid:
Also the proc cards need to be changed so that the p and j definitions do not include b b~

KTFeigelis:
I have updated the lxplus directory to remove b and b~ from all proc card files.

peterthomassen:
Do you need anything else from us / how do we proceed?

davidsheffield:
@bendavid can these cards be merged?

PR #718, opened 2015-11-11
added aMC@NLO cards for different low mass points for ggH, vbfH and V…

mmachet:
…H. The only thing that changes is the mass value


govoni:
@mmachet I don't recall by heart, is the width set by mg automatically?

mmachet:
I think yes, the width has to be specified just in powheg cards as far as I
know.

2015-11-12 8:06 GMT+01:00 Pietro Govoni <notifications@github.com>:

PR #720, opened 2015-11-12
hasan

hogul:
Hi, I am producing the "gridpack" for W+jets 13 TeV samples split in bins of jet multiplicity. I have the proc and run cards of 13 TeV samples. So I want to commit them here while jobs are running.

I am doing this first time, I hope I am doing everything properly. 

Hasan


govoni:
@hogul are you planning to merge these samples in the end?
If not, i think that the ickkw should be changed to 0, @bendavid could confirm it I guess.

bendavid:
These are intended to be combined I believe (and in that case ickkw=3 is correct)

hogul:
It was asked to me to produce the gridpacks with different jet multiplicity separately(I was just volunteer to produce them), so we can enhance statistics for wellnu012j... I believe these are intended to be combined.

hogul:
Hi @bendavid  and @govoni,

I just want to follow up this: I do not know how this process works but are there anything else I can do for approval of the pull request or should I just wait?

Best,
Hasan

PR #723, opened 2015-11-18
Fix lhapdf id in LHE file

yuanchao:
According to POWHEG source code, the PDF SET 'pdfsup' is set to '-1' as using internal herwig pdf's for showering. This patch gets PDF index from 'pdf_ndns' if the value of 'pdfsup' is '-1'. Though the PDF group index 'pdfgup' is untouched.


bendavid:
pdfgup should be set to zero when setting pdfsup to an lhapdf id (otherwise it could cause problems/confusion for programs trying to make automated use of this information)

covarell:
Sorry I did not know this. @yuanchao can you please open a new PR? The change should be trivial

yuanchao:
I didn't know either. The patch is now in #731.

PR #724, opened 2015-11-18
Replacing CT10 with CT14 for PDF reweights calculation

yuanchao:
The patch replaces the CT10 PDF weights calculations with CT14 PDF sets in the LHE production script. All three scripts to be packed in the gridpack tarball are modified.


govoni:
Do we have unit tests of all these procedures?

PR #725, opened 2015-11-18
Added support for condor cluster

kdlong:
Essentially this is just a two line change, but a bug in MadGraph
prevented cluster_queue from being set properly
(https://bugs.launchpad.net/mg5amcnlo/+bug/1517041). Also modified
cluster.py to print the error message when jobs are held. Have tested
with a few LO and NLO jobs on the Wisconsin condor cluster.


bendavid:
Why does cluster_queue need to be set to None for the condor case?

bendavid:
Also can you comment how well/not madgraph handles resubmission of failed jobs with condor?

kdlong:
I'm not sure that this is a general rule, but at Wisconsin we don't have specific queues. If you set this to a specific value, or leave the default value of "madgraph," for example, it adds a line like "(madgraph = True)" to the condor submission requirements which will never be satisfied.

bendavid:
Ah ok.
But then the present implementation wouldn't allow the user to specify a queue in case they needed to I guess.  (Ok, that can be improved later)

kdlong:
Agreed, I'm happy to take suggestions on organization. There is also a cluster_size variable that could be worth setting by the user. 

As far as "how well," MadGraph handles the resubmission of jobs, I don't know anything more refined to say than that it resubmits them up to the number of resubmit times you specify, I've never noticed anything not working properly here. It does immediately crash if a job gets held, however, which can be very annoying and is why I at least added the error message.

bendavid:
Ok thanks.  When condor puts a job in "HOLD" state, what happens to it?  Will it eventually fail, or is it possible that it starts running again?

kdlong:
Yes, it's possible to find out why the jobs where held, modify the conditions of the jobs (locate the files that aren't found, increase the memory request the job makes etc.), and then release them to run. Probably the most useful behavior here would be to notify the user, pause for the user to modify the job, and then check if it can continue. 

The particularly annoying thing MadGraph does is instantly delete the jobs from the queue before you can find our why the jobs where held.

kdlong:
I should add that a held job completely restarts once it is released, so crashing the program and restarting isn't always significantly more inconvenient than waiting for you to release the jobs.

bendavid:
Ok, got it, the important point is that this is not something which happens "randomly" because a worker node is misbehaving or whatever.  Those are the cases for which the resubmission is really needed.

kdlong:
Yes exactly.

covarell:
I think this needs to be rebased because of Josh's updates, sorry. It will not merge now

kdlong:
Hi Roberto, I'm not sure if I handled the rebase in the cleanest way possible, but the conflict should be resolved.

bendavid:
(Kenneth, sorry for producing this conflict btw.  I knew it would happen when I updated the madgraph scripts, but some of the other lsf issues fixed by those updates became rather urgent to fix)

kdlong:
No problem Josh. I guess the patch for the FxFx scale choice is now included in the new version of MG? 

For some reason the MadGraph authors haven't been able to produce the bug I was working around with the patch, but for me the change is necessary.

bendavid:
Yes, the update for the FXFX scale choice is included in mg_aMC@NLO 2.3.3, therefore it's removed from the cms patches.

covarell:
Re-base did not work, I see 1000 files changed. We will close this PR, please open a new one (sorry for the delay)

PR #732, opened 2015-11-24
Powhegv2 cards for t channel

maierbenedikt:


bendavid:
Why are you adding a runcmsgrid.sh here?

maierbenedikt:
because it features other PDFs (NF_4) than the ones stored in the default Powheg runcmsgrid.sh

maierbenedikt:
That's also why I asked when MMHT14_NF4 will be available in CMSSW. But we'll live with NNPDF and CT14 for now

covarell:
The CT14 weights should be already added in the default runcmsgrid.sh, if you take the head of genproductions. Do you confirm?

If so, please remove runcmsgrid.sh.

maierbenedikt:
Hmm the CT14 is, but the CT14_nf4 isnt. I also put in the CT10nlo_nf4 And the nf4 version of MSTW2008.

It is crucial to have those, because the ratio top over antitop is sensitive to the PDF, so we need as many of the important NF4 PDFs as possible

maierbenedikt:
Can this pull request be merged? 

Thanks,
Benedikt

covarell:
So I think in general it is not good/recommended to have separate runcmsgrid.sh for every process.
This script has been devised exactly to avoid this.

Would it be possible instead to have a runcmsgrid_nf4.sh in the main powheg directory, which covers reweighting in all cases that the user has selected a nf4 PDF as central value? (@maierbenedikt ou can start gridpack production in the meantime with your cards anyway, because they are correct, but I would like to keep the genproductions directory consistent for all users) @yuanchao @syuvivida

maierbenedikt:
Hmm my personal opinion is that this would lead to a proliferation of runcmsgrids for Powheg that one would have to maintain. How was it done until one when a Powheg processes needed additional/other PDFs compared to the runcmsgrid? 

But I will delete the runcmsgrid.sh from the pull request now. Can you then merge it? I would need the specific link to the Powheg cards in Github so that I can put that one in the external_tarball.dat file for aMCatNLO/MadSpin.

Thanks.

covarell:
In general we try to keep all PDF sets consistent for all samples (at least those produced by the same generator). If an analyser wants to use PDF reweighting for several samples, and each of them has a different weight pattern, it becomes very confusing.

In a future reprocessing of GENSIM, we may want to include new PDFs like the PDF4LHC15 sets, but there are no concrete plans for now.

So, I would be in favor of one set of reweightings for 5fs and one for 4fs (just two runcmsgrids)

PR #739, opened 2015-12-07
Adding cards for TTTT_5f_LO

heilman:
Cards for a Leading Order production of four top quark production.  These cards should be similar to the TTTT_5f_NLO cards but LO.


covarell:
Why do we need this in addition to NLO?

heilman:
The TOP PAG conveners have requested that we do a cross check with a LO sample as the behavior of the NLO samples with negative weights is still under question.  Also, since we are having difficulty properly understanding the negative weights of the NLO sample in BDTs, we wish to have a small LO sample to serve for comparison.

covarell:
In that case:
- model in proc_card must be sm_no_b_mass, not loop-sm_no_b_mass
- pdflabel must be set to "lhapdf" (as explained in the side comment) and lhaid should be 263000 (NNPDF30_lo_as_0130) as for all other LO samples, systematics at the bottom of the page should be changed accordingly

covarell:
Also in the run_card all maxjetflavor and similar occurrences must be changed to 5 (you are using 5-flavor scheme here)

heilman:
Relevant changes should be made.  Please let me know if I missed something.

PR #742, opened 2015-12-08
bbHBB hadronization fragment

OlivierBondu:
to go along with the [bbH pLHE production](https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/higgs/bbH_4FS_M125)


OlivierBondu:
up: is there anything wrong with the fragment or can we go ahead ?

bendavid:
JetMatching:nQmatch = 4',  should be removed, since this sample is not using jet matching.

covarell:
Olivier, I see setting which are typical of jet matching:

'TimeShower:nPartonsInBorn = 2',
 'JetMatching:nQmatch = 4',

but I see matching is not activated (as expected because there is one jet multiplicity)
Why these settings are needed?

bendavid:
TimeShower:nPartonsInBorn actually IS needed in this case (and 2 is the correct setting) since this also affects the parton shower FSR.

OlivierBondu:
I updated the previous fragment, and copied it to have also H > gamma gamma decay

PR #744, opened 2015-12-09
Fragments for WToLNu and WprimeToLNu for EXO-Summer15GS

davidsheffield:
From Begona De la Cruz


covarell:
I see several problems here:

The lines:
   'Main:timesAllowErrors = 10000',
+            'Tune:ee 3',
+            'Tune:pp 5',

must be removed everywhere.

Also, I do not see why you use pythia8 for Wlnu high mass. This will be inconsistent with all other CMS analyses which use MadGraph LO or NLO.

davidsheffield:
@bdelacruz would be able to respond to why they want to use pythia8. But that's what was produced in RunIIWinter15GS. It is line 408 on the rationalize spreadsheet from earlier this year https://docs.google.com/spreadsheets/d/16Z3rgM7XZzUMBxZr3DUXMldvGipeHwjDolUW5GSKn_s

bdelacruz:
We want to use Pythia8 as the signal (Wprime) is also generated with it.
We then use K-factors, both for signal and background, to consider NNLO 
cross sections.

It was used like this in the CMSSW_7_4_X campaign, I gues the question 
now comes back again for the CMSSW_7_6_X campaign?

Thank you,
Begoña

covarell:
"We want to use Pythia8 as the signal (Wprime) is also generated with it" is not a motivation. Most signals are generated using Pythia8, this does not mean we are using Pythia8 for all backgrounds. For SM, our suggestion is to use the best possible description (which would be NLO).

Second I do not understand why you need new cards if these samples were already produced for 7.4.X.

Physics is correct, so I will approve the request, but please consider suggestion above (for improving your analysis, not for making me happy)

bdelacruz:
Hi Roberto,

Concerning the answer, maybe it was not complete, you are right, being 
done in the rush for the Jamboree.

We know that NLO generators are a better description of the SM W 
production process than Pythia, and indeed for the "bulk" of the W 
production we use aMC@NLO W+jets samples.
We use Pythia for the high mass tail (M_W > 200 GeV and higher 
thresholds) , which is coveniently reweighted by K-factors which 
consider NNLO QCD cross sections from FEWZ and NLO EWK corrections from 
MCSANC. We believe this is enough to study the high transverse mass region.

Given your suggestion, and if possible, we would like to have the 
requests from Pythia (as sent to you) and one other sample, generated 
with aMC@NLO  (or Powheg) of W+jets, with M_W>1000 GeV, in order to 
compare the higher tails. It may well be that similar studies are 
already done, I will look for such comparisons.

Relative to our intention, sorry to deceive you, our answer was neither 
to make you happy, nor to upset you.

Best regards,
Begoña

PR #745, opened 2015-12-09
Fragments for gluino GMSB for EXO-RunIISummer15GS

davidsheffield:
From Begona De la Cruz


covarell:
Look ok to me, but is there someone more expert with SLHA tables?
@govoni @bendavid @vciulli @qili

davidsheffield:
Do the SLHA tables look good?

covarell:
In absence of any comment, I will just approve

PR #746, opened 2015-12-09
Pythia8 configs

mharrend:
* ttWJets
* ttZJets
* TTTo2L2NU Scale variation in PS
* TTToSemiLeptonic Scale variation in PS


covarell:
Is qcut = 10 a reasonable choice for ttV? It seems much lower than the ttbar one. 
Did you test DJR plots?

mharrend:
Hello Roberto,

the same config was used already in the Run I.
Furthermore, old studies done by one of my bachelor students showed that
a qcut matching scale of 10 to 60 GeV is a reasonable choice in LO.
Setting a higher value of the qcut is motivated if you have higher
analysis pT cuts anyway, while a smaller qcut value allows you in
principle to lower your pT cuts. I expect that the analysts who
requested the ttV samples are interested in rather low cuts since they
wanted also a m(l+l-) cut of 10 GeV.

Best regards,

Marco


-- 
Marco A. Harrendorf <marco.harrendorf@cern.ch>
KIT - IEKP 


Am Mittwoch, den 09.12.2015, 06:06 -0800 schrieb Roberto Covarelli:

bendavid:
Sorry, what is the xqcut of the madgraph configuration?  There needs to be sufficient headroom between xqcut and qcut, otherwise there will be missing phase space.

bendavid:
(ie if xqcut=10 then qcut should probably be at least 20)

mharrend:
Sorry to Roberto, you were of course right. I mixed it up and were refering to the xqcut value of 10 GeV but not thinking of the qcut value.
So I set now the qcut value to 20 GeV.

covarell:
I like 20 GeV more, bu if you had some 10k ttV GEN events at hand it would be worth checking:
https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideSubgroupMC#Measure_proper_value_of_qCut

For LO (as opposed to NLO) a wrong choice of this parameter can disrupt the cross-section significantly

mharrend:
@covarell I will check this. I still have to determine the matching efficiency anyway.
However, I will need to finish the GenSim configuration before I can start with measuring the matching efficiency and also checking the qcut value.
Since I would like to add the GenSim config by using the fragment tag and name of fragment field, can you please merge this configs for now.
I promise to change the config if necessary before asking to get the requests submitted. ;-)

mharrend:
Hello Roberto,

attached I am sending the DJR plots for a qcut of 20 GeV and 35 GeV for
the ttZ+0/1jet sample. In both cases I still observe a kink.

What qcut should I choose? Should I try out an even higher qcut value?

Best regards,

Marco
-- 
Marco A. Harrendorf <marco.harrendorf@cern.ch>
KIT - IEKP 


Am Mittwoch, den 09.12.2015, 06:06 -0800 schrieb Roberto Covarelli:

covarell:
Hi Marco,

Apart from kinks, do you have a clue of why the 1->2 DJR always stops abruptly at the qCut value? This seems unphysical to me. 

Of course if you try with a higher qCut value this problem will be "masked" because its value will fall outside the DJR spectrum... But in this case adding the 1-jet multiplicity at ME level becomes useless.

I wonder if Pythia is mystified by the fact that you mix up processes with different final states? 
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/TTZJets/TTZJets_5f_LO_MLM/ttZ01j_5f_proc_card.dat

Would you be able to run with 0+1 jet with only one final state?

@bendavid

mharrend:
Dear Roberto,

I am including Rebeca and Evan (our new second ttH MC contact) in this
discussion, since Rebeca maybe knows better why they used ttZ+0/1/2
jets in the run1 and since I will be on holidays from tomorrow on.

Honestly, I really wonder why this was not an issue with the old run 1
samples.

Best regards,

Marco

Am Freitag, den 11.12.2015, 00:47 -0800 schrieb Roberto Covarelli:

bendavid:
Optimal Qcut values may be quite different between pythia 6 and 8, and also change with cm energy.  For ttbar lo the 13tev sample was using qcut = 70 if I remember correctly.

PR #750, opened 2015-12-14
Xhh m750

OlivierBondu:
This PR is adding the mX = 750 GeV for (4 final states) * (radion, RS Graviton, Bulk Graviton) * 3 cards = 36 files. These are a simple mass change from the existing cards in [`bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson`](https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson)


covarell:
This is Jamboree-gossip driven!

PR #751, opened 2015-12-15
this sets the helicity flag for Pythia

mstoye:


covarell:
Sorry, are we using this same setting for 13 TeV? if not, why?

bendavid:
Yes the same setting is already in the 13TeV cards.

PR #760, opened 2015-12-31
For wgammajj pr

AndrewLevin:
this is a wgammaj ewk+qcd sample with vbf cuts since it is a background for the same sign wwjj analysis


govoni:
Hello Andrew,

are you ok with photons reaching only 2.5 in eta? CMS ECAL extends up to 3.0.
Also, are you ok with the minimal dr between jets an photons of 0.4 units?

pietro

AndrewLevin:
Hi Pietro,

We are only using electrons with |eta| < 2.5, but anyway I increased the etaa cut to 3.0, this increases the cross section by like 15%.

Indeed, I found that dra needs to be lowered since chanding it to 0.05 increases the cross section by like 30%.

I also changed dral to be 0.05.

0.05 is what is used in the sample

/WGToLNuG_TuneCUETP8M1_13TeV-madgraphMLM-pythia8/RunIISpring15MiniAODv2-74X_mcRun2_asymptotic_v2-v1/MINIAODSIM

so I assume we don't need to go even lower.

Andrew

PR #762, opened 2016-01-05
Remove filter from GMSB gen fragments

davidsheffield:


covarell:
What is the purpose of removing the filter?

mhwalker:
This was included by mistake. We don't need this filter.

PR #766, opened 2016-01-08
tt, WWbb, WWbb_notop for Les Houches

lorusso7:


lorusso7:
Config files for Sherpa. Sherpack  added in /afs/cern.ch/cms/generators/www/slc6_amd64_gcc481/13TeV/sherpa/2.1.1/

PR #768, opened 2016-01-10
Yuta for genproductions

gitytakahas:
we would like to produce dedicated DY->tautau samples with mass cut, m(tautau) > 150GeV.
You can find the reasoning here,

https://indico.cern.ch/event/472977/session/12/contribution/58/attachments/1208114/1761229/svFitMassVsGenMass_2016Jan04.pdf

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/DYJets_HT_LO_MLM/DYJets_HT_mll150/DYJets_HT-incl_mll150/DYJets_HT-incl_mll150_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/DYJets_HT_LO_MLM/DYJets_HT_mll150/DYJets_HT-incl_mll150/DYJets_HT-incl_mll150_proc_card.dat
@@ -30,8 +30,8 @@ define vl = ve vm vt
 define vl~ = ve~ vm~ vt~
 # Define multiparticle labels
 define p u c s d b u~ c~ s~ d~ b~ g
-define l+ = e+ mu+ ta+
-define l- = e- mu- ta-
+define l+ = ta+
+define l- = ta-
 define j  = p
 # Specify process(es) to run
 generate p p > l+ l- / h @0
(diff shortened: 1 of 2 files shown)


govoni:
why do you put the maximum rapidity of the photons at 2.5, not at three?

gitytakahas:
I'm using the same setting to that of the DY (LO) cards.
If I better change it to 3, I will do that. what do you think ?

Thanks,
Yuta

2016-01-10 9:44 GMT+01:00 Pietro Govoni <notifications@github.com>:

> why do you put the maximum rapidity of the photons at 2.5, not at three?
>
> —
> <https://github.com/cms-sw/genproductions/pull/768#issuecomment-170326766>
> .
>



-- 
Yuta Takahashi
LHC - CMS experiment,
CERN (European Organization for Nuclear Research)

vciulli:
I also find a bit strange the cut on the photon eta and I would change it to a large value, but it should have no effects since there isn't any photon in the final state 
Please note that using 
 15  = bwcutoff      ! (M+/-bwcutoff*Gamma)
the tau+tau- lepton pairs will not appear to come from a Z boson because m_ll > M_Z+15*Gamma_Z 
Apart from the history record, there is no difference in the results: just be aware not to search for a Z in the event history later on.

govoni:
I would put it at 3, which is the ECAL endcap maximal extension.

gitytakahas:
Dear Pietro, Vitaliano,

Thanks a lot for the prompt reply.
If there is no difference, I prefer to stay at 2.5, as these samples will
be used in combination with existing DY sample, and I don't want to change
settings, no matter how small it is ...
Do you still want to change it to 3.0 ?

Best,
Yuta


2016-01-10 10:12 GMT+01:00 Pietro Govoni <notifications@github.com>:

> I would put it at 3, which is the ECAL endcap maximal extension.
>
> —
> <https://github.com/cms-sw/genproductions/pull/768#issuecomment-170328877>
> .
>



-- 
Yuta Takahashi
LHC - CMS experiment,
CERN (European Organization for Nuclear Research)

govoni:
ok, it makes sense. Consistency is more important than truth :)

gitytakahas:
Thanks !

Yuta

2016-01-10 10:28 GMT+01:00 Pietro Govoni <notifications@github.com>:

> Merged #768 <https://github.com/cms-sw/genproductions/pull/768>.
>
> —
> <https://github.com/cms-sw/genproductions/pull/768#event-510413034>.
>



-- 
Yuta Takahashi
LHC - CMS experiment,
CERN (European Organization for Nuclear Research)

PR #770, opened 2016-01-13
Fixed nFinal to two

dnash86:
nFinal=1 caused a crash at the GS step, not the right number of final state particles.
Tests with nFinal=2 run successfully.


covarell:
In this case you should NOT overwrite a cff containing the string "1p" in the title, but rather create a new one with "2p" in the title

PR #776, opened 2016-01-15
VH with V to hadronic

benjaminkreis:
Needed by VBF Hbb analyses


perrozzi:
hi ben, upon pietro's request I looked at the cards and they look ok to me.
the only thing I don't know is if ggZH supports this decay, however it can be mostly neglected wrt VH.
Please once the gridpacks are prepared run O(1K) events and ensure that the V decay is handled properly (for both ZH and WH). 
thanks
luca

PR #777, opened 2016-01-18
VBF HWW 120 GeV powheg and JHU cards modified

lviliani:
The VBF HWW 120 GeV gridpack was faulty and I have redone it from scratch.
These are the new Powheg and JHUGen input cards.


govoni:
Do you need any actions to be taken on the existing samples, like removing them?

lviliani:
This sample has never been produced. There are only the wmLHE and GS requests.
As far as I understand the issue has been spotted during the validation of the wmLHE+GS chain.
I am going to reuse the same request but replacing the gridpack with the new one.

PR #778, opened 2016-01-20
wgstar by flavor and with pT cut

rafaellopesdesa:
More on the Wγ* issue...

Summary of what was done in the past and what is in this PR:
   - We had a madgraph (LO) Wγ* → ℓν ℓℓ with up to 2 parton at ME. This sample was failing validation. Upon further inspection, the failure was due to parts of the phase space not being sampled. In local test runs with 1000 events, only ~ 10% would be returned.
  - We tried to simplify the process by removing the τ from the γ* decays, by forcing the w as an intermediate state (madgaph decay notation), and by having only up to 1 parton at ME. Indeed, with this simpler phase space, a test run with 1000 events ran ok. However, upon even further checks, slight longer runs (~5000) would still fail to return all the requested events, indicating that some part of the phase-space was still not being sampled.
   - We noticed that in these tests all the γ* decays where to ee, due to divergent behavior of the amplitude at low mass. We now separate the gridpack in 3 (Wγ* → ℓν ee, Wγ* → ℓν μμ, Wγ* → ℓν ττ) and require the pT of the first and second leptons to be above 5 GeV, and low mll cuts of 0.001, 0.210 and 3.400 (the upper mll cut is always 4 GeV). I think this is a safe pT cut. I understand that this makes the sample hard to normalize. But...
   - In the analyses I imagine this sample being used we can select a 3 lepton control region to correct the normalization. Moreover, with this simplified final state phase-space it is conceivable that the aMC@NLO (NLO) gridpack would be generated without much problem (no, I haven't tested it yet, but I may). I would vote for generating this sample at LO at this moment, since we really need this sample. We are desperately late with it. The NLO would give a better normalization, I expect.
   - These 3 gridpacks seem to pass all local checks that I can imagine and have reasonable cross section (checked before matching). So I am asking for a PR for their cards. Luca Perrozzi is currently performing the validation of the EE sample in the production system. I can only hope we have no more problems...

Let me know what you think of this proposal and if further checks/modifications are necessary.

Thanks,
-- Rafael.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/WGstarToLNuEE012j_5f_LO_MLM/WGstarToLNuEE012j_5f_LO_MLM_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/WGstarToLNuEE012j_5f_LO_MLM/WGstarToLNuEE012j_5f_LO_MLM_proc_card.dat
@@ -1,0 +1,19 @@
+#switch to diagonal ckm matrix if needed for speed
+import model sm-ckm_lepton_masses_no_b_mass
+
+define p = p b b~
+define j = j b b~
+
+define ell+ = e+ mu+ ta+
+define ell- = e- mu- ta-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+
+generate p p > ell- vl~ e+ e- $$ t t~ h @0
+add process p p > ell+ vl e+ e- $$ t t~ h @1
+add process p p > ell- vl~ e+ e- j $$ t t~ h @2
+add process p p > ell+ vl e+ e- j $$ t t~ h @3
+add process p p > ell- vl~ e+ e- j j $$ t t~ h @4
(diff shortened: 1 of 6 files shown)


covarell:
@perrozzi , do you confirm a safe behaviour of these cards? If so, I would approve them straight away.
Probably the whole matter is worth discussing with MG authors, though...

rafaellopesdesa:
Hi @covarell ,

The MuMu and TauTau cards work fine. The EE has had more modifications (I should update the PR before merging). @perrozzi can confirm that the LHE step ran without problems.

The Madgraph authors have been contact, iirc, but I don't think that the generation of such low masses is really supported and it wasn't of much help.

perrozzi:
Ciao, I confirm that the gridpacks work properly, provided the modification for EE rafael will update.

rafaellopesdesa:
@covarell @perrozzi Ok, I updated the EE card to reflect exactly the version that went into production.

Best,
Rafael.

PR #785, opened 2016-01-28
add different Pythia tunes for NNPDF

gitytakahas:
Pythia fragment files for the ggH/VBF H->tautau with different Pythia tunes (needed for the systematic uncertainty studies)


covarell:
Ok to merge, but you will wait for a new 71X release before these can be used...

PR #787, opened 2016-01-28
bug fix 

perrozzi:
present while launching parallel grid production


covarell:
Luca and @yuanchao, does this revert to the original problem where the rest of the output were placed in a subdirectory and so runcmsgrid.sh did not run?

If not, I would approve the PR.

perrozzi:
I didn't check that yet, I will continue to the end and try to consistenly patch the script, if needed

perrozzi:
(thinking a bit more about it, my change is only acting on the bash script used to submit the job to create the grid, so shouldn't have any impact on the gridpack itself. I will anyway continue the debugging)

yuanchao:
This is a regression bug for grid production. I just tested the PR and it's fine. Please approve the PR. For runcmsgrid.sh, it is for event production and has been fixed previously.

PR #788, opened 2016-01-29
Powheg samples for the diphotons high-mass analysis

ferriff:
Few mass points with narrow (1.4e-4) and wide (5.6e-2) width approximation.


covarell:
Sorry, is this meant to be a fixed width or running width generation?
If the former is true (as it seems from the title commit) don't you need to select also "bwshape 2" according to POWHEG ggH_quark_mass manual?

covarell:
ping

ferriff:
Hi, thanks for checking this.
After investigation with experts (which caused the delayed answer) for the high mass Higgs the option 3 is used (Passarino’s complex pole scheme). So that should also be the most appropriate option in this case as well, do you agree?

covarell:
Complex pole scheme is indeed an "improved" running width generation but relies on the assumption of a scalar propagator. So, again, it depends which model you are targeting

covarell:
ping

ferriff:
@covarell: PR updated to `bwshape 3`, thank you.

ferriff:
ping

PR #789, opened 2016-01-29
Added support for condor cluster

kdlong:
Essentially this is just a two line change, but a bug in MadGraph
prevented cluster_queue from being set properly
(https://bugs.launchpad.net/mg5amcnlo/+bug/1517041). Also modified
cluster.py to print the error message when jobs are held. Have tested
with a few LO and NLO jobs on the Wisconsin condor cluster.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/patches/mgfixes.patch
+++ b/bin/MadGraph5_aMCatNLO/patches/mgfixes.patch
@@ -47,6 +47,54 @@ diff -ur -x '.*/models/.*' ../old/MG5_aMC_v2_3_3/madgraph/various/cluster.py MG5
  class Cluster(object):
      """Basic Class for all cluster type submission"""
      name = 'mother class'
+@@ -97,6 +106,7 @@
+         self.submitted_dirs = [] #HTCaaS
+         self.submitted_exes = [] #HTCaaS
+         self.submitted_args = [] #HTCaaS
++        self.hold_msg = ""
+ 
+         if 'cluster_queue' in opts:
+             self.cluster_queue = opts['cluster_queue']
+@@ -307,7 +317,8 @@
+             else:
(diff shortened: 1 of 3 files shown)


bin/MadGraph5_aMCatNLO/patches/mgfixes.patch
```diff
@@ -508,3 +556,16 @@ diff -ur -x '.*/models/.*' ../old/MG5_aMC_v2_3_3/Template/NLO/SubProcesses/MCmas
...
 +      mcmass(16)=0.0d0
        mcmass(21)=0.0d0
+diff -ur ../old/MG5_aMC_v2_3_3/madgraph/interface/madgraph_interface.py MG5_aMC_v2_3_2_2/madgraph/interface/madgraph_interface.py
+--- ../old/MG5_aMC_v2_3_3/madgraph/interface/madgraph_interface.py       2015-09-07 20:07:39.000000000 +0200
++++ MG5_aMC_v2_3_3/madgraph/interface/madgraph_interface.py    2015-11-17 10:50:56.985751642 +0100
+@@ -5851,6 +5851,9 @@
+         elif args[0] in ['cluster_local_path']:
+             self.options[args[0]] = args[1].strip()
+ 
++        elif args[0] in ['cluster_queue']:
```
bendavid: Can you remind what these two lines do?

kdlong: This was the fix for the bug I reported here: https://bugs.launchpad.net/mg5amcnlo/+bug/1517041

The cluster_queue needs to be set to the string "None," but elsewhere the code converts the string to the python None object, so I just added this to keep it as a string.

bendavid:
Thanks, this looks good to me.

bendavid:
@govoni @covarell I think this is ready to be merged.

PR #794, opened 2016-02-04
5TeV SMP

amarini:
5TeV ( sqrt s = 5020 )  cards production for SMP analysis:
* DY amcatnlo
* WJets amcatnlo 
* WW/WZ/TT powheg 
* ZZ pythia8


covarell:
I see several issues with these cards:
1) Why WW and WZ are done with POWHEG and ZZ (which has an identical setup in POWHEG) uses Pythia8 only? This is inconsistent.
2) All amcatnlo are named FXFX (and FXFX merging is activated in the run card) but the process card contains only inclusive processes. In this case ickkw = 0 should be used everywhere (and possibly there are other modifications, please check a sample run card with these settings)

amarini:
Should these comments be valid also for 13TeV ?  I copied the card from there.
For the ZZ we can produce the powheg too, but for our purposes its not necessary.

covarell:
Ah sorry went too fast.
Please produce ZZ with POWHEG indeed but amcatnlo cards are OK. I did not see you had separate cards for each jet bin, in this case it is correct.

amarini:
I added ZZ -> llnn and ZZ->llll

PR #796, opened 2016-02-06
ttbar + gamma both at parton and radiation level

jfernan2:


covarell:
I am not sure I understand these cards... top decay seems present both in the main and madspin cards. Which of the two programs should decay tops?

jfernan2:
Uhmm, the settings try to reproduce the 8TeV sample:

https://cmsweb.cern.ch/das/request?input=dataset%3D%2FTTGamma_TuneZ2star_8TeV-madgraph-tauola%2FSummer12_DR53X-PU_RD1_START53_V7N-v1%2FAODSIM&instance=prod%2Fglobal

but I guess you are right, madspin datacard is redundant since decay is requested at MadGraph level. In any case, I believe the madspin datacard is harmless

covarell:
Yes, can you please remove it then?

PR #797, opened 2016-02-11
Config fixed after gridpack generation

jfernan2:


covarell:
Do the cuts added imply something for the acceptance of this process in the detector, or is generation completely bias-safe?

jfernan2:
The cuts are in gammas radiated after top decays, those cuts were applied at the 8TeV sample. I had removed them at the beginning but then I figured out they were needed because I was getting significantly fewer events than asked for with the gridpack. I.e. without them it tryies to integrate over a divergence. Cuts consist mainly of a minimum pT>13GeV for the radiated photon.

https://answers.launchpad.net/mg5amcnlo/+faq/1987

PR #802, opened 2016-02-24
NLO samples with Vpt binning

idebruyn:
Z(vv)+jets
Z(ll)+jets
W(lv)+jets
photon+1jet


syuvivida:
@idebruyn 
 I would like to know more about these MC samples, for example, the number of events requested, etc.
The B2G diboson group is very interested in these samples. Could you please let me know how to contact you via CERN email address?
Thanks!!

I am Shin-Shan Eiko Yu (syu@cern.ch)

PR #803, opened 2016-02-25
Fragments for EXO PhotonIndBkgGGToLL for RunIISummer15GS

davidsheffield:
from Dimitri Bourilkov

### DIFF
--- a/python/ThirteenTeV/PhotonIndBkgGmGmToEE_M-015o060_TuneCUETP8M1_13TeV_pythia8_cff.py
+++ b/python/ThirteenTeV/PhotonIndBkgGmGmToEE_M-015o060_TuneCUETP8M1_13TeV_pythia8_cff.py
@@ -18,7 +18,7 @@ generator = cms.EDFilter("Pythia8GeneratorFilter",
          'PhotonCollision:gmgm2ee = on',
          'PhaseSpace:mHatMin = 15.',
          'PhaseSpace:mHatMax = 60.',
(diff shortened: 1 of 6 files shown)


covarell:
PDF setting should be removed from all cards

covarell:
unless there is a specific need for this (very old) PDF set. In any case I believe this is overwritten in the Pythia tune.

bouril:
This PDF is needed because only QED PDFs provide photons needed to calculate the photon induced background.

To avoid overwriting with the Pythia tune, I think the fragments should be modified:

PDF:useHard = LHAPDF6:MRST2004qed.LHgrid

This way Pythia will use a specific PDF for the hard process (the one we need here), and the standard PDF for the tune.

bouril:
Correction: after more testing, this is the setup to use a specific PDF for the hard process (the one we need here), and the correct (different) PDF for the tune.

PDF:useHard on
PDF:pHardSet  = LHAPDF6:MRST2004qed.LHgrid
PDF:pSet = <tune_PDF>

bendavid:
I believe that the default nnpdf23LO pdfset in pythia8 DOES have a photon pdf included.  Does it not work?

bouril:
Correct, it has QED included and works. The problem is that it gives suspiciously high cross sections at high mass 2-3 TeV, the MRST2004 PDF behaves more reasonably.

davidsheffield:
Can you approve these fragments?

PR #808, opened 2016-03-02
for high mass higgs to invisible

benjaminkreis:


covarell:
Invisible decay is not specified in POWHEG.
Are those the first cards ever with VH and high-mass H?

benjaminkreis:
Hi @covarell,

Similar to these existing samples, I am keeping the Higgs undecayed at the LHE step:
ggZH_HToInv_ZToLL_M125_13TeV_powheg_pythia8 (https://github.com/cms-sw/genproductions/blob/51d57507cb4efb9b4554beb636d97219f63da6b8/bin/Powheg/production/VH_from_Hbb/ggHZ_HanythingJ_NNPDF30_13TeV_M125_Vleptonic.input)
GluGlu_HToInvisible_M600_13TeV_powheg_pythia8 (https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/production/V2/13TeV/Higgs/gg_H_quark-mass-effects_NNPDF30_13TeV/gg_H_quark-mass-effects_NNPDF30_13TeV_M600.input)
VBF_HToInvisible_M600_13TeV_powheg_pythia8 (https://github.com/cms-sw/genproductions/tree/master/bin/Powheg/production/V2/13TeV/Higgs/VBF_H_NNPDF30_13TeV/VBF_H_NNPDF30_13TeV_M-600.input)
ZH_ZToEE_HToInvisible_M600_13TeV_powheg_pythia8 (https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/production/V2/13TeV/Higgs/HZJ_ZToEE_NNPDF30_13TeV/HZJ_ZToEE_HToInv_mH600_NNPDF30_13TeV.input)
ZH_ZToMM_HToInvisible_M600_13TeV_powheg_pythia8 (https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/production/V2/13TeV/Higgs/HZJ_ZToMM_NNPDF30_13TeV/HZJ_ZToMM_HToInv_mH600_NNPDF30_13TeV.input)

Then we would decay them to invisible at the GEN-SIM step.

As you can see above, these samples exist up to 600 GeV.  The Higgs conveners just asked for two higher masses -- 800 and 1000 GeV.

Ben

covarell:
OK. I wonder why these files do not use complex-pole-scheme (bwshape 3) and instead define a fixed width for high mass. This is not consistent with ZZ and WW high-mass samples, even if for an invisible decay it does not matter of course... My point is that new cards are not needed. Please check with @dsperka @lviliani

covarell:
Ok, I see now this is VH with high-mass Higgs (indeed, no one ever requested this, because the cross-section is too small), indeed they do not exist and do not need complex-pole scheme

PR #813, opened 2016-03-15
bug fixed.

yuchanggit:


covarell:
Hmmm... Is it sufficient to remove this line or should it be replaced with another "group combine"? Could you please check a MadGraph LHE file to find out?

yuchanggit:
updated.

PR #817, opened 2016-03-18
Update Epos LHC to 8TeV pPb

wouf:


wouf:
So strange path it has genproductions/python/HI/https/github.com/wouf/genproductions/blob/master/python/HI/ is it possible to move it to genproductions/python/HI/?

PR #818, opened 2016-03-20
setting the bwcutoff back to a reasonable value

AndrewLevin:
The madgraph authors have told us that using the decay chain syntax with such a large bwcutoff does not make sense. Also, we have checked that the results with bwcutoff=15 agree with what we get without decay chain syntax. More details can be found here:

https://indico.cern.ch/event/486671/session/0/contribution/3/attachments/1225694/1794216/VBS_GEN_9Feb16.pdf


bendavid:
Large bwcutoff with the decay chain syntax makes sense if and only if there are no other additional non-resonant diagrams.  (It makes sense for single W production for example)

AndrewLevin:
(In this case there are additional non-resonant diagrams. You can see them in the link.)

PR #822, opened 2016-03-29
Cards for T2bH

duanders:
MadGraph cards for T2bH simplified model.

(https://indico.cern.ch/event/504124/contribution/4/attachments/1235994/1814766/SUSY_Photons_HggRazor_MC_GEN_Feb29.pdf)


bendavid:
n.b. for GEN conveners, this is a special case which needs param cards, since the parameter changes used for SUSY signal production are not straightforward with the customize cards mechanism.

vciulli:
Do you mean that in case of a parameter scan you can not use the customize card mechanism or that the modification required in this case can not be made with the customize card?

bendavid:
No, this is not related to the parameter scans per se, just that the changes with respect to the default mssm parameter card are not straightforward to encode in a customize cards.
(Possibly some of these changes are ultimately not needed, but we're trying to stay coherent to some extent with the existing so-far plhe based susy signal production)

bendavid:
Anything holding this one up?

vciulli:
This is fine with me (just to make it clear, since I wrote a comment)

PR #823, opened 2016-03-29
genproductions for spin 0 to ttbar analysis

pequegnot:
This is the genproductions for the search for a heavy Higgs h,A->ttbar at 13 TeV.

The signal samples request has been discussed in the higgs-exo meeting: see [1].
A Madgraph5 model has been created for this particular signal generation: see details in [2].
As we have to tune the model "by hand", we are going to produce privately the LHE files. In this directory, you will find the param_card and run_card associated to this production. 

Before launching the LHE production, we wanted to be sure all the parameters setting in the different cards we are using are correct. Could you please check if everything is fine to you?

[1] [Slides from Higgs exo meeting](https://indico.cern.ch/event/489844/session/7/contribution/1/attachments/1246383/1835777/Higgs_exo_meeting_yazgan.pdf) 

[2] [Slides for details on Madgraph5 model](https://indico.cern.ch/event/484363/contribution/2/attachments/1215050/1774053/slides_HiggsMeeting_19Jan16_apequegnot.pdf)


bendavid:
Why do you need to produce the LHE's privately?

Even if you need to modify the model this can still be used to produce gridpacks.

pequegnot:
For technical reasons, we have to generate signal and interference parts separately. In Madgraph5, it is possible to keep the interference-only part in the generation using the squared order syntax. However, when we use this syntax, there is no support for decaying the particles in Madgraph neither in MadSpin as it has been reported on Madgraph [launchpad](https://bugs.launchpad.net/mg5amcnlo/+bug/1511378).  
The only way to get around this problem is to tune the model by hand (ie modify the matrix element by hand). This is the reason why it is easier to provide the LHE files.

covarell:
proc_card is missing...

PR #824, opened 2016-03-30
Pythia8 fragment for SUSYGluGluToHToAA_AToTauTau 8 mass points

chayanit:


govoni:
Hello @chayanit , I guess these setups are derived from already existing ones?

chayanit:
Hi @govoni, I don't think we have this setting before because it is the first request from gg->H->aa->4tau.

govoni:
@chayanit: the pythia settings are ok. 
May I ask why the particle 35 has the Higgs mass in your configuration, instead of the particle 25 as it usually is?

chayanit:
@govoni 
Yeah right. I was also checking this and found mass and ID of Higgs and A are switched. From my understanding, this shouldn't cause any problem as long as we set it clear that Higgs (id 35) decays to A (id 25) with correct masses implemented. What do you think? or does it matter in Pythia? Let me know if we need to fix this.

govoni:
@chayanit I don't know if it makes any difference, but I find that since the production did not start yet it might be worth setting the IDs in agreement with the usual notation.

chayanit:
@govoni ok. I agree. I will commit new changes.

covarell:
Sorry but this will work only in one way and not the other.
 '35:onIfMatch = 25 25',
will work if the decay 35 -> 25 25 exist in Pythia
 '25:onIfMatch = 35 35',
will work if the decay 25 -> 35 35 exist in Pythia
Of course the two cannot exist simultaneously in Pythia...

chayanit:
hi @covarell how can I check which one works? at least I tried producing events from 35 -> 25 25 as it shows on this fragment and it works.

covarell:
Then I would not modify

chayanit:
Ok thanks. So can you merge this PR? @covarell, @govoni

perrozzi:
@chayanit can you confirm that this PR can be merged?

chayanit:
@perrozzi yes, please.

PR #826, opened 2016-04-04
Add pair production mode for h++

dntaylor:
Add pair production mode for doubly charged Higgs production using Pythia8's LeftRightSymmetry process.


perrozzi:
what is the status of this PR?
At first glance the datacards look ok to me, could anybody please crosscheck?

PR #835, opened 2016-04-19
Added cards for WGGJets process at LO with MLM jet matching

mmachet:
For the run card I followed Josh instructions and I took the example found here:

https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/examples


bendavid:
For LO you probably shouldn't use madspin, but rather use the decay chain syntax

generate p p > w+ a a, w+ > l+ vl

(then you can also decide if you want the off-shell phase space, in which case you need to increase bwcutoff in the run cards)

bendavid:
Have a look at the LO WJets example cards here https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/examples/wellnu01234j_5f_LO_MLM

mmachet:
Hi Josh,

ok thanks. The  bwcutoff at 15000 is it ok?

2016-04-19 10:33 GMT+02:00 Josh Bendavid <notifications@github.com>:

bendavid:
So actually, stepping back for a minute, in this configuration (either madspin or decay chain) you will produce ONLY ISR photons.  If you want to also include fsr, then you need to instead do directly

generate p p > l+ vl a a
and similar.

mmachet:
Ok I changed the proc card in this way, and in the run card I put
15000 = bwcutoff

2016-04-19 10:46 GMT+02:00 Josh Bendavid <notifications@github.com>:

bendavid:
No if you use the p p > l+ vl syntax then bwcutoff can just go back to the default 15

mmachet:
done

2016-04-19 10:59 GMT+02:00 Josh Bendavid <notifications@github.com>:

bendavid:
The madspin card should be completely removed in this case.

mmachet:
Sorry forgot it. Done

2016-04-19 11:17 GMT+02:00 Josh Bendavid <notifications@github.com>:

mmachet:
Are there other things to change or the PR can be accepted?

2016-04-19 11:23 GMT+02:00 Martina Machet <martina.machet@cern.ch>:

PR #837, opened 2016-04-19
Revert "The LO is to compare with the BSM VBF HH samples"

govoni:
Reverts cms-sw/genproductions#836


perrozzi:
shall this be merged?

PR #848, opened 2016-05-02
Add a way to patch ME after generation

blinkseb:
If a file named `<name>_patch_me.sh` is found, it's executed right after the generation of the ME code.

This is useful if you need to patch the code generated by Madgraph before the gridpack generation starts. The script is provided with one argument, the absolute path of the current Madgraph working directory.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
+++ b/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
@@ -279,6 +279,11 @@ if [ ! -d ${AFS_GEN_FOLDER}/${name}_gridpack ]; then
   ########################
 
   ./$MGBASEDIRORIG/bin/mg5_aMC ${name}_proc_card.dat
+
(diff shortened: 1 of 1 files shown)


bendavid:
Can you point to an example patch_me.sh and a case where you would need to use this?

blinkseb:
I don't have an example script right now, but it will usually be something like

```sh
cd $1;
patch -p0 my.patch
cd -;
```

We need this mechanism to be able to generate only the interference part of our signal. Madgraph supports this out-of-the box using the `COUPLING^2` syntax, but the decay is not supported in this case. Hence we need to tweak the matrix element code to achieve the same thing.

With support directly inside `gridpack_generation.sh`, we could provide official gridpack, and we ensure reproducibility. IMO this is way better than providing directly LHE file (which is the only alternative).

bendavid:
Yes, agreed that this makes sense in that case.  @govoni @covarell I think this can be merged.

PR #852, opened 2016-05-04
Example card for gluon-fusion to narrow spin-0 to HH

OlivierBondu:
Title says it all.

the reason at why this PR is because we unfortunately just noticed that tau neutrinos are missing from the HH > bbllnunu datacards (which contain H > Z(ll)Z(nunu))... These cards are now only up to HH since the machinery to decay these in pythia now exist. Additionally, this also would mean easier setup for possible other newcomer HH final states.

Datacards for all mass points and deprecation of the `bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson/Spin-0/Radion_hh_hVVhbb_fullLep` folder would follow once this one is merged.


govoni:
@OlivierBondu how do you handle the HH decay if it's not done in mg5nlo ad it does not exist in pythia (what do you mean by that, by the way?)?

OlivierBondu:
@govoni : I am refering to https://github.com/cms-sw/cmssw/pull/12902

I meant this particular bit of machinery did not exist at the time when we produced the first HH samples. So now we should be able to have the Higgs decay to different final states at the pythia level, no need to redo a gridpack for each final state.

govoni:
@OlivierBondu tks

PR #856, opened 2016-05-06
add WZTo3lNu_mllmin01 fragment

lviliani:
Powheg card for WZTo3LNu.
This is the same card as:
https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/production/WZTo3lNu_NNPDF30_13TeV/WZ_lllnu_NNPDF30_13TeV.input
but we lowered the minimum value for mll (Z boson leptons) to 0.1 GeV (which is the default value).
In this way most of the events produced are actually Wgamma* events, and we can obtain a NLO Wgamma* sample.

### DIFF
--- a/bin/Powheg/production/WZTo3lNu_mllmin01_NNPDF30_13TeV/WZTo3lNu_mllmin01_NNPDF30_13TeV.input
+++ b/bin/Powheg/production/WZTo3lNu_mllmin01_NNPDF30_13TeV/WZTo3lNu_mllmin01_NNPDF30_13TeV.input
@@ -1,0 +1,71 @@
+! WZ boson production parameters
+mllmin 0.1d0         ! default 0.1 GeV this is minimum invar mass for Z leptons
+
+numevts NEVENTS   ! number of events to be generated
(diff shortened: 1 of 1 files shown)


bendavid:
Does POWHEG use massive leptons in the matrix element for this process?  Otherwise you would need to set the min mll to something larger than twice the muon mass to avoid unphysical contributions.

lviliani:
Looking at the LHE file I would say the mass of the leptons is taken into account. And I do not see events with mZ < 2*lepton mass. In any case the 0.1 GeV cut is the default value.
Also looking at the manual I do not see any issues.

bendavid:
Are the leptons themselves massive in the LHE file?  (I guess the question is whether this is included in the matrix element calculation, or inserted as a post-facto adjustment of the four vectors.  Maybe should just confirm with the authors that this is safe.)

lviliani:
Yes the leptons are massive in the LHE. I was indeed trying to understand if there is some kind of post-processing of the LHE file.
I will ask the confirmation to the authors.

PR #859, opened 2016-05-06
try to catch crashes in reweightings

perrozzi:
in principle, this change should be able to avoid cases like
https://hypernews.cern.ch/HyperNews/CMS/get/generators/2979.html
(although I couldn't reproduce the crash, so I am not 100% sure)

### DIFF
--- a/bin/Powheg/runcmsgrid_powheg.sh
+++ b/bin/Powheg/runcmsgrid_powheg.sh
@@ -335,6 +335,7 @@ then
     echo -e "\n finished computing weights ..\n" 
 fi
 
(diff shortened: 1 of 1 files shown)


perrozzi:
@covarell @govoni @bendavid what do you think?

bendavid:
This is good to have and should at least solve the problem for most cases.  On the other hand there are some (at least possible) border cases where the list of weights is truncated by the xml is technically valid maybe?

Does it work to also check the exit code of powheg after each step and immediately fail if one of the powheg steps has a non-zero exit code?  The xmllint could stay as well anyways.

perrozzi:
thanks, I implemented the suggestion

PR #860, opened 2016-05-06
Datacards for TpTpTotXtX M 550 - 1500 GeV

varuns23:
Datacards for TpTpTotXtX M 550 - 1500 GeV


perrozzi:
cards look ok to me, any comment from others?

PR #861, opened 2016-05-06
Tp lh

varuns23:
Datacards for tX_t_LH M 550 - 1500 GeV


perrozzi:
the datacards look ok

PR #865, opened 2016-05-10
More GF to spin-0 narrow to HH

OlivierBondu:
Follow-up on https://github.com/cms-sw/genproductions/pull/852, main changes:
- more mass points (all cards are the same as in the previous PR except for the name and mass parameter)
- deprecate `bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson/Spin-0/Radion_hh_hVVhbb_fullLep`


govoni:
@OlivierBondu may I ask why the jet pT cut is lowered to 2 GeV?

OlivierBondu:
mmm, I did not touch it wrt the previous card, must be a remnant

In any case there are only two H in the final state, no additional jet, as this is LO

OlivierBondu:
@govoni : so, do you want me to raise the cut ? These cards contain only H in the final state, the decay will be handled by pythia, so it should not matter what the jet pT cut is.

PR #866, opened 2016-05-11
Little bug corrected for Powheg gridpack submission in multiprocessor…

pgunnell:


perrozzi:
Ciao Paolo,
thanks a lot. I also had corrected already a similar bug here
https://github.com/pgunnell/genproductions/commit/2364c133065071341fef6d8a2c251c0f95a4d38d
and I now noticed that there might be also other places in the code where this change is needed, but should be checked to make sure...
In the meanwhile I merge this fix

PR #867, opened 2016-05-15
added aMC@NLO cards for ZllGGJets at LO

mmachet:
In the run card I put the cut mll = 50, as discussed in the past with @bendavid  and analysers.


perrozzi:
cards look ok

PR #873, opened 2016-05-18
Example hadronizer card for HH decay

OlivierBondu:
Based on https://github.com/cms-sw/cmssw/pull/12902

NB: need https://github.com/cms-sw/cmssw/pull/14555 to be integrated in CMSSW first before injection in MCM


bendavid:
https://github.com/cms-sw/cmssw/pull/12902 is already in CMSSW_7_1_22 in fact.

OlivierBondu:
that's why the nota bene was only on the second link :wink:

bendavid:
Yes, sorry, read too fast.

OlivierBondu:
(but since you're around, many thanks again for that feature, _very_ much appreciated)

python/ThirteenTeV/Higgs/HH/ResonanceDecayFilter_example_HHTo2B2L2Nu_madgraph_pythia8_cff.py
```diff
@@ -0,0 +1,53 @@
...
+            '25:onIfMatch = 24 -24',
+            '24:mMin = 0.05',
+            '23:mMin = 0.05',
+            'ResonanceDecayFilter:filter = on',
+            'ResonanceDecayFilter:exclusive = off', #off: require at least the specified number of daughters, on: require exactly the specified number of daughters
+            'ResonanceDecayFilter:eMuAsEquivalent = off', #on: treat electrons and muons as equivalent
+            'ResonanceDecayFilter:eMuTauAsEquivalent = on', #on: treat electrons, muons , and taus as equivalent
+            'ResonanceDecayFilter:allNuAsEquivalent = on', #on: treat all three neutrino flavours as equivalent
+            #'ResonanceDecayFilter:mothers =', #list of mothers not specified -> count all particles in hard process+resonance decays (better to avoid specifying mothers when including leptons from the lhe in counting, since intermediate resonances are not gauranteed to appear in general
+            'ResonanceDecayFilter:daughters = 5,5,11,11,12,12',
```
bendavid: This is not safe in case you decay HH events produced with jet matching and extra ME jets, since this could allow b-quarks from QCD rather than the higgs decay.  You should probably explicitly add
'ResonanceDecayFilter:mothers = 25' to protect against this (even if initially the fragment will be used with LO tree level HH events where this won't happen)

bendavid: Sorry, should actually be
'ResonanceDecayFilter:mothers = 25,23,24'

since the leptons and neutrinos will actually come from W's or Z's in the event history

OlivierBondu: mmm, but in your comment you mention not specifying W's and Z's in case of leptonic decays ?

bendavid: This is a special case.  Since the W"s and Z's would come from pythia decays, they ARE guaranteed to be included the event history.

My comment is a warning in case the hard process includes lepton production at the lhe level.

OlivierBondu: ok, I understand

actually I realized I should also have the `exclusive = on` if I don't want HH -> Z(bb)Z(ll)Z(ll)Z(nunu) decays possibly laying around...

bendavid: Yes that is also true.

bendavid: (but you can and should turn off the Z->qqbar decay modes for this config anyways)

OlivierBondu:
@bendavid: card updated

OlivierBondu:
@bendavid : card updated

OlivierBondu:
actually re-reading your comment I'm not sure if you were talking about `23:onIfMatch` or `ResonanceDecayFilter:daughters`, if it's the later then I'm not sure how to ensure 6 final objects for the hard process and having all lepton flavours...

bendavid:
I was indeed talking about 23:onIfMatch, so this looks ok now (for really maximum efficiency could do similar to switch on only the needed higgs and W decay modes, but ok probably this is pretty irrelevant, since it will only change the number of retries needed for the decays and should have a pretty small impact on cpu time)

OlivierBondu:
@bendavid  card updated:
- reshuffling the ordering of the forced decays for easier reading
- W now forced to decay leptonically
- open only leptonic tau decays (not quite sure of the interplay with the `ResonanceDecayFilter` here, feedback is welcome !)
- `onIfMatch` -> `onIfAny` for W, Z, tau

bendavid:
This looks ok to me.

Note that the ResonanceDecayFilter checks the decay state after the Higgs/W/Z/Top/etc decays but BEFORE the tau decays.  So there is no interplay here, and the taus are counted by the ResonanceDecayFilter if so configured, but the decay products of the tau never are.

OlivierBondu:
Thanks !
@covarell @perrozzi @govoni this is ready to be merged

PR #876, opened 2016-05-19
Fix LHAPDF iteration id; refine code style

yuanchao:
This fix the bug on CT14 PDF as discussed on HN.
https://hypernews.cern.ch/HyperNews/CMS/get/generators/3014.html

Also as suggested by Luca, the coding style on the while loops is modified and re-indented.


perrozzi:
there were some other off by one, fixed in https://github.com/cms-sw/genproductions/pull/877

PR #882, opened 2016-05-25
New mass points for HZZ_2L2Nu at 3000GeV both in ggH and VBF

amagitte:
Hi all,

I need urgently this request to be approved..
Sorry and thanks!

Alessio


govoni:
Is is expected that Higgs mass and width are the same?

amagitte:
Hi Pietro,

Yes, we have discussed it with the theorist (Maltoni). 
We are producing, starting from 1TeV up 3TeV, all samples with a width equal to the value of the mass in such a way to have the Phase-space completely covered by a flat shape.
This procedure allow us to study every model that predict a scalar boson only by reweighting to the model we want to test without producing centrally any BSM model.
Is it clear, more or less?

Alessio

govoni:
yes it's clear, thanks.
Are you ok with the complex-pole scheme in this case?

amagitte:
yes

PR #883, opened 2016-05-26
cards for new Z'->ttbar+1jet model

peiffer:
Notes: 

The process p p > v1 > t t~ has cross section 0. 

The LO diagrams of the process require already one top loop in the production, thus the [QCD] option has to be enabled but a LO PDF is used.

References for this model:
http://arxiv.org/pdf/1604.07421.pdf
http://arxiv.org/pdf/1410.6099.pdf

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/Top_Philic/TopPhilicZprimeToTTbar_M1000/TopPhilicZprimeToTTbar_M1000_customizecards.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/Top_Philic/TopPhilicZprimeToTTbar_M1000/TopPhilicZprimeToTTbar_M1000_customizecards.dat
@@ -1,0 +1,3 @@
+set param_card mass 6000055 1000
+set param_card mass 6 172.5
+set param_card yukawa 6 172.5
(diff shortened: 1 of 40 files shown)


perrozzi:
hi, just to understand: the process is effectively NLO but LO pdf are used?
If so, may I ask why?

peiffer:
This process does not contain any valid diagrams without loops because the v1 particle can only be produced via a top loop. Thus, it is a LO process and also a LO pdf should be used. The [QCD] option is only needed to enable the loop diagram, otherwise no diagrams would be found.

perrozzi:
Hi, thanks for the clarification. In the case of gg -> zh which is also a loop induced process, I used [*]
generate g g > h ell+ ell- [noborn=QCD] @0
I wonder if the same syntax should be used, I don't know what the difference would be. Do you know?

[*] https://github.com/cms-sw/genproductions/blob/b4be9d5f7ab6e4db859bebf02d14f89b5426ae42/bin/MadGraph5_aMCatNLO/cards/production/13TeV/higgs/ggzh01j_5f_LO_MLM_M125_ZToLL/ggzh01j_5f_LO_MLM_M125_ZToLL_proc_card.dat

peiffer:
I noticed that  p p > v1 > t t~ j does not work. I asked the author of the model and he said that aMCatNLO cannot handle the v1 decay correctly. I am going to configure the v1 decay in Pythia now.

I also switched to [noborn=QCD] instead of [QCD], but the only difference is that a warning is skipped claiming that no born diagrams can be found. The result is exactly the same.

perrozzi:
yes, as far as I know also madspin is not yet supported for loop induced processes, so you will need to use pythia (you will loose spin-correlations, but that's it). the rest looks ok, going to merge

bendavid:
[noborn=QCD] is preferred (and is actually required in case both loop-induced and non-loop induced processes exist for the given process).  In case only loop-induced diagrams exist then they should be equivalent.

PR #884, opened 2016-05-27
vbf and Zh productions for h->aa

nadjieh:
Samples for h->aa->mmbb


perrozzi:
cards look ok, merging

PR #890, opened 2016-06-02
DYJets cards binned in Lepton Transverse (LT) energy bins

gfasanel:


perrozzi:
@gfasanel just to understand: do you have such a long _cut.f  because you are interested in isolated photons and need to compute lots of global event stuff?

bendavid:
No the cuts.f is modified from the standard one, and includes lots of stuff by default.

Has to be like this unless madgraph authors would really split out only the user defined stuff into a separate file.

perrozzi:
ok thanks a lot. I had forgotten that, it is clear now

PR #891, opened 2016-06-06
Correct 5-flavor process definition for WAJJ

kdlong:
Remove tt~ propogators and correctly add b to proton definition. Also
switching to non-diagonal CKM and loosing lepton kinematic cuts


govoni:
@kdlong for my own education, is the minimal angular cut between jets and leptons removal safe in terms of divergences? I don't see any apparent cases where it should not, just checking whether you performed any tests.

kdlong:
Hi Pietro,

I wondered about this as well. This was the suggestion from Qiang. Qiang,
can you comment?

Thanks,

Kenneth

bendavid:
I don't think this cut is needed at all though, and in general would be better to avoid it as it makes the sample less inclusive.  What was the original motivation to have it?

kdlong:
I don't think there was a particular reason, I suspect it is a MadGraph default or was in whatever reference card this sample was based on. I agree that we want to make it more general, and I see that the Wgamma QCD sample also has it set to 0, so I think this is the best option.

qliphy:
Just saw this message.  Yes, we refer to Wgamma sample and would like it to be more inclusive.
And I think there is no singularity region between jet and lepton in this case.

PR #892, opened 2016-06-07
Settings for ttbar with mtop 8Xwidth central sample

jfernan2:


perrozzi:
general question, out of curiosity: I see that you have
withnegweights 0
and not phase space folding.
Which POWHEG code is this? Does it feature MiNLO?

jfernan2:
I cannot tell since I'm far from a POWHEG expert, but we are not using MiNLO at all. However the settings (and in particular "withnegweights 0") are standard to the TTbar POWHEG configurations used so far in CMS in POWHEG V2, 
E.g.
https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/production/TT_hdamp_NNPDF30_13TeV/TT_hdamp_NNPDF30_13TeV_powheg.input

PR #895, opened 2016-06-08
Cards for additional masses of EXO SeesawTypeIII

davidsheffield:
From Maxi Heindl.


govoni:
@davidsheffield I guess you're ok with 4 flavour scheme?

davidsheffield:
The analyzers used the same scheme for their previous samples at other mass points.

PR #901, opened 2016-06-09
High mass hbb

benjaminkreis:
New cards for narrow, high mass Higgs to BB


perrozzi:
cards look ok, merging

PR #902, opened 2016-06-10
HZJ, HWminusJ and HWplusJ cards

lviliani:
Added cards for HZJ, HWminusJ and HWplusJ processes for 120 and 130 GeV.
The cards are the same as the 125 GeV one, I have just changed the Higgs mass and width.


perrozzi:
talked with lorenzo, datacards already existing, closing the PR

PR #906, opened 2016-06-13
Mass cards for 500 GeV Higgs

pallabidas:
Adding the three input cards for 500 GeV for creating the corresponding gridpacks


perrozzi:
cards look ok to me, merging

PR #909, opened 2016-06-23
hadronizer for GJet NLO sample

idebruyn:


bendavid:
It's discouraged to have the full expanded configuration here.  You should just allow the common blocks to be imported without expanding their contents like with the other fragments.

idebruyn:
Like this?

bendavid:
Also the amcatnlosettings should not be expanded.

idebruyn:
Ok!

bendavid:
This looks fine now.

PR #911, opened 2016-06-27
Ttdilep

violatingcp:
Cards for use in ttbar dilepton dark matter high stats production


perrozzi:
things look ok, merging

PR #913, opened 2016-06-28
mass cards for Higgs Exo

pallabidas:


perrozzi:
just to check: are you using the SM widths?

pallabidas:
yes I am using SM widths according to the link you provided

pallabidas:
Thanks

PR #915, opened 2016-06-28
some additional fixes and example for multiple pdf weights

bendavid:
So for only wplustest_5f_NLO is updated for multiple pdf weights.

Central value is left as nnpdf for now.

Included sets are
292200	NNPDF30_nlo_nf_5_pdfas
90000	PDF4LHC15_nlo_mc_pdfas
90200	PDF4LHC15_nlo_100_pdfas
90400	PDF4LHC15_nlo_30_pdfas
13100	CT14nlo
13164	CT14nlo_as_0117
13166	CT14nlo_as_0119
25200	MMHT2014nlo68clas118
25260	MMHT2014nlo_asmzsmallrange
11000	CT10nlo
11067	CT10nlo_as_0117
11069	CT10nlo_as_0119

If the central value is to shift to pdf4lhc15 then we can just change the order.  (The sample is generated with the first in the list for the central value.)

Important point (up for discussion) is that in this list the version of MMHT included is the one with alpha_s = 0.118, consistent with the other sets.  This is the one which is used as input to pdf4lhc15.  There is also a different set (25100 MMHT2014nlo68cl) which uses alphas=0.120, which I believe is the value preferred by the MMHT fit and/or authors.  Worst case scenario would be to add this in addition.

Note also that the 4fs case will need a slightly different list, keeping in addition at least the PDF4LHC15 5fs sets to allow for evaluation of correlations.  Possibly also other 5fs sets, to be discussed.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/examples/wplustest_5f_NLO/wplustest_5f_NLO_run_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/examples/wplustest_5f_NLO/wplustest_5f_NLO_run_card.dat
@@ -10,6 +10,9 @@
 #   Lines starting with a hash (#) are info or comments                *
 #                                                                      *
 #   mind the format:   value    = variable     ! comment               *
+#                                                                      *
+#   Some of the values of variables can be list. These can either be   *
+#   comma or space separated.                                          *
 #***********************************************************************
 #
 #*******************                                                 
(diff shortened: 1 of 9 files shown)


perrozzi:
by chance, have you also already checked that it runs smoothly (and doesn't slow down dramatically) with ~700 weights?

bendavid:
Yes, the technical overhead is about 0.7s/event.  (But needs to be verified also for more complex processes and a larger number of events.)

There is also the longstanding issue, which is made worse now that all the pdf's get stored inside the gridpack tarball rather than read directly from CVMFS.  This seems not fatal (simple gridpacks increase in size from 50MB to 150MB), but I will still try to fix it.  Madgraph already has some switches to avoid this, but we had disabled them because they interfered with smooth lsf handling in the past.

perrozzi:
thanks for the tests.
@yuanchao could you please update the powheg scripts to comply with this list and make some tests with a couple of processes to ensure that it runs smoothly?

yuanchao:
Ok. Let me see how to handle an optional local LHAPDF copy.

bendavid:
Well I don't think a local pdf copy is needed or desired.  This is something we are trying to remove from madgraph in fact.

PR #932, opened 2016-07-27
Enable SVN info LHE hardcoded

yuanchao:
This patch is to turn on Powheg's built-in routine that writes the svn info into the output LHE files. Currently tested on DMGG, gg_H_quark_mass_effects, HJ, HZJ, WZ processes.


perrozzi:
I assume that this change is general, in the sense that is not needed to test every single process, right?

yuanchao:
I tried to make it general, but the actual patch is applied on the Makefile in the individual process directory.

perrozzi:
ok, I assume that the modification in the Makefile is always the same,
so in this sense is general, right?

yuanchao:
Makefiles for each processes are different, but the modification is always the same. I've tested more processes: dijet, trijet, dislepton-jet, DMV, DMS, ggHZ, HJ, hvq, HW, ttb, VBF_H, Wgamma, Wenu, Z2jet, Zj... So far no problem.

PR #933, opened 2016-07-27
d/l cteq6m for W_enu example card

yuanchao:
Auto d/l the additional file "cteq6m" for the example card file of W_enu.


perrozzi:
can you pleasout of curiosity, can you please specify why is needed?

yuanchao:
Originally, for processes need extra card files, run_pwg.py will d/l them automatically. But I just found that it was missed for the W_enu example card. So this is just to fix this missing part.

PR #951, opened 2016-09-15
New ttbar tunning CUETP8M2T4

epalencia:


perrozzi:
out of curiosity, is there any parameter of the matrix element specifically used for CUETP8M2T4 and different from previous datacards?

epalencia:
Yes, hdamp is the only one that has changed in the datacards: from m_top to 1.581*m_top.

PR #966, opened 2016-10-05
Nnlops hj

lenzip:
Fix for running JHUGen
- checks out the proper decay weights for WW and ZZ decays, so that they are available in the gridpack (for recent gridpacks users had to unpack, add the weights and repack)
- also fixes the way the JHU input file is handles. Previous version of the code was deleting JHUGen.input before doing anything with it.

Fixes for HJ + NNLOPS:
- fixes the input file name for nnlops reweighter 
- fixes a problem running HJ + CT14 (code was trying to access member 1 for alphas variation, when there is no member 1 for those variations, only member 0 in LHAPDF 6.1.6)

### DIFF
--- a/bin/Powheg/run_pwg.py
+++ b/bin/Powheg/run_pwg.py
@@ -985,7 +985,7 @@ if __name__ == "__main__":
         os.system('rm -rf JHUGen.input')
         inputJHUGen = '/'.join(powInputName.split('/')[0:-1])+'/JHUGen.input'
         if not os.path.exists(inputJHUGen) :
-            os.system('wget --no-check-certificate -N http://cms-project-generators.web.cern.ch/cms-project-generators/'+inputJHUGen)
+            os.system('wget --quiet --no-check-certificate -N http://cms-project-generators.web.cern.ch/cms-project-generators/'+inputJHUGen)
 
             if os.path.exists('JHUGen.input') :
                 os.system('cp -p JHUGen.input '+args.folderName+'/.')
--- a/bin/Powheg/run_pwg.py
+++ b/bin/Powheg/run_pwg.py
@@ -494,6 +494,12 @@ fi
 
 echo 'Compiling pwhg_main...'
 pwd
+if [ "$process" = "HJ" ]; then
+  echo "fixing q2min determination for HJ"
+  # avoid accessing member 1 for q2min determination. Use member 0, as member 1 may not be available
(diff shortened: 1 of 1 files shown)


bin/Powheg/run_pwg.py
```diff
@@ -329,6 +329,10 @@ def runGetSource(parstage, xgrid, folderName, powInputName, process, tagName) :
 if [[ -s ./JHUGen.input ]]; then
   jhugen=$(expr $jhugen + 1)
   echo "JHUGen activated!"
+  #for decay weights in H->WW and H->ZZ
```
perrozzi: naive question: so far we produced already HZZ and HWW gridpacks, why is this addition needed?

bin/Powheg/run_pwg.py
```diff
@@ -968,10 +982,13 @@ def runhnnlo(folderName, njobs, QUEUE):
         os.system('mkdir -p '+rootfolder+'/'+args.folderName)
         os.system('cp -p '+args.inputTemplate.split('/')[-1]+' '+args.folderName+'/powheg.input')
 
-        os.system('rm -rf JHUGen.input')
-        inputJHUGen = '/'.join(powInputName.split('/')[0:-1])+'/JHUGen.input'
+        #os.system('rm -rf JHUGen.input')
+        #inputJHUGen = '/'.join(powInputName.split('/')[0:-1])+'/JHUGen.input'
```
perrozzi: was this really an empty string before JHUGen.input ?!?

lenzip:
As far as I know people used to unpack the gridpack, add this stuff and 
repack it.
Now it should end up in the gridpack automatically.
Cheers
Giulio

lenzip:
I did not quite understand this.
first thing it did was to remove the user provided  JHUGen.input and 
then try to download a generic one.
If others understand how this was supposed to work I'm happy to change 
it as needed.
Cheers
Giulio

perrozzi:
@yuanchao  could you please check these modifications wrt yours?

bin/Powheg/run_pwg.py
perrozzi: can you please restore this, according to yuan chao's comment?

lenzip: It should be done now.
Let me know if you are able to merge now.
Thanks!
Giulio

PR #970, opened 2016-10-06
Fix input card path; update source to Sep 2016

yuanchao:
* Fixing the powheg input card file path issue
* Update the powheg source tarball to ver. Sep 2016 after tests done by Y.-H. Chang

### DIFF
--- a/bin/Powheg/run_pwg.py
+++ b/bin/Powheg/run_pwg.py
@@ -190,7 +190,7 @@ def runSingleXgrid(parstage, xgrid, folderName, nEvents, powInputName, seed, pro
 
 #    runCommand('mv -f powheg.input powheg.input.temp')
 #    sedcommand = 'sed "s/parallelstage.*/parallelstage ' + parstage + '/ ; s/xgriditeration.*/xgriditeration ' + xgrid + '/" ' + powInputName + ' > ' + folderName + '/powheg.input'
-    sedcommand = 'sed "s/NEVENTS/' + nEvents + '/ ; s/SEED/' + seed + '/" ' + powInputName.split('/')[-1] + ' > ' + folderName + '/powheg.input'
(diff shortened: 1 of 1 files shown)


perrozzi:
@lenzip could you please check these modifications wrt yours

lenzip:
I still don't see why the first thing it does with the JHUGen.input is 
to remove it.
https://github.com/cms-sw/genproductions/pull/970/commits/9f2ed923d7f2abb95db333f810e26f3c4294ff92#diff-ecd8ac3748a3baeb09846ae599287a66R974
This might work only if the supplied JHU input file (for which an 
additional input parameter has been added in this PR) is not named 
JHUGen.input.
Does this make sense?
Other changes seem orthogonal to my PR 966.
Cheers
Giulio

yuanchao:
There used to a problem that non-JHUGen jobs got JHUGen mode turned on due to the left over JHUGen.input in the working directory. The previous modification that with "args.inputTemplate.split('/')[-1]" breaks if you are referring the input card from other directories.

perrozzi:
ok, merging this first

lenzip:
can you please send me the command line that is supposed to work with 
this modification?
Imagine I have JHUGen.input in my current directory, and that I do want 
this to be a JHU job, I don't see how this can work.
Cheers
Giulio

yuanchao:
@lenzip : Sorry that I don't have good solution for your use case at the moment. I used to thought it's more convenient in that way, but I regret afterwards. For JHUGen.input kept in the same directory of powheg.input (other than the present directory), run_pwg.py will copy and use it automatically. Otherwise, one should use a different filename instead. This is to avoid non-JHUGen jobs accidentally pick up a left-over JHUGen.input in the present directory.

lenzip:
Ok, I understand now that the JHU input file is not supposed to be named 
JHUGen.input.
I'll modify my pull request to make it compatible with yours.
Cheers
Giulio

yuanchao:
@lenzip Yes. Thank you very much! Sorry for the inconvenience.

PR #978, opened 2016-10-13
Dalitz cards updates

andreypz:
In this update:
* Added mH=120 and 130 names in extramodels.dat. The updated parameter cards themselves were loaded to [cms-project-generators](https://cms-project-generators.web.cern.ch/cms-project-generators) web page.
* Commented out decays of the Z and W in VH production. This is because Madspin would not preserve the correct spin of those particles due to *spinmode=none* setting. The idea now is to decay W/Z in Pythia 8 wile doing the hadronization. **Please advise if that is possible?**
* Added readme file.


perrozzi:
it is always possible to decay the particles in pythia8, I don't see any problem with that

perrozzi:
thinking a bit about it, keep in mind that (as far as I know) also pythia will not preserve the correct spin correlations

PR #980, opened 2016-10-14
Settings for hdamp reweighting using TuneT4 as being used for Moriond…

jfernan2:


perrozzi:
last question: this will affect only top production or any other process is using this script for production?

jfernan2:
Sorry I should have clarified from the beginning: This script is used only if the person creating a gridpack explicity renames this script as runcmsgrid.sh  (default name for the gridpack running script) and packs it inside the gridpack. Hdamp reweighing is only being used by TOP, so we are the only consumers for this script so far. I just wanted to bookkeep it on git since this is what we are using in Moriond17.

PR #981, opened 2016-10-14
Adding Madgraph cards for Codex DarkMatter signals

abdollah110:
Please find the Madgraph cards for producing the signal samples for Dark Matter in Coannihilation Codex.
The gridpacks are located in:
/afs/cern.ch/user/a/abdollah/public/GridPack_CodexDarkMatter

And the model is already uploaded in : 
/afs/cern.ch/cms/generators/www/DarkMatter_Codex.zip


perrozzi:
has the gridpack been uploaded to cvmfs?

bin/MadGraph5_aMCatNLO/cards/production/13TeV/DarkMatter_Codex/Codex_LQ1000_DM_400_X_420_gen2/Codex_LQ1000_DM_400_X_420_gen2_run_card.dat
```diff
@@ -0,0 +1,283 @@
...
+#*********************************************************************
+# Beam polarization from -100 (left-handed) to 100 (right-handed)    *
+#*********************************************************************
+     0.0     = polbeam1 ! beam polarization for beam 1
+     0.0     = polbeam2 ! beam polarization for beam 2
+#*********************************************************************
+# PDF CHOICE: this automatically fixes also alpha_s and its evol.    *
+#*********************************************************************
+     nn23lo1    = pdlabel     ! PDF set                                     
+     230000    = lhaid     ! if pdlabel=lhapdf, this is the lhapdf number
```
perrozzi: I am not 100% sure which pdf is nn23lo1, if corresponds to 
230000    = lhaid
then according to https://lhapdf.hepforge.org/pdfsets.html
that is NNPDF23_nlo_as_0119
which is a NLO PDF with 5F
while below you specify 
4 = maxjetflavor
which means 4F, and also you import the model sm which is 4F.
Can you please clarify which PDF you want to use?

abdollah110:
Hi @perrozi
The gridpacks are in my public area and not uploaded to cvmfs.



On Sat, Oct 15, 2016 at 1:32 AM, perrozzi <notifications@github.com> wrote:

> *@perrozzi* commented on this pull request.
> ------------------------------
>
> In bin/MadGraph5_aMCatNLO/cards/production/13TeV/DarkMatter_
> Codex/Codex_LQ1000_DM_400_X_420_gen2/Codex_LQ1000_DM_400_
> X_420_gen2_run_card.dat
> <https://github.com/cms-sw/genproductions/pull/981#pullrequestreview-4367887>
> :
>
> > +#                                         3=photon from electron     *
> +#*********************************************************************
> +     1        = lpp1    ! beam 1 type
> +     1        = lpp2    ! beam 2 type
> +     6500.0     = ebeam1  ! beam 1 total energy in GeV
> +     6500.0     = ebeam2  ! beam 2 total energy in GeV
> +#*********************************************************************
> +# Beam polarization from -100 (left-handed) to 100 (right-handed)    *
> +#*********************************************************************
> +     0.0     = polbeam1 ! beam polarization for beam 1
> +     0.0     = polbeam2 ! beam polarization for beam 2
> +#*********************************************************************
> +# PDF CHOICE: this automatically fixes also alpha_s and its evol.    *
> +#*********************************************************************
> +     nn23lo1    = pdlabel     ! PDF set
> +     230000    = lhaid     ! if pdlabel=lhapdf, this is the lhapdf number
>
> I am not 100% sure which pdf is nn23lo1, if corresponds to
> 230000 = lhaid
> then according to https://lhapdf.hepforge.org/pdfsets.html
> that is NNPDF23_nlo_as_0119
> which is a NLO PDF with 5F
> while below you specify
> 4 = maxjetflavor
> which means 4F, and also you import the model sm which is 4F.
> Can you please clarify which PDF you want to use?
>
​I guess ​
 nn23lo1
​ is the default PDF of MG_amc@NLO. I have not change it when I get the
cards by running the MG_amc@NLO. I have google it and found that people are
using this PDF set already.
<https://www.google.com/search?q=nn23lo1&oq=nn23lo1&gs_l=serp.12...0.0.0.30792.0.0.0.0.0.0.0.0..0.0....0...1c..64.serp..0.0.0.qN1HRGLZRv8>

Cheers
Abdollah


​

abdollah110:
Any progress on this issue?

Best
abdollah

abdollah110:
Dear all

Would you please comment on this. We need to proceed as soon as possible as
its getting a bit late with these samples


Thanks
Abdollah

perrozzi:
@abdollah110 my only interest is that the gridpack has the correct settings.
@bendavid, @kdlong could you please crosscrosscheck?

kdlong:
Hi Abdollah, all,

I believe when you set "nn23lo1    = pdlabel" you use the NNPDF2.3 LO PDF set which is stored internally to MadGraph. With this setting the "lhaid" setting does nothing. Typically what is done for CMS productions is to use "lhapdf = pdlabel" which then allows you to use a very large number of PDFs through the LHAPDF libraries. For leading order samples, we typically use NNPDF30_lo_as_0130, which has LHAPDF ID 263000 as you can see here:

You can see an example configuration with these settings here: https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/examples/wellnu01234j_5f_LO_MLM/wellnu01234j_5f_LO_MLM_run_card.dat#L51-L52

This PDF set is for a 5 flavor scheme. If you want to use a 4F scheme you should use 263400. I don't see a strong argument in favor of using 4F in this case, and we have typically used 5F unless there is good reason to favor 4F. To move to 5F, you'll need to use the model sm-no_b_mass and include b-quarks in the proton definition, as done here:

https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/examples/wellnu01234j_5f_LO_MLM/wellnu01234j_5f_LO_MLM_proc_card.dat#L3-L6

It's up to the GEN conveners and experts to give the final word (@bendavid, @perrozzi, @govoni), but I think it would be best to regenerate the gridpacks with more standard settings, starting with a CMS example rather than the MadGraph default.

Best,

Kenneth

govoni:
We agree with Kenneth about the gridpacks regeneration, could it be done please? Luca and Pietro

abdollah110:
Dear Pitero, Luca and Kenneth

Thanks for the information. I will regenerate the gridpacks and will let
you know, hopefully by the end of today.
Do I need to upload them somewhere as well?

Cheers
Abdollah

perrozzi:
Dear Abdollah,
the recommended procedure is to update the datacards, wait for them to be approved, and only then produce the gridpacks with the approved datacards.
Otherwise you risk to regenerate the gridpacks several times...

Thanks
Luca

abdollah110:
Dear Luca

I have updated the datacards.
The only changes are as following:
Updating the pdf as was suggested in the run_card.dat
Adding the 5FS in the proc_card.data

Would you please have a look and let me know if it is OK with you.
I am in the process of generating the gridpack. I sometimes see that some of the jobs fail randomly and I need to resubmit them again. Is such failures expected?
I am following the procedure describe in 
https://twiki.cern.ch/twiki/bin/viewauth/CMS/QuickGuideMadGraph5aMCatNLO#Quick_tutorial_on_how_to_produce


Thanks
Abdollah

bin/MadGraph5_aMCatNLO/cards/production/13TeV/DarkMatter_Codex/Codex_LQ1000_DM_400_X_420_gen2/Codex_LQ1000_DM_400_X_420_gen2_proc_card.dat
```diff
@@ -34,6 +34,11 @@ define l+ = e+ mu+
 define l- = e- mu-
 define vl = ve vm vt
 define vl~ = ve~ vm~ vt~
+
+#import model sm-no_b_mass
```
perrozzi: ciao, all the import commands are commented out, so you are not importing...
please fix

abdollah110:
Fixed the import issue

kdlong:
Hi Abdollah, all,

Looking closer at the cards, I realize I gave some misleading information. Adding import sm-no_b_mass to your cards does nothing, as this is later overwritten by the second import statement: import lq_met. 

Since you use the model lq_met, the flavor scheme is dependent on this model. Looking at it briefly, it looks like it uses a 4F scheme, so the easiest thing to do would be to use a 4F PDF (263000). If you want to use 5F instead, you can modify the model to have massless b-quarks (or by adding a restrict_no_b_mass.dat file to the model). If you do use a 4F scheme, setting "4 = maxjetflavor" is probably a good idea for clarity, but I think this is only used if you use MLM matching in any case.

Best,

Kenneth

abdollah110:
Dear all

This is getting so long. Can we converge on the cards and I start the
gridpack production?
Is current cards incompatible with SF5?

Cheers
Abdollah

abdollah110:
On Fri, Oct 21, 2016 at 9:28 AM, perrozzi <notifications@github.com> wrote:

> import model sm-no_b_mass



​Done


Cheers
Abdollah
​

kdlong:
Hi Abdollah,

Yes, the model is incompatible with the 5F scheme unless it is modified. The easiest thing to do is to go back to 4F (so remove b from the proton definition) and use PDF 263400.

Best,

Kenneth

abdollah110:
Hello Kenneth


On Mon, Oct 24, 2016 at 4:44 AM, Kenneth Long <notifications@github.com>
wrote:

> Hi Abdollah, all,
>
> Looking closer at the cards, I realize I gave some misleading information.
> Adding import sm-no_b_mass to your cards does nothing, as this is later
> overwritten by the second import statement: import lq_met.
>
​Initially I had commented this line, but I have been asked to add that
line again.​


> Since you use the model lq_met, the flavor scheme is dependent on this
> model. Looking at it briefly, it looks like it uses a 4F scheme, so the
> easiest thing to do would be to use a 4F PDF (263000). If you
> ​d​
> want to use 5F instead, you can modify the model to have massless b-quarks
> (or by adding a restrict_no_b_mass.dat file to the model). If you do use a
> 4F scheme, setting "4 = maxjetflavor" is probably a good idea for clarity,
> but I think this is only used if you use MLM matching in any case.
>
> ​I have discussed the issue of 4F v.s. 5F with the theorists (author of
the Codex paper and the model lq_met). They mention it is OK to use 5FS.
So having said that, do I still need change the current  ​
​data cards?

Cheers
Abdollah
​

abdollah110:
I have switched back from 5FS to 4FS  and switched to 263400 as PDF

abdollah110:
On Tue, Oct 25, 2016 at 1:39 PM, Kenneth Long <notifications@github.com>
wrote:

> 263400



​Thanks Kenneth

I have switched back to 4FS and updated the PDF to 263400
Please let me know in case there is still any issue, otherwise please
accept the pull request and I will start producing the gridpack.

Cheers
Abdollah



​

bin/MadGraph5_aMCatNLO/cards/production/13TeV/DarkMatter_Codex/Codex_LQ2000_DM_900_X_945_gen2/Codex_LQ2000_DM_900_X_945_gen2_run_card.dat
```diff
@@ -0,0 +1,283 @@
...
+#*********************************************************************
+# Beam polarization from -100 (left-handed) to 100 (right-handed)    *
+#*********************************************************************
+     0.0     = polbeam1 ! beam polarization for beam 1
+     0.0     = polbeam2 ! beam polarization for beam 2
+#*********************************************************************
+# PDF CHOICE: this automatically fixes also alpha_s and its evol.    *
+#*********************************************************************
+     'lhapdf'    = pdlabel     ! PDF set                                  
+     263000 = lhaid ! if pdlabel=lhapdf, this is the lhapdf number
```
perrozzi: I assume this needs to be 263400 too...

abdollah110:
I actually had removed /Codex_LQ2000_DM_900_X_945  and /Codex_LQ2000_DM_900_X_1035  from my request lists and added two new "Low" mass points, so this is the reason that the above 2 mass points had not been updated. I will 'git rm ' them and commit again.

perrozzi:
the rest looks ok to me, merging

abdollah110:
Thanks

abdollah110:
I have finished producing the Gridpacks  and they are located here :

/afs/cern.ch/user/a/abdollah/public/DM_Codex

Would you please let me know what the next step is?


Thanks
Abdollah

perrozzi:
which PAG is asking for these requests? Check with the corresponding MC contacts https://twiki.cern.ch/twiki/bin/view/CMS/GeneratorContactPersons

PR #984, opened 2016-10-24
Explicitly compile all subprocess in rwgt directory

kdlong:
Following suggestion from MadGraph authors, manually compile libraries
for all subprocesses used by reweight. By default, subprocesses are only
compiled as they are encountered during the reweight run. This change
forces all to be pre-compiled. Note: Requires the lines
"change rwgt_dir rwgt" to be in the reweight_card.dat (currently the
directory name "rwgt" is hardcoded in the compilation commands).

### DIFF
--- a/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
+++ b/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
@@ -513,6 +513,16 @@ else
       export LIBRARY_PATH=$LD_LIBRARY_PATH
       cd madevent
       bin/madevent reweight pilotrun
+      # Explicitly compile all subprocesses
+      for file in $(ls -d rwgt/rw_me/SubProcesses/P* | head -4); do
+        echo "Compiling subprocess $(basename $file)"
+        pushd $file > /dev/null
+        for i in 2 3; do
+            MENUM=$i make matrix${i}py.so >& /dev/null
+            echo "Library MENUM=$i compiled with status $?"
+        done
+        popd >& /dev/null
+      done
       cd ..      
   fi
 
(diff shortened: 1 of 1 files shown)


perrozzi:
I assume it has been tested and only affects the generation if the reweight_card.dat is present, right?

kdlong:
To be honest I didn't explicitly check yet, thought the if statement should only be entered if the reweight_card.dat exists. Likewise the for loop is simply skipped if the appropriate directory structure isn't found.

I wanted to make the pull request so I could reference it in the meeting today, but I do think it should be tested a little more before being merged.

perrozzi:
what about these datacards? do they need this feature?
https://github.com/cms-sw/genproductions/pull/983/files#diff-1af8c6902564c074f374857955761a39R25

perrozzi:
@jasperlauwers can you also comment?

jasperlauwers:
Yes, I had to use this version of the gridpack_generation.sh script to be able to generate the gridpack.

perrozzi:
what's the status here? still testing? can be merged?

kdlong:
I've used it a few times and others have as well. I think it's ready to be
merged.

Note that it doesn't work with MadSpin currently, which is supposed to run
after reweight, while the scripts put it before. There's still an issue
with this in the mg240 version that I just now realized actually, the order
is corrected in the generation scripts but not in the runcmsgrid.sh script.
Hopefully this is a trivial fix.

vciulli:
Merging will not break the usage of MadSpin without reweighting right? 
The trivial fix you are mentioning is for mg240?
Isn’t this already our default mg release in CMSSW? 
It was validated some time ago and though on the longer term we plan to move to mg250 for the next production I thought we moved to mg240 for the present one (but I might be wrong).

Vitaliano

kdlong:
Hi Vitaliano,

Merging shouldn't affect the performance without reweighting. 

I'm talking about needing another fix in MG to make reweight+madspin work. This might be possible in v2.3.3, but I was planning to make the change in 2.4.2 because I was planning to move to using it. As for 2.4 it being the default version, I think it is considered validated, but it's not yet used by the master branch. My understanding is that requests with both versions are currently supported, @bendavid might have more info.

Actually it seems like there's a bug affecting  2.4.2 which I'm trying to sort out with the authors now, hopefully we'll have a fix soon.

vciulli:
Thanks Kenneth for the reply
I have no objection to merge this pull request. 
V.

PR #986, opened 2016-10-26
Add run_pwg_parallel.py to steer MiNLO gridpack production

perrozzi:
allows to create the MiNLO gridpack (in general, the powheg gridpacks requiring parallelization) in one go. based on an example from @pfs 
automatic resubmission not yet included (very rarely needed)


perrozzi:
NNLOPS handling not included (yet)

PR #989, opened 2016-10-27
Modifications to allow reweight in MG2.4 branch

kdlong:
Allow cluster queue to be set properly for condor running. Originally bug affecting set cluster_queue None is now fixed, it seems that there is now an issue that only affects LSF bash. Seems to be working properly for condor submission (https://bugs.launchpad.net/mg5amcnlo/+bug/1517041).

Add precompilation of MEs (same as https://github.com/cms-sw/genproductions/pull/984) also move reweight to run before madspin.


bendavid:
Looks good to me.

PR #990, opened 2016-10-27
Update ZH120_012j_NLO_HtoMuMuGamma_madspin_card.dat

andreypz:
Commenting out Z decay: missed that for 120 GeV mass point


perrozzi:
do you plam to use pythia for the z decay? why? to reuse the same gridpack with different decay options?

andreypz:
Yes, that is the plan. Because the Z spin would not be correct with this MadSpin card (spinmod=none).
I thought that it will be correct with Pythia-8 decay.  And if it's not correct there too, than what's the difference?
Without Z decays, the gridpacs are quicker to produce.

bendavid:
Agreed that in this case there is no advantage to using madspin for the Z decays over pythia.

The only other possibility would be to have the z decay specified as part of the process definition (but this would require removing one or both of the additional jets because of the extra complexity)

andreypz:
Yes, I thought of that too. I concluded that the extra jets of NLO is a nicer feature if the Z decay is done in pythia.

PR #994, opened 2016-10-31
Turns out the set cluster_queue command is broken...

kdlong:
Turns out the set cluster_queue is broken for condor as well. Apparently I made a mistake testing it earlier. Sorry about that. Now adopting the technique to explicitly set it in the configuration file after generation as used for LSF.


bendavid:
Looks fine (and makes sense that this would be similarly broken for lsf and condor).

kdlong:
Right, though I was thrown off by the authors suggesting it worked fine for them... In any case it works now.

PR #998, opened 2016-11-03
DY to tau tau cards

l-cadamuro:
This PR contains the cards for DY production in 2/3/4 LO jet multiplicity and B/BB jets, forcing the production of a tau pair.
It targets bbtautau analysis: jet-binned samples are used because they provide the most efficient coverage of this particular event topology.
Cards were created with minimal changes with respect to the existing ones, corresponding to same jet multiplicities but with decays into e/mu/tau pairs
[1] : https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/DYJets_jetbinning_LO_MLM
[2] : https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/DYBJetsToLL_M-50_5f_LO_MLM
[3] : https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/DYBBJetsToLL_M-50_5f_LO_MLM

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/DYJetsToTauTau/DY2JetsToTauTau_madgraph_5f_LO/DY2JetsToTauTau_madgraph_5f_LO_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/DYJetsToTauTau/DY2JetsToTauTau_madgraph_5f_LO/DY2JetsToTauTau_madgraph_5f_LO_proc_card.dat
@@ -1,0 +1,40 @@
+#************************************************************
+#*                        MadGraph 5                        *
+#*                                                          *
+#*                *                       *                 *
+#*                  *        * *        *                   *
+#*                    * * * * 5 * * * *                     *
+#*                  *        * *        *                   *
+#*                *                       *                 *
+#*                                                          *
+#*                                                          *
+#*         VERSION 1.5.11                2013-06-21         *
+#*                                                          *
+#*    The MadGraph Development Team - Please visit us at    *
+#*    https://server06.fynu.ucl.ac.be/projects/madgraph     *
+#*                                                          *
+#************************************************************
+#*                                                          *
+#*               Command File for MadGraph 5                *
+#*                                                          *
+#*     run as ./bin/mg5  filename                           *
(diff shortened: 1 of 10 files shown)


govoni:
@l-cadamuro could you please clean up duplications in the definition of particles?
@bendavid @kdlong @perrozzi do you see any issues with these datacards?

kdlong:
Hi all,

The only comment I have is that 'ptj = 0.01' seems ridiculously low. I'm not sure if such a low pt jet even makes sense for a LO calculation, and I expect such events to fail the matching anyway. It would be more efficient to move this up to pt > 10 or so (but not greater than xqcut). 

In general cards look fine to me though, looks they were directly copied from samples which were already produced in any case. 

Best,

Kenneth

l-cadamuro:
Hi,

yes, I took the very same cards as DY 2/3/4 and B/BB jets and only changed the decay in the process. I avoided any change to the settings as the idea is to use these samples together with the DY->ll and only increase statistics for the tautau decays.
I can change the ptj value to 10 if needed, but should it then be changed also in the DY->ll samples to have a coherent output?

@govoni : MadGraph experts confirmed that duplicated declarations are not a problem as the last one overrides the previous [1]. If you want I can remove them but it won't affect the generation.

Thanks,
Luca

[1] : https://answers.launchpad.net/mg5amcnlo/+question/403756

govoni:
@l-cadamuro ok, then leave the cards as they are :)
For the time being let's also keep the cfg coherent, we'll not forget to fix the ptj for the next iteration. @kdlong do you agree?

kdlong:
My expectation is that it won't make any difference at all in the final distributions to move the ptj cut up to 10, since all the lower pt events will fail the matching. This is what the auto_ptj_mjj cut would do anyway. You could check this by generating in both setups and seeing if (matching efficiency)*(initial cross section) is equal in each case. 

But if it's too much trouble, the only hit is in efficiency of the generation. It's up to you.

perrozzi:
Kenneth is right, however it is true that also previous datacards were created like this...
I think the net effect should be a lower matching efficiency, but should be a small effect. merging

PR #1002, opened 2016-11-06
add neutrino pt (i.e. MET) filtered w+jets samples (100-200)

mialiu149:


perrozzi:
can you rename the PR? instead of pt filtered, MET filtered for example

PR #1006, opened 2016-11-08
High-statistics ZZ+jets samples

Werbellin:
Sampels to be used in ZZ+jets analysis, only on-shell Z bosons


perrozzi:
what is "p" in ZZTo4lp1j ?

Werbellin:
p=plus

PR #1010, opened 2016-11-11
Patch for 'change rwgt_dir' in 2.4.2

kdlong:
Adding patch from Olivier Mattalear to fix bug affecting 'change rwgt_dir' in 2.4.2, which is critical for using precompiled MadSpin. Also moving MadSpin to run after reweight.

@bendavid I wasn't sure how you managed to format the patch files in the way you did. Let me know if it should be done differently.

Correspondence with Olivier:

I actually
fixed many bug in 2.5.0 related to “change rwgt_dir rwgt”
but I guess mainly for the NLO re-weighting and not that many for the LO
re-weighting.

Here is the full difference in those file (in case they can be usefull)
    http://bazaar.launchpad.net/~madteam/mg5amcnlo/series2.0/revision/267#madgraph/interface/reweight_interface.py
http://bazaar.launchpad.net/~madteam/mg5amcnlo/series2.0/revision/268#madgraph/interface/reweight_interface.py

Just for information here is a small timeline of the re-weighting package:
2.0.0 (2013) first implementation (only same model/same process) based on MadSpin idea (fortran executable called on the flight)
2.3.2 (2015) new implementation via fortran library linked to python (f2py). possibility to change model/process definition.
                    possibility to perform a LO accurate re-weighting on NLO sample.
2.4.0 (2016) proper NLO re-weighting of NLO sample.
2.5.0 (2016) Possibility to run the re-weighting on multiple core
2.x.y (2017) change of structure.
                    - having only two (four at NLO) fortran library to be linked to python (faster compilation/better memory footprint)
                    - the fortran code would be modified such that the python code will extract information from the fortran library directly.
                      This allows more flexibility for modifying the re-weighitng code by hand.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/patches/0009-fix-to-change-rwgt_dir-for-LO-reweight.patch
+++ b/bin/MadGraph5_aMCatNLO/patches/0009-fix-to-change-rwgt_dir-for-LO-reweight.patch
@@ -1,0 +1,27 @@
+diff --git a/interface/reweight_interface.py b/interface/reweight_interface.py
+--- a/madgraph/interface/reweight_interface.py
++++ b/madgraph/interface/reweight_interface.py
+@@ -1612,18 +1612,21 @@
+         to_save['rwgt_dir'] = self.rwgt_dir
+         to_save['has_nlo'] = self.has_nlo
+         to_save['rwgt_mode'] = self.rwgt_mode
++        to_save['rwgt_name'] = self.options['rwgt_name']
+ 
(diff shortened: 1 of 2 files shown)


kdlong:
Hi all, have you had a chance to look at this?

perrozzi:
ping @bendavid @vciulli (don't remember Qiang username)?

vciulli:
adding @qili 
I had a quick look and if @kdlong tested it I am ok to merge it
@bendavid: do you want to have a closer look ?

PR #1023, opened 2016-11-23
change t-channel cards for the heavy neutrino search

helee:
We change model cards with fixed version, change process cards and run cards.


perrozzi:
hi, since the diff is very heavy, could you elaborate a bit more about the changes? thanks

helee:
We decided to decay particles(i.e. heavy neutrinos) using madspin,
so we added thd madspin cards and split the channels according to letopn's charge.
We want to see sama-sign leptons from the decay of heavy neutrinos.
Also, we adjust kinematic cuts to optimize for the signals.

helee:
If there is no special problem with the new cards, could you merge as soon as possible?
I urgently change the cards while the official production is ongoing.

PR #1024, opened 2016-11-24
adding WZ,WV aTGC cards

senka:


perrozzi:
can you please confirm that the model uses 4F?

senka:
I use two different restrict cards with the model. One is 4f (cmass=0) and another is 5f (cmass=bmass=0). I use 4f for WW and 5f for WZ. If you want to have a look I put the model and all restriction cards here:

/afs/cern.ch/user/s/senka/public/EWdim6Model

Senka

senka:
Any more questions or comments about the cards?

Senka

PR #1025, opened 2016-11-28
Contact Interactions Fragments for EXO group

jotadram6:


jotadram6:
I missed 20 fragments to be included in the pull request, should I revert this pull request or should I do a new one?

govoni:
as you wish

perrozzi:
may I suggest to simply add the missing cards instead of creating a commit to revert and another to add them all? probably it is a bit cleaner, although you will have 2 commits for the datacards (I suggest use the same name and use pt2 or so)

jotadram6:
Ok, no problem

PR #1031, opened 2016-12-01
MajoranaNeutrinoToTriLeptons gridpack cards added

jedori0228:
MajoranaNeutrinoToTriLeptons gridpack cards added.

https://docs.google.com/spreadsheets/d/1nBAko5DgS9dMslxqQcQOs2TYcpWkxmuQ2VBg5BsphBU/edit#gid=0

PR #1037, opened 2016-12-06
MajoranaNeutrinoToTriLeptons gridpack cards added

jedori0228:
MajoranaNeutrinoToTriLeptons gridpack cards added.

https://docs.google.com/spreadsheets/d/1nBAko5DgS9dMslxqQcQOs2TYcpWkxmuQ2VBg5BsphBU/edit?usp=sharing

reviesed commit of https://github.com/cms-sw/genproductions/pull/1031.
- Unnecessary lines are removed in proc_card. (lines that imorts sm model cards)


perrozzi:
do I see correctly that the model is NLO and the PDF are LO (263000 instead of 260000)?
(cfr https://lhapdf.hepforge.org/pdfsets.html)
also, is the model 5F or 4F?

jedori0228:
It is a NLO model, but our process is LO, so we use LO pdf.
And this is a five-flavour model.
Model details are expained here : http://feynrules.irmp.ucl.ac.be/wiki/HeavyN

PR #1052, opened 2016-12-14
Nonres hh example (no decay)

OlivierBondu:
Similar as https://github.com/cms-sw/genproductions/pull/1051: example card without dihiggs decay, more cards to follow if this template is approved. The physical content is exactly the same as stuff eg in https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/NonRes_hh/4b_node_sm , but Higgs are left undecayed


govoni:
If the gridpack for this card is going to replace an existing one, please handle the versioning of gridpacks with subfolders (v1, v2, ...)

OlivierBondu:
I would rather delete the old cards being not used any longer from the `master` branch, for now this would concern only the https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/NonRes_hh/vvbb_fullLep_node_* folders, would that be fine ?

Then of course, we still would put the new gridpacks in a 'v2' directory on cvmfs and add a v2 also to the sample naming...

govoni:
why do you want to delete the old cards? they serve as reference for the analyses which have been using them.

OlivierBondu:
github has memory, nothing is forgotten :) 

in this particular case (bbVV analyzers) the proper cards are linked in MCM, see [this link](https://cms-pdmv.cern.ch/mcm/edit?db_name=requests&prepid=HIG-RunIIWinter15wmLHE-00604&page=0) which point to [this folder](https://github.com/cms-sw/genproductions/tree/aa244debe3894ae9c80c846342db3366f7cacb45/bin/MadGraph5_aMCatNLO/cards/production/13TeV/NonRes_hh/vvbb_fullLep_node_sm) which features the exact commit hash, so if you are digging from the sample itself you can always find back the reference

govoni:
yeah, but it is very cumbersome to consult... my personal preference is to keep is as plain as CVS would be, @perrozzi what do you think?

OlivierBondu:
(IMHO not more cumbersome than understanding where the hell are the cards in this huge repo, and why would there be some v1 and not v2 for everyone, etc. But you're the boss boss ! :smile: )

govoni:
:) I am old, and still refusing to use svn...

PR #1056, opened 2016-12-14
Adding two data cards for ttH and TT futures with 14TeV settings

3wolf3:


perrozzi:
what about hdamp here?

3wolf3:
hdamp is set to the 272 value we use currently in ttH

PR #1062, opened 2016-12-16
cards for single VLQ production with variable width

peiffer:
these cards have been created originally by Alexandra Carvalho, see https://github.com/acarvalh/singleVLQ_wide


peiffer:
Some of the cards are not yet ready and have to be modified

PR #1064, opened 2016-12-16
Fix for bug in runcmsgrid_powheg*.sh scripts  

jfernan2:
which caused GenXsecAnalyzer to report null cross-section after matching in Pythia8

### DIFF
--- a/bin/Powheg/runcmsgrid_powheg.sh
+++ b/bin/Powheg/runcmsgrid_powheg.sh
@@ -360,18 +360,18 @@ if [ -s pwgstat.dat ]; then
   mv pwgstat.dat pwg-stat.dat
 fi
 
-if [ -s pwg-stat.dat ]; then
-  XSECTION=`cat pwg-stat.dat | grep total | awk '{print $7}'`
-  XSECUNC=` cat pwg-stat.dat | grep total | awk '{print $9}'`
-  head=`cat   cmsgrid_final.lhe | grep -in "<init>" | sed "s@:@ @g" | awk '{print $1+1}' | tail -1`
-  tail=`wc -l cmsgrid_final.lhe | awk -v tmp="$head" '{print $1-2-tmp}'`
-  tail -${tail} cmsgrid_final.lhe                           >  cmsgrid_final.lhe_tail
-  head -${head} cmsgrid_final.lhe                           >  cmsgrid_final.lhe_F
-  echo "  "$XSECTION"   "$XSECUNC"  1.00000000000E-00 10001" >>  cmsgrid_final.lhe_F
-  echo "</init>"                                           >>  cmsgrid_final.lhe_F
-  cat cmsgrid_final.lhe_tail                               >>  cmsgrid_final.lhe_F
-  mv cmsgrid_final.lhe_F cmsgrid_final.lhe
-fi
+#if [ -s pwg-stat.dat ]; then
+#  XSECTION=`cat pwg-stat.dat | grep total | awk '{print $7}'`
+#  XSECUNC=` cat pwg-stat.dat | grep total | awk '{print $9}'`
+#  head=`cat   cmsgrid_final.lhe | grep -in "<init>" | sed "s@:@ @g" | awk '{print $1+1}' | tail -1`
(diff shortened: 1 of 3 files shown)


jfernan2:
Copy of the message in the hypernews:

I have just found that latest Powheg gridpacks have a bug when collecting the cross-section and modifying the final lhe file.

This change in the runcmsgrid script [1] followed by this other [2] cause that, (specially) the non inclusive gridpacks (such as e.g. TTTo2L2Nu or TTToSemiLepton) the total cross-section of the process (hvq in the previous examples) is taken from pwg-stat.dat (which saves the inclusive process) and inserted in the LHE file replacing the exclusive one from pwgevents.lhe.

Moreover, the fixed 10001 at the end of the line does not reflect the process name and decay, being different for every process (e.g. 322222 for inclusive hvq, 311111 for semileptonic hvq, 322200 for dileptonic hvq,...). 

The consequence[3] of this is that, when runing Pythia8 over the gridpack, the GenXsecAnalyzer gets confused and reports a final null cross-section after matching (inclusive - inclusive = 0). For the excluisve gridpacks the behaviour is the same and the exclusive cross-section value is not reported.

The fix[4] for the bin/Powheg/runcmsgrid_powheg.sh is apparently easy, just commenting the section which modifies the line in the final lhe file. But I ignore if those lines were introduced for other purpouse, so I let the experts to comment and discard/modify the PR if so.

Thanks,
Javi

[1] https://github.com/cms-sw/genproductions/commit/29d42b632e916bb1356caad9c36c288bf5468ef0#diff-27073c98e429b1b16c3d90c356b74e8cL310

[2] https://github.com/cms-sw/genproductions/commit/d60cd3eb381b16fad9bbe37ee023e8e57f072114#diff-27073c98e429b1b16c3d90c356b74e8cL317

[3]
GenXsecAnalyzer:
------------------------------------
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Overall cross-section summary
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Process         xsec_before [pb]                passed  nposw   nnegw tried  nposw    nnegw   xsec_match [pb]                 accepted [%]     event_eff [%]
0               8.645e+02 +/- 6.791e-01         0       0       0       0      00       0.000e+00 +/- 0.000e+00         -100.0 +/- -100.0       -100.0 +/- -100.0
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Total           8.645e+02 +/- 6.791e-01         0       0       0       0      00       -8.645e+02 +/- 0.000e+00                -100.0 +/- -100.0       -100.0 +/- -100.0
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Before matching: total cross section = 8.645e+02 +- 6.791e-01 pb
After matching: total cross section = -8.645e+02 +- 0.000e+00 pb
Filter efficiency (taking into account weights)= (1000) / (1000) = 1.000e+00 +- 0.000e+00
Filter efficiency (event-level)= (1000) / (1000) = 1.000e+00 +- 0.000e+00
After filter: final cross section = 0.000e+00 +- 0.000e+00 pb

perrozzi:
I would like @violatingcp to comment on the reason why this modification was added, affecting the whole powheg production for what appears a specific use-case...

violatingcp:
This commit was done 1.5 years ago when these scripts were first being
adopted (there was only one other user at the time). I do not recall Powheg
properly reporting the cross sections for a number of processes. The fix
used the appropriately reported pwg-stat.dat file to extract the cross
sections and update the lhe file. Powheg has a rich history of not
reporting the right cross section to the LHE header, it comes as a surprise
that they are considering doing it now.

Additionally, it has always been recommend by the gen group that the header
cross sectoin not be used for any official reporting of numbers. For a real
improvement to this script, the updater should validate this on a number of
processes that currently use Powheg.

jfernan2:
What's causing the bug is not the replacement of cross-section value itself, but rather the last process number added at the end, which is fixed in the script (10001) and it is actually variable depending on the physics process. This is what makes GenXsecAnalyzer crazy. I would propose to fix the script so that the index is taken from the LHE file or anywhere else.

jfernan2:
I have added to this PR a possible solution for the process dependent number

PR #1065, opened 2016-12-19
Cards for DY + Jets LO

amagitte:
New cards for DY+0/1/2/3 jets at LO and MLM scheme


perrozzi:
ciao, why only 3 additional jets? if I remember correctly we usually do it up to 4.
also, there is a second unrelated commit in the PR, I assume is still relevant to merge it, right?

amagitte:
Ciao Luca,

Actually I have just done a chat with SM people and I will remove DY directory..
I am cloning their request.
the second commit in the PR request is relevant because it is a fix in a parameter of JHUGen input file.

PR #1073, opened 2017-01-09
HJJ M300 input cards

amagitte:
New input cards for HJJ at 300 GeV for HZZ group


perrozzi:
did you use single core or parallel?
did you have a brief look at the integration grids?
(in particular, nbound seems to be a bit on the low end side)

amagitte:
Hi Luca,

The grid-pack was prepared by David.
He used a single core, and here[*] you can find the plots of the grid, which looks ok.

[*] /afs/cern.ch/user/d/dsperka/public/Run2MC/HJJ_300/pwg-btlgrid.top

PR #1077, opened 2017-01-12
Decay fragment for HH->bbWW (semi-leptonic W decay)

OlivierBondu:
Content as per title, actual fragment constructed taking inspiration from [HH->bbZ(ll)Z(jj)](https://github.com/cms-sw/genproductions/blob/master/python/ThirteenTeV/Higgs/HH/ResonanceDecayFilter_example_HHTo2B2L2J_madgraph_pythia8_cff.py) as well as [HH->bbVV->bbllnunu](https://github.com/cms-sw/genproductions/blob/master/python/ThirteenTeV/Higgs/HH/ResonanceDecayFilter_example_HHTo2B2L2Nu_madgraph_pythia8_cff.py) decay fragments

May be useful 'soon' as ATLAS is most likely coming with the analysis of this channel for Moriond (no analysis started in CMS as of now)

tagging:
- @DavidMorse and @bendavid for their knowledge of the ResonanceDecayFilter
- @govoni and @giacomoortona for information since we discussed this

PR #1078, opened 2017-01-12
forgot to switch on ResonanceDecayFilter:udscbAsEquivalent

OlivierBondu:
follow up on https://github.com/cms-sw/genproductions/pull/1077

tested with this fix on a dozen of events: work as expected


govoni:
do you plan to inject requests for these samples? it might be worth profiting of resources getting free

PR #1086, opened 2017-01-18
bb4l

intrepid42:
Further changes to make bb4l work

Running might be very slow with full number of weights

Step3 jobs crashed when waiting for first job to create the fullgrid*.dat files, added option to do a step3 pilot run and submit full number of jobs afterwards


intrepid42:
2:16h run time for 100 events with complete set of scale/pdf weights

perrozzi:
let us know when is ok to merge

intrepid42:
Should be ok to merge :)

Should we add somewhere in the twiki that it should run with 64 simultaneous jobs?

Also, I would propose to modify the instructions in the twiki and replace these
```
    cp $genpath/bin/Powheg/*.py .
    cp $genpath/bin/Powheg/*.sh .
    cp -r $genpath/bin/Powheg/patches .
```
by `ln -s ...` so that the tarballs are created with the newest versions after updating the code from git.

perrozzi:
feel free to add a dedicated section for this process to https://twiki.cern.ch/twiki/bin/viewauth/CMS/PowhegBOXPrecompiled
thanks

PR #1091, opened 2017-01-24
Charged Higgs to tau nu

amarini:
High mass points for H+- -> tau nu.
Mass points [180,3000] GeV:
* new model (2HDMtII_NLO)
* H+ and H- simultaneously 
* every 100 GeV up to 800 GeV, every 200GeV up to 3TeV
Mass points (500, 3000] GeV:
* H+ decay in pythia8 (not in madspin)
* lhe are identical for H+ -> tb, since H+ is not decayed


perrozzi:
why this change of name?
https://github.com/cms-sw/genproductions/pull/1091/files#diff-af8d8ca05bfdb08d9ff426a60e6a4240

amarini:
Hi Luca, 
these renaming are there in order to remove madspin configuration for mH > 500 GeV.
This step is necessary since MadSpin (to my understanding) do not probe in the test run all the processs, and therefore fails later.

The strategy to avoid this, for the time being, is to let pythia decay the charged Higgs, as we (me, you and Josh)  discussed in an email tread the 5-6th May 2016, and later summarized in the HExo meetings the 23rd
[link](https://indico.cern.ch/event/515356/?filterActive=1&showDate=2016-05-23&showSession=7)

The choice to have mv -> xxx, reflects the hope that this problem will be fixed in the future. 
Consider also that the madspin files, are identical, therefore git may associate this moving uncorrectly between the directories. (eg, mv 3000 -> 1200) 

Eg,
> diff /tmp/amarini/genproductions//bin/MadGraph5_aMCatNLO/cards/production//13TeV/ChargedHiggs_heavy_M{1200,3000}/XXX_madspin_card.dat 
> echo $?
> 0

-Andrea

perrozzi:
ok thanks, if it's just a trick to keep the files for further updates without using them, I think is fine

PR #1093, opened 2017-01-24
Add low mass WZ EWK and QCD MLM cards

kdlong:


perrozzi:
are the different jet multiplicities meant to be used together? if so, is there any specific setting in the datacards?
why different datacards for different multiplicities?

kdlong:
Hi Luca,

Yes, the different multiplicities need to be used together. There are separate cards because there is a separate sample and therefore separate gridpack for each multiplicity. The intention for these samples is to be used as backgrounds for VBS, so the statistics requested are much higher in the 2 and 3j bins. There's no special flag for this in MadGraph (just ickkw=1), its controlled by the maxJet/minJet settings in Pythia.

PR #1097, opened 2017-02-01
Chaging sherpack location to match serpack fetcher of old CMSSW release

jotadram6:


perrozzi:
will you change the directory every time you need new sherpa release?
have you though at a general strategy for cmssw-dependent fragments?

jotadram6:
This change is necessary due to the specific cmssw release requested by analyzers. You can check the recommendation from sherpa expert here: https://hypernews.cern.ch/HyperNews/CMS/get/generators/3277/1/2/1.html
I don't know if the sherpa experts have an strategy to avoid to change sherpack directory when changing cmssw release.

vciulli:
adding @pmillet

I think is good to have both the gcc and the sherpa version in the path of the sherpack as they are not supposed to work with a different versions from those used to create them. 
However the dependence on the cmssw version is only due to changes on the above libraries (and openloops or blackhat if they are used)

pmillet:
As vitaliano said, the sherpacks are bound to a specific gcc <-> sherpa combination. The changes for this specific commit were needed, since for CMSSW version prior to CMSSW_7_1_21, an old mechanism (fn-fileget) is used to get the sherpacks and thus the path need to point to afs instead of cvmfs.

PR #1104, opened 2017-02-10
Settings for ttbar with mtop width change

jfernan2:
Settings for ttbar with mtop width x8, x0.5 and x0.2 samples with TuneCUETP8M2T4
Settings for ttbar with mtop width x0.5 and x0.2 samples with TuneCUETP8M1


govoni:
out of curiosity, would it be in principle possible to implement these modifications through weights?

jfernan2:
Honestly I don't know, but I don't believe it is possible since it will affect kinematics of the decay products, right?

govoni:
my naive understanding of the reweighting is that you change the population of the linewidth so to deform it: if you want to make a wider lineshape, your weights will be larger than one away from the pole, while they will be smaller if you want to narrow the width. If your "wanna-be reweighted sample" populates a place which is not covered by the main production, this is not possible (for example, to change the pole mass more than the width) otherwise I don't see why not. @perrozzi, how do you do it with W samples?

perrozzi:
In my case the W width was larger than the mass scan, so that a single generation was sufficient to account for the full reweighting, in the top case one might need to generate a quite large width and pay some inefficiency price that will vanish the effort (but it's a strategy that should be at least pursued with a back of the envelope calculation)

jfernan2:
Not sure if I fully understand you Luca, nevertheless for the top case the width is not larger than the mass scan...

govoni:
I think we can merge this request, and we suggest to check how you could use poweheg reweigting also in this case. @perrozzi, do you have an example of usage in the W mass study?

PR #1105, opened 2017-02-10
DY for TDR studies at 14 TeV

amagitte:
Card to be used in production of the DY at 14 TeV for TDR studies.


govoni:
@vciulli @kdlong can we merge these?

kdlong:
Looks good to me

PR #1121, opened 2017-03-07
exclusive WW

Mohammed2:


perrozzi:
may I ask why this sample and, would be the difference wrt the inclusive pythia8 (statistics?). thanks

rebassoo:
This sample is quite different from the inclusive sample. It is an exclusive process where the protons stay intact and they radiate photons that collide and produce two W bosons. This is the signal for our analysis, which we published for the 7 and 8 TeV data, and are beginning to investigate for 13 TeV. Our background is inclusive WW.

Thanks,
Finn Rebassoo

PR #1122, opened 2017-03-08
Adding back b_bbar_4l, fix d/l input card bug, update to Mar 2017 source tarballs.

yuanchao:
The Vres version of PowhegBox requires gcc 520. Since it supports the new RWGT scheme, we can use the same script as this gcc530 one. So the b_bbar_4l process is added back (leave out the patch to make it compile-able in gcc48). Also now pointing the source tarball to the Mar. 2017 version. A bug for d/l directly the input card is also fixed.


perrozzi:
things look ok to me, @intrepid42 please crosscheck the b_bbar_4l related part

intrepid42:
Thanks a lot, I think this should work :)

PR #1127, opened 2017-03-13
Add automated scripts to compute cross sections for existing datasets

perrozzi:
documented in 
https://twiki.cern.ch/twiki/bin/view/CMS/HowToGenXSecAnalyzer#Automated_scripts_to_compute_the
could be useful in particular to fill the cross section database


perrozzi:
@kdlong did you try by chance to use these scripts and have any feedback?

perrozzi:
@syuvivida can you please comment if is ok with you and you can briefly test it, in case (of course this PR doesn't include the modification to cmssw to display the fraction of negative weights and the equivalent luminosity for 1M events, which is added as git cms-merge-topic in the twiki)

kdlong:
@perrozzi Very nice! Will give it a closer look and test carefully in the next few days. I also have some code that I wrote quite a while back which could be useful to extend it (for example to do the scale and/or PDF uncertainties from the EDM file).

One quick comment. A lot of the modules loaded by ana.py aren't actually needed (looks like it was copied from cmsDriver at some point). I'd been in favor of making it more minimal, e.g. https://github.com/kdlong/TheoreticalUncertainties/blob/master/External/genXsec_cfg.py

syuvivida:
@perrozzi, I am sorry but I am kind of busy with administrative/teaching work in the university these two days, if @kdlong can help, it would be great!

perrozzi:
added a few minor updates by kenneth, basically ready to be integrated

PR #1129, opened 2017-03-13
Small fixes

covarell:
- Add simple script to create POWHEG-RES source
- Remove completely hdamp script (not suggested by POWHEG authors)


perrozzi:
ciao roberto, can you remind me if/how we keep track of the svn version in the source file since we remove all the/svn directories? thanks

covarell:
ciao Luca, this is a simple copy of the original script for POWHEG V2 by @syuvivida .
I think that, if we want to keep svn information, there is no problem in including back the svn directories.

covarell:
P.S. Since the svn ID is unique to the full package (and not by process) would it make sense to name the source tarballs like powhegboxXX_<svnID>.tgz, rather than powhegboxXX_<month/year>.tgz ?
The month/year is not super-meaningful

syuvivida:
@covarell, I agree. I can make this change later today.

syuvivida:
@covarell oops, sorry I thought this was another script. I would not do any change here.

covarell:
hmmm... ok. So @syuvivida , how would you propose to keep track of svn version?

syuvivida:
I am trying to make a change in this shell script so that the SVN version is reflected in the name of the tar ball.
https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/source/create_source.sh

covarell:
ok ok, then I would just copy it in the other one (it just differs by the powhegbox svn source)

syuvivida:
Hello,
 here is the pull request I made. https://github.com/cms-sw/genproductions/pull/1130

perrozzi:
ok, I made a comment on the 1130. May I ask why you would not change anything here? It is a different script targeting a different version of POWHEG (namely powhg-res) so also here could be useful to add the revisio, or am I wrong?

covarell:
We DO want to. But I was waiting for the final update by Eiko. I will now push another commit on top

perrozzi:
@covarell can you please remove the exclusion of the .svn folders according to https://github.com/cms-sw/genproductions/pull/1130/files#diff-f6b5f5d9a29b3714a20b519e171ab17fL45
thanks

perrozzi:
(and according to also this https://github.com/cms-sw/genproductions/pull/1131)

covarell:
Done!

PR #960, opened 2016-09-29
Nnlops

lenzip:
this pull request should enable HJ+NNLOPS

PR #1130, opened 2017-03-14
include the SVN revision number in the name of powheg tar ball

syuvivida:
Instead of using any random name for the powheg tar ball, we include the SVN revision of POWHEG BOX V2 in the name.

### DIFF
--- a/bin/Powheg/source/create_source.sh
+++ b/bin/Powheg/source/create_source.sh
@@ -6,8 +6,8 @@ EXPECTED_ARGS=1
 
 if [ $# -ne $EXPECTED_ARGS ]
 then
-    echo "Usage: `basename $0` outputtarball_prefix"
-    echo "Example: `basename $0` powhegboxV2_Oct2015"
+    echo "Usage: `basename $0` workdir_prefix"
+    echo "Example: `basename $0` my"
     exit 1
 fi
 
@@ -16,8 +16,8 @@ echo "         Running Powheg  "$basename"                         "
 echo "   ______________________________________________________    "
 
 topdir=$PWD
-output=$1
-workdir=$topdir/temp_$output
+prefix=$1
+workdir=$topdir/temp_$prefix
 
 if [[ -e ${workdir} ]]; then
   fail_exit "The directory ${workdir} exists! Please clean up your work directory before running!!"
@@ -30,6 +30,10 @@ cd $workdir
 svn checkout --username anonymous --password anonymous svn://powhegbox.mib.infn.it/trunk/POWHEG-BOX-V2 POWHEG-BOX
 
 powhegdir=$workdir/POWHEG-BOX
+cd $powhegdir
+version=`svn info | grep -a "Revision" | awk '{print $2}'`
+cd -
+output=powhegboxV2_rev${version}
 
 ### Check out user process
 svn co --username anonymous --password anonymous svn://powhegbox.mib.infn.it/trunk/User-Processes-V2 
--- a/bin/Powheg/source/create_source.sh
+++ b/bin/Powheg/source/create_source.sh
@@ -32,8 +32,9 @@ svn checkout --username anonymous --password anonymous svn://powhegbox.mib.infn.
 powhegdir=$workdir/POWHEG-BOX
 cd $powhegdir
 version=`svn info | grep -a "Revision" | awk '{print $2}'`
+date=`date +%Y%m%d`
 cd -
-output=powhegboxV2_rev${version}
+output=powhegboxV2_rev${version}_date${date}
 
 ### Check out user process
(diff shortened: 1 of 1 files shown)


syuvivida:
@covarell , @yuanchao I've modified create_source.sh

perrozzi:
thanks. would it make sense to have both the svn revision and the month/date when the source is created?

syuvivida:
@perrozzi, date information is added. now the tarball name is 
powhegboxV2_rev${version}_date${date}

perrozzi:
thanks a lot. a last comment. If I look at this PR by @yuanchao  https://github.com/cms-sw/genproductions/pull/932/files I see that the svn version should be somehow used in the code. 
However while creating a gridpack recently I observed a few messages like
svn: '.' is not a working copy
and I suspect that without including the .svn directories the versioning will never work.
I personally think that the svn version is important both on the tarball and somewhere in the gridpack, so I kindly ask to clarify the latter is available or requires some fix.

perrozzi:
btw, following this PR from Giulio, as per request from Pietro in the discussion
https://github.com/cms-sw/genproductions/pull/960/files#diff-ecd8ac3748a3baeb09846ae599287a66R357
currently a file called VERSION is created in the gridpack, containing the name of the source tarball.
As far as the name of the tarball will contain the svn version too, the information will be propagated.
But in any case I want to clarify if the PR from Yuan Chao is effectively working or not.
Thanks

syuvivida:
@perrozzi ,
 I don't remember why I excluded .svn directory in the tar ball. I am double checking again. If the size of tar ball doesn't increase too much, I will not exclude ".svn" directory in the tar ball.

syuvivida:
@perrozzi 
  I checked. If I include only the SVN of POWHEG BOX directory, the size of tar ball is increased by 6%. If I include also the .svn directories of each physics process in User-Processes-V2, the size is increased by a factor of 2. If its OK, I could simply add back the .svn directory of POWHEG-BOX directory.

perrozzi:
Thanks @syuvivida. 
I think that there are a few important points here, let me stress them.
In principle the powheg svn version needs to be accessible from (at least either of)
1) the powheg source
2) the powheg gridpacks
3) the LHERunInfoProduct
and should be quoted in the papers (ATLAS started doing this long time ago, and I find it very important).
**I am afraid that currently this information is not stored/accessible anywhere**.
1) The source strips the .svn folders and doesn't provide reference to the version before this commit.
2) The gridpack only contains the name of the source (starting from end of september after
https://github.com/cms-sw/genproductions/pull/960) which however doesn't contain the svn version)
3) The gridpacks, in principle, should contain the svn info after this commit from last August https://github.com/cms-sw/genproductions/pull/932 as suggested in the documentation but in reality having stripped the .svn folders it only returns "no svn repository found" (cfr attached file
[powheg_LHERunInfoProduct.txt](https://github.com/cms-sw/genproductions/files/844109/powheg_LHERunInfoProduct.txt) ). Sorry if I sound harsh, but I had expected the the PR to be properly tested...
From the log, it can be seen that the svn revision is checked both in the main powheg version and in the process folder:

SVN status Begin
[...]/POWHEG-BOX
no svn repository found
[...]/POWHEG-BOX/Zj
no svn repository found
SVN status End

so I assume is important to check both.
If we strip the .svn folders in the process folders, this will not be possible.
Can we do a quick test to verify whether the main folder and process folder svn versions are different?

Or, simply, I would suggest to add back all the .svn folders.
There is a very remote but possible bad thing that can happen with the .svn folders, i.e. that a user types svn update before to use the repository. But we use scripts to create gridpacks, so I expect this probability to be negligible.
I don't think that the size of the source is an issue. It is currently O(100MB) I think we can afford twice as much, given the limited number of sources per year.

I hope things are clear and we can finally solve this important issue.
Let me know if you have comments
Luca

yuanchao:
It could be me that suggested to exclude the .svn directories as it actually much reduced the size by a factor ~2. However, we later added the corresponding patch to turned on the version prompt in PowHeg, we need this directory of "every" sub-processes for it. (I added it back in my local version of the source tarball script) With this built-in option (patch), pwg_main will print out the SVN version during run time. The VERSION file in gridpack is handy, but may be not necessary.

perrozzi:
@yuanchao can you please push to the central repository any modification used to build the source tarball?
It is important to keep track of the evolution of the scripts centrallyto avoid any source of confusion

syuvivida:
@yuanchao and @perrozzi 

  I just did a check. At least for the current download, the SVN version under POWHEG-BOX and the SVN version in every subprocess (within UserProcess) are all 3370, i.e. they are all the same.
This was not the case before. I checked some of my old powheg directories, in some cases, some subprocess can have an older revision number than the revision in POWHEG-BOX.

perrozzi:
thanks @syuvivida I think is ok to keep all the .svn directories then and have the 2 separate numbers in case they are different

PR #1131, opened 2017-03-15
add .svn directories of each subprocess

syuvivida:
@perrozzi ,
 sorry there was mis-understanding. My last commit in the previous pull only include .svn directory of the POWHEG-BOX but not those of the user processes. I was not able to make additional change before it was merged. Therefore I open this pull request again.


perrozzi:
my fault, thanks

PR #1132, opened 2017-03-15
added RPVSquarkTo4J cards

elliot-hughes:
I'm adding RPVSquarkTo4J cards for my analysis's signal samples in preparation for official production.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/RPVSquarkTo4J/RPVSquarkTo4J_Msq-100_Mneu-75/RPVSquarkTo4J_Msq-100_Mneu-75_customizecards.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/RPVSquarkTo4J/RPVSquarkTo4J_Msq-100_Mneu-75/RPVSquarkTo4J_Msq-100_Mneu-75_customizecards.dat
@@ -1,0 +1,3 @@
+set param_card msu3 100
+set param_card mneu3 75
+set param_card mneu4 75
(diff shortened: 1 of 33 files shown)


perrozzi:
is the model using 4F PDF or 5F PDF? (you are specifying  4 = maxjetflavor which means 4F)
can you remind me which PDF is selected when nn23lo1 is used?

elliot-hughes:
I believe `nn23lo1` corresponds to an internal MadGraph PDF called `NNPDF23_lo_as_0130_qed_mem0.grid`. If you want to look at it, you can find it in MadGraph5's `Template/LO/lib/Pdfdata` directory.

perrozzi:
and which PDF flavor scheme is the model supposed to use?

elliot-hughes:
As I understand it, `NNPDF23_lo_as_0130_qed_mem0.grid` is a "variable-flavor number" PDF, so it's not fixed to either 4-flavor or 5-flavor. And you're correct in pointing out that my run cards are setting a 4-flavor scheme. I'm not very knowledgeable about PDFs, but by my understanding I'm using an acceptable setup. Please enlighten me if I'm mistaken.

perrozzi:
I am not sure I agree with this statement, sorry...
you must investigate which flavor scheme is using the model you are importing RPVMSSM_UFO
(@bendavid @kdlong @vciulli etc can you please I am not sure which is the default pdf flavor scheme, can you comment?)

kdlong:
@elliot-hughes, is there any reason to prefer this PDF? It's best to use an up to date set like NNPDF 3.0 unless there's a good reason to use an older one.

perrozzi:
ping

kdlong:
As for the flavor scheme, this depends on the model. Is the model in MadGraph by default or are you loading it from somewhere?

perrozzi:
ping. what shall we do here?

elliot-hughes:
Hey, sorry about dropping out of communication. I've been delayed in sorting this out. Let's close this PR, and I'll open another after I make some revisions and test locally.

PR #1133, opened 2017-03-20
Powheg M60 and M80 gg_H_quark-mass-effects cards

gourangakole:
Cards are similar to M70 GeV. 
hfact is calculated using formula:
0.1*mH + 47.5


govoni:
@chayanit @benjaminkreis @covarell would you please check these cards as well?

benjaminkreis:
I can confirm that the only differences with respect to the [mass=125 Gev gridpack](https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/production/V2/13TeV/Higgs/gg_H_quark-mass-effects_NNPDF30_13TeV/gg_H_quark-mass-effects_NNPDF30_13TeV_M125.input) are hmass, hwidth, and hfact.

The hfact now follows the formula 0.1*hmass+47.5, which I believe is correct.

But there is something wrong with the widths.  They are both set to 1.81D-03.  I'm actually not sure where to get the M=60 GeV width.  M=80 GeV is included [here](https://arxiv.org/pdf/1307.1347.pdf#page=284).

chayanit:
There was a discussion on HN (https://hypernews.cern.ch/HyperNews/CMS/get/generators/3348/1.html) and I saw Roberto commented that the hwidth is not available for mH < 80 GeV in CERN YR so any number is fine. Maybe @covarell can confirm again here. Other than this I think the cards look fine.

gourangakole:
I have committed the hwidth for M=80 GeV as @benjaminkreis pointed the doc.

covarell:
Cards are ok for me to merge.

govoni:
thank you all for the cross-checks!

PR #1135, opened 2017-03-23
add 500 GeV mass points for BulkGraviton_ZZ_ZhadZhad_narrow and BulkG…

clelange:
…raviton_WW_inclu_narrow

These will be requested via B2G, gridpacks have been created and will be uploaded by B2G MC contacts.

PR #1137, opened 2017-03-24
add additional command line parameter to allow simple inclusion of ex…

clelange:
…ternal models for MadGraph5_aMCatNLO gridpack generation.

This is what I referred to in #1135. I actually just found that something in the line of that is already possible (but unfortunately not documented):
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/gridpack_generation.sh#L217-L240

The external models need to be available from https://cms-project-generators.web.cern.ch/cms-project-generators/ though - which is not the case for the model in #1135.

The changes I implemented simply allow to provide a directory containing the model as additional parameter (`$4`) after the choice of the batch queue. I agree that all models that are used in CMS should be stored centrally, so maybe the existing approach is better. However, mind that for basically all BulkGraviton samples `*_extramodels.dat` files need to be created, otherwise the gridpack generation won't work.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
+++ b/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
@@ -32,9 +32,12 @@ carddir=${2}
 # which queue
 queue=${3}
 
+# provide new model files
+modeldir=${4}
+
 if [ -z "$PRODHOME" ]; then
   PRODHOME=`pwd`
-fi 
+fi
 
 if [ ! -z ${CMSSW_BASE} ]; then
   echo "Error: This script must be run in a clean environment as it sets up CMSSW itself.  You already have a CMSSW environment set up for ${CMSSW_VERSION}."
@@ -75,6 +78,9 @@ echo "System release " `cat /etc/redhat-release` #And the system release
 echo "name: ${name}"
 echo "carddir: ${carddir}"
 echo "queue: ${queue}"
+if [ -n "${modeldir}" ]; then
+    echo "modeldir: ${modeldir}"
+fi
 
 cd $PRODHOME
 git status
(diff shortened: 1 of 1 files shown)


perrozzi:
Hi Clemens,
sorry I am not sure I understand what you are proposing (better, why).
The UFO support is documented here
https://twiki.cern.ch/twiki/bin/view/CMS/QuickGuideMadGraph5aMCatNLO#Preparation_of_the_cards
and yes, requires an additional cfg and models centrally uploaded to cmvfs.
This ensures reproducibility of the results. You can create the additional cfgs with a script.
Why would you like to pass a parameter with a local directory?

Thanks
Luca

clelange:
Hi Luca,

Sorry, I missed the UFO support part, it is actually documented. My addition to the script would allow testing a new model without uploading it to the generator web repository. It's often the case that the MC contacts ask analysts to create gridpacks, and they do not have the possibility of uploading files to the repository afaik. Currently, a large number of gridpack config in the repository just won't work without this option since the extramodels files and models themselves are missing. Anyway, I don't have a strong opinion on that, I just think it's a nice feature to have and if you like it, you can accept it or just close the PR.

Cheers,
Clemens

perrozzi:
@kdlong @vciulli (can't remember Qiang's username) can you please comment?

perrozzi:
@qliphy

bendavid:
Given how this is supposed to work, sounds to me like the correct thing to do would be to upload the missing models to the central repository, and update the problematic cards with the proper extramodels.dat

perrozzi:
any other comment? btw, is there any central mg5_amc@nlo model repository?

clelange:
I agree with Josh that it's of course better to make all the cards in the repository work, but this will be quite some work (could probably be automated though). The additions I implemented should be used for testing, not for official production, but I think they are also useful for people who want to perform a private production of a model that is not in the generator repository. 
As far as I know, there is no central mg5_amc@nlo model repository, you often only find models attached to the MG twiki or on the author's webpage. I guess a collection of some models is linked from https://cp3.irmp.ucl.ac.be/projects/madgraph/wiki/Models

perrozzi:
shall we close this PR?

kdlong:
I tend to agree that it should be closed. I'm sure centrally produced samples have been produced with this type of modification, but it would be better to enforce uploading models. Otherwise reproducibility is impossible.

PR #1140, opened 2017-03-28
Moving cards for WpTpb to the right directory

jsiado:
Moving the cards from genproduction to /geproduction/.../WpToTpBottom


perrozzi:
can you please (re)move https://github.com/cms-sw/genproductions/tree/master/WpToTpBottom ?

jsiado:
I tried to do it but I could not do it

perrozzi:
From inside the folder genproductions type:
git rm -r WpToTpBottom
then
git commit -m "Remove duplicated cfg"
git push master

does it work?

jsiado:
It looks like they have been removed. Do I need to create another pull request?

perrozzi:
perfect, now the PR shows the file as if they were moved from the old to the new destination.
that's what I meant, thanks a lot

PR #1144, opened 2017-03-30
Update for mar 2017 powheg tarball

yuanchao:
It looks like for the powheg source checked out in Mar 2017, some code needs gfortran 5.1 are introduced. The current patch by res_gfortran.patch needs some modifications. Also I update run_pwg.py for some clean up on the input cards handling as well as the gridpack driving script bug found by Gouranga Kole.


gourangakole:
Hello ..is this PR going to merge soon?
Ref to the problem was: https://hypernews.cern.ch/HyperNews/CMS/get/generators/3357/1/1/1.html

perrozzi:
@covarell @yuanchao has it been tested?

covarell:
Just for me to understand... The changes to python script and runcmsgrid seem OK, but what is the purpose of res_gfortran48.patch? Running POWHEG-RES with gfort 4.8? In the long run, we DO NOT want to do that, right?

perrozzi:
as far as I remember the ttbar process of resonance aware POWHEG version is native gcc5 (which requires CMSSW >=80x) but TOP wanted it to be produced with 2016 conditions (which require CMSSW 71x) only supporting gcc4.
A patch was provided by the POWHEG authors to roll back the code to gcc4 which then requres this patch. @intrepid42 can comment more

For the longer run, we asked computing to provide a CMSSW version with 2016 conditions and supporting gcc5. Very likely will be a CMSSW 80x version but this is not an issue as far as the MC campaign is updated). you can follow up this github issue https://github.com/cms-sw/cmssw/issues/18175

intrepid42:
Hi, yes, we need the res_gfortran48.patch only for running in CMSSW 71X.

But if this update requires gfortran 5.1, the pull request should better go to https://github.com/cms-sw/genproductions/tree/powheg_gcc530?

gourangakole:
Just let me comment:
I just took "runcmsgrid_powheg.sh" and "run_pwg.py".
used old "res_gfortran48.patch" to test.
I see gridpack is not converging.
See details: https://gkole.web.cern.ch/gkole/public/MC_Contact/Powheg/04_04_2017/

yuanchao:
Sorry for the late reply.

The "normal" version of powheg now needs gcc5 to be compiled. (tested with the source checked out in early Mar.) With some modification of the res_gfortran48 patch (included in this PR), I can compile the source w/o problem if applying the patch. I only successfully tested the DMGG and ggH processes, the rest will be finished by Yu-Hsiang. (not yet)

@gourangakole : the original res_gfortran48.patch won't work here.

gourangakole:
@yuanchao  I have tried with 3 latest files in this PR: 
"runcmsgrid_powheg.sh", "run_pwg.py" and "res_gfortran48.patch" to test.
But seems it does not work for me, Could you please have a look:
my details: https://gkole.web.cern.ch/gkole/public/MC_Contact/Powheg/04_04_2017/v5/

yuanchao:
Ok. Thank you @gourangakole. It seems some problem with the patched pwhg_io_interface.f. Let me further check what went wrong.

perrozzi:
@yuanchao the current CMSSW 71X doesn't support gcc5 so if you switch genproduction default branch to use POWHEG with gcc5, it will not be possible anymore to produce gridpacks that run under 71X...
For the time being we need to have 2 separate branches, one for gcc4 (master) and another for gcc5  https://github.com/cms-sw/genproductions/tree/powheg_gcc530
We are interacting with computing to solve this situation, but currently we need to live with that.
I hope Is clear now, let me know in case of questions

intrepid42:
And let me add that you obviously don't need `res_gfortran48.patch` for gcc5 because it should work without the patch then!

covarell:
It seems there is some confusion here.

This PR cannot be merged in master because it needs gcc >= 5.3.0 to work. In master, for the moment we keep the old version for CMSSW71X (gcc = 4.8.3).

So this needs to be split in two PRs.
- one must compare and PR to the powheg_gcc530 branch, NOT the master branch (everything related to updates of powheg to Mar2017 and removing patches for gcc 4.8.3)
- one with JUST the fixes needed for @gourangakole 's issues (I did not understand what the source is exactly...)

Can we close this PR and open two new ones with the requirements above? @yuanchao @perrozzi

perrozzi:
thanks Roberto, I am closing this PR

yuanchao:
It turned out that the problem comes from res_gfortran48.patch. There are several print out statements that exceeds the line length limit. Ex.
`
+c4.8.5               call pwhg_io_enlarge_buffer
+              print*,"unimplemented, send email to tomas.jezo@physik.uzh.ch"
`

I'll prepare another PR.

Not sure how this was triggered as I replaced the source tarball with a later checked out in late Mar. Just wondering why I didn't meet the problem and neither for the RES version case?

PR #1150, opened 2017-04-06
Fix powheg.input missing if using URL and runcmsgrid_powheg.sh error

yuanchao:
* fix a problem that powheg input card is missing if using a URL path.
* fix runcmsgrid error if process name contains '-'.


bin/Powheg/runcmsgrid_powheg.sh
```diff
@@ -373,8 +373,8 @@ if [ -s pwg-stat.dat ]; then
   tail -${tail} cmsgrid_final.lhe                           >  cmsgrid_final.lhe_tail
   head -${head} cmsgrid_final.lhe                           >  cmsgrid_final.lhe_F
   proclin=`expr $head + 1`
-  PROCESS=`sed -n -e ${proclin}p  cmsgrid_final.lhe |  awk '{print $4}'`
-  echo "  "$XSECTION"   "$XSECUNC"  1.00000000000E-00 "$PROCESS >>  cmsgrid_final.lhe_F
+  proc=`sed -n -e ${proclin}p  cmsgrid_final.lhe |  awk '{print $4}'`
```
perrozzi: hi, for my curiosity: why $PROCESS would not work while $proc would instead?

yuanchao: @perrozzi : If you look at the beginning of the script, there's a tag with the same name. It will be replaced with the powheg process name by run_pwg.py. As "ggH_quark-mass-effect" has '-' in the process name, it will be a problem. But actually, this variable will be an empty string which does nothing after all, this won't affect the result even if not fixed.

Just that, I'm not sure if this evaluation yields an empty string is correct or not.
PROCESS=`sed -n -e ${proclin}p  cmsgrid_final.lhe |  awk '{print $4}'`

perrozzi: ok, so I think I uderstand (?) more or less what could be happening, and should be related to this PR
https://github.com/cms-sw/genproductions/pull/1064 that was done to fix an hard-coded value for the process number (?) https://github.com/yuanchao/genproductions/commit/835e5c0ce66828e7e4803f323ee6c3d53fee809b#diff-8cf10f8734b1bb4d2bcecd3878fdbc28L389
causing null xsec in genXsecAnalyzer. Thinking again about it, I am surprised that genXsecAnalyzer uses information put by hand in the lhe file.

So, as you say when the word "PROCESS" was used to name the variable, it was overwritten with the process name. I am not sure why when this variable name contains an underscore it should complain.

May I ask to do a check:
- produce a lhe file with this fix using ggH_quark-mass-effect, and post it somewhere for us to inspect
- convert the lhe file to edm 
- use GenXsecAnalyzer to check that a non-null value is returned

Thanks a lot
Luca

yuanchao: Hi @perrozzi , #1064 is just a synchronization copy from runcms_powheg.sh to runcmsgrid_powhegjhugen.sh and runcmsgrid_powheg_hdampgrid.sh. Maybe I shouldn't do the modification at all?

The problem is caused by the dash '-' not underscore '_'. For most cases, there are no dashes in the process name.

yuanchao: @perrozzi please point us some usage guide for GenXsecAnalyzer, I can let Yu-hsiang do some tests here.

perrozzi: thanks @yuanchao now it makes sense, is the "-" in the variable name, which is not supported by bash.
to run the GenXsecAnalyzer you can use 
https://github.com/cms-sw/genproductions/blob/master/test/calculateXSectionAndFilterEfficiency/genXsec_cfg.py
and run it with the command
cmsRun genXsec_cfg.py inputFiles="file.root" maxEvents=1000
further instructions here
https://twiki.cern.ch/twiki/bin/viewauth/CMS/HowToGenXSecAnalyzer

perrozzi:
ping

yuanchao:
Thanks a lot for the reminder. I asked Yu-hsiang to do this test and I need to ping him again.

yuanchao:
During the test, Yu-Hsiang found a problem. There seems to be a limit on the numbers of PDF weights to be processed. The runcmsgrid.sh script will stop during the processing of MMHT2014nlo68clas118. There is a patch on this to increase 'lhrwgt_maxnheader' from 200 to 400 but seems not effected. I'm changing the 'maxids' too and see if this can be fixed.

perrozzi:
ok let us know

yuchanggit:
Hi all,
The validation is done.

I can get the GenXsecAnalyzer message in GEN-SIM production after the cmsRun running is finished. And the cross section value is identical to the value shown in POWHEG production (pwg-stat.dat) 

I can also give more detail if you wish.

perrozzi:
I think we can merge the PR now. @yuanchao @covarell ok with you?

covarell:
Ok for me, remember that it needs to be forward-ported to branch powheg_gcc530

perrozzi:
@yuanchao please port to powheg_gcc530

yuanchao:
@perrozzi : Ok, sure.

PR #1156, opened 2017-04-23
Update aqgc

kdlong:


kdlong:
Updating aQGC points for WZjj EWK sample to scan values with sensitivity for current dataset (some mistakes were made in estimating the parameters for the original samples).

PR #1152, opened 2017-04-12
Wp to tp bottom bp top t hb

knash:
adding MG cards for W' -> Tb / tB -> tHb production.

PR #1158, opened 2017-04-26
Wp to tp bottom bp top t hb

knash:
Same as this PR
https://github.com/cms-sw/genproductions/pull/1152#pullrequestreview-32354666

But with LH signals and wider T' to coordinate with 
https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/WpToTpBottom

Also I dropped the 4TeV point for higher statistics on the more realistic mass points


perrozzi:
ok to merge?

knash:
Actually...wait until early next week.  There are some discussions ongoing

knash:
Ok, this should be final

PR #1159, opened 2017-04-27
Charged resonance missing cfgs

shchauha:


govoni:
for what production are those cfg files intended?

shchauha:
These are for Charged resonance Pythia samples to be produced with Moriond setting. Analysers had requested Madgraph and Pythia samples earlier. Madgraph ones are already produced. Pythia configs had some issues that time, they are fixed now by analysers.

govoni:
ok. Please remember that for 2017 productions the tune will have to be changed in each of the files.

shchauha:
Okay Thanks Pietro.

PR #1161, opened 2017-05-03
Update parsing.py

efeyazgan:
This is to add a warning for the parton shower to be specified in run card for NLO configurations.


perrozzi:
@lorusso7 please note this add-on and let us know if you have comments

PR #1163, opened 2017-05-09
Settings for TTWW 5FS LO

jfernan2:


perrozzi:
hi, why NLO PDF?

jfernan2:
There's no LO 5FS pdf, right?

perrozzi:
generally speaking, when no flavor is specified, it's 5F. 
You can see that 263000 doesn't specify 5F and the other close-by PDFs specify 3F, 4F, 6F
263000	NNPDF30_lo_as_0130	101	2	
263200	NNPDF30_lo_as_0130_nf_3	101	2	
263400	NNPDF30_lo_as_0130_nf_4	101	2	
263600	NNPDF30_lo_as_0130_nf_6	101	2
so you can use 263000

jfernan2:
OK. I believe you are right. I have updated the settings

jfernan2:
Sorry, I screwed up the PR, I'll close and create a new one

PR #1167, opened 2017-05-10
single top t-channel Vtd=Vts=0.5 production cards

ckmackay:
single top t-channel Vtd=Vts=0.5 production cards


perrozzi:
naive question: where did you set Vtd=Vts=0.5 ?

ckmackay:
Hi Luca,

As this is for production they are actually set in the powheg cards, which are noted in the customize_card.dat, so in fact the other files are no different from the standard ones (I hadn't thought about that at the time). So the only one that is different is customize_cards as it gives the addresses of the gridpack and the cards used to make teh gridpack with Vtd=Vts=0.5. Hope that makes sense.

Kate

PR #1172, opened 2017-05-17
adding Z aTGC VBF cards

senka:


perrozzi:
does this gridpack implies compiling code at execution time? (@kenneth)

kdlong:
Ah yes I think so. So we need to address this. Though I believe Senka is using MG2.4, was the accidental use of all cores only an issue in 2.5?

senka:
I am using MG 2.4.2

perrozzi:
@kdlong my only worry is not to abuse this feature due to recent happening, in case this request is not very different from other succesfully produced, I would say please go ahead (but try to address this with the mg5_amcnlo authors asap in parallel)

kdlong:
@perrozzi I was wrong previously, these are LO samples so the process should be fully precompiled. The .f files are removed before the gridpack is produced so it's completely impossible for them to compile on the cluster.

With that said, it looks to me like something went from for the WToLNuJJ sample and the aQGC weights weren't produced.

perrozzi:
@senka I see a new commit, can you please comment?

senka:
I also added W VBF cards. So now the request includes Z VBF and W VBF. The setup is the same for both samples.

perrozzi:
@kdlong but the weights don't work yet? can the pr be merged or needs further updates?

senka:
i fixed the problem with weights for W VBF samples and redid gridpacks. From my side this is fixed now. Kenneth can confirm from his side.

perrozzi:
@kdlong ping

kdlong:
Looks good to me. Both of these gridpacks will have the weights precompiled.

PR #1179, opened 2017-05-30
Fix bbH compilation and enable NNLOPS reweighting a posteriori

covarell:
Several changes/fixes @perrozzi @yuanchao @lenzip 
- bbH is now compiled with -O0 (neverending with -O2)
- NNLOPS reweighting to be done a posteriori: runcmsgrid does NOT run nnlops but just sends a warning. 
- Two new scripts: run_lhe_parallel.sh just sends many jobs e.g. for HJ with standard naming; run_nnlops.sh merges these jobs and runs the reweighting (needs complicated LHE manipulation because of stupid limit on total weights in POWHEG)


perrozzi:
at a first look seems good to me

bendavid:
Is there any prospect for solving the NNLOPS reweighting issue in a nicer way, e.g. by precomputing the grids for the weights as part of the gridpack generation, or pre-generating some events for this purpose?

perrozzi:
authors said "not in the short term" unfortunately...

PR #1180, opened 2017-05-30
mg25x

pmandrik:
Switch MG_amc@nlo gridpack generation from v2.5.1 to v2.5.4 and update related patches.


bin/MadGraph5_aMCatNLO/gridpack_generation.sh
```diff
@@ -119,8 +119,8 @@ if [ ! -d ${AFS_GEN_FOLDER}/${name}_gridpack ]; then
 #   export SCRAM_ARCH=slc6_amd64_gcc472 #Here one should select the correct architechture corresponding with the CMSSW release
 #   export RELEASE=CMSSW_5_3_32_patch3
 
-  export SCRAM_ARCH=slc6_amd64_gcc481
-  export RELEASE=CMSSW_7_1_23
+  export SCRAM_ARCH=slc6_amd64_gcc530
```
perrozzi: ah, so is intended that we switch to gcc5 also for MG5_aMC@NLO. This is an even stronger case to urgently discuss with computing how to handle the production of MC events for 2016 data analysis in the future...

bin/MadGraph5_aMCatNLO/gridpack_generation.sh
```diff
@@ -119,8 +119,8 @@ if [ ! -d ${AFS_GEN_FOLDER}/${name}_gridpack ]; then
...
+  export RELEASE=CMSSW_9_0_2
```
perrozzi: is there a specific reason why 902 is used and not, for instance 920?

perrozzi:
things look ok to me, let's wait for some other comment

bendavid:
Is there a particular reason to change the default release and scram_arch?

pmandrik:
Dear all, there is no specific reason of release and scram_arch changes.
It was made in order to try to bypass the bug in 2.5.1-2.5.2, didn't help and left in the code.
Are there any reasons to prefer previous versions?

bendavid:
Yes, 71x and the corresponding scram_arch is strongly preferred to maintain compatibility with existing campaigns.

perrozzi:
I assume is ok to merge now

PR #1194, opened 2017-06-15
Update condor management to work also on lxplus

kdlong:
Small updates to the condor interface to be more robust across different condor versions. Tested and working on both new (installed on lxplus) and old versions of condor.

@khurtado, @lorusso7, would it be straightforward to test this version with CMS Connect?

@huohuo21, would you be able to run some test gridpacks at Fermilab or another Tier 2? The lxplus afs issues are not yet solved.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/patches/0005-add-rsync-based-submit2-functionality-for-lsf-plus-v.patch
+++ b/bin/MadGraph5_aMCatNLO/patches/0005-add-rsync-based-submit2-functionality-for-lsf-plus-v.patch
@@ -1,7 +1,7 @@
 From 556b9a5288351fb2730cc9b71ea69a583201c4b7 Mon Sep 17 00:00:00 2001
 From: Josh Bendavid <Josh.Bendavid@cern.ch>
 Date: Mon, 27 Jun 2016 19:10:11 +0200
-Subject: [PATCH 5/8] add rsync-based submit2 functionality for lsf, plus various lsf robustness fixes and minor condor fixes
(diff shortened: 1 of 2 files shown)


khurtado:
@kdlong : We could test, here is a couple of points to consider though:

- I remember `mg25x` had issues with the `set cluster_type condor` where even if declared it ran everything on the submission node in multicore mode on CMS Connect. This was related to the version of Madgraph `mg25x` uses, which is why I moved to the `master` branch. Was that also a problem on `lxplus` machines at some point? If not, that would be the first thing to re-test.
- Assuming the above works, I can make a pull request with the CMS Connect changes I added to the `master` branch and either @lorusso7 or me could test ( I would actually need to add an [extra patch](https://github.com/khurtado/genproductions/commit/53253fe5c9eef0cf5fac3fde4e81335c788cd009) in @lorusso7 has been testing that fixes an issue with CMS Sites using Singularity )

Does that sound reasonable?

kdlong:
You're right. I realized this problem was still there shortly after making this pull request. It's a bug in MadGraph itself, though I think a work around should be trivial. I'll hopefully have an update shortly.

menglu21:
@kdlong sure, i will do that

menglu21:
@kdlong , submit_condor_gridpack_generation.sh works fine at Fermilab

kdlong:
Thanks @menglu21! Though I suspect it was running in multicore mode and not on the cluster, because of the problem @khurtado mentioned (if you look at the log file and see that the exact same number of jobs were always running at the same time, it would give you a strong indication that this was the case).

I'll push a quick fix to this problem shortly and we can try again.

PR #1196, opened 2017-06-15
Work around for run_mode not getting properly set

kdlong:
set run_mode 1 is apparently broken, manually add it to the run card.

@khurtado @lorusso7 @huohuo21 can you see if it works for you now?


khurtado:
@kdlong I can confirm the workaround works on CMS Connect.

perrozzi:
I assume can be merged now, let us know if there is anything else to be tested before to do it

PR #1202, opened 2017-06-22
add scram_arch and cmssw versions as parameters

perrozzi:
ongoing work, PR to let people start commenting in case


perrozzi:
@bendavid what do you think?

bendavid:
All this does is makes the release more easily configurable at gridpack generation time.  We rather wanted to modify the runtime scripts I thought.

perrozzi:
indeed, getting there soon

perrozzi:
this should be it https://github.com/cms-sw/genproductions/pull/1202/commits/e6cf46eaa1226d37965f3013c1cdab12cd29f991
unless we change also https://github.com/cms-sw/cmssw/blob/CMSSW_7_1_X/GeneratorInterface/LHEInterface/data/run_generic_tarball_cvmfs.sh 
right?

perrozzi:
(need to source from cvmfs, of course...)

perrozzi:
updated and in synch with https://github.com/cms-sw/genproductions/pull/1206

kdlong:
One thing I don't understand, why is it necessary to specify CMSSW version and SCRAM_ARCH? Why wouldn't you always just take the scram_arch from the CMSSW version?

perrozzi:
others might comment better than me, but afaik to avoid any problem you should always set SCRAM_ARCH first and then source CMSSW

bendavid:
Also a given release may exist in different scram_archs, and one might explicitly like to override this in the future (eg if older releases are rebuilt/repackaged for a newer Linux distribution)

PR #1206, opened 2017-06-27
add scram_arch and cmssw versions as parameters for mg25x

perrozzi:


perrozzi:
this PR should enforce a consistent scram_arch and cmssw release during the gridpack creation and in the worker node producing LHE events.
scram_arch and cmssw release are optional arguments, the default values are slc6_amd64_gcc481 and CMSSW_7_1_28

@bendavid what do you think?

PR #1210, opened 2017-06-28
bb4l: use relative lib paths, ship all needed libs

intrepid42:
* Use relative lib paths, otherwise the gridpack works only for the creator...
* Ship openloops and other libs (modified by Powheg authors -> errors when falling back to libs in cmssw)


perrozzi:
this has been tested, right?

intrepid42:
yes :-)

PR #1208, opened 2017-06-28
Additional Bulk and WPrime cards

dmorang:
adding new mass points for low mass ZV analysis

PR #1215, opened 2017-07-03
Revert "Additional Bulk and WPrime cards"

govoni:
Reverts cms-sw/genproductions#1208


govoni:
could you please remove duplications and useless definitions from the cards?

perrozzi:
more than revert the PR, would it be better to ping the author of the original PR to modify the existing files?

PR #1216, opened 2017-07-03
Cards and hadronizer fragments for bbH production in madgraph5_amc@nlo

ajgilbert:
Just to note that the gridpacks are currently not produced by the standard gridpack_generation.sh script. The instructions we were given from the theory contacts in the LHC Higgs XSWG are here:

https://twiki.cern.ch/twiki/bin/view/CMS/HiggsExoticsTheory#MSSM_SusHi_ggF_bbh_FeynHiggs_mas

which involves downloading a pre-prepared process directory. I confirmed that the process directory that can be generated from the proc_card.dat is not identical to the contents of the pre-prepared directory.

I included the cards in this PR anyway for reference, but instead the gridpacks have been prepared manually, in such a way as to have the same structure as expected by the runcmsgrid.sh script.


perrozzi:
thanks for the details

PR #1221, opened 2017-07-04
Make MG  v2.5.5 default for 25x branch

kdlong:
All patches work in the current form. Tested a simple NLO gridpack without problems. Did not test any more complicated features. @bendavid did you have anything else in mind?


bendavid:
Not sure we want to have 0j FXFX cards in the examples (since these actually won't produce correct physics results)

kdlong:
Oops right. Fixed the name and matching. I added it because it's nice to have something quick to run when testing things but can remove it if others don't find it useful.

bendavid:
Ok, that was supposed to be the purpose of wplustest_5f_NLO and wplustest_4f_NLO in fact.  (These are even faster to run because they are 2->1.

Otherwise looks fine to me.

perrozzi:
@kdlong @vciulli @qliphy please confirm when ready to merge

perrozzi:
what's the status here?

kdlong:
From my perspective it's ready, but I can remove the DY cards if others prefer them to not be there.

bendavid:
I think it's better to remove the drell yan cards as it may confuse people.  Or at least remove 0j from the name to be consistent with the usual naming scheme for non-merged samples.

kdlong:
Replaced by #1231

PR #1224, opened 2017-07-06
corret bugs releated to -z

qliphy:
passing right contents to scram_arch and cmssw_version


perrozzi:
thanks for spotting this! needs to be propagated ~everywhere, I will prepare another PR

PR #1232, opened 2017-07-11
NNLOPS in new reweighting scheme

covarell:
NNLOPS for powheg in new reweighting scheme.

Tested OK, but still needs editing resulting LHE files with scripts if the main MINLO-HJ run has a lot of weights (a lot means > 10...)

### DIFF
--- a/bin/Powheg/make_rwl.py
+++ b/bin/Powheg/make_rwl.py
@@ -11,7 +11,7 @@ import os
 m_outfile = 'pwg-rwl.dat'
 
 m_factor = ['1d0', '2d0', '0.5d0']
-m_idx = 1000
+m_idx = 1001
 
 fout = open(m_outfile, 'w')
 
@@ -26,7 +26,7 @@ for m_rensc in m_factor :
 		  
 fout.write("</weightgroup>\n")
 
-m_idx = 2000
+m_idx = 2001
 
 fout.write("<weightgroup name='PDF_variation1' combine='gaussian' >\n")
 
@@ -38,12 +38,12 @@ for idx in range(260001, 260101) :
   
 # computing weights for NNPDF 3.0 nlo alphas=0.117 variation
 # computing weights for NNPDF 3.0 nlo alphas=0.119 variation
-fout.write("<weight id='2101'> lhapdf=265000 </weight>\n")
-fout.write("<weight id='2102'> lhapdf=266000 </weight>\n")
+fout.write("<weight id='2102'> lhapdf=265000 </weight>\n")
+fout.write("<weight id='2103'> lhapdf=266000 </weight>\n")
 
 fout.write("</weightgroup>\n")
 
-m_idx = 3000
+m_idx = 3001
 
 # computing weights for 56+1 CT14 nlo PDF variations
 fout.write("<weightgroup name='PDF_variation2' combine='hessian' >\n")
@@ -54,13 +54,13 @@ for idx in range(13100, 13157) :
 
 # computing weights for CT14 nlo alphas=0.117 variation
 # computing weights for CT14 nlo alphas=0.119 variation
-fout.write("<weight id='3057'> lhapdf=13164 </weight>\n")
-fout.write("<weight id='3058'> lhapdf=13166 </weight>\n")
+fout.write("<weight id='3058'> lhapdf=13164 </weight>\n")
+fout.write("<weight id='3059'> lhapdf=13166 </weight>\n")
 
 # computing weights for CT10 nlo central values
-fout.write("<weight id='3059'> lhapdf=11000 </weight>\n")
+fout.write("<weight id='3060'> lhapdf=11000 </weight>\n")
 
-m_idx = 4000
+m_idx = 4001
 
 # computing weights for 50+1 MMHT2014nlo68clas118 PDF variations
 
@@ -74,7 +74,7 @@ for idx in range(25260, 25265) :
   fout.write("<weight id='"+str(m_idx)+"'> lhapdf="+str(idx)+" </weight>\n")
   m_idx = m_idx + 1
 
-m_idx = 5000
+m_idx = 5001
 
 # computing weights for 56+1 CT14 nnlo PDF variations
 
--- a/bin/Powheg/run_pwg.py
+++ b/bin/Powheg/run_pwg.py
@@ -600,33 +600,33 @@ nnlofiles
 
 <initrwgt>
 <weightgroup name='nnl'> 
-<weight id='nnlops-11-1'> combines 'nn-mtmb-11' with '1000' </weight> 
-<weight id='nnlops-11-2'> combines 'nn-mtmb-11' with '1001' </weight> 
-<weight id='nnlops-11-3'> combines 'nn-mtmb-11' with '1002' </weight> 
-<weight id='nnlops-11-4'> combines 'nn-mtmb-11' with '1003' </weight> 
-<weight id='nnlops-11-5'> combines 'nn-mtmb-11' with '1004' </weight> 
-<weight id='nnlops-11-6'> combines 'nn-mtmb-11' with '1005' </weight> 
-<weight id='nnlops-11-7'> combines 'nn-mtmb-11' with '1006' </weight> 
-<weight id='nnlops-11-8'> combines 'nn-mtmb-11' with '1007' </weight> 
-<weight id='nnlops-11-9'> combines 'nn-mtmb-11' with '1008' </weight> 
-<weight id='nnlops-22-1'> combines 'nn-mtmb-22' with '1000' </weight>          
-<weight id='nnlops-22-2'> combines 'nn-mtmb-22' with '1001' </weight>         
-<weight id='nnlops-22-3'> combines 'nn-mtmb-22' with '1002' </weight>         
-<weight id='nnlops-22-4'> combines 'nn-mtmb-22' with '1003' </weight>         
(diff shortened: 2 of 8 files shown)


govoni:
ciao Roberto, what does it mean that editing is still needed?

covarell:
It means the merging of the resulting LHE files is done using a trick, so that NNLOPS is blind to the hundreds of PDF weights and just sees the scale variations needed to run (you can have a look at the run_nnlops.sh script). This is what I also used to produce current NNLOPS samples.

govoni:
is this preventing an automated running of the NNLOPS with the usual gridpack scheme in a wmLHE campaign?

covarell:
Obviously. I believe this is a well established fact by now (regardless of the POWHEG version). 

NNLOPS reweighting is now understood to work only on a minlo LHE file containing a few 100k's of events and it is not possible to instruct wmLHE to produce so many events in one shot.

govoni:
:D :D

bin/Powheg/run_pwg.py
```diff
@@ -595,33 +600,33 @@ def runGetSource(parstage, xgrid, folderName, powInputName, process, tagName) :
...
-<weight id='nnlops-0505-1'> combines 'nn-mtmb-0505' with '1001' </weight>
-<weight id='nnlops-0505-2'> combines 'nn-mtmb-0505' with '1002' </weight>
-<weight id='nnlops-0505-3'> combines 'nn-mtmb-0505' with '1003' </weight>
-<weight id='nnlops-0505-4'> combines 'nn-mtmb-0505' with '1004' </weight>
-<weight id='nnlops-0505-5'> combines 'nn-mtmb-0505' with '1005' </weight>
-<weight id='nnlops-0505-6'> combines 'nn-mtmb-0505' with '1006' </weight>
-<weight id='nnlops-0505-7'> combines 'nn-mtmb-0505' with '1007' </weight>
-<weight id='nnlops-0505-8'> combines 'nn-mtmb-0505' with '1008' </weight>
-<weight id='nnlops-0505-9'> combines 'nn-mtmb-0505' with '1009' </weight>
+<weight id='nnlops-11-1'> combines 'nn-mtmb-11' with '1000' </weight>
```
perrozzi: may I ask what is the reason to shift this?

covarell: No idea. It is like this in the make-pwg-wl.py script and I followed the new convention.
@yuanchao  ?

yuanchao: I don't understand either. Also that I'd prefer all weighting stuff should go to make-pwg-wl.py.

bin/Powheg/run_pwg.py
```diff
@@ -338,7 +338,7 @@ def runGetSource(parstage, xgrid, folderName, powInputName, process, tagName) :
 fi
 
 ### retrieve the powheg source tar ball
-export POWHEGSRC=powhegboxV2rwgt_Mar2017.tar.gz
+export POWHEGSRC=powhegboxV2_rev3407_date20170711.tar.gz
```
yuanchao: The powhegbox source package naming would need some discussion? Here 'date' is a bit redundant. Surely 'rwgt' is not needed anymore.

bin/Powheg/run_pwg.py
covarell: @yuanchao 
In this script (written by you)
https://github.com/cms-sw/genproductions/blob/powheg_gcc530/bin/Powheg/make_rwl.py
the weights start all from a round number (e.g. 1000, 2000 etc.) because the counter is increased at the end of the loop. While in the past they used to start from a round number + 1. Did you mean to make this change? Otherwise, please modify the script and I will eliminate the shift from NNLOPS weights.

bin/Powheg/run_pwg.py
covarell: rwgt is in fact removed. 
The new naming is given automatically by the create_source script (of which I am also not an author)

covarell:
I have shifted all weights by 1 in the make_rwl script in the last commit since it seems we all agree on starting from 1001, 2001 etc.

Is there a final agreement already on the PDF sets for 2017 NLO and NNLO samples? If so, we can make standard script for the general use.

bin/Powheg/run_pwg.py
```diff
@@ -679,10 +684,6 @@ def runGetSource(parstage, xgrid, folderName, powInputName, process, tagName) :
   sed -e 's/PROCESS/'${process}'/g' ${WORKDIR}/runcmsgrid_powheg.sh > runcmsgrid.sh
 fi
 
-
-sed -i s/SCRAM_ARCH_VERSION_REPLACE/${SCRAM_ARCH}/g runcmsgrid.sh
```
perrozzi: could it be that you started modifying the script before updating to the central repository?
I recently introduced these lines to ensure the back-/forward- compatibility of the gridpacks in production,
unless there is a specific reason to remove them they need to be there :-)

bin/Powheg/run_pwg.py
```diff
@@ -816,14 +817,10 @@ def createTarBall(parstage, folderName, prcName, keepTop, seed, scriptName) :
 
 sed -i 's/pwggrid.dat ]]/pwggrid.dat ]] || [ -e ${WORKDIR}\/pwggrid-0001.dat ]/g' runcmsgrid.sh
 
-if [ "$process" = "HJ" ]; then
-  cat runcmsgrid.sh  | gawk '/produceWeightsNNLO/{gsub(/false/, \"true\")};{print}' > runcmsgrid_tmp.sh
-  mv runcmsgrid_tmp.sh runcmsgrid.sh
-fi  
-
-
-sed -i s/SCRAM_ARCH_VERSION_REPLACE/${SCRAM_ARCH}/g runcmsgrid.sh
```
perrozzi: same here

bin/Powheg/run_pwg.py
```diff
@@ -854,12 +851,11 @@ def createTarBall(parstage, folderName, prcName, keepTop, seed, scriptName) :
 
 if [ $keepTop == '1' ]; then
     echo 'Keeping validation plots.'
-    echo 'Packing...' ${WORKDIR}'/'${folderName}'_'${SCRAM_ARCH}'_'${CMSSW_VERSION}'_'${process}'.tgz'
-    tar zcf ${WORKDIR}'/'${folderName}'_'${SCRAM_ARCH}'_'${CMSSW_VERSION}'_'${process}'.tgz' * --exclude=POWHEG-BOX --exclude=powhegbox*.tar.gz --exclude=*.lhe --exclude=run_*.sh --exclude=*.log --exclude=*temp --exclude=pwgbtlupb-*.dat --exclude=pwgrmupb-*.dat
```
perrozzi: same here

bin/Powheg/run_pwg.py
```diff
@@ -854,12 +851,11 @@ def createTarBall(parstage, folderName, prcName, keepTop, seed, scriptName) :
...
+    echo 'Packing...' ${WORKDIR}'/'${folderName}'_'${process}'.tgz'
+    tar zcf ${WORKDIR}'/'${folderName}'_'${process}'.tgz' * --exclude=POWHEG-BOX --exclude=powhegbox*.tar.gz --exclude=*.lhe --exclude=run_*.sh --exclude=*.log --exclude=*temp --exclude=pwgbtlupb-*.dat --exclude=pwgrmupb-*.dat
 else
-
-    echo 'Packing...' ${WORKDIR}'/'${folderName}'_'${SCRAM_ARCH}'_'${CMSSW_VERSION}'_'${process}'.tgz'
```
perrozzi: same here

govoni:
@covarell the agreement about pdfs is here:
https://docs.google.com/spreadsheets/d/1YjhruLPvmwqoA1YZPgmU5H9T3ELnYnriHgO9DrJQoxk/edit#gid=0

covarell:
Ok, fixed wrong backports (@perrozzi, one of the comments, the one about setting produceWeightsNNLO should not be changed, otherwise weights are calculated on the fly) and updated pdf sets

govoni:
thanks. Are the weights ok for any powheg production already?

bin/Powheg/make_rwl.py
```diff
@@ -20,47 +20,51 @@
 
 for m_rensc in m_factor :
   for m_facsc in m_factor :
-    fout.write("<weight id='"+str(m_idx)+"'> lhapdf=260000 renscfact="+ \
+    fout.write("<weight id='"+str(m_idx)+"'> lhapdf=305800 renscfact="+ \
```
perrozzi: I believe we want to go to NNLO for the central value, there is a mistake in the google spreadsheet, therefore 306000 and similar

covarell: What do you mean? NNLO for NLO samples???

govoni:
@covarell yes, after a long discussion during GEN meetings and pdf4lhc forum meetings.

bin/Powheg/make_rwl.py
perrozzi: and LO samples too :-)

govoni:
@perrozzi can we merge?

perrozzi:
the main purpose of the PR is to update stuff for NNLOPS, which is done. PDF are not fully updated yet but can be done in a future PR.
so yes I think we can merge

govoni:
ok. Roberto, we assume there will be a dedicated PR for the PDFs, then.
It looks they deserve one :)

PR #1236, opened 2017-07-14
diphoton+2bjets

gourangakole:


perrozzi:
@pmillet can you please crosscheck?

pmillet:
some comments:
- check wether process in L21-L25 is commented on purpose or by accident
- does not specify any variations (is this wanted?)
- the WRITE_MAPPING_FILE 3; was removed was this on purpose?
can't comment on the values of the cuts etc though ...

michelif:
Hi Philipp,
thanks. I can reply to some points:
= L21-25 commented on purpose to save time, we already have samples with just two photons and no jets generated with sherpa.
= what is variations?pdf and scale variations? not an expert on how it works.
= WRITE_MAPPING_FILE 3 was not in the datacard of gg + jets from sherpa authors with the fix to avoid the crash for this process. I tested the workfolow with sherpa interface with this datacard and it works.

pmillet:
ok sounds good to me

if you want pdf/scale variations you can have a look at [1] 

[1] https://twiki.cern.ch/twiki/bin/view/CMS/SherpaInterfaceAdditionalInfo

perrozzi:
@michelif if you are not interested in PDF and scale systematics (check with Pasquale and Mauro) we can merge

michelif:
Ciao,
no we are not interested in PDF and scale systematics, we will  use this MC for background studies, not limit extraction, if everything ok for me it can be merged.

Issue #1212, opened 2017-06-29
[mg] best way to distribute the condor python bindings for mg5_amc@nlo


khurtado:
Hello,
During @lorusso `tt012j_5f_ckm_NLO_FXFX` test, I have noticed most madgraph jobs in that test run ran for less than 8 hours, with a very few going up to 24 to 35 hours.

This can be problematic while submitting to the Global Pool, because jobs are matched to currently running pilot jobs (pilots live for up to 48 hours and multiple real jobs run inside them) in the pool and you need to tell in advance how long a job is going to last at most (or it can be evicted when the pilot it is running at dies otherwise). The default in CMS Connect for this is 8 hours, which means a job can match to pilots that have been alive between 0 to 40 hours. Increasing the number to e.g 47 hours would mean a job can only match to very new pilots, which means matching to less resources. For the `tt012j_5f_ckm_NLO_FXFX`  run tests, one of the jobs ran 5 times on 5 different machines (first 4 got evicted after 8 to 24 hours) before completion. which translated into +120 hours wall clock time for that job since it started.

To fix this, I am thinking on implementing a list of “promised times” for a job. This is, something like: MaxWallTimeMins = [ 8*60, 24*60, 45*60]. Which means, if a job gets evicted after 8 hours, try 24 hours next and 45 next. The numbers in the list could be changed by the user and have the above by default.

Do you think that is reasonable? Otherwise, is there a way to tell if some particular jobs are going to take way longer than the others in madgraph so that we can set bigger times for those in advance?

kdlong:
Hey @khurtado, as you know the issue with widely distributed run times is one we are basically stuck dealing with. I think the solution you propose is a quite a good one. We could also think about whether such an approach is possible outside of CMS connect. By the way, for faster responses, feel free to tag me, Qiang, and Vitaliano in your posts.

vciulli:
I agree with @kdlong the proposed solution looks nice
Unfortunately I don't think there is any way to know in advance how long a specific job will take

khurtado:
Thanks! Sounds good. I will implement this then.

perrozzi:
any news on this?

khurtado:
@perrozzi : I'm almost done with this, I ended up replacing `condor_q` external commands with condor python API schedd queries in madgraph first so that the code is more robust for this sort of thing. Will most likely make a PR with this plus other features by tomorrow.

kdlong:
Adding this comment here since it might be the better location:

[...]

PR #1237, opened 2017-07-14
Changes for additional CMS Connect and condor compatibility

khurtado:
These are bug fixes and features added based on @lorusso7 run tests.

Bug fixes:
-  Madgraph uses `getenv` in condor to copy the cvmfs environment to the worker node. Tier Sites using Singularity containers don't allow this for `LD_LIBRARY_PATH` for linux security reasons. Patch madgraph job templates to append `SRT_LD_LIBRARY_PATH_SCRAMRT` when this is set and is not already part of `LD_LIBRARY_PATH` to workaround this.
- Fix location of `condor_chirp` for singularity-contained jobs.
- Do not execute `git status` from the gridpack_generation script when running from CMS Connect, but do it on its own script instead. Otherwise, `git status` would be run on the worker node where there is no genproductions git repository.
- Fix dashboard reporting.

Features:
- Replace `condor_q` command line subprocesses with the condor python API schedd query. Make sure the python binding directory is defined in the PYTHONPATH (otherwise bare python can still detect it from the system libraries, but somehow the python binaries from CVMFS don't after executing scram/cmsenv)
- Add functions to Madgraph to automatically release hold codes (so that they run again somewhere else) that are known to be transient (e.g: Hitting a Worker node with CVMFS problems, transient networking issues while transferring files,  etc). This is only triggered when the following environment variables exist: `CONDOR_RELEASE_HOLDCODES, CONDOR_RELEASE_HOLDCODES_SHADOW_LIM`. The first variable defines the list of known problematic hold codes, the second imposes a limit to the number of shadows (default 10) for automatic release of a job (default values present in the submit_cmsconnect script). 
- Add a proxy watcher. This is a cron job that checks the user's proxy remaining lifetime and sends an e-mail reminders to renew the proxy in the login node.
- Progressive Walltimes: This  should fix issue #1212. For this to work, `CONDOR_SET_MAXWALLTIMES` needs to exist in the environment (default values present in the submit_cmsconnect script).


perrozzi:
great, thanks a lot!
do you suggest merging and then testing, or hold and wait for further feedback?

khurtado:
@perrozzi : Well, I tested these changes on CMS Connect and went just fine. The only components that could give any issues in other environments (like `submit_condor` on condor 8.6.x with its corresponding python bindings) are all in the madgraph patches directory, which are isolated changes and can be easily removed if needed without modifying the code, so I think just merging and then testing is okay.

perrozzi:
what is the status of cms connect and mg5_amc 2.5.x ?
would you mind porting this also to https://github.com/cms-sw/genproductions/tree/mg25x ?

khurtado:
Will do.

kdlong:
Thanks @khurtado, this is really nice! I think we should also consider talking to the MadGraph authors about whether they would be interested in merging this into their master code.

How do you envision enforcing the requirement that the python bindings should be available at the site? For lxplus/Wisconsin I see it in /usr/lib64/python2.6/site-packages/. It doesn't appear to be an issue to import it and use in Pythia 2.7, however.

PR #1243, opened 2017-07-18
2017 pdfs 242

kdlong:
My understanding from @vciulli and @qliphy was that we preferred to first make the changes in 2.4.2 using the old approach before moving to the new "systematics" approach for 2.5. This pull request is an initial implementation of this, though there are outstanding issues as described below.

- For LO: basically identical to the old approach, with the PDF set names changed
- For NLO: Adopt the list style in the run_card.dat to specify all sets to be consider (currently only in a single example file). Specifying "False" or "True" for the corresponding entry in reweight_PDF tells whether to include the whole set or only the central member. With this approach these two lines need to be included for all samples. A more automated approach will be possible using "systematics" with 2.5.

NOTE: The public version MadGraph copies all PDF sets to the PDFsets directory, which is ~1.5 GB for these 28 sets. It was simple change to use symlinks in the tarball instead (part of this commit). 

Consequence of this: CURRENTLY NOT WORKING ON CONDOR! 

However, in the general workflow all the data files for all PDF sets are copied to each cluster machine. This is slightly broken with the current pull request. It's an easy fix which I've made but haven't pushed yet, because this much file transfer puts a huge load on the cluster and slows things down dramatically. I suspect it will just be impossible on a larger scale, but I don't see an obvious way around this without a script that creates the symlinks expected on the machine before the main job runs. @khurtado, did you ever look at this?

I expect the systematics program can easily get around this by just doing the weighting step locally. My impression is that this will be simpler to figure out than a better way to access the PDF sets without file transfer. @vciulli @qliphy @perrozzi @govoni @efeyazgan is it acceptable to focus on getting this into 2.5 or is it badly needed in 2.4 first?

TESTING: I only tested things from a technical standpoint. I looked to see that LHE files were filled and the weights seemed unique. I did not look at any distributions. You can see the log files and LHE files for the NLO and LO W-->lv samples here: http://kelong.web.cern.ch/kelong/MadGraph2017PDFTests

ISSUE 1: Not working on condor as described above. I didn't test on LSF, but I don't think it depends on the file transfer in the same way as MadGraph.

ISSUE 2: As you can see in the log files, for the NLO samples I get the follow errors in generation time: 

WARNING: Could not access LHAPDF to compute uncertainties for HERAPDF20_NLO_VAR
WARNING: Could not access LHAPDF to compute uncertainties for HERAPDF20_NNLO_VAR
WARNING: Could not access LHAPDF to compute uncertainties for CT14qed_inc_proton

However, the weights seem to appear in the LHE file. At LO, the same sets don't show any clear issues.


bendavid:
I think we should work with the authors to try to get the pdf sets accessed directly from the lhapdf installation on the fly, such that they are neither copied to the worker nodes, nor included in the gridpacks.  In principle this might already be possible with proper use of the cluster_local_path parameter, but I never made it work/fully understood how it was supposed to.

bendavid:
Also I don't see why we should bother implementing or testing this in 2.4.x.

If we want to use the old method with pdf lists specified in the run card this also still works in 2.5.x (the python method is optional, not mandatory)

kdlong:
The idea behind having it working in 2.4 is to have a contingency plan in case any of the various outstanding issues in 2.5 don't converge quick enough.

During the initial "gridpack" step are all the PDFs necessary? Are the grids PDF independent? If so I think the best solution would be to just store symlinks to the cvmfs paths in the tarball and to do all the reweighting locally. The reweight step is fast and even without the PDF transfer transferring the LHE files back is overhead that isn't necessarily worth it.

bendavid:
The grids don't depend on the additional pdfs, but it IS necessary to produce some events at gridpack generation time in order to initialize madspin, etc.  If the on the fly method is used then this will require accessing the pdfs.

Switching to the python method might indeed avoid this issue.

govoni:
During the last coordination meeting, @vciulli estimated the validation time of the 2.5 release to be 90 days at least. Therefore, the production will start with the 2.4 and switch to 2.5 once the latter is ready.
Josh, how fast do you evaluate, the possibility to have a patch to 2.4  to get the pdf sets accessed directly from the lhapdf installation on the fly?

kdlong:
Actually I realized there is already support for this, using the cluster_local_path variable in the run card. It's seems to not quite work with all the PDF sets (I think it expects to find a directly where the .dat files are located, not a directory of directories with .dat files) but I may have a workaround for this already which I'm testing now. Regardless, it's looks a lot simpler than I initial though.

bendavid:
Ok, I thought the gridpacks needed for 2.5 validation were already done aside from the problematic w+jets ones which we concluded were using the wrong cards?

khurtado:
@kdlong: Never looked into the PDF sets. I agree that reading these sets straight from CVMFS makes much more sense than copying or symlinking at the WN. I will take a look at the `cluster_local_path` option.

perrozzi:
that was fast! ready to merge, let me know if you prefer to wait

kdlong:
From my perspective this is ready to be tested more extensively. I have tested the LO and NLO examples on top of Kenyi's fixes and see no issues so far.

PR #1247, opened 2017-07-20
Add authors fix to bug in 3 body decays with FxFx matching

kdlong:
Patches from here:
http://bazaar.launchpad.net/~mg5core1/mg5amcnlo/2.5.6/revision/276

Discussion of issues here:
https://answers.launchpad.net/mg5amcnlo/+question/631090


perrozzi:
ready to be merged?

bendavid:
This looks good to me (and is the last non-technical change foreseen, such that validation could go ahead full steam).

(If systematic weights are later migrated to python this will of course also have to be checked, but probably a single process is sufficient in that case)

Issue #1223, opened 2017-07-05
Update MadGraph examples to 2017 PDF recommendation 


kdlong:
Implement list of PDF sets in [1] into example cards and test performance.

For NLO MadGraph: Change this line to appropriate LHAPDF numbers 

https://github.com/cms-sw/genproductions/blob/mg25x/bin/MadGraph5_aMCatNLO/cards/examples/wplustest_5f_NLO/wplustest_5f_NLO_run_card.dat#L65

For LO MadGraph: Change calls in Syscalc/update to new "systematics" tool.

https://github.com/cms-sw/genproductions/blob/mg25x/bin/MadGraph5_aMCatNLO/runcmsgrid_LO.sh#L94-L120

https://cp3.irmp.ucl.ac.be/projects/madgraph/wiki/Systematics

govoni:
a curiosity: how does ng5nlo know that for some pdfs we want only the central value, while for others we want also the uncertainty variations?

bendavid:
There's a syntax for this, but I forget what it is at the moment...

govoni:
@kdlong @vciulli @qliphy : do you happen to know how to do it? This we need, for the tests of the PDF sets configuration, which is here https://docs.google.com/spreadsheets/d/1YjhruLPvmwqoA1YZPgmU5H9T3ELnYnriHgO9DrJQoxk/edit#gid=0

kdlong:
From https://cp3.irmp.ucl.ac.be/projects/madgraph/wiki/Systematics#NLO, it looks like something like 29000@0 is supported, but I haven't yet tried this out.

govoni:
Ok, thanks. So how far are we, from having an example that implements the pdf weights we planned to include in the production?

perrozzi:
according to the link posted by kdlong EVERY request will need to implement the list of reweightings... probably this should be automatized inside gridpack_generation.sh right?
as far as I understand/remember also for LO the same trick will be needed.
Can anybody confirm?

bendavid:
For NLO at least I think we will continue to use the "on the fly" method, where everything is specified in the run card as before.

bendavid:
Ah, but indeed since it is possible to use the python method also for NLO, maybe this is preferred in order to have more coherence between LO and NLO.

In this case I think the thing to do first is to migrate the LO systematics from syscalc to the new python method to understand how to implement it (then decide what to do for NLO).

perrozzi:
[...]

PR #1248, opened 2017-07-21
Allow jobs to read PDF sets directly from cvmfs

khurtado:
This pull request fixes the `cluster_local_path` in madgraph.
Just like other settings, `set cluster_local_path <value>` doesn't work, so I had to workaround that by directly setting that into the configuration. This setting is used twice: during the code-generation step, some fortran templates are modified according to this value... then it is use in the integrate step too. The first part failed because our cvmfs paths are too long in string, so I had to fix that too.

I tested on CMS Connect with the `wplustest_4f_NLO` gridpack and worked fine. Haven't tested further.

This is related to #1223


perrozzi:
@kdlong I meant ready to merge this before #1243 unless there are other comments

kdlong:
Thanks @khurtado! @perrozzi there's no reason it couldn't be merged before  #1243. Let me test it and get back to you.

kdlong:
Currently testing with #1243 and it seems to be working. However, with all these PDF sets the jobs seem to be taking quite a bit longer at NLO and haven't yet finished.

kdlong:
I confirm that this is working with the full list of PDF sets. Looks ready to merge to me.

PR #1250, opened 2017-07-24
[Mono-Higgs] Add higher Z' mass points for A0 mass=300,500,800 GeV

syuvivida:
Given the increase of luminosity, now we can set limits on higher Z' mass points. Therefore, we would like to add more mass point for the Zp2HDM model. It's a 3*4 mass grid with 3 A0 mass points and 4 Zp mass points.

mA0=(300,500,800) GeV
mZp=(2750,3000,3500,4000) GeV

Note, there is some difference with respect to the cards of lower mass points.
1) adding the extramodels card (since this card was not required two years ago)
2) set decay width to auto (it was not working with MG5 2.3.3 in the gridpack mode, now it works with MG 2.4.x). We have cross-checked the cross sections are identical for MG5 2.3.3 and MG5 2.4.2.


govoni:
does it make sense to modify all the gridpacks with the new configuration, for consistency in view of the next round of production?

syuvivida:
Among the two differences, 1. is simply due to the difference of gridpack_generation.sh framework, 2. we have cross-checked also the kinematic distributions are consistent. For now (2016 data analysis), we don't plan to update the full mass points yet. But for the 2017 analysis, we may do that.

govoni:
ok, thanks for the answer.

PR #1258, opened 2017-07-26
Fix mistake in PDF warning printout

kdlong:
Fixing a bug introduced in the last commit.


perrozzi:
ok so you are introducing the keywords DEFAULT_PDF_....
but they will be checked only after submission, and the user needs to go and check in the log to spot this.
Could it be enforced/checked in the submit_.....sh script requiring the used to acknowledge by pressing a key, before to submit the gridpack production?

perrozzi:
(this can be merged anyhow while thinking of a more "stringent" solution)

kdlong:
@perrozzi, sure I can make the check a bit more restrictive than just the printout. It can also be checked in Lorenzo's scripts. My hope is still that this can be replaced by the "systematics" module where this will happen automatically.

perrozzi:
no rush

Issue #1252, opened 2017-07-25
CMSConnect in MadGraph v2.5.5


kdlong:
Port changes in master branch allowing gridpack generation with CMSConnect to 2.5 branch.

@khurtado, any projections on this? I think it would be very useful to have for testing the Z/Wgamma samples which are very resource intensive.

khurtado:
I have this on my todo list, but haven't been able to work on it yet, hope to have it for the middle of next week though.

khurtado:
@kdlong Just made a pull request for this:  #1261

kdlong:
Great, thanks! Testing at T2_US_Wisconsin now.

khurtado:
Thanks! Let me know how that goes.

kdlong:
It worked without any problems. We just updated to 8.6, so that also doesn't seem to be a problem. Maybe we could add the line 

PYTHONPATH=/usr/lib64/python2.6/site-packages:$PYTHONPATH

into the submit_condor_gridpack_generation.sh script since this seems to be the expected location, and if the packages are located elsewhere this line shouldn't cause any harm. What do you think?

bendavid:
This could break things if there are a mix of python versions.  Better to add only the path for the needed condor bindings package.

khurtado:
@kdlong: I would prefer not to hardcode that directory in the PYTHONPATH every time if possible. 
If any change is necessary, it should be added into the `source_condor.sh` script which `source_condor / source_cmsconnect` depend on. This script currently finds the PATH for the python bindings in the local system as long as it is available in the and python 2.x is the default version in the system, which is the usual in EL6/7 based systems and will probably be the case until 2020 (end of support of python 2.7.x). If this becomes a problem we should push to have these dependencies included in CVMFS ( #1212 ) . We could also easily include these 2 libraries in the github repository, but then somebody would need to keep track of updates when needed, so I would rather not go that route.

 @bendavid The problem is the condor bindings only have its own directory when they are installed from the condor tarball binaries. When condor and the bindings are  installed from RPMS, `htcondor.so` and `classads.so` are installed in the parent directory on EL6/7-based systems, so the whole `site-packages` will be included in most cases.
In any case, since we setup the CMS framework before doing any actual work in madgraph and CVMFS PATHs are prepended in the PYTHONPATH, a module will only tried to be used from the local system if it's not avaialble in cvmfs... and at this point the only candidate for that would be the condor python bindings.

PR #1261, opened 2017-07-27
Port CMS Connect/Condor changes to mg25x branch

khurtado:
This basically has everything that has been added to the `master`  branch so far for CMS Connect.
Shortly tested with `wplustest_4f_NLO` on CMS Connect. Since this involves madgraph changes to condor and the way jobs read the PDF sets (from cvmfs rather than copying to the worker node scratch area), can anybody test the submit_condor script on `lxplus` and/or Tier2 Sites?

 Changes listed below:
Genproductions scripts:
- `gridpack_generation.sh` modified to split work in different steps (CODEGEN, INTEGRATE, ALL)
- Added `submit_cmsconnect_gridpack_generation.sh` for submission to CMS Connect
  -  `CODEGEN` step in CMS-Connect is submitted as a condor job and output is transferred back via XRootD or condor-chirp, then extracted and adapted properly for the `INTEGRATE` step 
  - Submits a proxy watcher condor cron job to send notifications when proxy certificate is about to expire (48 hours in advance, actually).
  - Automatic dashboard reporting for jobs
  - Automatically release hold codes (so that they run again somewhere else) that are known to be transient (e.g: Hitting a Worker node with CVMFS problems, transient networking issues while transferring files, etc). This is only triggered when the following environment variables exist: CONDOR_RELEASE_HOLDCODES, CONDOR_RELEASE_HOLDCODES_SHADOW_LIM. The first variable defines the list of known problematic hold codes, the second imposes a limit to the number of shadows (default 10) for automatic release of a job (default values present in the submit_cmsconnect script).
  - Progressive Walltimes: Defined in #1212. For this to work, `CONDOR_SET_MAXWALLTIMES` needs to exist in the environment (default values present in the submit_cmsconnect script).

Madgraph Changes:
- Add cvmfs paths to `LD_LIBRARY_PATH` on Tier Sites using singularity. Madgraph uses getenv in condor to copy the cvmfs environment to the worker node. Tier Sites using Singularity containers don't allow this for LD_LIBRARY_PATH for linux security reasons. Patch madgraph job templates to append SRT_LD_LIBRARY_PATH_SCRAMRT when this is set and is not already part of LD_LIBRARY_PATH to workaround this.
- Replace condor_q command line subprocesses with the condor python API schedd query. Make sure the python binding directory is defined in the PYTHONPATH (otherwise bare python can still detect it from the system libraries, but somehow the python binaries from CVMFS don't after executing scram/cmsenv)
- Add functions to Madgraph to automatically release certain condor hold codes defined on environment user settings.
- Add walltimes classads support, which are used by the CMS Global Pool
- Allow jobs to read PDF sets directly from cvmfs by fixing `cluster_local_path`


perrozzi:
what's the status of this?

khurtado:
@perrozzi : This is being discussed in #1252. The summary is that  @kdlong tested the changes and they worked without any issues on both CMS Connect and T2_US_Wisconsin (in the latest, using the condor submit script on condor 8.6.x).

Since this requires a new dependency (python condor bindings) we are still discussing what the best way to guarantee the availability of these bindings are (so that's related to #1212 also), but since changes work on CMS Connect, T2s and lxplus (similar environment as T2_US_Wisconsin test, all of them providing the bindings locally), I don't think we need to resolve that issue before being able to merge this, unless @kdlong @bendavid or any other person think otherwise.

Are there any other opinions?

kdlong:
Looks good to me. We could bring up the distribution of the python bindings to the computing group via email or in a future meeting, but at the moment we're not aware of this causing any issues.

perrozzi:
great, merged

PR #1264, opened 2017-07-31
gridpacks for ggh->aa->4tau, and qqh->aa->2b2tau

cecilecaillol:
Higgs EXO subgroup
Gridpacks for ggH->aa->4tau, for mH=125 GeV, and 15 < ma < 60 GeV per steps of 5 GeV
Gridpacks for qqH->aa->2b2tau, for mH= 125 GeV, and 10 < ma < 60 GeV per steps of 5 GeV
Code inspired from ggH->aa->2mu2tau gridpacks in same repository

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/higgs/ggh01_M125_Toa01a01_M10_Tobbtautau/ggh01_M125_Toa01a01_M10_Tobbtautau_run_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/higgs/ggh01_M125_Toa01a01_M10_Tobbtautau/ggh01_M125_Toa01a01_M10_Tobbtautau_run_card.dat
@@ -1,0 +1,271 @@
+#*********************************************************************
+#                       MadGraph5_aMC@NLO                            *
+#                                                                    *
+#                     run_card.dat MadEvent                          *
+#                                                                    *
+#  This file is used to set the parameters of the run.               *
+#                                                                    *
+#  Some notation/conventions:                                        *
+#                                                                    *
+#   Lines starting with a '# ' are info or comments                  *
+#                                                                    *
+#   mind the format:   value    = variable     ! comment             *
+#*********************************************************************
+#
+#*******************                                                 
+# Running parameters
+#*******************                                                 
+#                                                                    
+#*********************************************************************
+# Tag name for the run (one word)                                    *
+#*********************************************************************
+  tag_1     = run_tag ! name of the run 
+#*********************************************************************
+# Run to generate the grid pack                                      *
+#*********************************************************************
+  .false.     = gridpack  !True = setting up the grid pack
+#*********************************************************************
+# Number of events and rnd seed                                      *
+# Warning: Do not generate more than 1M events in a single run       *
+# If you want to run Pythia, avoid more than 50k events in a run.    *
+#*********************************************************************
+  1000 = nevents ! Number of unweighted events requested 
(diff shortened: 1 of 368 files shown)


cecilecaillol:
Could you please approve this pull request?

bin/MadGraph5_aMCatNLO/cards/production/13TeV/higgs/ggh01_M125_Toa01a01_M10_Tobbtautau/ggh01_M125_Toa01a01_M10_Tobbtautau_run_card.dat
```diff
@@ -0,0 +1,271 @@
...
+#*********************************************************************
+# Beam polarization from -100 (left-handed) to 100 (right-handed)    *
+#*********************************************************************
+        0     = polbeam1 ! beam polarization for beam 1
+        0     = polbeam2 ! beam polarization for beam 2
+#*********************************************************************
+# PDF CHOICE: this automatically fixes also alpha_s and its evol.    *
+#*********************************************************************
+ 'lhapdf'    = pdlabel     ! PDF set
+  263000    = lhaid     ! if pdlabel=lhapdf, this is the lhapdf number
```
perrozzi: I believe you need to use 263400 which is 4F, since the process is 4F too

cecilecaillol: Where can I find the recommended numbers for lhaid? Would using 263400 make a non negligible difference? All H->aa samples (for example https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/higgs/ggh01_M125_Toa01a01_M11_Tomumutautau/ggh01_M125_Toa01a01_M11_Tomumutautau_run_card.dat, but there are many more) have been produced with 263000. I would prefer to keep the same settings for all final states if possible. If lhaid makes a difference we need to reproduce dozens of samples officially.

perrozzi:
I am afraid you might need to reproduce dozens of official samples then...
The first question to ask is whether the model you are importing (NMSSMHET) uses 4F or 5F.
Your cards are setup to use 4F (proton defined without the b, maxjetflavor = 4) but you use a 5F PDF.
This is inconsistent, and although I never tried explicitly to size the difference, it will be sizeable since in the 4F case the b quark PDF is absorbed into the gluon PDF, in the 5F the b quark PDF will be separated.
In other words, the most notable problem is that you are sampling a gluon PDF which is largely underestimated calling a 5F PDF in a 4F calculation.
Both distributions and cross section will be wrong...
I strongly suggest to investigate further with the analysis proponents what is the model using and consider reproducing the old samples in case of need

bin/MadGraph5_aMCatNLO/cards/production/13TeV/higgs/ggh01_M125_Toa01a01_M10_Tobbtautau/ggh01_M125_Toa01a01_M10_Tobbtautau_run_card.dat
perrozzi: the list of LHAPDF sets is here https://lhapdf.hepforge.org/pdfsets.html

cecilecaillol:
Can you please refuse this pull request? I am going to submit a new one with the correct lhaid.

PR #1268, opened 2017-07-31
Settings for ttbar Tune validation using NNPDF3.1 NNLO as central PDF

jfernan2:
Notes:
- Powheg crashes when asking for reweighting of NNPDF3.1_LO_as0118
- FxFx only allows consecutive pdf lha ids reweighting, since NNLO, NLO and LO ids are not consecutive, only NLO will be used for reweighting and NNLO as central PDF
- MLM has NNPDF3.1 NNLO as central PDF, and any other PDF as reweighting thanks to syscalc

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/TuneValidation/ttbar_FxFx/tt012j_5f_ckm_NLO_FXFX/tt012j_5f_ckm_NLO_FXFX_run_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/TuneValidation/ttbar_FxFx/tt012j_5f_ckm_NLO_FXFX/tt012j_5f_ckm_NLO_FXFX_run_card.dat
@@ -55,7 +55,8 @@ average = event_norm ! Normalize events to sum or average to the X sect.
 # PDF choice: this automatically fixes also alpha_s(MZ) and its evol.  *
 #***********************************************************************
  lhapdf    = pdlabel   ! PDF set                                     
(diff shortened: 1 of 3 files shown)


bendavid:
That limitation for FXFX should not be present anymore in 2.4.x (there should already be some updated examples right?)

jfernan2:
Hi Josh, sorry but I did not find those updated examples...

bendavid:
See here
https://github.com/cms-sw/genproductions/commit/8b60ca68ccb2c931c6321213fe2f755ca5d2443a

jfernan2:
OK, I will use that, however afaics it only allows central pdfs (so no hessian variations) and NNPDF3.1 LO (315000) is excluded for some reason in the tube_validation branch...

bendavid:
So with that syntax ~all combinations of central value and variations should be possible.  Indeed I think it was decided to remove the LO weights from the default set.  @kdlong any further comments?

kdlong:
For 2.4.2 you can either include the full set or only the central value of the set. There's a hardcoded limit of 30 PDF sets at the moment.

As Josh pointed out, if you use 

$DEFAULT_PDF_SETS = lhaid

The script will substitute the combination of sets/central values recommended for 2017 production that came out of all the discussions over the past few months and recorded in the spreadsheet in [1]. I don't really remember all the reasoning behind why these choices were made though.

[1] https://docs.google.com/spreadsheets/d/1YjhruLPvmwqoA1YZPgmU5H9T3ELnYnriHgO9DrJQoxk/edit#gid=0

PR #1274, opened 2017-08-03
Cards for h->aa, VBF and ggH production, bbtt tttt and mmtt final states

cecilecaillol:
This is a correction of the previous pull request. The lhaid has been changed to 4F as the process is. The theorists responsible of the NMSSMHET model confirmed we can use 4F.


cecilecaillol:
The files that had problems in the previous requests have been removed (which is the reason there are several commits in this PR).

perrozzi:
Ciao, I see only new files here, while I would have expected also modification to existing files. Am I missing something?

cecilecaillol:
I have removed the old files, and replaced them by new ones in a cleaner structure (subdirectories with various masses per production and decay modes). The previous pull request was not accepted so I have rewritten the requests from scratch in the new structure instead of modifying them.

perrozzi:
Maybe I was not clear, let me re-state it.
I am not talking about old files in the previous PR that were not merged into cms-sw:master.
I am talking about files that should be already present in cms-sw:master since long time and should have been used to produce the "dozens of samples" which had inconsistent settings and, I assume, need to be modified here. Are these files already in cms-sw:master? Do they need to be modified?

cecilecaillol:
I have modified the files I am responsible for. I have contacted the analysts for the other final states (Rachel Yohay and Nadjieh Jafari), and they should make pull requests soon with modifications to the current cards. After this is done we will check that there is no outdated file left.

PR #1275, opened 2017-08-03
For pdf scheme

nadjieh:
This is to correct the LHEID for the correct use of pdf in higgs production.
It is changed as
 263000 to 263400.


perrozzi:
you need to solve the conflicts to be able to merge

nadjieh:
Sorry but what is the conflict? These are exactly those files that I changed the lheid in!

perrozzi:
I have no idea, just pull from cms-sw:master and see what it is

nadjieh:
It looks like it was based on old commits not the head version of genproduction in CMSSW . Have no clue why but OK.
I will close this pull and open another one which is based on the latest version.

PR #1279, opened 2017-08-08
s-channel 14GeV cards

ckmackay:


ckmackay:
Sorry, that should read s-channel 14 TeV cards.

perrozzi:
@ckmackay just in case: you can always rename a PR

PR #1285, opened 2017-08-13
DijetPowheg

pgunnell:
Upload of two cards for dijet production in POWHEG with NNPDF31 (NLO and NNLO sets)


efeyazgan:
Should these be commented out?
+! To be set only if using internal (mlm) pdfs
+! 131 cteq6m
+ndns1 131            ! pdf set for hadron 1 (mlm numbering)
+ndns2 131            ! pdf set for hadron 2 (mlm numbering)

perrozzi:
when the line starts with `!` is equivalent to `#` and is interpreted as commented out,
concerning the internal pdf I agree that should be commented out explicitly to avoid any confusion and problem

pgunnell:
Thanks a lot Luca and Efe! Now should be ok, with your comments.

PR #1286, opened 2017-08-14
2017 PDF sets using systematics module for MadGraph 2.5.5

kdlong:
This is probably not quite ready to be merged, but I'm starting the pull request so it can be referenced and so people can start to comment.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/examples/wplustest_5f_NLO/wplustest_5f_NLO_run_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/examples/wplustest_5f_NLO/wplustest_5f_NLO_run_card.dat
@@ -62,7 +62,7 @@
 # PDF choice: this automatically fixes also alpha_s(MZ) and its evol.  *
 #***********************************************************************
  lhapdf = pdlabel ! PDF set
- 292200, 90000, 90200, 90400, 13100, 13164, 13166, 25200, 25260, 11000, 11067, 11069 = lhaid   ! If pdlabel=lhapdf, this is the lhapdf number. Only 
+ 303600 = lhaid   ! If pdlabel=lhapdf, this is the lhapdf number. Only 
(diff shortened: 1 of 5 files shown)


perrozzi:
any news here?

bin/MadGraph5_aMCatNLO/cards/examples/wplustest_5f_NLO/wplustest_5f_NLO_run_card.dat
```diff
@@ -92,20 +92,20 @@
 #*********************************************************************** 
 # Reweight variables for scale dependence and PDF uncertainty          *
 #***********************************************************************
- 1.0, 2.0, 0.5 = rw_rscale ! muR factors to be included by reweighting
- 1.0, 2.0, 0.5 = rw_fscale ! muF factors to be included by reweighting
+ 1.0 = rw_rscale ! muR factors to be included by reweighting
+ 1.0 = rw_fscale ! muF factors to be included by reweighting
  True = reweight_scale ! Reweight to get scale variation using the
```
bendavid: Would it make sense to have False here as well?

bendavid:
This pull request should probably also remove the installation/compilation of syscalc from the main script.

bin/MadGraph5_aMCatNLO/cards/examples/wplustest_5f_NLO/wplustest_5f_NLO_run_card.dat
kdlong: Annoyingly 1 of the 2 between reweight_PDF and reweight_scale have to be set to true for the weights to be stored. I asked Olivier about this and it seemed intentional

kdlong:
@bendavid agreed, done now. Should also update the PDFs (and maybe add a way for users to customize the PDFs through cards)

kdlong:
I'm not sure if the PDF list is yet the "final" one but otherwise I think this can be merged.

PR #1296, opened 2017-08-25
DY To mumu mll105 160

amarini:
DY gridpacks to MuMu and MLL requirement 105 -- 160
* MLM LO
* AMC@NLO


bin/MadGraph5_aMCatNLO/cards/production/13TeV/DYJets_HT_LO_MLM/DYJetsToMuMu_HT_mll105_160/DYJetsToMuMu_HT_mll105_160_proc_card.dat
```diff
@@ -0,0 +1,39 @@
...
+#************************************************************
+#*                                                          *
+#*               Command File for MadGraph 5                *
+#*                                                          *
+#*     run as ./bin/mg5  filename                           *
+#*                                                          *
+#************************************************************
+
+import model sm-ckm_no_b_mass
+define p = g u c d s u~ c~ d~ s~
```
perrozzi: naive question: aren't you missing the b quark here?

amarini: @perozzi, I actually used these ones as examples: https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/DYJets_HT_LO_MLM/DYJets_HT_mll50/DYJets_HT-100to200
but I think you are right, we should have a b jets there since it is asked the 5-flav scheme.
Were these cards the one used for the 2016 Moriond17 production?

perrozzi: I checked [*] and am afraid so........
(@bendavid can you have a look too?)

[*] https://cms-pdmv.cern.ch/mcm/public/restapi/requests/get_fragment/SUS-RunIIWinter15wmLHE-00030
pointing to
/cvmfs/cms.cern.ch/phys_generator/gridpacks/slc6_amd64_gcc481/13TeV/madgraph/V5_2.2.2/DYJets_HT_LO_MLM/DYJets_HT-100to200/V1/DYJets_HT-100to200_tarball.tar.xz
for which the log contains

Defined multiparticle p = g u c d s u~ c~ d~ s~
[...]
generate p p > l+ l- / h @0

perrozzi: @kdlong @vciulli @qliphy can you have a look too and comment please?

amarini: Any news ? I can change and produce the one with mll 105-160 with a b inside, but you may want to understand if this is correct before the new production.

amarini:
@perrozzi, is there anything witholding the amc@nlo production ?

perrozzi:
as I said, the datacards are wrong, no matter that they were used as they are also for other productions.
I kindly ask again @kdlong @vciulli @qliphy @bendavid to comment

kdlong:
What's the use case here? I think in general we would prefer 5 flavor for Drell-Yan, so unless there is a strong argument to be consistent with the HT-binned samples I would update these to 5F. Other than the HT-binned samples most LO and NLO samples were done with the 5FS for 2016.

perrozzi:
Apparently also the HT binned samples are bugged because they are 5F but the b is not part of the incoming proton! cfr https://github.com/cms-sw/genproductions/pull/1296#discussion_r135560406
I propose to fix it in this datacard (can you confirm that is a bug?), we can discuss the HT-binned elsewhere.

amarini:
The DY MLM LO with the bjets in the proton is here (datacards committed):
~~~
~amarini/public/DYJetsToMuMu_HT_mll105_slc6_amd64_gcc481_CMSSW_7_1_28_tarball.tar.xz
~~~

perrozzi:
please move the datacards to
https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/pre2017/

bendavid:
Recent versions of madgraph (definitely from 2.4.x at least) will actually automatically take care of making the proton and jet definitions consistent at model import time.  Of course if this is overidden with the inconsistent definition after the model import, then it will indeed be wrong.

Probably we should update the example cards to remove the explicit proton and jet definitions since this should only be used/needed in special cases from now on.

perrozzi:
how can we know then if the used PDF is correct or not?
It can be relatively straigthforward for the most used models, but for others will be a nightmare...
cfr https://github.com/perrozzi/genproductions/pull/14

kdlong:
@bendavid @perrozzi: If this is done by editing the proton multiparticle definition, then the check I started to set up here should still work (then choose the PDF based on this): https://github.com/kdlong/genproductions/blob/pdfsync/bin/MadGraph5_aMCatNLO/gridpack_generation.sh#L347-L353

bendavid:
Yes, the multiparticle definition of p and j are changed automatically as part of the "import model" call, so I agree this check should still work. (If I read the regexp correctly it's also not sensitive to the order in which the quarks appear in the multiparticle definition, which is also good)

amarini:
@perrozzi, Your link is broken, I cannot follow https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/pre2017/ . 
Do you want me to commit there anyhow ? 
Is the pre2017 the compressed file ? 
If so, every PR to that file is going to conflict.

amarini:
Is there anything withholding the start of the production of these samples ?

perrozzi:
just move the files in the pre2017 directory (you need to create it)

perrozzi:
(this PR got "trapped" in a major revision of the repository needed for the forthcoming Fall17 production)

amarini:
moved to pre2017 directory.

perrozzi:
still wrong path, use
bin/MadGraph5_aMCatNLO/cards/production/pre2017/
not
bin/MadGraph5_aMCatNLO/cards/pre2017/

amarini:
(y)

PR #1297, opened 2017-08-29
A few more HGCal samples for the official TDR production

apsallid:


perrozzi:
Why the datacards are called DoublePi and the comment (as well as the filter) Single Pi ?
Also, for better clarity in the datacard name use "p" to indicate the decimal dot, for instance "1p8" to indicate 1.8

apsallid:
This is the convention we followed because we thought that adding the
antiparticle means that we should name the filter DoubleSomething. From your comment I
understand that I should rename the files to SingleSomething right? And keep the comment
to single. Double must refers to shooting two particles (which we don't) and adding the antiparticle should make no change to the name, right?

PR #1299, opened 2017-09-02
Store all log files (needed for CMS connect) and data cards

kdlong:
At the moment the codegen file for CMSConnect isn't stored. This fixes that problem.


perrozzi:
would you mind adding also ALL the datacards that are provided as input to mg5_amc when creating the gridpack?

perrozzi:
great, thanks!

PR #1300, opened 2017-09-02
Store all log files (needed for CMS connect) and data cards

kdlong:
At the moment the codegen file for CMSConnect isn't stored. This fixes that problem.


perrozzi:
are the WA cards meant to be stored or are they spurious?

perrozzi:
(also in this case, would you mind adding all the datacards in the gridpack?)

kdlong:
The WA cards are spurious to this change I guess, but they since they were needed to remake the gridpack in this version I think it's reasonable to leave them in. I'll push the change to add the data cards shortly.

PR #1303, opened 2017-09-06
Important pdf  and weight fixes

covarell:
- Insert a check (which can be bypassed) for standard 2017 PDFs, rather than silently forcing these
- Insert a check for anomalus weights

@yuanchao


perrozzi:
nice, thanks

perrozzi:
let us know if it can be merged or needs to wait for further input (I assume you tested)

covarell:
Modifications are tested.

What is not tested is whether all POWHEG processes work properly with new reweighting (can be done only after creating gridpacks for all of them): so the new features are for MC contacts to check their gridpacks.

Issue #1287, opened 2017-08-15
[2017 MC v2] Finalize pdf weights for 2017


bendavid:
These would be useful for alpha_s extraction studies.

Pdf sets are 

322500 | NNPDF31_nnlo_as_0108
322700 | NNPDF31_nnlo_as_0110
322900 | NNPDF31_nnlo_as_0112
323100 | NNPDF31_nnlo_as_0114
323300 | NNPDF31_nnlo_as_0117
323500 | NNPDF31_nnlo_as_0119
323700 | NNPDF31_nnlo_as_0122
323900 | NNPDF31_nnlo_as_0124

Each includes full uncertainties (101 replicas) but it would be sufficient to just have the central set for each (ie 8 additional weights)

efeyazgan:
Would this be a template analysis?
Won't you consider using a Fixed Order calculation (e.g. analogous to TOP-17-001)?

bendavid:
This is a general request for the moment to leave open the possibility of doing something at detector level also.
(since it's only 8 additional weights the cost is not so high)

syuvivida:
Hello,
  Could we add the following two PDF sets in the list of PDF weights? I think just the central values are OK.
263000    NNPDF30_lo_as_0130          --> to compare with ATLAS DM cross sections
263400    NNPDF30_lo_as_0130_nf_4 --> to cross check/compare with 2015/2016 MC samples

If possible, it would be nice to add these two PDF sets in the default runcmsgrid_LO.sh. If we can only add them for a subset of the samples, below is the list of samples to add:

               ZprimeToA0hToA0chichih*_2HDM_MZp-*_MA0-*_13TeV-madgraph*
               ZprimeToZhToZinvh*_2HDM_M-*_13TeV-madgraph
               MonoH*_ZpBaryonic_MZp-*_MChi-*_13TeV-madgraph
               MonoH*_Scalar_MZp-*_MChi-*_13TeV-madgraph
               ZprimeToWW*_narrow_M-*_13TeV-madgraph
               ZprimeToWW*_width*_M-*_13TeV-madgraph
               ZprimeToZH*_narrow_M-*13TeV-madgraph-pythia8
               ZprimeToZh*_narrow_M-*_13TeV-madgraph
               WprimeToWhToWlephtata_narrow_M-*_13TeV-madgraph
               WprimeToWZ_width*_M-*_TuneCUETP8M1_13TeV-madgraph-pythia8
[...]

PR #1306, opened 2017-09-11
Pdf last sync before production

perrozzi:
following https://github.com/cms-sw/genproductions/issues/1287


kdlong:
Some of these PDFs (32XXXX) are not in the LHAPDF set in CMSSW_7_1_28.

kdlong:
I don't see them in any current CMSSW version. I vote we exclude these from the list and then add them in 1-2 weeks when the LHAPDF version can be updated.

@bendavid, are there specific samples that should be put on hold until we have this?

perrozzi:
btw, are ALL the other PDFs available both in 71x and 92x/93x ?

bendavid:
Well I think the alphaS variations are in principle desirable for all V+jets and ttbar samples...

kdlong:
For 5F, yes I think so. For 4F, no.

```grep 320900 /cvmfs/cms.cern.ch/slc6_amd64_gcc530/*/lhapdf/*/share/LHAPDF/pdfsets.index``

Gives no matches.

perrozzi:
https://www.hepforge.org/archive/lhapdf/pdfsets/6.2/ being added to cvmfs, it should have them all

perrozzi:
includes #1308

perrozzi:
@covarell @yuanchao this PR contains substantial reorganization of make_rwl.py (use dictionaries and a single loop to store PDFs, added new PDF sets, allow 4F sets) and integration inside run_pwg.py (i.e. pwg-rwl.dat produced automatically at process compilation time checking 4F and 5F by process name, few checks and warnings added)

please test

yuanchao:
Hi, Just did a quick test with DMGG process... I don't see the corresponding codes in master. So I try to checkout from perrozzi:pdfsync. A show stopper is that the make_rwl.py is done AFTER the PDF checks. So the script fails anyway...

`
if [[ $is5FlavorScheme -eq 1 ]]; then
  echo "INFO: The process $process uses the 5F PDF scheme"
else
  echo "INFO: The process $process uses the 4F PDF scheme"
fi

# enforce standard pdfs?
echo "INFO: PDF set for central value: ${teststandpdf}"
if ([[ $is5FlavorScheme -eq 1 ]] && [[ ${teststandpdf} -ne 306000 ]]) || ([[ $is
5FlavorScheme -eq 0 ]] && [[ ${teststandpdf} -ne 320900 ]]) ; then
   if [ $noPdfCheck == '0' ] ; then
      fail_exit "WARNING: The input card does not have the standard 2017 PDF (NN
PDF31 NNLO, 306000 for 5F, 320900 for 4F). Either change the card or run again w
ith -d 1 to ignore this message. Exiting now..."
   else 
      echo "WARNING: FORCING A DIFFERENT PDF SET FOR CENTRAL VALUE wrt standard 
2017 (NNPDF31 NNLO, 306000 for 5F, 320900 for 4F)"
   fi
fi

cd $WORKDIR
python make_rwl.py ${is5FlavorScheme} ${teststandpdf}
cd ${name}
`

> INFO: The process DMGG uses the 5F PDF scheme
> INFO: PDF set for central value: 260000
> WARNING: The input card does not have the standard 2017 PDF (NNPDF31 NNLO, 306000 for 5F, 320900 for 4F). Either change the card or run again with -d 1 to ignore this message. Exiting now...

perrozzi:
Hi @yuanchao, I am not sure I understand. 
1) `perrozzi:pdfsync` is merged into `cms-sw:master` (i.e. this PR) so you should be able to get the modifications, please try again and let me know if you still have problems
2) I am not sure what you refer to as "show stopper". The PDF check, originally implemented by Roberto, in your case "correctly" fails because
`WARNING: The input card does not have the standard 2017 PDF (NNPDF31 NNLO, 306000 for 5F, 320900 for 4F). Either change the card or run again with -d 1 to ignore this message. Exiting now...`
as suggested in the WARNING message, you are running a 5F process and have different central PDF (260000) wrt the default 2017 one (306000). You can force the script to continue if you really want to use them (in your case is a test so doesn't matter) adding the argument `-d 1`. I tested extensively and it works, for instance compiling with:
`python ./run_pwg.py -p 0 -i powheg.input -f test -m DMGG -d 1`
Please try again and let me know if you still have problems

yuanchao:
Sorry it seems that I mixed up with the stuff that checked out days before the merge is done. I just clean up everything and try again. The make-rwl part is working with the cms-sw:master.

As to PDF check, the error only shows up after the generated shell script is running. This means that if the check is failed, I'll know it only after the submitted jobs exits. Would it be better to be done in the driving python script, run_pwg.py?

perrozzi:
good idea, can you please do it?

yuanchao:
Give me some time.

perrozzi:
@yuanchao when do you think you will have it ready?

yuanchao:
CF #1339

PR #1308, opened 2017-09-12
Script to checkout legacy cards and modify the PDF set

kdlong:
Adding bash script to checkout legacy cards from the pre2017 tag of the repository, modify the PDF set, and commit them. Is not super robust but has some error checking. Should be a good place to start for most 2017 production.

Would appreciate some validation and comments.


kdlong:
Anything needed here? I'm trying to document it so it would be good to have it in the master branch

perrozzi:
in principle I merged this commits here
https://github.com/cms-sw/genproductions/pull/1306
because of concurring modifications.
Shall I merge that and close this?
We can open new PRs in case of problems with currently missing PDFs, what do you think?

kdlong:
Ah right. I think it makes sense to merge and then correct later if we want to have a (mostly) working and documented recipe as soon as possible.

PR #1309, opened 2017-09-12
Cleanup example cards and LHAPDF support, add NLO bias example

bendavid:
-Remove explicit p and j definition in example cards, since madgraph now handles the 4f and 5f cases automatically

-remove workarounds for older CMSSW releases using LHAPDF5 from scripts (this was needed to support CMSSW_5_3_X, but with environment switching support, scripts no longer need to support this release)

Add example card for NLO bias weighting (example is drell yan FXFX with bias weighting in pT and njets)

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/examples/WW_aTGC_reweight/WW_aTGC_reweight_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/examples/WW_aTGC_reweight/WW_aTGC_reweight_proc_card.dat
@@ -1,6 +1,4 @@
 import model EWdim6-EWdim65f
-define p = g u c d s b u~ c~ d~ s~ b~
-define j = p
 define l+ = e+ mu+ ta+
 define l- = e- mu- ta-
 define vl = ve vm vt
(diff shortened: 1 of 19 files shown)


bendavid:
@kdlong

kdlong:
Very nice! So I assume the bias weighting is working at generation time but hasn't been validated on event samples yet?

The only comment I have is that the central PDF could also be updated in the example cards, to 306000 or using $DEFAULT_PDF_SETS/$DEFAULT_PDF_MEMBERS.

bendavid:
Yes it works at event generation time, and I get weights which make sense qualitatively.
I didn't actually try it in a gridpack yet, or make any kind of tests at the level of distributions.

(also the functional form I put there for the z pt is really random, one should fit the pt from the inclusive sample with some kind of power law, and come up with something which makes sense, probably also forcing the bias to 1.0 below a certain pt threshold to avoid introducing strange behaviour as pt->0)

bendavid:
Well this is 2.6.x so we might as well use the systematic module right?
(ie I should follow what has already been done in https://github.com/cms-sw/genproductions/blob/mg26x/bin/MadGraph5_aMCatNLO/cards/examples/wplustest_5f_NLO/wplustest_5f_NLO_run_card.dat )

bendavid:
(and I think the $DEFAULT_PDF_SETS was not implemented yet in the 2.6.x branch)

bendavid:
This should be ok now (and tested up to the level of producing a gridpack and then producing events from it)

perrozzi:
seems ok to me, @kdlong give me a final feedback

qliphy:
@bendavid  Do we have also an example card for LO? Or is it similar as NLO?  If these are ready, I think we can ask the MG26 team (Hoa, Meng, Petr...) to start check this.

bendavid:
For LO the procedure is somewhat different, and I didn't look in detail how to do it.  I think we can ask Valentina and co. to add that in a separate pull request.

bendavid:
Any reason not to merge this?

PR #1311, opened 2017-09-12
insert Phantom in the production chain as wmLHE

govoni:
the right code now :)

### DIFF
--- a/bin/Phantom/cards/examples/phantomProduction_example.cfg
+++ b/bin/Phantom/cards/examples/phantomProduction_example.cfg
@@ -15,12 +15,10 @@
 CMSSW = CMSSW_9_2_6_patch1
 
 # phantom version to be used
-package = /afs/cern.ch/work/b/ballest/public/phantom/phantom_1_2_8_nc1_lowopt.tar.gz
-# this page should give a reasonable association between architecture and release:
+package = https://cms-project-generators.web.cern.ch/cms-project-generators/phantom/phantom_1_3_slc6_amd64_gcc530.tar.gz
+# this page should give a reasonable association between architecture and CMSSW release:
 # https://cmssdt.cern.ch/SDT/cgi-bin/ReleasesXML?anytype=1
 ARCH = slc6_amd64_gcc530
-# FIXME when I compile in CMSSW, the name of the compiled version has to have the architecture in it
-#       so that the version is picked up with the right architecture
 
 # phantom r.in file to be used (provided in the genproduction github)
 # this is expected to be found in the same folder from where the submission script
@@ -32,7 +30,6 @@ ARCH = slc6_amd64_gcc530
 # foldername.tgz
 foldername = gridpackTesting_SS
 
-#FIXME how to choose between EWK, QCD, ALL?
 
 # ---- ---- ---- ---- ---- ---- ---- ---- ---- ---- ---- ---- ---- ---- ----
 
--- a/bin/Phantom/cards/examples/phantomProduction_example.cfg
+++ b/bin/Phantom/cards/examples/phantomProduction_example.cfg
@@ -18,7 +18,7 @@ CMSSW = CMSSW_9_2_6_patch1
 package = /afs/cern.ch/work/b/ballest/public/phantom/phantom_1_2_8_nc1_lowopt.tar.gz
 # this page should give a reasonable association between architecture and release:
 # https://cmssdt.cern.ch/SDT/cgi-bin/ReleasesXML?anytype=1
-# ARCH = slc6_amd64_gcc530
+ARCH = slc6_amd64_gcc530
 # FIXME when I compile in CMSSW, the name of the compiled version has to have the architecture in it
 #       so that the version is picked up with the right architecture
 
@@ -30,7 +30,7 @@ package = /afs/cern.ch/work/b/ballest/public/phantom/phantom_1_2_8_nc1_lowopt.ta
 
 # folder name containing the gridpack, the gridpack name will be
 # foldername.tgz
-foldername = gridpackTesting_OS
+foldername = gridpackTesting_SS
 
 #FIXME how to choose between EWK, QCD, ALL?
 
@@ -47,8 +47,8 @@ queue     = 1nw
 [generation]
 # what generation has to be done, with which configuration parameters
 # production channel, the list of leptons is expected
-# channel            = mu e vm_ ve_
-channel            = mu e_ vm_ ve
+channel            = mu e vm_ ve_
+# channel            = mu e_ vm_ ve
 
 # number of top/antitop quarks required in the
 # final state. If the option is not set any number of
--- a/bin/Phantom/cards/examples/phantomProduction_example.cfg
+++ b/bin/Phantom/cards/examples/phantomProduction_example.cfg
@@ -14,13 +14,13 @@
 # CMSSW environment for the gridpack creation
 CMSSW = CMSSW_9_2_6_patch1
 
-# this page should give a reasonable association between architecture and release:
-# https://cmssdt.cern.ch/SDT/html/showIB.
-# or search for the release here https://cmssdt.cern.ch/SDT/cgi-bin/ReleasesXML?anytype=1
-ARCH = slc6_amd64_gcc530
-
 # phantom version to be used
 package = /afs/cern.ch/work/b/ballest/public/phantom/phantom_1_2_8_nc1_lowopt.tar.gz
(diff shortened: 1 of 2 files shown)


bin/Phantom/cards/examples/phantomProduction_example.cfg
```diff
@@ -0,0 +1,177 @@
...
+#                                                                                        #
+# ---- ----  ---- ----  ---- ----  ---- ----  ---- ----  ---- ----  ---- ----  ---- ---- #
+
+
+[general]
+# where the zipped version of the compiled phantom code can be found
+# and where the gridpacks will be created
+
+# CMSSW environment for the gridpack creation
+CMSSW = CMSSW_9_2_6_patch1
```
perrozzi: suggest to use CMSSW_9_2_11 as we just did for POWHEG (cfr https://twiki.cern.ch/twiki/bin/viewauth/CMS/PowhegBOXPrecompiled#Download_all_the_necessary_scrip)

bin/Phantom/cards/examples/phantomProduction_example.cfg
```diff
@@ -0,0 +1,177 @@
...
+
+# phantom version to be used
+package = https://cms-project-generators.web.cern.ch/cms-project-generators/phantom/phantom_1_3_slc6_amd64_gcc530.tar.gz
```
perrozzi: I suggest to use the existing convention, cfr
https://cms-project-generators.web.cern.ch/cms-project-generators/slc6_amd64_gcc481/powheg/V2.0/src/
can you change to 
https://cms-project-generators.web.cern.ch/cms-project-generators/slc6_amd64_gcc530/phantom/V1.3/phantom_1_3_slc6_amd64_gcc530.tar.gz
Also, can you rename phantom_1_3.tar.gz to phantom_1_3_src.tar.gz to make it clearer that is the source tarball?

bin/Phantom/cards/examples/phantomProduction_example.cfg
```diff
@@ -0,0 +1,177 @@
...
+
+[parameters]
+# parameters for the phantom config file
+# parameters in this cfg file will be replaced in the phantom cfg template.
+
+# NECESSARY PARAMETER
+# the full list of available pdf is found in the folder stored 
+# in the $LHAPDF_DATA_PATH environment variable of the CMSSW release setup
+# working with LHAPDF6 only
+PDFname        = NNPDF30_nnlo_as_0118
```
perrozzi: this should be NNPDF31_nnlo_hessian_pdfas (306000)

bin/Phantom/submit_phantom.py
```diff
@@ -0,0 +1,468 @@
...
+def modifySubmitfileIntelCompiler (submitfilename, pdfgridfolder, phantompdflib):
+    submitfile = open (submitfilename, 'read')
+    lines = submitfile.readlines ()
+    submitfile.close ()
+    submitfile = open (submitfilename, 'write')
+    submitfile.write ('#!/bin/bash\n\n')
+    submitfile.write ('source /afs/cern.ch/sw/IntelSoftware/linux/setup.sh intel64\n')
+    submitfile.write ('source /afs/cern.ch/sw/IntelSoftware/linux/x86_64/xe2016/compilers_and_libraries/linux/bin/compilervars.csh intel64\n')
+    submitfile.write ('export LHAPDF=' + pdfgridfolder + '\n')
+    submitfile.write ('export PDFLIBDIR=/afs/cern.ch/work/b/ballest/public/phantom/LHAPDF-6.1.5_work/lib\n')
```
perrozzi: Why private path?

bin/Phantom/cards/examples/phantomProduction_example.cfg
govoni: Ciao Luca,
I prefer not to adhere to that convention, since it's not given that I will compile a version for slc6_amd64_gcc481, while it's sure there will be a phantom generator, so it makes more sense to me to have a phantom folder containing the precompiled versions which exist in CMS

bin/Phantom/cards/examples/phantomProduction_example.cfg
govoni: Ok, that's not a problem, this is just the example for testing matters. Whatever is the release I will use for the test in genproduction (and then for the actual production) I will put it there

bin/Phantom/cards/examples/phantomProduction_example.cfg
govoni: The source name is the one I get from the authors, this is why I did not change it

bin/Phantom/cards/examples/phantomProduction_example.cfg
govoni: ok

bin/Phantom/submit_phantom.py
govoni: that's for old versions of phantom, which where compiled by the authors with an Intel compiler. The last version I am using (which is also the first one for the wmLHE production) is compiled with gFortran: if you check the script, you will see that the function for the intel compiler is not called in the main

bin/Phantom/cards/examples/phantomProduction_example.cfg
govoni: (how do I change something before the merge, in this PR? do I have to delete the PR and make a new one, or what???)

bin/Phantom/submit_phantom.py
govoni: (I leave the old function for the smart user who might need to use an old release and would like to profit of the scripts anyhow)

bin/Phantom/cards/examples/phantomProduction_example.cfg
govoni: cool, I managed to ;) the pdfs and cmssw release are in the cfg file

bin/Phantom/submit_phantom.py
perrozzi: I don't refer to the Intel compiler but to the PDFLIBDIR.
Since AFS is not accessible in production this can create problems

bin/Phantom/cards/examples/phantomProduction_example.cfg
perrozzi: I suggest to keep the CMSSW version "fixed" once for all and use the trick in runcmsgrid_phantom.sh similarly to [POWHEG](https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/runcmsgrid_powheg.sh#L31-L60) to override the CMSSW version in production. This way you can use the same compiled version for 2016 and 2017 MC productions, for instance

bin/Phantom/cards/examples/phantomProduction_example.cfg
perrozzi: ok for the folder convention. if you want to keep the source tarball with the same name, please put it under a folder called "src"

bin/Phantom/cards/examples/phantomProduction_example.cfg
perrozzi: +1

bin/Phantom/submit_phantom.py
govoni: I understood, the two things are linked and this function is NOT used in the main, so it will not create any problems by construction. The function remains there for any users who wants to produce LHE files or calculate cross sections with the old versions, privately, using the same input cfg file as the new configuration.

bin/Phantom/cards/examples/phantomProduction_example.cfg
govoni: ok, I can add to the cfg file that it has not to be changed. The script is already compliant with that, for it sources the CMSSW as indicated in the config file before generating the events.

bin/Phantom/cards/examples/phantomProduction_example.cfg
govoni: ok, let me test the modifications, there's now two folders: source and compiled.

govoni: it works, PR updated

perrozzi:
please remove bin/Phantom/cards/examples/.gitkeep

govoni:
should be gone, with other residual vestigial things

perrozzi:
seems good to go, let me know if you have other commits or want to merge

govoni:
I think this is a reasonable first question. New things will come, but I would like to test it in the official production soon, so I'd like it to be merged.

PR #1325, opened 2017-09-19
Script to test compilation of POWHEG processes

perrozzi:
Test recursively all (or a subset) of the POWHEG processes.
Useful when a new source tarball is created, before to distribute it


perrozzi:
@covarell @yuanchao fyi

PR #1334, opened 2017-09-22
gridpack_generation: Fix sed expression.

AndreasAlbert:
The sed expression to remove any lines containing "$DEFAULT_PDF_MEMBERS"
from run_cards in case of LO does now handle additional spaces between around the
equal sign.

This seems like nitpicking, but please note that the example on the twiki [1] actually has an extra space, so the old replacement failed for this example.

[1] https://twiki.cern.ch/twiki/bin/view/CMS/QuickGuideMadGraph5aMCatNLO#PDF_Choice_for_2017_production

### DIFF
--- a/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
+++ b/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
@@ -492,7 +492,7 @@ if grep -q -e "\$DEFAULT_PDF_SETS" -e "\$DEFAULT_PDF_MEMBERS" $CARDSDIR/${name}_
             # 4F PDF
             sed "s/\$DEFAULT_PDF_SETS/320900/" $CARDSDIR/${name}_run_card.dat > ./Cards/run_card.dat
         fi
-        sed -i "s/ *\$DEFAULT_PDF_MEMBERS = .*//" ./Cards/run_card.dat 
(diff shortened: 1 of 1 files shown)


perrozzi:
Hi, thanks for reporting this. I fixed right away the example. However, this looks too delicate to me.
I would even remove the required extra spaces and just use something like
> sed -i "s/ *\$DEFAULT_PDF_MEMBERS*=.*//" ./Cards/run_card.dat
can you check and update your PR?
Thanks a lot

AndreasAlbert:
I do not follow.

In the way the code is written right now, the spaces are not *required*. The replacement simply works no matter whether there are 0, 1 or N spaces. I believe that to be consistent with the rest of the card syntax.

If I get your suggestion right, you want it explicitly work only for 0 spaces? That is quite non-standard as normal syntax of the card does not explicitly depend on the number of spaces in a line. Maybe I misunderstand.

Also, I your proposal is missing a "\" in front of the "$" sign, which is probably an accident.

AndreasAlbert:
If you put the little accents around your code fragment, I think it will not mess up the special characters: ``
(or click on the "code" environment icon: image )

AndreasAlbert:
Okay, I see your updated version and now I get what you want. You are missing a dot after MEMBERS though. The way you wrote it, the start does not work properly.

`sed -i "s/ *\$DEFAULT_PDF_MEMBERS.*=.*//" ./Cards/run_card.dat`

PR #1336, opened 2017-09-22
avoid Powheg job failure due to zero weights

agrohsje:
Removed check of zero weights in Powheg event generation scripts (runcmsgrid*.sh) as the problem can be traced back to PDFs that give zero. The check causes the whole job to fail and will thus lead to high job failure rates.

### DIFF
--- a/bin/Powheg/runcmsgrid_powheg.sh
+++ b/bin/Powheg/runcmsgrid_powheg.sh
@@ -156,7 +156,8 @@ xmllint --noout ${file}_final.lhe > /dev/null 2>&1; test $? -eq 0 || fail_exit "
 
 grep ">        NaN</wgt>" ${file}_final.lhe; test $? -ne 0 || fail_exit "Weights equal to NaN found, there must be a problem in the reweighting"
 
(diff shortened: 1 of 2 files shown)


bin/Powheg/runcmsgrid_powheg.sh
```diff
@@ -156,7 +156,8 @@ xmllint --noout ${file}_final.lhe > /dev/null 2>&1; test $? -eq 0 || fail_exit "
 
 grep ">        NaN</wgt>" ${file}_final.lhe; test $? -ne 0 || fail_exit "Weights equal to NaN found, there must be a problem in the reweighting"
 
-grep ">0.00000E+00</wgt>" ${file}_final.lhe; test $? -ne 0 || fail_exit "Weights equal to 0 found, there must be a problem in the reweighting"
+# uncomment to avoid problems in production as whole job would fail and problem traced back to neg. PDFs
```
perrozzi: uncomment --> comment ?

perrozzi:
@yuanchao @covarell

covarell:
Ok

PR #1337, opened 2017-09-22
Settings for TTTT NLO Fall17 production

jfernan2:


efeyazgan:
please cleanup commented lines

PR #1338, opened 2017-09-22
Settings for ttbar Powheg

jfernan2:
Preliminary settings, hdamp value still pending tuning

### DIFF
--- a/bin/Powheg/production/2017/TT_hdamp_NNPDF31_NNLO_dilepton/powheg.input
+++ b/bin/Powheg/production/2017/TT_hdamp_NNPDF31_NNLO_dilepton/powheg.input
@@ -1,0 +1,91 @@
+! TTbar production parameters
+!randomseed 352345 ! uncomment to set the random seed to a value of your choice.
+                   ! It generates the call RM48IN(352345,0,0) (see the RM48 manual).
+                   ! THIS MAY ONLY AFFECTS THE GENERATION OF POWHEG EVENTS!
+                   ! If POWHEG is interfaced to a shower MC, refer to the shower MC
+                   ! documentation to set its seed.
+
+!Heavy flavour production parameters
+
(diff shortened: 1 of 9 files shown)


perrozzi:
@jfernan2 is the PR intended to wait for the new hdamp values?

jfernan2:
Yes, we need it for final settings...

jfernan2:
hdamp value set to 237.8775 (i.e 1.379mtop)
Please sign

perrozzi:
Hi, there are a few "problems" with the naming conventions of this PR.
- The sub-folder for the center of mass energy is missing 
- a common folder called "TT" or similar can be added after 2017/ to avoid the proliferation of the number of files/folders in the main directory. 
- each folder currently contains a single file named poheg.input while the usual convention was to use what is now the directory name as filename.

More generally speaking, probably we should create a folder for each powheg process to avoid any source of confusion (@kdlong, @covarell , @yuanchao what do you think?)

perrozzi:
following https://hypernews.cern.ch/HyperNews/CMS/get/generators/3606.html
indeed can you please cleanup the datacards? there are a lot of commented lines, other than removing the last lines which are added automatically
https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/run_pwg.py#L861-L865

@agrohsjean @jfernan2

kdlong:
@perrozzi so the cards should be in production/2017/13TeV...? Is there any production that would qualify as both 2017 and not 13 TeV?

agrohsje:
@jfernan2 should I do it or are you already working on it ?

perrozzi:
@kdlong yes, for instance 5TeV pp run later this year. 
In any case 2017 referers more to the conditions than the data taking year.

jfernan2:
Sorry, the real PR is #1360
I will update there

PR #1339, opened 2017-09-22
For PDF checks done in run_pwg.py

yuanchao:
Now the check of PDF set in powheg.input for 2017 production defaults are done in run_pwg.py. Tests done locally. Also there is an update on the JHUGen.input handling to avoid creating an empty file.

### DIFF
--- a/bin/Powheg/run_pwg.py
+++ b/bin/Powheg/run_pwg.py
@@ -317,13 +317,6 @@ if [ -s ../${cardInput} ]; then
   cp -p ../${cardInput} powheg.input
 fi
 
-# same pdfs for the 2 protons?
-teststandpdf=`grep "lhans1" powheg.input | awk '{print $2}'`
-teststandpdf2=`grep "lhans2" powheg.input | awk '{print $2}'`
-if ! [[ $teststandpdf == $teststandpdf2 ]]; then
-    fail_exit "ERROR: PDF settings not equal for the 2 protons: ${teststandpdf} vs ${teststandpdf2}... Please check your datacard"
-fi
-
 # 5F
 is5FlavorScheme=1
 defaultPDF=306000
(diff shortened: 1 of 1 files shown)


bin/Powheg/run_pwg.py
```diff
@@ -1138,7 +1121,9 @@ def runhnnlo(folderName, njobs, QUEUE):
         prepareJob(tagName, '', '.')
 
         if not os.path.exists(args.inputTemplate) :
-            os.system('wget --quiet --no-check-certificate -N http://cms-project-generators.web.cern.ch/cms-project-generators/'+args.inputTemplate+' -O '+args.folderName+'/powheg.input')
+            m_ret = os.system('wget --quiet --no-check-certificate -N http://cms-project-generators.web.cern.ch/cms-project-generators/'+args.inputTemplate+' -O '+args.folderName+'/powheg.input')
```
perrozzi: question: these lines make it such that when no input file neither passed nor present, a default version is downloaded, right? Would it make sense to prevent that no datacard argument is passed (i.e. raise an exception if this happens), and instead add another argument to explicitly use the default cards (which are there usually for testing purpose)?

bin/Powheg/run_pwg.py
```diff
@@ -1148,16 +1133,53 @@ def runhnnlo(folderName, njobs, QUEUE):
...
         os.system('rm -rf JHUGen.input')
-        inputJHUGen = '/'.join(powInputName.split('/')[0:-1])+'/JHUGen.input'
+        inputJHUGen = args.inputJHUGen
+        if args.inputJHUGen == "":
+            inputJHUGen = '/'.join(powInputName.split('/')[0:-1])+'/JHUGen.input'
+
         if not os.path.exists(inputJHUGen) :
-            os.system('wget --quiet --no-check-certificate -N http://cms-project-generators.web.cern.ch/cms-project-generators/'+inputJHUGen+' -O '+args.folderName+'/JHUGen.input')
+            m_ret = os.system('wget --quiet --no-check-certificate -N http://cms-project-generators.web.cern.ch/cms-project-generators/'+inputJHUGen+' -O '+args.folderName+'/JHUGen.input')
+            if ((m_ret>>8) & 255) != 0 :
```
perrozzi: what is this doing?

perrozzi: ok, it avoids creating an empty file

bin/Powheg/run_pwg.py
```diff
@@ -1148,16 +1133,53 @@ def runhnnlo(folderName, njobs, QUEUE):
...
+            test_pdf2 = 0
+
+            default_pdf = "306000"  # for 5 flavours
+
+            if args.prcName=="ST_tch_4f" or args.prcName=="bbH" or args.prcName=="Wbb_dec" or args.prcName=="Wbbj" :
+                default_pdf = "320900"  # for 4 flavours
+
+            for line in open(args.folderName+'/powheg.input') :
+                if 'lhans1' in line and len(line.split(' ')) >= 2:
+                    test_pdf1 = line.split(' ')[1].strip()
```
perrozzi: these conditions 
```
test_pdf1 = line.split(' ')[1].strip()
test_pdf2 = line.split(' ')[1].strip()
```

are not sufficient. if the line is like
```
lhans1   306000
```
i.e. contains multiple spaces, it will not work!
I will merge then push a fix (i.e. simply use `split()` )

yuanchao: Ah, ok. I should use [-1] instead. The strip() is simply to get rid of possible new line character at the tail.

bin/Powheg/run_pwg.py
yuanchao: Right. Currently run_pwg.py will produce an empty JHUGen.input during the trails. This is to avoid it.

bin/Powheg/run_pwg.py
yuanchao: Ya, surely just use split() will prevent the issue.

Issue #1333, opened 2017-09-22
[mg] default lhaid list too long for NLO mode


AndreasAlbert:
In the NLO 5F case, we put a list of 37 IDs into the run_card by default. It seems madgraph cannot handle more than 30 (error message below). In the NLO 4F case, we only give 27, so it's no problem.
Does this work for anyone else?

Error message:
InvalidRunCard : Length of list for 'lhaid' too long: max is 30.

https://github.com/cms-sw/genproductions/blob/b2883cc58eeb31c03918a3b96b44996999362b11/bin/MadGraph5_aMCatNLO/gridpack_generation.sh#L480

perrozzi:
thanks for reprting this. I am afraid we will need to strip off a few IDs...

perrozzi:
cfr https://github.com/cms-sw/genproductions/pull/1346

kdlong:
Note that its a totally trivial change to make MadGraph support 37 PDF sets. The original default was 10.

https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/patches/0008-allow-for-reweighting-with-up-to-30-PDF-sets-for-NLO.patch#L39

Though I set the max to 30 rather than 100 or so because this seemed excessive, so I'd still be in favor of reducing the selection.

perrozzi:
can you simply set to 40 for the moment?

kdlong:
Sure

kdlong:
#1349

perrozzi:
PR merged. @AndreasAlbert please try again and let us know

perrozzi:
as described in https://github.com/cms-sw/genproductions/pull/1371#issuecomment-332699433
likely the last thing to do is to patch these files
```
Template/NLO/MCatNLO/include/HwU.inc:      parameter (max_wgts=1024)
Template/NLO/FixedOrderAnalysis/HwU.inc:      parameter (max_wgts=1024)
Template/NLO/SubProcesses/c_weight.inc:      parameter (max_contr=128,max_wgt=1024,max_iproc=100)
```
substituting 1024 with 1100 or so. This modification can be added to the existing patch
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/patches/0008-allow-for-reweighting-with-up-to-40-PDF-sets-for-NLO.patch

@qliphy could you please try this?

qliphy:
[...]

PR #1346, opened 2017-09-24
remove 7 pdf sets to comply with mg5_amc limit when NLO

perrozzi:
cfr
https://github.com/cms-sw/genproductions/issues/1333
https://github.com/cms-sw/genproductions/issues/1287#issuecomment-331470543


perrozzi:
useless, superseeded by https://github.com/cms-sw/genproductions/pull/1349

PR #1351, opened 2017-09-25
equivalent JHUGen cards for 2l2q and 2l2nu

hroskes:
similar to the 4l cards already there

### DIFF
--- a/bin/JHUGen/cards/decay/ZZ2l2nu_notaus.input
+++ b/bin/JHUGen/cards/decay/ZZ2l2nu_notaus.input
@@ -1,0 +1,1 @@
+DecayMode1=0 DecayMode2=3 Seed=SEED
--- a/bin/JHUGen/cards/decay/ZZ2l2nu_withtaus.input
+++ b/bin/JHUGen/cards/decay/ZZ2l2nu_withtaus.input
@@ -1,0 +1,1 @@
+DecayMode1=8 DecayMode2=3 Seed=SEED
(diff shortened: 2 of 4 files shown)


perrozzi:
hi, are these used for 2016 MC, 2017 MC or the distinction is irrelevant?

hroskes:
There are equivalent cards used in 2016, but here I collect them in one place to be used for both future 2016 and 2017.

hroskes:
@perozzi For example this card
https://github.com/cms-sw/genproductions/blob/pre2017/bin/Powheg/production/H_ZZ_2l2nu_withJHUGen_NNPDF30_13TeV/VBF_H/JHUGen_VBF_ZZ2l2nu_M400GeV.input
is equivalent to the new card ZZ2l2nu_notaus.input (many of the options there are the same as the defaults, so not needed), but I think it's simpler to maintain the basic ones in one place.

We're planning to use the cards primarily going forwards in 2017, but also for some samples that were missed in 2016 MC.

hroskes:
Hi, is there any objection to merging this?  I don't see anything likely to be controversial, but I'm happy to answer more questions if there are any.

perrozzi:
do all the JHU cards need to be updated after
https://github.com/cms-sw/genproductions/pull/1399
?!?

hroskes:
I don't think there are any backwards incompatibilities in the command line from v7.0.2 to v7.0.9.  (There are some new options, but none removed).

The only difference is that I put Seed=${rnum} in the script and removed it from the cards (no need to specify it multiple times).  That's already done in #1399.

PR #1352, opened 2017-09-25
update bbH, H->bb with narrow width

chayanit:
We have updated the fragment after discussed with Steve to get narrow width resonance and already used for official production in Moriond17 but never committed to github. For the future record, please merge this change. Thank you very much.

### DIFF
--- a/python/ThirteenTeV/SUSYGluGlu/SUSYGluGluToBBHToBB_M-100_TuneCUETP8M1_13TeV-pythia8_cfi.py
+++ b/python/ThirteenTeV/SUSYGluGlu/SUSYGluGluToBBHToBB_M-100_TuneCUETP8M1_13TeV-pythia8_cfi.py
@@ -7,19 +7,22 @@ generator = cms.EDFilter("Pythia8GeneratorFilter",
     filterEfficiency = cms.untracked.double(1.0),                                                                                
     pythiaHepMCVerbosity = cms.untracked.bool(False),                                                                            
     comEnergy = cms.double(13000.0),                                                                                             
-    crossSection = cms.untracked.double(518.3),                                                                                  
+    crossSection = cms.untracked.double(1.0),                                                                                  
     maxEventsToPrint = cms.untracked.int32(1),                                                                                   
     PythiaParameters = cms.PSet(                                                                                                 
             pythia8CommonSettingsBlock,                                                                                          
             pythia8CUEP8M1SettingsBlock, 
                     processParameters = cms.vstring(                                                                     
-                                                    'Higgs:useBSM = on',
-                                                    'HiggsBSM:gg2A3bbbar = on',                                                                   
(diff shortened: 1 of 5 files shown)


efeyazgan:
Could you please move them to https://github.com/cms-sw/genproductions/tree/pre2017 ?

perrozzi:
question: is this fragment meant to be used for Morionf17 or Moriond18 (aka MC 2017 v2 aka Fall17) ?

chayanit:
Hi Luca, these fragments won't be used for Fall17 as we plan to move to MG NLO but I need to commit this for a future record. 

Could you guide me how I can make PR to the tag?

Cheers,
Chayanit

perrozzi:
thanks for the clarification. unfortunately can't be made to the tag :-)
In principle the request could be merged as it is, although we are thinking to adopt a similar pre2017-2017 also for pythia only requests.
I have a more general question: I suppose that if the corrected fragments are uploaded only now, you used "hand-written" cfg in McM and didn't link to genproductions. if so, why? (just to understand)

chayanit:
ok hmm I have already closed this PR. Is it ok to re-open another one? 

Yeah, it was urgently needed during that time for ICHEP16 so I didn't manage to upload the fragments to github and used hand-written on McM. I found this way is quicker and more user-friendly than waiting until they are merged to get fragment tag. Sorry :-(

efeyazgan:
Yes, it should be ok to re-open another one.

perrozzi:
(for the record, this PR can't be re-opened because you deleted the branch `unknown repository` otherwise it would have been possible to re-open this one itself.
Concerning the fragment wvs hand-written cfg, I understand quicker (although without review errors can appear more often), why also user-friendly? we can try to make it more user-friendly if needed

chayanit:
Personally, I found only under /Thirteen sub-directory we already have thousands of pythia fragment and sometimes it's hard to find corresponding fragments, looking for hand-written fragment directly on McM (specific requests) and make changes on fragment box is more user-friendly but I understand it's not good for the record.

perrozzi:
Thanks for the feedback.
I fully agree that is a problem to have thousands of files in the same folder.
For this reason I spent quite some time trying to rationalize things
https://github.com/cms-sw/genproductions/tree/master/python/ThirteenTeV
and we will watch the new PR to always have a subfolder to avoid this problem.
Does it sound a good compromise/solution? Let us know if you have other suggestions or comments

perrozzi:
> I see it looks more promising now with subfolder but I think it will end up with too many folders according to processes again. How about subfolder with sub-groups (POG/PAG)? 

This is probably not efficient because different pog/pag can produce similar samples (for instance TTbar)

> Since all current fragments need to update with the new Tune, why don't you move everything to tags/pre2017 and restart with subfolder according to sub-group? Just a suggestion :-)

Yes this is something we are considering, thanks

PR #1355, opened 2017-09-26
Update run_pwg.py

intrepid42:
Fix call to make_rwl.py


perrozzi:
@yuanchao please triple-check the code before to commit, this bug has been introduced in https://github.com/cms-sw/genproductions/pull/1339

PR #1356, opened 2017-09-26
Qcd ht lo mlm

ShevchenkoRostyslav:
Nominal QCD HT cards for the gridpacks with updated PDFs


bin/MadGraph5_aMCatNLO/cards/production/2017/QCD_HT_LO_MLM/QCD_HT2000toInf/QCD_HT2000toInf_proc_card.dat
```diff
@@ -0,0 +1,17 @@
+import model sm-ckm_no_b_mass
+define p = g u c d s u~ c~ d~ s~
```
perrozzi: remove p and j definitions, remove duplicated lepton definitions

perrozzi: (p and j definitions are automatically taken into account when importing a given model)

perrozzi:
Please move the cards to
[...]/cards/production/2017/**QCD**/QCD_HT_LO_MLM/QCD_HT...dat

ShevchenkoRostyslav:
Thanks for your comments and explanations. Corresponded changes have been committed.

perrozzi:
PR looks ok to me, @efeyazgan ?

chayanit:
@perrozzi just my curiosity, is tau lepton normally included by default? or we have to define it if we want to include tau lepton/neutrino?

perrozzi:
Thanks for asking. I think is included, although I am not 100% sure :-)
If so, we could definitely avoid the re-definition in the datacards.
One can make a quick test and look at the log
@bendavid @kdlong and others might know by heart

chayanit:
Hi Luca, thanks for guiding so I checked the LHE file from this gridpack https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/2017/QCD_bEnriched_HT_LO_MLM/QCD_bEnriched_HT200to300/QCD_bEnriched_HT200to300_proc_card.dat and found the default definition of leptons/neutrinos is below. 
define l+ = e+ mu+
define l- = e- mu-
define vl = ve vm vt
define vl~ = ve~ vm~ vt~

kdlong:
I think it's normally left out by default and should be explicitly included. You can check with ```display multiparticles``` in the MadGraph command prompt.

By the way, should the path instead be 2017/13TeV/QCD...? At the moment the copyOldCardsTo2017.sh script is copying them without the 13TeV...

perrozzi:
yes please, 13TeV would be beneficial

bin/MadGraph5_aMCatNLO/cards/production/2017/QCD/QCD_HT_LO_MLM/QCD_HT1000to1500/QCD_HT1000to1500_proc_card.dat
```diff
@@ -0,0 +1,11 @@
+import model sm-ckm_no_b_mass
+define vl = ve vm vt
```
perrozzi: it seems to me that for these datacards the definition of the neutrinos and leptons is not needed, since only jets are produced

ShevchenkoRostyslav: Indeed, also the definition of neutrinos and leptons is provided by Madgraph: 
MG5_aMC>display multiparticles
Multiparticle labels:
all = g u c d s u~ c~ d~ s~ a ve vm vt e- mu- ve~ vm~ vt~ e+ mu+ t b t~ b~ z w+ h w- ta- ta+
l- = e- mu-
j = g u c d s u~ c~ d~ s~
vl = ve vm vt
l+ = e+ mu+
p = g u c d s u~ c~ d~ s~
vl~ = ve~ vm~ vt~

So, in any case this lines can be safely removed.

ShevchenkoRostyslav:
Should I move the QCD_bEnriched_HT_LO_MLM cards to 13TeV/QCD/ folder as well ?

efeyazgan:
Yes, please.

PR #1360, opened 2017-09-26
Moved TTbar settings to 13TeV folder

jfernan2:


perrozzi:
why are you adding the pwg-rwl.dat? It is created on the fly, automatically, for each gridpack

perrozzi:
Sorry for being so picky and thanks for your patience.
Since yours is basically the first commit with POWHEG datacards is good if we try to set a standard.
- I wonder if TT_hvq is the best way to name the folder. Unfortunately hvq can be confusing. However a convention like "human-readable"_"powheg-process" could be confusing becase "powheg-process" often contains an underscore
- I wonder if a directory with a single file is worth or is better to put all the files in the same folder

Any comments?

efeyazgan:
I don't have a strong opinion but I think TT_hvq is a convenient one since we understand immediately the ME without going into the config. Later on hopefully we will have other TT ME methods which we can keep separate.

perrozzi:
thanks, for me is good to go. @efeyazgan merge or suggest other modifications

PR #1362, opened 2017-09-26
patch for mg5_242 to solve the lhapdf621 boost issue

alberto-sanchez:
This should fix the boost and pdfset issue observed in mg5 2_4_2 and lhapdf 6.2.1

### DIFF
--- a/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
+++ b/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
@@ -207,7 +207,7 @@ if [ ! -d ${AFS_GEN_FOLDER}/${name}_gridpack ]; then
 
   LHAPDFINCLUDES=`$LHAPDFCONFIG --incdir`
   LHAPDFLIBS=`$LHAPDFCONFIG --libdir`
-  BOOSTINCLUDES=`scram tool tag boost INCLUDE`
+  export BOOSTINCLUDES=`scram tool tag boost INCLUDE`
 
   echo "set auto_update 0" > mgconfigscript
   echo "set automatic_html_opening False" >> mgconfigscript
--- a/bin/MadGraph5_aMCatNLO/patches/0015-boost-setpdf.patch
+++ b/bin/MadGraph5_aMCatNLO/patches/0015-boost-setpdf.patch
@@ -1,0 +1,45 @@
+diff --git a/Template/LO/Source/PDF/pdf_lhapdf6.cc b/Template/LO/Source/PDF/pdf_lhapdf6.cc
+index 35b34d5..77daf67 100644
+--- a/Template/LO/Source/PDF/pdf_lhapdf6.cc
++++ b/Template/LO/Source/PDF/pdf_lhapdf6.cc
+@@ -12,6 +12,11 @@
+ #include "LHAPDF/Version.h"
+ #include "LHAPDF/LHAGlue.h"
+ 
++#include <boost/shared_ptr.hpp>
++#include <boost/foreach.hpp>
++#include <boost/algorithm/string/case_conv.hpp>
++#include <cstring>
++
+ using namespace std;
+ 
+ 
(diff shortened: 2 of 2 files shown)


perrozzi:
Thanks Alberto. Out of curiosity, have you already tested it somehow?

alberto-sanchez:
@perrozzi  yes. At least I have run the examples

./gridpack_generation.sh wplustest_4f_LO cards/examples/wplustest_4f_LO 1nd
./gridpack_generation.sh wplustest_4f_NLO cards/examples/wplustest_4f_NLO 1nd

perrozzi:
great, very good to know. @kdlong please sign off to merge

kdlong:
@perrozzi it seems like the boost includes are taking from the local installation, so we should fix this, probably by just removing all boost dependencies in exchange for C++11 as proposed by @bendavid

perrozzi:
yes indeed. 
@alberto-sanchez are you still looking into this and can update the PR at some point (based on the example provided by josh) ? Otherwise let us know...

alberto-sanchez:
@perrozzi looks like adding the needed includes, helps

bendavid:
This looks ok.  Do we have a way to test that in case both CMSSW and system boost is available that the CMSSW one is used?

bendavid:
(and can you also make the corresponding pull request to the mg26x branch?)

kdlong:
Provided it's different on the system and in CMSSW (or testing 7_1_30 and 9_3_0, where it's different) this should work: http://www.boost.org/doc/libs/1_44_0/libs/config/doc/html/boost_config/boost_macro_reference.html#boost_config.boost_macro_reference.boost_informational_macros

perrozzi:
Not sure I understand: ready to be merged? Needs confirmation that the correct boost is called?

bendavid:
I would like to see a test with the method proposed by @kdlong before this is merged at least.

alberto-sanchez:
do you mean using 7_1_30 and 9_3_1?. at the moment I have run that on lxplus and in my vm(no boost) without problem.

perrozzi:
is this considered as sufficient? other tests to propose?

bendavid:
Alberto did you confirm that indeed different boost versions are used when running in 71x Vs 93x?

alberto-sanchez:
In lxplus, I have not found a way to do so. My debugging does not produce anything. In my VM I am sure, since no other boost is used. Still looking on this.

perrozzi:
I guess you can simply put a
echo $BOOST_VERSION
somewhere in the execution script, right?

kdlong:
I think the thing to do is add

```#include <boost/version.hpp

std::cerr << BOOST_VERSION;```

to pdf_lhapdf6.cc and check that the output changes with boost version (e.g. CMSSW version). I'm looking at this now, hope to have something shortly.

alberto-sanchez:
Well, I was trying to print the actually part where BOOST is used in the compilation of the PDFs.
Otherwise the correct version

kdlong:
Hi all, sorry this took me much longer than anticipated, but I was finally able to add some output statements to Source/PDF/pdf_lhapdf6.cc and confirm that I see the boost version of my local installation, of CMSSW_7_1_30, and of CMSSW_9_3_0 depending on if I apply the fix and what CMSSW version I run from. You can see the patch I made to test this in [1]. So I think this is working properly and is ready to merge.

One additional comment: when I tried to save the .f/.cc files and then edit these with the change, I couldn't get the Source directory to recompile due to a problem linking with LHAPDF. Is this expected? I guess there must be some environment variables set in gridpack_generation.sh which affect the linking but I couldn't figure out which. I guess this isn't an issue as the code isn't meant to be recompiled, but I found it pretty confusing and frustrating. Any ideas?

perrozzi:
Thanks a lot to everybody for the fast turnaround, merging this PR then!
I am not sure I understand the source problem, which environment variable were you looking at or searching for?

alberto-sanchez:
great to know!. If nothing else is needed here. I will produce the patch for MG 26x as suggested

perrozzi:
@alberto-sanchez sounds like a good plan, please proceed

PR #1371, opened 2017-09-27
Remove spaces in PDF list, update card copy script

kdlong:
1) Remove spaces between PDF and members lists, which MadGraph doesn't like
** NOTE: Moving to a more readable format with list of PDFs in a file is already in progress. Expected in a few days.
2) Update copyOldCardsTo2017.sh script to put cards in 13TeV subfolder
3) Move exisiting cards to 13 TeV subfolder


bin/MadGraph5_aMCatNLO/gridpack_generation.sh
```diff
@@ -478,11 +478,11 @@ if grep -q -e "\$DEFAULT_PDF_SETS" -e "\$DEFAULT_PDF_MEMBERS" $CARDSDIR/${name}_
...
-            sed -i "s/\$DEFAULT_PDF_MEMBERS/True,  False, False, False, False, False, False, False, False, True,  True,False,False,True, False,False,False,True, True, False,True, True, True, True, True, True, True, True, True, True, True, True,  False, False, False, False, False/" ./Cards/run_card.dat 
+            sed -i "s/\$DEFAULT_PDF_MEMBERS/True,False,False,False,False,False,False,False,False,True,True,False,False,True,False,False,False,True,True,False,True,True,True,True,True,True,True,True,True,True,True,True,False,False,False,False,False/" ./Cards/run_card.dat 
         else
             # 4F PDF
-                  sed "s/\$DEFAULT_PDF_SETS/320900,11082,13091,13191,13202,23100,23300,23490,23600,23790,25410,25510,25570,25605,25620,25710,25770,25805,25840,92000,320900,320500,260400,262400,263400,292000,292400/" $CARDSDIR/${name}_run_card.dat > ./Cards/run_card.dat
```
perrozzi: in fact, in powheg I added 306000 (as first dataset) as per Josh's suggestion
https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/make_rwl.py#L104

kdlong:
Just realized this is still failing for the 5F case. With the error

```ERROR too many weights in reweight_pdf        1025        1024```

My guess is there is a hard coded cap at 1000 weights. I'll try to track down how this can be increased but if anyone has a good idea to reduce the number to < 1000 that would be very nice

File: bin/MadGraph5_aMCatNLO/gridpack_generation.sh (hunk 1 above)
Comment (perrozzi) [reply bin/MadGraph5_aMCatNLO/gridpack_generation.sh]: no, just replace the first 320900 occurrence with 306000 and keep the second occurrence of 320900, instead of removing the latter

**Comment by perrozzi** [comment]
I tracked it down to these 3 instances:
```
Template/NLO/MCatNLO/include/HwU.inc:      parameter (max_wgts=1024)
Template/NLO/FixedOrderAnalysis/HwU.inc:      parameter (max_wgts=1024)
Template/NLO/SubProcesses/c_weight.inc:      parameter (max_contr=128,max_wgt=1024,max_iproc=100)
```
Currently we store 1071 weights for 5F, we could decide to remove one set like
MMHT2014nlo68clas118 (51)
or
HERAPDF20_NLO_EIG (29) + HERAPDF20_NLO_VAR (14)
I am sure somebody will complain...

File: bin/MadGraph5_aMCatNLO/gridpack_generation.sh [hunk 2]
```diff
@@ -478,20 +478,15 @@ if grep -q -e "\$DEFAULT_PDF_SETS" -e "\$DEFAULT_PDF_MEMBERS" $CARDSDIR/${name}_
...
-            sed -i "s/\$DEFAULT_PDF_MEMBERS/True,  True, False,False,False,True, True, True, True, True, True, True, True, True, True, True, True, True, True, True, True,  True,  True,  False, False, True,  False/" ./Cards/run_card.dat 
+                  sed "s/\$DEFAULT_PDF_SETS/320600,11082,13091,13191,13202,23100,23300,23490,23600,23790,25410,25510,25570,25605,25620,25710,25770,25805,25840,92000,320900,320500,260400,262400,263400,292000,292400/" $CARDSDIR/${name}_run_card.dat > ./Cards/run_card.dat
+            sed -i "s/\$DEFAULT_PDF_MEMBERS/True,True,False,False,False,True,True,True,True,True,True,True,True,True,True,True,True,True,True,True,True,True,True,False,False,True,False/" ./Cards/run_card.dat 
         fi
     elif [ "$isnlo" -eq "0" ]; then
-        if [ $is5FlavorScheme -eq 1 ]; then
-            # 5F PDF
+        # Same central PDF for both 4 and 5F, variations added at runtime
```
perrozzi: I am afraid you misunderstood, read above

kdlong:
I should also point out that this is still failing for the 5F case with the "too many weights" error. I thought I had found the fix but apparently not.

kdlong:
@bendavid do you have any idea on this?

bin/MadGraph5_aMCatNLO/gridpack_generation.sh
perrozzi: this is not correct, the central pdf will ebe 4f.
in the original version, for the 4f weights, just replace the first 320900 occurrence with 306000 and keep the second occurrence of 320900, instead of removing the latter

perrozzi:
@kdlong your first error log was pointing to 1024 weights, 
so I suspect you need to change also these lines as I wrote here
https://github.com/cms-sw/genproductions/pull/1371#issuecomment-332699433

```
Template/NLO/MCatNLO/include/HwU.inc:      parameter (max_wgts=1024)
Template/NLO/FixedOrderAnalysis/HwU.inc:      parameter (max_wgts=1024)
Template/NLO/SubProcesses/c_weight.inc:      parameter (max_contr=128,max_wgt=1024,max_iproc=100)
```

bin/MadGraph5_aMCatNLO/gridpack_generation.sh
kdlong: Still not clear to me. Can you just list what you want the weights to be or make the change?

kdlong:
Ok, good suggestion. I can't check and validate this at the moment. I would still prefer to remove ~25 weights so that the master version is working.

bin/MadGraph5_aMCatNLO/gridpack_generation.sh
perrozzi: please revert to the original format

kdlong: I tried again.

perrozzi:
seems ~ok now, what about the number of pdf?

perrozzi:
Not yet finished but merging this in the meanwhile

PR #1372, opened 2017-09-28
resolving issues with lhapdf and boost in mg26x

alberto-sanchez:
it has been tested with LO. and making sure the correct boost is used


perrozzi:
thanks a lot. let's wait for @bendavid and/or @kdlong, @qliphy to comment, in case, before to merge

bendavid:
I tested this for NLO as well and it looks fine, so I think it can be merged.

PR #1332, opened 2017-09-21
Single top t-channel 4 FNS for 2017 MC production

Andrej-CMS:
please check the powheg input files for the 2017MC production

PR #1376, opened 2017-09-29
Powheg cards for single top tW 5f inclusive

ckmackay:
Please could you check these are correct for the Fall17 production? Thanks

### DIFF
--- a/bin/Powheg/production/2017/13TeV/st_tW_5f_dr_ckm_13TeV/ST_tW_DR_5f_ckm_antitop_13TeV.input
+++ b/bin/Powheg/production/2017/13TeV/st_tW_5f_dr_ckm_13TeV/ST_tW_DR_5f_ckm_antitop_13TeV.input
@@ -1,0 +1,64 @@
+! ST-wtchannel production parameters
+
+! GENERAL POWHEG PARAMETERS
+
+numevts NEVENTS    ! number of events to be generated
+ih1   1           ! hadron 1 (1 for protons, -1 for antiprotons)
+ih2   1           ! hadron 2 (1 for protons, -1 for antiprotons)
+ebeam1 6500d0     ! energy of beam 1
+ebeam2 6500d0     ! energy of beam 2
+
+iseed    SEED    ! initialize random number sequence 
+
+! To be set only if using LHA pdfs
+lhans1 306000      ! pdf set for hadron 1 (LHA numbering)
+lhans2 306000      ! pdf set for hadron 2 (LHA numbering)	
+
+! Parameters to allow or not the use of stored data
+use-old-grid    1 ! if 1 use old grid if file pwggrids.dat is present (<> 1 regenerate)
+use-old-ubound  1 ! if 1 use norm of upper bounding function stored in pwgubound.dat, if present; <> 1 regenerate
+
+ncall1 50000   ! number of calls for initializing the integration grid
+itmx1    5     ! number of iterations for initializing the integration grid
+ncall2 50000   ! number of calls for computing the integral and finding upper bound
+itmx2    5     ! number of iterations for computing the integral and finding upper bound
+foldcsi   1    ! number of folds on csi integration
+foldy     1    ! number of folds on  y  integration
+foldphi   1    ! number of folds on phi integration
+nubound 50000  ! number of bbarra calls to setup norm of upper bounding function
+xupbound 2d0   ! increase upper bound for radiation generation
+
+
+withdamp       1
+hdamp 172.5   
+
+
+! mandatory production parameters
+ttype       -1          ! 1 for t, -1 for tbar
+
+
+! mandatory parameters used in decay generation
+topdecaymode 11111   ! decay mode: the 5 digits correspond to the following
+                     ! top-decay channels (l,mu,tau,u,c) 
+                     ! 0 means close, 1 open
+wdecaymode 11111     ! decay mode: the 5 digits correspond to the following
+                     ! primary-w-decay channels (l,mu,tau,u,c) 
+                     ! 0 means close, 1 open
+tdec/elbranching 0.108  ! W electronic branching fraction
+
+! optional production parameters 
+! (defaults defined in init_couplings.f)
+topmass      172.5
+wmass        80.419
+sthw2        0.2226459
+alphaem_inv  132.3384323
+
+CKM_Vud   0.97462
+CKM_Vus   0.22530
+CKM_Vub   0.00338
+CKM_Vcd   0.22530
+CKM_Vcs   0.97462
+CKM_Vcb   0.04101
+CKM_Vtd   0.00862
+CKM_Vts   0.04101
+CKM_Vtb   1.0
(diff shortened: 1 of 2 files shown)


efeyazgan:
Here https://github.com/cms-sw/genproductions/pull/1332/files for single top t-channel, @Andrey-CMS changed the CKM matrix elements to PDG values. Please check with @jfernan2 or analyzers.

agrohsje:
This is using v1 or am I mistaken ?

efeyazgan:
Is there a v2 for this?

agrohsje:
v2 is what I was discussing with Emanuele in the thread with ATLAS. There is the open issue of the transverse b momentum that he wanted to follow up: http://www.desy.de/~agrohsje/hidden/validate_tW_top/MC_SingleTop_Truth/index.html  . Just wondering if we want to go ahead now or wait a little bit ?

ckmackay:
Thanks, yes there is now a V2, the old cards must have been V1. I'll change them.
Kate

ckmackay:
BVT have an urgent request for tW, I'm not sure how long they can wait. Efe do you know?

efeyazgan:
I don't. Please directly check with them. thanks.

ckmackay:
Actually what do I need to change in the cards for V2? I was just using the normal genproductions from github which will be V2 and it ran fine. Do I need to change the cards in some way?
From: Catherine Mackay
Sent: 29 September 2017 14:55
To: cms-sw/genproductions
Subject: RE: [cms-sw/genproductions] Powheg cards for single top tW 5f inclusive (#1376)

Thanks, yes there is now a V2, the old cards must have been V1. I'll change them.
Kate

agrohsje:
Yes. Maybe best. I mean it would be a little bit of a waste if we could get it in 2 weeks. But ok you never know maybe it is not 2 weeks. At least Emanuele promised yesterday to follow up asap.

agrohsje:
I checked with Andrej, tW is 16 Mio. We could also redo them once v2 is validated? Efe, what do you think?

agrohsje:
@ckmackay The reweighting info is automatically added by the run script. I see that Javi has
in https://github.com/cms-sw/genproductions/tree/master/bin/Powheg/production/2017/13TeV/TT_hvq
pdfreweight 1
storeinfo_rwgt 1 ! store weight information
but as far as I understand from the Powheg documentation this is obsolete. pdfreweight would allow you to store id,x,... to calculate reweighting afterwards and storeinfo_rwgt should be superseeded when using the new weighting scheme 

@efeyazgan  @perrozzi  what did we agree on about withnegweights 1 ? 
I thought that we wanted to turn them on in general? Maybe I am misremembering?

efeyazgan:
We agreed to have it as "withnegweights 1".

ckmackay:
@agrohsje The twiki still says to include
pdfreweight 1
storeinfo_rwgt 1 ! store weight information
I'll change the withnegweights to 1 and check about the values of the CKM matrix.
Thanks

agrohsje:
@yuanchao @covarell can one of you confirm that 
pdfreweight 1 
storeinfo_rwgt 1
is obsolete now. just to be sure.

perrozzi:
Ciao, I might be wrong but don't think it's obsolete. A few other parameters are added to use the new weight format
https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/run_pwg.py#L860-L865

agrohsje:
I quickly checked standalone production for 10 events with and without these two options:  
diff pwgevents.lhe  kept.lhe  | more 
> pdfreweight 1
> storeinfo_rwgt 1    ! store weight information
> #pdf  21 21 0.83394365E-02 0.10212137E+00 0.18946452E+03 0.94170749E+01 0.79018417E+00
> #pdf  21 21 0.83394365E-02 0.10212137E+00 0.18946452E+03 0.94170749E+01 0.79018417E+00
So it is indeed just this pdf line that we don't want to use because of the born-only reweighting.

covarell:
Indeed this is the only difference.

I remember once someone from sintheta w analysis  (aleko k..? ) who asked us to keep this line anyway...

agrohsje:
Despite that we have the proper re-weighting or despite that the info is not useful as the results are born-only and wrong? The later was my motivation to remove it so that nobody accidentally makes use of it.

jfernan2:
I may comment for ttbar, since production has already been launched with :
pdfreweight 0
storeinfo_rwgt 1    ! store weight information
and
withnegweights 0    ! default 0

I have checked that:
- pdfreweight and storeinfo have no effect since the fast reweighting options we set later override them

- withnegweights: my understanding was that we wanted ONLY positive weights, to have full valid useful stats and since the negative weight fraction is very small for ttbar. Anyway the option seems to be ignored and the pwg-stat.dat output yields in both cases:
negative weight fraction:   3.9370622427248548E-003

However, despite this, the weights in cmsgrid_final.lhe are different, and the nominal weight is NOT unity in the negative weights option case (there's a cross-section normalization), and we don't want that at this point:

neg< withnegweights 1    ! default 0
pos> withnegweights 0    ! default 0
neg<      2212     2212  6.50000E+03  6.50000E+03      0      0 306000 306000     -4      1
pos>      2212     2212  6.50000E+03  6.50000E+03      0      0 306000 306000 3      1
neg<      13 320000  8.07759E+00  5.72461E+01 -1.00000E+00  1.33889E-01
pos>      13 320000  1.00000E+00  5.72461E+01 -1.00000E+00  1.33889E-01
neg < <wgt id='1001'>0.80776E+01</wgt>
pos> <wgt id='1001'>0.10000E+01</wgt>
and following weights therein...

So I'd rather prefer to leave ttbar as it is (no negative weights, i.e. no cross-section normalization) since we are normalizing to NNLO cross-section anyway, we don't want a tedious sample as aMC@NLO, and the difference is apparently non-existing.

All outputs and gridpacks of this test are on:  /afs/cern.ch/user/j/jfernan/public/ttGenFall17/CMSSW_9_3_0/src/genproductions/bin/Powheg/SettingsCheck
Thanks

perrozzi:
Generally speaking, I have a different idea, as the negative weights must be included and we know that they can be used as +/-1 and one renormalizes anyway. 
In this particular case the fraction is negligible, but is not always the case so the analysis frameworks must handle negative weights properly....
concerning pdfreweight I would also suggest to keep it on as I can't imagine how somebody could use in first principle and use it wrongly unless is really needed and one knows what is doing. Am I missing anything?
Ciao
Luca

agrohsje:
@perrozzi I agree with what you wrote about neg. weights and also about hvq where it seems practically meaningless. For the PDFs, I am not sure. Do people really know the x1,x2 problem. It was heavily used in the past. On the other hand, who could need more PDFs than the 1000 we have? The people that shouldn't use just x1/x2.

covarell:
I agree pdf reweighting is obsolete. For me it is OK to drop it.

efeyazgan:
So, what is the conclusion in general? Do we ask all to have
pdfreweight 0
storeinfo_rwgt 0    ! store weight information
withnegweights 1    ! default 0
?

covarell:
It is ok for me, I will do a PR (also including Paolo Nason's VV fix)

efeyazgan:
thanks

perrozzi:
is the PR ready then?

ckmackay:
I've added the following to both cards as agreed in the discussion above:
pdfreweight 0
storeinfo_rwgt 0 ! store weight information
withnegweights 1 ! default 0
They should be ready for PR now, thanks.

perrozzi:
I assume is ready to be merged, @efeyazgan and others what do you think?

efeyazgan:
BTW has this ever been discussed with single top people that we use hdamp 172.5 for single top samples and not the value optimized from ttbar (even when we switched to the new tune)? @jfernan2, any thoughts?

jfernan2:
Sorry for the delay, I missed this thread: we have not discussed this within the ST crew, however for the moment we should stick to that damping value of 172.5 as for Moriond17 production, since the new (and old) hdamp tuning has been made with ttbar and ST is not sensitive to extra njet distribution used for the tuning

PR #1381, opened 2017-10-01
Fix5 flavor cms connect

kdlong:
The problem arose from the fact that CMS connect divides the codegen and integration step into two separate runs. In this case the flavor scheme should be read from the original codegen log file.

Also changed LO example cards to use default PDF sets.


kdlong:
Realized I made some mistakes in testing, not yet ready to merge.

PR #1383, opened 2017-10-02
GJets HT-binned MLM

yiiyama:
Cards for GJets_HT-* and GJets_DR-0p4_HT-*


efeyazgan:
please remove p and j definitions

yiiyama:
OK, I removed all p and j definitions - hopefully that's what you meant.

efeyazgan:
Yes, thanks. Could you please also remove the multiple lepton definitions (keep the one you would use)?

yiiyama:
Sure! In fact the proc cards only have
p p > a j {j} {j} {j}
so I got rid of all 'define's.

PR #1384, opened 2017-10-02
Moving towards more readable PDF weight format

kdlong:
Starting to setup a common interface between MadGraph5_aMC@NLO LO and NLO and POWHEG so all PDF sets can be read from a common, clearly readable place before being converted into the favored format of the generator by a specific module.

~done for MadGraph NLO, central set done for LO, should be easily extendable to POWHEG.

Not yet ready for merge but adding here so it can be easily referenced.


perrozzi:
this PR has now conflict

kdlong:
I think we should just revisit this in MG 2.6.0, where it's already merged

kdlong:
No, it's already implemented in 2.6.0 so we'll just wait for that to move to this setup

PR #1386, opened 2017-10-02
Python cff for ttbb_4FS_OpenLoops_13TeV_NNPDF30

Andrej-CMS:
Please check the cff that corresponds to the ttbb_4FS_OpenLoops_13TeV_NNPDF30 gridpack


efeyazgan:
@pmillet could you please take a look at this as well?

pmillet:
Since sherpacks are bound to a gcc version, I think it would be wise to have the scram architecture somwhere in the path (as before).

I can't comment on the physics settings of the card. It does not have pdf/scale variations (which is of course ok if this is intentional). If it was created with the newest Sherpa version the path should better point to 2.2.4 instead of 2.2.2.

Andrej-CMS:
Hello @pmillet ,
I was under the impression that the weights in sherpa production do not work for now. This card is basically a copy of the Moriond17 prodcution with updated b quark mass, to make it match the definitions in other cards (4.8 GeV).

This sample is mainly used by ttH and ttbb for comparison with the nominal Powheg+Pythia8 samples. Usually, only the mean values get compared. The systematic uncertainty on the prediction gets derived from the parton shower and matrix element uncertainties of the nominal sample.

I will update the path, since the gridpack was uploaded in an updated folder.

Please let me know if you have any further question and comments!

Kind regards,
Andrej

Andrej-CMS:
please check again the naming and the link to the gridpack

pmillet:
@Andrej-CMS -
Looks ok to me. Gen conveners should check if they like the naming scheme of the path.
Concerning the weights: They should work in general.

efeyazgan:
Right, the path. 
Could you please change the path to genproductions/bin/Sherpa/production/2017/13TeV/
?

PR #1393, opened 2017-10-05
Update POWHEG version and enforce pdf weights off/negative weights on

covarell:


bin/Powheg/run_pwg.py
```diff
@@ -848,15 +848,21 @@ def createTarBall(parstage, folderName, prcName, keepTop, seed, scriptName) :
...
 sed -i "s/^xgriditeration/#xgriditeration 1/g" powheg.input
 
+# turn off obsolete stuff
+grep -q "pdfreweight" powheg.input; test $? -eq 0 || printf "\\n\\npdfreweight 0\\n" >> powheg.input
+grep -q "storeinfo_rwgt" powheg.input; test $? -eq 0 || printf "\\nstoreinfo_rwgt 0\\n" >> powheg.input
+grep -q "withnegweights" powheg.input; test $? -eq 0 || printf "\\nwithnegweights 1\\n" >> powheg.input
+  
+sed -i "s/^pdfreweight.*/#pdfreweight 0/g" powheg.input
+sed -i "s/^storeinfo_rwgt.*/#storeinfo_rwgt 0/g" powheg.input
+sed -i "s/^withnegweights/#withnegweights 1/g" powheg.input
```
perrozzi: I sincerely hope that people will always use the correct normalization... this enforcement needs to be explcitly clarified in the twiki unless already present/updated

covarell:
Hi Luca, I understand your concern only partially... 
- if the fraction of negative weights for a process is not negligible, then they must be activated, therefore people will have to know how to normalize (even in previous campaigns)
- if the fraction of negative weights for a process is negligible, then it means that incorrect normalization will bring to a negligible mistake :-)

perrozzi:
Ciao Roberto,
I always hear from TOP that they want to keep the withnegweights to 0 because is negligible and the sample is easier to handle.
But if a framework uses other samples, some of which feature negative weights, needs to support this feature properly and it should be transparent to the user.
So I don't understand the statement that the sample is easier to handle and the reason to put withnegweights to 0.
I think the aim should be to minimize the risk of making mistakes, so always enforcing withnegweights to 1 in my opinion is the way to go, even if the fraction is negligible (the worst it can happens is a negligible mistake as you say :) )

covarell:
The PR *is* putting withnegweights to 1.

perrozzi:
yes indeed

PR #1398, opened 2017-10-09
Samples for 14TeV PhaseII TDR HGCal production

jfernan2:


perrozzi:
I believe the main path for these cards should be
bin/MadGraph5_aMCatNLO/cards/production/pre2017/14TeV

jfernan2:
If you don't mind, I made the change in another PR since there were more files:
https://github.com/cms-sw/genproductions/pull/1403
I believe the same applies to a lot of settings in 2017 folder which are not in the 13TeV folder, so we don't know the ecm.
Thanks

perrozzi:
thank you

PR #1399, opened 2017-10-09
JHUGen v7.0.9

hroskes:
Upgrading to JHUGen v7.0.9.

Also clean up the script (for both plain JHUGen and POWHEG+JHUGen) so that the seed is input in the script and doesn't have to be specified in every input card.

And prepare for some process which require MCFM libraries linked.


perrozzi:
my question was more of the kind "will the seed be specified multiple times and in case, is this a problem?"

hroskes:
Ah, I see.

Specifying the seed multiple times in general is not a problem (it just takes the second), but there will be a bug here if the old cards are used as is, since it will try to parse Seed=SEED and fail.

I can add an line for backwards compatibility if you'd like (something along the lines of sed "s/Seed=SEED//g")

perrozzi:
yes please

PR #1289, opened 2017-08-17
14 TeV datacards for Single tt h wide

dmajumder:
Update from [PR#1209](https://github.com/cms-sw/genproductions/pull/1209) for upgrade studies

PR #1408, opened 2017-10-12
Ttot h 14 te v

skhalil:
@perrozzi and @dmajumder, I am creating this new PR compare to #1289.


perrozzi:
please merge the two branches and use only #1289

skhalil:
Please accept this branch and Devdatta should cancel the PR #1289 .

perrozzi:
please move all the folders in a common folder called tHbj_TL or what you prefer.

skhalil:
Done. Please check again.

perrozzi:
I am afraid you copied and not moved all the datacards.
However, since this PR consists of 300 files, probably is worth doing this: assuming that you created them all with a script, would it make sense to upload only a representative example and the script used to produce them all?
Sorry for the request but we are trying not to flood the repository with too many files, otherwise git will slow down enormously

skhalil:
I moved all of them into respective parent directories. Each sample directory contains 5 files, so yes they are many. I copied the templates from Alexandra, e.g: https://github.com/acarvalh/singleVLQ_wide/tree/master/tHbj_TL and then using a python script, produced them. Do you just like to commit the python script and two template folders as tHbj_TL and tHbj_TR ? I have to then add instructions to run the script for you, but I don't think so that its a proper why to send request to gen group and git is pretty fast to accommodate all the card files as we have for all 3 TeV samples. Why found 14TeV will make a big difference?

PR #1409, opened 2017-10-12
replace automatically maxjetflavor to avoid inconsistencies

perrozzi:


perrozzi:
@qliphy @kdlong I have checked that it works properly, can you please crosscheck briefly?

qliphy:
@perrozzi   Tests with wplustest_LO_4/5f work well.  For NLO, there is no 'maxjetflavor' in run_card.dat so this might not be needed.

perrozzi:
updated, thanks a lot

bendavid:
Mm, I believe this setting should also be there for NLO.

perrozzi:
I interpret the fact that is not there as being set automatically, isn't it?
would you prefer to have an explicit setting anyway, as for LO?

bendavid:
From which card was it drawn the conclusion that it's not present?

qliphy:
@bendavid    I checked [1] contains "maxjetflavor" while [1] not.     

I then checked with standalone MG242 and generated process like  
generate p p >  w+ [QCD]
or the FxFx one
generate p p >  w+ [QCD]
add process p p > w+ j [QCD]
In the output run_card.dat, I don't find "maxjetflavor", so I think it is not needed.  And in [1], it might just copied from elsewhere?

[1] https://github.com/cms-sw/genproductions/blob/mg26x/bin/MadGraph5_aMCatNLO/cards/examples/dyellell012j_4f_NLO_FXFX/dyellell012j_4f_NLO_FXFX_run_card.dat

[2] https://github.com/cms-sw/genproductions/blob/mg26x/bin/MadGraph5_aMCatNLO/cards/examples/wplustest_5f_NLO/wplustest_5f_NLO_run_card.dat

bendavid:
It could be that the default is 5 and so it wasn't included in the 5f cards (to be double checked)

qliphy:
@bendavid   Yes, it seems the 4f w NLO card contains "maxjetflavor"

@perrozzi  So maybe it is more safe to include "maxjetflavor" for NLO case, although from standalone MG it seems not needed.  

[1] https://github.com/cms-sw/genproductions/blob/mg26x/bin/MadGraph5_aMCatNLO/cards/examples/wplustest_4f_NLO/wplustest_4f_NLO_run_card.dat

perrozzi:
ok then. question: shall I only replace it when it exists or also add it if is missing?
(same question is in principle valid for LO too)

qliphy:
@perrozzi     "also add it if is missing" sounds good to me.

perrozzi:
https://github.com/cms-sw/genproductions/pull/1418 (not tested yet, can you please do?)

qliphy:
@perrozzi    

" +      echo "${nFlavorScheme}\ =\ maxjetflavor" >> ./Cards/run_card.dat"
**should be**
" +      echo "${nFlavorScheme} = maxjetflavor" >> ./Cards/run_card.dat "

After correcting this,  tests with LO/NLO 4f/5f work well.

perrozzi:
thanks a lot, please check now

qliphy:
@perrozzi   Checked. It works now.

PR #1415, opened 2017-10-13
heavy neutrino samples

sihyunjeon:
Sorry I did the pull request to the master branch.
I expect no problem here maybe..?


perrozzi:
are these datacard meant to be used with which version of mg5_amc? 2.4.2 or 2.6 ?
also, I wonder if is worth it putting 600 files in the repository, differing for a single number. Assuming that you created them all with a script, would it make sense to upload only a representative example and the script used to produce them all?

sihyunjeon:
It would be used with 2.6
Then would upload the minimal number of repositories (I guess 8) and uploading the script be okay?

perrozzi:
that would be great, thanks

PR #1417, opened 2017-10-13
Updating boosted Z and inclusive (m>50GeV) DYJets cards for 2017 (MadGraph5_aMCatNLO)

afanfani:
Banner updated and neutrino defintions removed. However I didn't remove the proton and jet definitions because the parsing code complains without them.


perrozzi:
we need to fix the parsing code then, please remove jet and proton definitions :-)

afanfani:
Ok. p and j definitions removed.

PR #1419, opened 2017-10-15
Hadronizer settings for ttbb_4FS_ckm_amcatnlo_madspin_pythia8

Andrej-CMS:
please check the hadronizer settings


perrozzi:
please add "CP5" to the fragment name

perrozzi:
I see that now the old tune file is removed.
Would it be worth leaving it after having added CUEP8M1 to the old filename?

Issue #1420, opened 2017-10-15
gridpack generation problem while using CMSConnect


gourangakole:
Hello Experts
card I am using: 
https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/pre2017/14TeV/vh012j_5f_NLO_FXFX_M125_14TeV_VToAll

I see there is an error
++ tee /srv/vh012j_5f_NLO_FXFX_M125_14TeV_VToAll.log
gridpack_generation.sh: line 105: /dev/fd/62: No such file or directory
++ tee /srv/vh012j_5f_NLO_FXFX_M125_14TeV_VToAll.log

genproduction master upto PR #1405

I have used command like below:
nohup ./submit_cmsconnect_gridpack_generation.sh  <name of process card without _proc_card.dat> <folder containing cards relative to current location> > somefile.debug 2>&1 &

khurtado:
I think this is because singularity has issues with process substitution >()
Just replaced it with an alternative and made a PR in #1421, but I haven't tested it.

@gourangakole : Could you test this PR?

gourangakole:
@khurtado thanks a lot. I am running with the updated gridpack_generation.sh script for last couple of hours.
Till now I see its running fine.
Once my gridpack is done I shall update.

gourangakole:
To update this issue:
I see my gridpack generation is finished by at the end I see a gridpack but there is some error while running madspin:

INFO: Restrict model /stash2/user/gkole/genproductions/bin/MadGraph5_aMCatNLO/vh012j_5f_NLO_FXFX_M125_14TeV_VToAll/vh012j_5f_NLO_FXFX_M125_14TeV_VToAll_gri\
dpack/work/MG5_aMC_v2_4_2/models/loop_sm with file ../MG5_aMC_v2_4_2/models/loop_sm/restrict_ckm_no_b_mass.dat .
INFO: Run "set stdout_level DEBUG" before import for more information.
^[[1;31mError when reading /stash2/user/gkole/genproductions/bin/MadGraph5_aMCatNLO/vh012j_5f_NLO_FXFX_M125_14TeV_VToAll/vh012j_5f_NLO_FXFX_M125_14TeV_VToA\
ll_gridpack/work/processtmp/tmp9AH5Vd/temp_decay/SubProcesses/P0_wp_cdx/G1/results.dat^[[0m
^[[1;31mCommand "launch -n pilotrun" interrupted with error:
[...]

PR #1421, opened 2017-10-16
Update gridpack_generation.sh

khurtado:
Tier Sites with singularity don't like process substitution <() format, so replacing it with an alternative.

Should solve #1420, but it is still untested.


khurtado:
@kdlong, @perrozzi : I noticed this issue happens only when the CODEGEN step in CMS Connect gets to run on a Singularity enabled Site (integrate step runs fine on singularity sites). There weren't many singularity sites months ago, so I didn't notice (as this is only 1 job per card), but Sites are starting to implement singularity more and more, so this issue became noticeable now.

The solution above should fix it. stdout/stderr redirection to the logfile should work the same this way, the difference is that the process substitution method writes to the logfile AND prints the logfile into the screen stdout. This method won't print the logfile into the screen, but I don't think that matters for non-interactive jobs.
It would need to be pushed into the 26x branch too.

PR #1423, opened 2017-10-16
first try in sintax: the V decqys should be inclusive (not only-hadro…

acarvalh:
This should be HH > tau tau VV , where V is either  W/Z with SM branching ratios. 
V should decay inclusively (hadronic+leptonic), please check this sintax is correct


python/ThirteenTeV/Higgs/HH/ResonanceDecayFilter_example_HHTo2T2V_madgraph_pythia8_cff.py
```diff
@@ -0,0 +1,54 @@
...
+                         filterEfficiency = cms.untracked.double(1.0),
+                         pythiaHepMCVerbosity = cms.untracked.bool(False),
+                         comEnergy = cms.double(13000.),
+                         PythiaParameters = cms.PSet(
+        pythia8CommonSettingsBlock,
+        pythia8CUEP8M1SettingsBlock,
+        processParameters = cms.vstring(
+            '23:mMin = 0.05',
+            '23:onMode = off',
+            '23:onIfAny = 1 2 3 4 5 11 12 13 14 15 16', # inclusive
```
perrozzi: I wonder if there is a smarter switch for the inclusive decay, probably "Any"

perrozzi:
things look fairly ok to me, I would kindly ask for another pair of eyes to have a look before to merge

python/ThirteenTeV/Higgs/HH/ResonanceDecayFilter_example_HHTo2T2V_madgraph_pythia8_cff.py
acarvalh: I could even not specify, however I think I need to add this to be able to write the ResonanceDecayFilter:daughters

python/ThirteenTeV/Higgs/HH/ResonanceDecayFilter_example_HHTo2T2V_madgraph_pythia8_cff.py
```diff
@@ -0,0 +1,54 @@
...
+            '25:onIfMatch = 24 -24',
+            'ResonanceDecayFilter:filter = on',
+            'ResonanceDecayFilter:exclusive = on', #off: require at least the specified number of daughters, on: require exactly the specified number of daughters
+            'ResonanceDecayFilter:eMuAsEquivalent = off', #on: treat electrons and muons as equivalent
+            'ResonanceDecayFilter:eMuTauAsEquivalent = on', #on: treat electrons, muons , and taus as equivalent
+            'ResonanceDecayFilter:allNuAsEquivalent  = off', #on: treat all three neutrino flavours as equivalent
+            'ResonanceDecayFilter:udscAsEquivalent   = off', #on: treat udsc quarks as equivalent
+            'ResonanceDecayFilter:udscbAsEquivalent  = on',  #on: treat udscb quarks as equivalent
+            'ResonanceDecayFilter:mothers = 25,23', #list of mothers not specified -> count all particles in hard process+resonance decays (better to avoid specifying mothers when including leptons from the lhe in counting, since intermediate resonances are not gauranteed to appear in general
+            'ResonanceDecayFilter:daughters = 15,-15,1,1,1,1',
```
OlivierBondu: I'm not 100% sure of the setup you are proposing here, the daughters in this case are only leptons, which are not equivalent to quarks ? Have you checked the output of this setup on some events ?

acarvalh:
The part of inclusive VV decays for Higgs was fixed provided this pull request https://github.com/cms-sw/cmssw/pull/21132  
It was explicitly checked that specify the 23 particle as child does not constrain it to be on shell == this is equivalent to specify the full decay chain in the  @filtering.

PR #1425, opened 2017-10-16
Adding STARLIGHT to generator repository

kurtejung:
This commit adds the STARLIGHT generator (https://starlight.hepforge.org) to the toolkit for official MC generation.  The script is loosely based on the POWHEG gridpack setup script and downloads the latest STARLIGHT version from SVN, compiles, and generates events based on some user input.  Commands include:

Production Type (1-4, allowing for production of diphoton or coherent/incoherent vector meson production)
Production Channel (defined on the main site, showing the primary production EM channel)
Number of Events
Random Seed

More commands and interfaces may be added as needed, but this first draft of STARLIGHT should be sufficient for basic production (without PYTHIA/DPMJet showering)

Pinging some Heavy Ion participants: @mverwe @cmironov @gsfs


perrozzi:
ok so the idea is to produce lhe and convert to hepmc and then directly to root.
this seems to be suited for private production only.
is there any interest to use central production?

kurtejung:
The plan was to have central production as well.  Can you advise on what the script should output to conform with the official production chains?

Thanks

kurtejung:
Also, part of the problem is that STARLIGHT doesn't produce LHE output - just a nonstandard text list which must be converted to LHE or HepMC formats.  That's the primary reason for all the conversion.

gsfs:
There's a standard procedure in central production for generators that produce LHE output. Most of them, like powheg and madgraph, start from gridpacks (which are not created in official production) but in all cases you need an LHE producer. For the cases using gridpacks, it looks like this:

externalLHEProducer = cms.EDProducer("ExternalLHEProducer",
    args = cms.vstring('/cvmfs/cms.cern.ch/phys_generator/gridpacks/slc6_amd64_gcc530/8TeVpPb/Powheg/V2/TT_hdamp_pPb_CT14_EPPS16_8.16TeV_powheg_hvq.tgz'),
#    args = cms.vstring("PATH/TO/TARBALL/GRIDPACK.tgz"),
    nEvents = cms.untracked.uint32(1),
    numberOfParameters = cms.uint32(1),
    outputFile = cms.string('cmsgrid_final.lhe'),
    scriptName = cms.FileInPath('GeneratorInterface/LHEInterface/data/run_generic_tarball_cvmfs.sh')
)

There's an automated procedure for getting the gridpacks from user accounts into cvmfs for access by central production. So, for STARLIGHT, you would presumably produce the equivalent of the gridpack privately and then use a variation of the gridpack script to produce LHE in central production. From then on, it's all very standard what to do.

perrozzi:
ciao, a few comments:
- indeed the easiest (and currently only) way to get things in central production is through LHE.
- externalLHEProducer needs to be touched only when preparing the central request.
- Currently bin/Starlight/create_lhe_starlight.sh doesn't produce LHE but HepMC.
- The gridpack should be an archive containing the compiled program, the input datacard and a bash script called runcmsgrid.sh which is expected by the externalLHEProducer.
- The gridpack is usually created via a bash or python script (in powheg's case [run_pwg.py](https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/run_pwg.py)) that downloads the generator from the repository, duplicates/edit a pre-existing local version of runcmsgrid.sh (in powheg's case [runcmsgrid_powheg.sh](https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/runcmsgrid_powheg.sh) for instance), runs the phase space integration step whenever applicable, then packs the necessary files

kurtejung:
Thanks for the info.  The primary issue is that STARLIGHT doesn't output anything in a standard format - only a list of particle momenta and PDGs.  I found and validated a STARLIGHT -> HepMC converter which seems to work ok, but I can't seem to find a tool to convert it to a LHE format.  I've contacted the authors, but in the meantime is there a CMS tool to convert HepMC -> LHE?

perrozzi:
I am not aware of an "official" HepMC to LHE converter, unfortunately.
However, this might (is) not the only case where a direct injection of HepMC files in the system would be good to have in central production. This is not technically possible yet, but only few things are missing.
If you are interested we can have a presentation on this (HepMC in central production) during a GEN meeting

kurtejung:
I've updated this PR to use a gridpack-style format now.  The create_lhe_starlight.sh sets up the gridpack by grabbing the Starlight code from SVN, compiling and tarring, then the runcmsgrid_starlight.sh script can be passed to the McM framework for production.

This has been debugged locally, but it's ability to integrate properly with the McM framework has not - if anyone has some good ideas for how to do this, please let me know.

Thanks

kurtejung:
I've also taken a stab at a starlight -> LHE converter.  Starlight is very basic and does not provide any details on cross-section, global parameters, etc so most of the info is set to -1 by default, but it hadronizes properly through Pythia

perrozzi:
looks good to me, McM testing can be done

perrozzi:
I assume this can be merged, right?

kurtejung:
Sorry please hold for a couple more days - we're finalizing the coupled implementation of STARLIGHT+DPMJet

kurtejung:
Hi all - this should be able to be merged now. The scripts now support bare STARLIGHT and STARLIGHT+DPMJet production modes and can create gridpacks or produce events locally. I've checked the interface with the McM generation style and it seems to work well. Please let me know if there are remaining tasks before merging this PR. Thanks.

perrozzi:
Hi, I have 2 comments:
* I would remove dpmjet3.0-5.tar and put it under the standard web repository (path as soon as the twikis come back to life). I have to remember to squash merge to avoid this file being present in history
* in runcmsgrid_starlight.sh it is advisable to have a scram_arch/cmssw environment settings like this
https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/runcmsgrid_powheg.sh#L29-L61
where `SCRAM_ARCH_VERSION_REPLACE` and `CMSSW_VERSION_REPLACE` are replaced with the ones used to create the gridpack in the gridpack creation script. This mechanism is useful to ensure compatibility across releases and campaigns

kurtejung:
hi @perrozzi - your changes should be implemented now - thanks

kurtejung:
please wait ~1 hr on merging - pending a bug fix with particle status

kurtejung:
thanks - done. Please merge when ready.

PR #1429, opened 2017-10-17
Critical fixes for HNNLO and MadLibs

covarell:
Two fixes:
- Call PDF initialization by name in HNNLO as suggested by @agrohsje --> affects HJ+NNLOPS
- Remove weird fortran options (fixed line length causes problems in gfortran6.3 on lxplus)  --> affects trijet and possibly others

@gourangakole @pgunnell 

@bendavid please let us know if HNNLO fix is somehow clashing with your dev version

### DIFF
--- a/bin/Powheg/run_pwg.py
+++ b/bin/Powheg/run_pwg.py
@@ -424,9 +424,9 @@ if [ "$process" = "bbH" ]; then
    sed -i -e "s#O2#O0#g" Makefile
 fi
 
-## Not needed anymore
-# sed -i -e 's#$(PDFPACK#lhefread.o pwhg_io_interface.o rwl_weightlists.o rwl_setup_param_weights.o \\ \
-#	$(PDFPACK#' Makefile
+# Remove strange options in fortran (fixed line length, multiCPU compilation)
+sed -i -e "s#132#none#g" Makefile
+sed -i -e "s#make -j FC#make FC#g" Makefile
 
 # hardcode svn info
 sed -i -e 's#^pwhg_main:#$(shell ../svnversion/svnversion.sh>/dev/null) \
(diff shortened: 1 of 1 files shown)


covarell:
Adding also @yuanchao

perrozzi:
I assume it has been tested and provided the ok from josh it can be merged, right?

yuanchao:
@covarell: is it just the "-ffixed-line-length-132" option caused the crash, or it's the multi-process option of "make -j" or both? I'm still puzzled why it works w/o modification on my cluster but not on lxplus.

covarell:
I could compile successfully only when applying both patches

yuanchao:
With "-j 8" it works for me on lxplus.

covarell:
I can try but is this really important? Do we need parallel compilation for trijet? In my case it took 15-20 minutes at most.

covarell:
H->tautau are already setting up NNLOPS requests, other subgroups should follow, so maybe we can merge this and rethink when the new workflow will be ready? @bendavid

agrohsje:
@covarell @gourangakole 
I had a look at the PDF problem in hj_minlo. InitPDFsetByName works ok, the problem is that the argument is wrong. In   
hnnlo-v2.0/src/Need/pdfset_lhapdf.f 
this line: 
call InitPDFsetByName('PDFsets/'//PDFname)                                                                                                                                                                   
should be changed to        
call InitPDFsetByName(PDFname)
Moreover, I remove //'.LHgrid' in 
if (oldPDFname(i:i+6) .eq. '.LHgrid') then
          PDFname=oldPDFname(1:i-1)//'.LHgrid'                                                                                                                                                                                              
else
The same should be true for the line below with  //'.LHpdf
Fixing these lines solves the problem. I can run the code without seg fault.

perrozzi:
thanks alexander!
please anybody prepare a PR asap :-)

PR #1436, opened 2017-10-20
Gridpack generation support for T2_BE_IIHE's cream02 cluster

UAEDF-tomc:
With these changes you can run the gridpack_generation.sh script also on the T2_BE_IIHE's cream02 cluster. This is also very useful for the Belgian users, so I make the pull request in order to share it. As it simply adds "cream02" for the queue option, it will not affect other users.

The patch in madgraph/various/cluster.py makes an essential change to the PBSCluster class: it sources the CMSSW environment such that the needed fortran libraries are available.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/patches/0017-fix-pbs-cluster.patch
+++ b/bin/MadGraph5_aMCatNLO/patches/0017-fix-pbs-cluster.patch
@@ -5,6 +5,17 @@ diff --git a/madgraph/various/cluster.py b/madgraph/various/cluster.py
 -    maximum_submited_jobs = 2500
 +    maximum_submited_jobs = 1000
 @@ -1244,2 +1244,3 @@
-+        text += "source $VO_CMS_SW_DIR/cmsset_default.sh;eval `scram runtime -sh`;"
++        text += "source /cvmfs/cms.cern.ch/cmsset_default.sh;eval `scram runtime -sh`;"
          if not os.path.isabs(prog):
              text += "./%s" % prog
+@@ -1281,6 +1282,9 @@
+-        for line in status.stdout:
++        for line in (status.stdout + status.stderr):
+             line = line.strip()
+             if 'cannot connect to server' in line or 'cannot read reply' in line:
+                 raise ClusterManagmentError, 'server disconnected'
+             if 'Unknown' in line:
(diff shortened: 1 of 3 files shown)


bin/MadGraph5_aMCatNLO/patches/0017-fix-pbs-cluster.patch
```diff
@@ -0,0 +1,10 @@
+diff --git a/madgraph/various/cluster.py b/madgraph/various/cluster.py
+--- a/madgraph/various/cluster.py
++++ b/madgraph/various/cluster.py
+@@ -1216,1 +1216,1 @@
+-    maximum_submited_jobs = 2500
++    maximum_submited_jobs = 1000
+@@ -1244,2 +1244,3 @@
++        text += "source $VO_CMS_SW_DIR/cmsset_default.sh;eval `scram runtime -sh`;"
```
perrozzi: out of curiosity, isn't the "central" cmsenv sufficient? why?

UAEDF-tomc:
@perrozzi You are right, it seems we can simply use /cvmfs/cms.cern.ch/cmsset_default.sh instead of $VO_CMS_SW_DIR/cmsset_default.sh, using $VO_CMS_SW_DIR was a habit from the past.

Additionaly, I added another change to the patch: checking the status of pbs jobs sometimes fails (randomly, because too busy use of the pbs system we think) and gives an "Invalid credential" message. Madgraph would interpret this as all the jobs are failed and resubmit them. The patch simply waits a bit before trying again to check the status with qstat.

perrozzi:
things look ok to me, I ask @kdlong @qliphy to have a final look

kdlong:
Looks good to me. 

@pgunnell, @Andrej-CMS  and Alexander (can't find his username to tag him) don't you use a PBS cluster at DESY as well? Did you ever use it with MadGraph? If so, it might allowing the queue to be manually set rather than hardcoding the Belgium specific option.

perrozzi:
@agrohsje

agrohsje:
It is SGE. You don't need to specify the queue in our case.

kdlong:
Thanks. In that case I'd just merge as is and modify it to be more general later if someone expresses interest.

PR #1438, opened 2017-10-20
CMSConnect: Allow customization of retry input parameters for condor queries

khurtado:
A gridpack [1] was having issues with the cluster manager due to (I think) the scheduler query timeouts being too short (20s). Queries for group of jobs greater than 1K can take longer than that and fail. I increased the timeout tolerance in the scheduler, but am also customizing the number of query retries in the manager.

I also added a similar retry mechanism while querying the CODEGEN exit code in the cmsconnect submit wrapper, just in case.

[1] https://hypernews.cern.ch/HyperNews/CMS/get/generators/3648.html


khurtado:
@qliphy : Noticed you are running  DYJets, so you might want to apply this PR if it gives you the issue above.

qliphy:
@khurtado  Thanks for the reminding.  I will try the PR once the current run is finished.

khurtado:
@kdlong Any objections to merge this?

kdlong:
Sorry I didn't test yet. Do you recommend something specific, or should I just double check that it's working?

khurtado:
Just double checking it. I tested it myself and works fine in connect.
Shouldn't break any other submit method, so I think should be ready to
merge unless you can think of a reason not to.

qliphy:
@khurtado   I tried with this PR for a complicated process, i.e. dy01234jets MLM, it worked well and almost ran to the end, but unfortunately it stopped due to the CMS Connect maintenance in late Oct.

I am rerunning this process and so far it looks good.

gourangakole:
@khurtado  Just to let you know.
I am running dyellell012j_5f_NLO_FXFX.
I have reported back on hn [1]. I did not get any suggestion. So running with this PR. 
But it has not finished yet. 

[1]
https://hypernews.cern.ch/HyperNews/CMS/get/generators/3675.html

khurtado:
@qliphy Thanks for the update!
@gourangakole : Alright! I just replied to your thread in the HNs. The first error in the thread was due to infrastructure issues that started happening after the upgrade, that should be fixed now. The second seems to be related to the madgraph process itself, since the job ran and returned the output from the worker node just fine.

efeyazgan:
Now it's OK. See https://hypernews.cern.ch/HyperNews/CMS/get/generators/3675/1/1/1/1/1/1/1.html

PR #1439, opened 2017-10-22
[Mono-Higgs] cards for the scan of MA and Ma masses in the 2HDM+a model

syuvivida:
This is a new mono-Higgs model discussed in the LHC DMWG. 

The cards committed have fixed model parameters, including sinp, lambda, tanbeta, and mXd. The mass of MA is set to be identical to those of MH and MH+-. We perform a scan of the masses of the two pseudo scalar bosons in the model: MA and Ma. There are 55 mass points.

The production of gg and bb~ must be generated separately because gg process is loop-induced while bb~ is a simple tree diagram.


efeyazgan:
Instead of having 55 mass points here, please upload 1 representative card, and if it exists the script to produce the other mass points from the representative one.

efeyazgan:
Are these for Fall17 (=Moriond18) production?

syuvivida:
Hi Efe,

  This is Moriond 2017 production (and that's why they are in the pre-2017 folder).
Do you mean now the rule is to put only a set of example cards for one mass point and the script (it available)?

Thanks!!

Eiko

perrozzi:
Ciao Eiko,
we made a cleanup of the repository because too many files slow down git.
Given that this PR alone has 440 files, we would kindly ask to upload the datacards of one (or more) representative points of the scan, together with the script used to produce them all.

syuvivida:
Hello,
 now I only leave one mass point for bb~ and gg process and a text file of mass points. I didn't use a script (simply use awk and sed command).

perrozzi:
Hi Eiko, thanks.
are you using update_ATLAS_mass only for gg or also for bb?
in the latter case, please move the file to the parent folder

syuvivida:
Hi Luca,
  file is moved to the parent folder as this file is used by both gg and bb~ processes.

Eiko

perrozzi:
another, final request to improve the clarity: together with the list of mass points I would kindly ask to upload a script containing the awk and sed command you used, to guide anybody willing to reproduce the datacards

AndreasAlbert:
@syuvivida Is there a reason not to use the updated UFO file? [1]

[1] https://gitlab.cern.ch/lhc-dmwg-material/model-repository/tree/master/models/Pseudoscalar_2HDM

syuvivida:
@AndreasAlbert
   We have checked the updated one and the v1 gave the same cross sections for our model. ATLAS mono-Higgs group is also using the old one. In addition, we want to use 5F PDF, but by default v2 model files have non-zero masses for u,c,d,s, e, mu. Just for simplicity and to reduce the number of lines in the cutomized cards, we chose to use v1 model files.

BTW, we did not succeed in using restrict card for this model. If you manage to do so, please let us know.

syuvivida:
@perrozzi
The scripts and necessary files are all committed.

bin/MadGraph5_aMCatNLO/cards/production/pre2017/13TeV/monoHiggs/2HDM+a/sinp_0p35_tanb_1_mXd_10/bbProduction/bbProduction_proc_card.dat
```diff
@@ -0,0 +1,6 @@
+import model Pseudoscalar_2HDM
+define p = p b b~
```
perrozzi: one last thing, we recommend to remove the proton and jet definition,
as this is taken care automatically depending on the input model

syuvivida: @perrozzi 
  My understanding is that the flavor scheme is automatically taken care of only in the gridpack_generation.sh in the master branch.

While in my case, I am generating samples for EXO-16-050, a 2016 analysis and Moriond 2017 MC samples. To be consistent with other mono-Higgs samples, I am using Madgraph 2.4.2 and the gridpack_generation.sh in mg240 branch.

Note, I did ask in the past if I should commit this kind of MC cards in mg240 branch or master branch in the pre-2017 folder. I was told to commit in the pre-2017 folder.

syuvivida: @perrozi,

 Hello, are the cards OK? Can they be merged?

perrozzi: Hi Eiko, I think is not a good idea to use the mg240 branch but the master one instead, even if the samples are meant to be used for 2016 MC (in any case you would get MG5_aMC v2.4.2 but in the mg240 branch you would not get all the updates and fixes to the production scripts etc).

agrohsje:
Shouldn't one use the restrict_bbMET_5FS for 5FS? The default has massive b's no?

syuvivida:
@agrohsje 
 We have tried to use the restrict card, but we always got this error of 
 KeyError : 'mdl_MXd'
 
 That's why we are setting b quark mass to zero in the parameter card instead.

syuvivida:
@agrohsje 
 We have tried to use the restrict card, but we always got this error of 
 KeyError : 'mdl_MXd'
 
 That's why we are setting b quark mass to zero in the parameter card instead.

agrohsje:
funny. it is working for me ok. anyway, if you adjust it by hand, that should be fine as well.

syuvivida:
@agrohsje

   Can you point me to your full setup? Which mono-X are you generating?
Thanks!!

Eiko

agrohsje:
I am looking into ttDMDM.

PR #1446, opened 2017-10-23
madspin for powheg+madpsin+pythia8 single antitop 4f tch

Andrej-CMS:


efeyazgan:
Why are these under MadGrapg5_aMC@NLO folder?

agrohsje:
The tarball (and the scripts) is a MG5_aMC@NLO one and Powheg is included as an external tarball in the tarball.

perrozzi:
seems good to me

PR #1449, opened 2017-10-24
Modify madspin card using the new sintax

lviliani:
Change the madspin card according to the new sintax, as suggested by Kenneth.


perrozzi:
was it a github or HN thread or private conversation?

lviliani:
It was a private conversation with @kdlong .
The point is that without the "set onlyhelicity True" command, madspin causes a crash when generating events from the gridpack.
Kenneth can probably comment further.

kdlong:
MadSpin in this case is only used to trigger the inclusion of some spin-related info to be passed to Pythia to be used for the tau decays. In the past we triggered this by includin a MadSpin card where the w decays are just dummy decays so that this info would be included, but now that they've added the onlyHelicity option this seems to crash if the decays aren't relevant to the process.

So for all processes where taus are included but MadSpin isn't actually decaying a particle present in the final state of the proc card, this should be added.

PR #1457, opened 2017-10-26
small improvement to POWHEG source compilation check

perrozzi:
remove DYNNLOPS from POWHEG source compilation checks


covarell:
+1

PR #1459, opened 2017-10-26
Fix processes already using bookhist-multi-new

covarell:
Some processes using already new dynamical histo-booking should not also link against old one


covarell:
That should be the last fix for compilation in gcc5.3/6.3 (except for SUSY processes that no one uses in CMS)

PR #1471, opened 2017-10-31
Update PDF list for MG 2.6 and move to more flexible setup

kdlong:
Copy over some additional updates from master branch, and move to more modular setup for PDF weights (to be back-ported to master branch as well)

NOTE: Some further testing would be beneficial. In addition, for the wplustest NLO samples the reweighted cross sections all differ from a factor of ~2 from the central value, which is not understood. Example output for wplustest_5f_NLO:

```
# original cross-section: 103354.4079
#     scale variation: +-46.6% -55%
# PDF variation: +500% -500%
#
#PDF PDF4LHC15_nnlo_100_pdfas: 50974.7 +1.27% -1.27%
#PDF MMHT2014nlo68clas118: 48742.3 +0.721% -0.733%
#PDF NNPDF30_nlo_nf_5_pdfas: 48323.9 +0.968% -0.968%
#PDF LUXqed17_plus_PDF4LHC15_nnlo_100: 50753.1 +1.06% -1.06%
#PDF NNPDF31_nlo_hessian_pdfas: 50275 +0.919% -0.919%
#PDF HERAPDF20_NLO_EIG: 52729.7 +0.445% -0.493%
#PDF CT14nlo: 49356.3 +1.24% -1.41%
```


perrozzi:
Thanks. A couple of comments
* I think it would be better not to move those cards located in the main folder in view of merging with the master branch ,from which they were already removed (if I remember correctly)
* Also, this probably can be merged and then tested, because without this no consistent pdf reweighting is available for mg26x
* the factor 2 in the xsec for reweighted sets is scary...

kdlong:
@perrozzi 

1. 1. Should I just delete the cards then?
2. 2. Yes, it's not super rigorously tested but a many of the changes here are badly needed (fixing the PDF list, fixing the environment setting...)
3. Agreed. I will try to reproduce this in standalone MadGraph and ask what the problem is to the authors.

Also added changes to fix things that broke CMSConnect and to port (a lot of) changes which were made in the master branch but not propagated here. Adopting a new folder structure is a hassle but my gaol is to reduce the degree to which all these things are interconnected, which is making supported two branches of the code a huge pain.

kdlong:
Asking authors about reweighted cross sections: https://answers.launchpad.net/mg5amcnlo/+question/660414

perrozzi:
I see that Olivier replied, probably the test he proposes will make things clearer (looks like there is a bug).
Shall this PR be merged in the meanwhile?

kdlong:
It doesn't matter to me if it's merged now or if we wait for the systematics situation to become clear. Indeed it looks like a major bug. I tried the test he suggested but it fails in other ways... Will post an update to the launchpad shortly.

perrozzi:
The PR is becoming very big and starts to contain many different modifications, I think is better to merge and makr new PRs for future fixes. Thanks a lot

PR #1472, opened 2017-11-01
st_tW_5f_dr_ckm_13TeV run_pwg.py required for V2 validation

ckmackay:


perrozzi:
would it be possible to add if statements checking for the process name in the standard script, instead of a new script?

perrozzi:
(can you provide a diff wrt standard cfg for a quick check?)

ckmackay:
Hi,

It is just one line added at line 562, here are the instructions I received:

There is a small caveat. I needed to modify run_pwg.py by inserting the sed command below.  Search for QCDLoop and you will find the following lines that you can extend by the sed in between:
>   mv QCDLoop-1.96 QCDLoop-1.9
>   cd QCDLoop-1.9
>   sed -i -e 's#/Users/ellis/QCDLoop#./QCDLoop#' ff/ffinit_mine.f
>   make

Thanks, Kate

perrozzi:
if so I would strongly advise to modify the original script more than adding another, can you please do it? thanks

ckmackay:
OK, I'll close this request and open another for the new run_pwg.py

PR #1474, opened 2017-11-02
adding cards for WJetsToLNu sample

deepakcern:
I would like to add the cards for the WJetsToLNu_13Tev sample.


perrozzi:
@deepakcern may I ask what is the use case for this sample?

perrozzi:
why do you include only w- ?
why both the 1j and the 2j start at 0j ?
not sure I understand...

deepakcern:
Sorry for mistake, i talked to Anirban and then compare with previous cards https://github.com/cms-sw/genproductions/tree/pre2017/bin/MadGraph5_aMCatNLO/cards/production/13TeV/WJets_jetbinning_LO_MLM
and i found that we need to include W+ and W- , we don't need to include 0j, 1j and 2j.
I will make some changes according to previous cards.

anirban1982:
Yes, Deepak has to change the lhaid as well, he will do the changes and make fresh PR soon.

PR #1476, opened 2017-11-02
add correct cp5 minbias cfg

perrozzi:


perrozzi:
@bendavid @pgunnell please crosscheck

pgunnell:
Correct from my point of view! Thanks Luca!

PR #1487, opened 2017-11-06
POWHEG+JHUGen mass scan cards for 2017 MC

hroskes:
Each one with a template, a script to make the cards, and a gitignore


bin/Powheg/production/2017/13TeV/HWJ_NNPDF31_13TeV/.gitignore
```diff
@@ -0,0 +1 @@
+HW*J_HanythingJ_NNPDF31_13TeV_M*.input
```
perrozzi: I don't think this file should be committed

bin/Powheg/production/pre2017/13TeV/HZJ_NNPDF30_13TeV/.gitignore
```diff
@@ -0,0 +1 @@
+HZJ_HanythingJ_NNPDF30_13TeV_M*.input
```
perrozzi: neither this

perrozzi:
the rest looks ok to me

hroskes:
@perrozzi 
I was putting those there so that when you run the script and then do git status, you don't see all the cards there (which are not supposed to be committed).  Is that a problem?

PR #1488, opened 2017-11-07
CP5 tune WToENu

alsusy:


efeyazgan:
Config files seem OK to me but the files should be moved to genproductions/python/ThirteenTeV/WToENu/
@perrozzi please comment on the file location.

perrozzi:
@alsusy may I ask what is the use case for these cards and why pythia8 standalone is used instead, for instance, of mg5_amc?
(in any case I agree on the wrong file location)

alsusy:
These are pythia samples for Fall17 and clones of WToENu*TuenCUETP8M1*.    Is the right location under genproductions/python/ThirteenTeV  just like WToENu*TuenCUETP8M1* ?   Then should I commit again ?

efeyazgan:
I think genproductions/python/ThirteenTeV/WToENu/ should still be better. But could you please clarify the purpose of such a sample?

AndreasAlbert:
@efeyazgan @perrozzi These samples are used by the lepton+MET (Wprime) resonance search (e.g. EXO-16-033). They apply k-factors on top of the bare samples to get the shape right, which in the end seems to be doing a good job of describing the background.

perrozzi:
Andreas, can you move the cards to the suggested folder?

AndreasAlbert:
@perozzi: I don't think I can since this is not my PR

@alsusy: Please move the cards!

alsusy:
Hi,  I run  those cards with "git mv ... " and "git commit ... " to  genproductions/python/ThirteenTeV/WToENu/  .    could you check  if it is OK or  I missed any procedure ?

perrozzi:
then you probably still need to do git push

alsusy:
"git push "  is done.

perrozzi:
last thing: can you please delete similar 4C requests like https://github.com/alsusy/genproductions/blob/8ed094cf5811ddfa76d3c443f17b9e7403cf4428/python/ThirteenTeV/WToENu_M_500_Tune4C_13TeV_pythia8_cfi.py and move the the same folder you just created the CUETP8M ones?

alsusy:
Just to be sure, you'd like to delete WToENu*4C* and to move WToENu*CUETP8M* to WToENu folder ?

perrozzi:
yes. you can even move them all into the WToENu folder but I don't think the 4C tune is relevant

alsusy:
It's done.  Let me know if anything else is needed.

PR #1492, opened 2017-11-07
new version, working with the patched phantom release, the submit has…

govoni:
… also a test option to produce events locally

### DIFF
--- a/bin/Phantom/cards/examples/phantomProduction_example.cfg
+++ b/bin/Phantom/cards/examples/phantomProduction_example.cfg
@@ -14,13 +14,15 @@
 # CMSSW environment for the gridpack creation
 # this is NOT necessarily the one used in the detector simulation and reconstruction,
 # it just sets the environment for the phantom running
-CMSSW = CMSSW_9_2_11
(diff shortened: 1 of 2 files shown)


efeyazgan:
Please remove the unnecessary commented lines.

govoni:
which ones?

bin/Phantom/cards/examples/phantomProduction_example.cfg
```diff
@@ -14,13 +14,15 @@
 # CMSSW environment for the gridpack creation
 # this is NOT necessarily the one used in the detector simulation and reconstruction,
 # it just sets the environment for the phantom running
-CMSSW = CMSSW_9_2_11
+# CMSSW = CMSSW_9_2_11
```
efeyazgan: Commented inputs like this

govoni:
these are not unnecessary: associating scram_arch and cmssw release is always unnecessarily painful for cms, as that info in hidden. In this way, in the example we have an example for both a 2016 and 2017 configuration.

PR #1431, opened 2017-10-18
Wh01 m125 toa01a01 tobbbb

aminghiasi:
This pull request will add "Wh01 m125 toa01a01 tobbbb" Madgraph cards for central production. We have already produced private MC samples using these cards and gave two updates in the HiggsExo meeting:
https://indico.cern.ch/event/668522/contributions/2741419/attachments/1537379/2409199/haa4b_georgia091017.pdf
https://indico.cern.ch/event/663418/contributions/2709637/attachments/1521025/2396927/haa4b_higgsexo110917.pdf

PR #1498, opened 2017-11-08
Adding 2017 MadGraph cards for Zh01_m125_toa01a01_tobbbb

aminghiasi:
After looking at Wh production (https://github.com/cms-sw/genproductions/pull/1431), we plan to look at h->aa->bbbb channel in Zh production. This pull request adds the needed MadGraph cards for the MC central production for 2017.


efeyazgan:
Could you please leave only representative cards (i.e. one mass point)? And if available, add a script to generate the cards for different mass points. thanks.

aminghiasi:
I removed all cards except for one mass point.

PR #1501, opened 2017-11-09
Fix for Powheg+Madpsin gridpacks

jfernan2:
Following:

https://hypernews.cern.ch/HyperNews/CMS/get/generators/3662/1/1/1/1/1.html

Not sure if this should be the solution for other gridpacks than Powheg + Madspin

### DIFF
--- a/bin/MadGraph5_aMCatNLO/runcmsgrid_NLO.sh
+++ b/bin/MadGraph5_aMCatNLO/runcmsgrid_NLO.sh
@@ -104,8 +104,8 @@ if [ ! -e $LHEWORKDIR/header_for_madspin.txt ]; then
 #else handle external tarball
 else
     cd $LHEWORKDIR/external_tarball
-
-    ./runcmsgrid.sh $nevtjob $rnum $ncpu
+    #for Powheg gridpacks + Madspin to have same amount of events
+    ./runcmsgrid.sh $nevt $rnum $ncpu
 
 #splice blocks needed for MadSpin into LHE file
     sed -i "/<init>/ {


perrozzi:
thanks
have you tried to check whether this has an impact on the other gridpacks?
(@kdlong @qliphy )

efeyazgan:
Any news?

bin/MadGraph5_aMCatNLO/runcmsgrid_NLO.sh
```diff
@@ -104,8 +104,8 @@ if [ ! -e $LHEWORKDIR/header_for_madspin.txt ]; then
 #else handle external tarball
 else
     cd $LHEWORKDIR/external_tarball
-
-    ./runcmsgrid.sh $nevtjob $rnum $ncpu
```
perrozzi: btw, it looks to me that if you remove nevtjob here, it is not used anywhere else, so I would suggest to remove it everywhere (cfr L71)

jfernan2: I may be wrong, but I'm afraid it is used for pure MG5_aMC@NLO gridpacks. That's the purpouse of that variable.

perrozzi: if I click on "view" I can't see this variable used anywhere, unless I am missing something
https://github.com/jfernan2/genproductions/blob/7eb379735966dcce8786bb43c8171fb6b7bb8a9d/bin/MadGraph5_aMCatNLO/runcmsgrid_NLO.sh

jfernan2: It is stored on runscript.dat and then used in L96

PR #1502, opened 2017-11-09
update cards to 2017

mmozercern:
These are signal samples for the all-hadronic diboson searches.
They are based on the previous pruduction with the following changes:
-> explicit pdf numbers replaced by placeholders
-> pdfwgt set to T

This is only the most urgent subset of configurations


efeyazgan:
There are too many files. Could you please keep one example from each (i.e. one mass point) and instead put the script that you use to produce the cards with different parameters?

mmozercern:
Tricky, I didn't set up the old cards, so I don't have the appropriate scripts, just the old git.
I'll check to see if I can find who produced these originally.

efeyazgan:
Script is not mandatory but please only have example representative cards for each config.

mmozercern:
Ok, I've cleaned things up.

PR #1508, opened 2017-11-10
Heavyneutrino s and tchannel

sihyunjeon:
Hi all,

I am using default mg26x way of pdf weightings
and modified the changes.

Are there more that I should change?


efeyazgan:
Could you please put one mass point for each case? thanks.

sihyunjeon:
They are with different settings (decay channels are different, and there is "spinmode onshell" used only on several mass points.) Are these differences negligible?

efeyazgan:
No, sorry, then they should stay like that.

sihyunjeon:
This should be good to go for merge?

PR #1520, opened 2017-11-14
Add DYJetsToMuMu mll105to160 cards (2017 version)

UAEDF-tomc:


perrozzi:
the cards look ok, but why there is "HT" in the name? it looks inclusive to me...

UAEDF-tomc:
Indeed, there is no reason to have the HT, I removed it from all the cards

perrozzi:
(then on the main folder too, I assume...)

UAEDF-tomc:
Main folder is also renamed now.

efeyazgan:
I guess we should also have "T = auto_ptj_mjj"

UAEDF-tomc:
Yes, I changed it.

PR #1528, opened 2017-11-15
cards for decay to 2l2nu and 2l2q with reweighting

hroskes:


perrozzi:
in principle we provide instructions to run from any folder, suggesting to copy the files
https://twiki.cern.ch/twiki/bin/viewauth/CMS/PowhegBOXPrecompiled#Download_all_the_necessary_scrip

hroskes:
Oh, I didn't see that.  I guess this wasn't necessary, although it does make it a little easier.

Do you want me to remove this commit and just keep the new JHUGen cards?

perrozzi:
for the moment yes, thanks

hroskes:
@perrozzi done (and changed the PR name so it actually makes sense)

PR #1527, opened 2017-11-15
ttHjets aMC@NLO FXFX

Andrej-CMS:
Dear experts,
please check the cards.

Kind regards,
Andrej

PR #1533, opened 2017-11-20
Moved Higgs Processes to Higgs Folder

Andrej-CMS:
Dear experts,
I moved the Higgs processes to the Higgs folder as discussed in https://github.com/cms-sw/genproductions/pull/1527
Kind regards,
Andrej

EDIT: I also fixed vh012j_5f_NLO_FXFX_M125_VToAll_run_card.dat , the PDF reweighting was defined twice before


perrozzi:
thanks a lot. as far as I can see, Zh01_M125_Toa01a01_Tobbbb_LO_MLM has one extra folder layer, can you please remove it?

Andrej-CMS:
You're welcome!
I think that the extra layer is supposed to be for the folders for mass variations, like it is done in the WH folder. Should I anyway remove it?

PR #1535, opened 2017-11-20
Fixing mkfifo problem with functions

kdlong:
An obvious way to avoid the need to use either mkfifo or process substitution -- which both seem to fail on a few machines -- is to instead just use tee for a specific function. This has the added advantage of making the code organization a little more sane, which I would really like to do in any case.

This has the disadvantage of making the git change history a bit of mess. Sorry about that. It's a huge pain to make changes that should just be a restructuring and test that they don't have any unintended consequences, but in principle the only change here is restructuring the code + 1 fix to an incorrect path in submit_condor_gridpack_generation.sh. In my tests it seems that things are working correctly.

I'm making it here so we can do a few more tests before putting it into the master version. But I plan to make the same change and a few more so that the code organization between the two is more similar, which will be a big help when we got to merge 26x to master.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
+++ b/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
@@ -381,7 +381,7 @@ make_gridpack () {
     #Add PDF info and copy run card #
     #################################
     script_dir="${PRODHOME}/Utilities/scripts"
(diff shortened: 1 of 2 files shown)


perrozzi:
ready to be merged?

kdlong:
According to me, yes

PR #1537, opened 2017-11-21
remove mgrwgt block

qliphy:
Related to https://hypernews.cern.ch/HyperNews/CMS/get/prep-ops/4526/1/1/1/1/1.html

Only affect NLO not LO

### DIFF
--- a/bin/MadGraph5_aMCatNLO/runcmsgrid_NLO.sh
+++ b/bin/MadGraph5_aMCatNLO/runcmsgrid_NLO.sh
@@ -143,7 +143,7 @@ fi
 
 cd $LHEWORKDIR
 gzip -d ${runname}_final.lhe.gz
-
+sed -i -e '/<mgrwgt/,/mgrwgt>/d' ${runname}_final.lhe 
 ls -l
 echo
 


perrozzi:
this can be merged, right?

qliphy:
@perrozzi   Yes, I think so.  Already tested from my side.

bendavid:
Is this still needed given the fix to the duplicate header issue?

qliphy:
@bendavid    I think this is different from the duplicate header case (you meant last week's fix, right?).  There is no multiple runs and merge.pl will not be called.   

The problem here is  due to a new block "mgrwgt" appear in LHE with mg26X [1] which triggers bugs when pythia tries to hardonize these events.  Currently one simple solution is to remove this block...

[1]https://hypernews.cern.ch/HyperNews/CMS/get/prep-ops/4526/1/1/1/1/1.html

bendavid:
No I wasn't referring to the merging issue, but rather the handling of headers from multiple lhe files at the CMSSW level, as addressed here
https://github.com/cms-sw/cmssw/pull/18846

(so my question is whether this change is still needed given the above pull request)

Issue #1534, opened 2017-11-20
[MG] CMS Connect jobs: subjobs fail


AndreasAlbert:
Hi,

I am trying to create some gridpacks with CMS Connect. Relevant info (cards, etc) below.
The short summary is that I can successfully run these cards locally or on LSF, but not on CMS Connect.

The generation ultimately fails because one or two of the subjobs fail and the main job cannot read the results.dat (log at [1]). Looking at the corresponding subprocess log [2], there are a lot of "Error done early" lines in it that do not appear in succesful subprocesses. Seeing how the production works on other architectures, I think there is a configuration problem somewhere (possibly Madloop?).

I have no idea at all how to debug this. Could anyone advise what to look at?

Best

Andreas

Cards in my  [my genproductions fork](https://github.com/AndreasAlbert/genproductions/tree/monoz_2hdm_production_clean_oct30/bin/MadGraph5_aMCatNLO/cards/production/2017/2HDM_5F/Pseudoscalar2DHM_MonoZLL_mScan_5F_mh3-200_mh4-500)

Submission command is

`name="Pseudoscalar2DHM_MonoZLL_mScan_5F_mh3-200_mh4-500"; nohup ./submit_cmsconnect_gridpack_generation.sh ${name} ./cards/production/2017/2HDM_5F/${name} > ${name}.debug 2>&1 &`

[1]` /afs/cern.ch/work/a/aalbert/public/2017-11-20_cmsConnect/Pseudoscalar2DHM_MonoZLL_mScan_5F_mh3-200_mh4-500.debug`
(error appears around line ~900)

[2]` /afs/cern.ch/work/a/aalbert/public/2017-11-20_cmsConnect/Pseudoscalar2DHM_MonoZLL_mScan_5F_mh3-200_mh4-500/Pseudoscalar2DHM_MonoZLL_mScan_5F_mh3-200_mh4-500_gridpack/work/gridpack/process/madevent/SubProcesses/P0_gg_xdxdxll/G2.26_1/run1_app.log`

kdlong:
Hi Andreas,

I agree that the error seems to come from these couple of failed subprocesses. However, I have no idea how to interpret all the ```Error done early``` messages. Could you post this message to the MadGraph launchpad to ask the authors? Without any indication of what could cause this I don't know how to debug it as a CMSConnect issue.

AndreasAlbert:
Hi Kenneth,

[...]

PR #1544, opened 2017-11-23
Set TMPDIR to condor scratch area when not defined

khurtado:
Update patch 0018 to make sure TMPDIR does exist in the environment.

Most cluster configurations define TMPDIR as the batch manager scratch area in the worker node environment. When this doesn't happen, it falls back to `/tmp`. This patch set TMPDIR to the condor scratch area if:
- TMPDIR is not set in the environment (LSF jobs and local condor jobs should have this set already, but not necessarily on all sites in the global pool)
- _CONDOR_SCRATCH_DIR does exist (this should be the case for all condor jobs, local or grid)

otherwise, it falls back to `/tmp`, which is the default.

This is also related to #1534


perrozzi:
I assume it can be merged right away, correct?

PR #1546, opened 2017-11-23
Fix for dilepton decay en ttbar

jfernan2:


efeyazgan:
This doesn't make it decay only to electrons but to other leptons as well, right?
 +                     ! 20000    both top go into e l nu (with the appropriate signs)

jfernan2:
@efeyazgan Only to electrons, as in the semileptonic case...
See line 32...

efeyazgan:
But then is it correct?

jfernan2:
Now it is, before it wasn't: only electrons, no mu and taus

perrozzi:
sorry, not sure I understand yet: does 
`topdecaymode 22200`
include electrons, muons and taus?

jfernan2:
@perrozi Yes, 22200 includes e, mu and tau

perrozzi:
ok so is the pr ready to be merged? (better to be pedantic)

jfernan2:
Yes. Sorry again

efeyazgan:
Ok I see that there was "!". So all fine.

PR #1555, opened 2017-11-23
spinmode onshell fixes

sihyunjeon:
Hi all,

The formats aren't correct I think... Could you guys take a look at this?
These are fixes that need to be applied to use spinmode onshell in madspin (mg26x)

And sorry for taking too long...

### DIFF
--- a/bin/MadGraph5_aMCatNLO/patches/0011-fix-f2py-path.patch
+++ b/bin/MadGraph5_aMCatNLO/patches/0011-fix-f2py-path.patch
@@ -1,0 +1,11 @@
+--- a/madgraph/iolibs/export_v4.py 
++++ b/madgraph/iolibs/export_v4.py
+@@ -1739,7 +1739,7 @@
+             f2py_compiler = ''
+         # Try to find the correct one.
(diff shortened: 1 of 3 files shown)


perrozzi:
@kdlong @qliphy

agrohsje:
did you try : 
git format-patch master --stdout > nameofpatch.patch

perrozzi:
what shall we do with this?

perrozzi:
any news on this?

sihyunjeon:
@perrozzi  Hi I am really sorry for late reply. There are some more problems on using this "spinmode onshell" so I am still checking some things. I will at least provide these patches after checking they are properly applied this weekend.
@agrohsje  Thanks, I think I found out how to do it thanks to your help.

kdlong:
Hi, the reason the tabs didn't work is because the tabs are wrong. I actually fixed them before the break and forgot to make a pull request. Will do this now.

sihyunjeon:
Thanks, then I will close this in few hours

kdlong:
Fixed in #1618

PR #1564, opened 2017-11-29
force mg5_amc LO to produce the requested number of LHE events

perrozzi:
@kdlong

### DIFF
--- a/bin/MadGraph5_aMCatNLO/runcmsgrid_LO.sh
+++ b/bin/MadGraph5_aMCatNLO/runcmsgrid_LO.sh
@@ -78,7 +78,7 @@ run_counter=0
 # with the same seed across the workflow
 run_random_start=$(($rnum*10))
 # otherwise don't change the seed and increase number of events as 10000 if n_evt<50000 or n_evt/9 otherwise
-if [  $run_random_start -gt "89999999" ]; then
+if [  $run_random_start -gt "89999990" ]; then
     run_random_start=$rnum
     max_events_per_iteration=$(( $nevt > 10000*9 ? ($nevt / 9) + ($nevt % 9 > 0) : 10000 ))
 fi
@@ -90,8 +90,8 @@ while [ $produced_lhe -lt $nevt ]; do
   # increase the iteration counter
   let run_counter=run_counter+1 
   
-  # don't allow more than 9 iterations
-  if [  $run_counter -gt "9" ]; then
+  # don't allow more than 90 iterations
+  if [  $run_counter -gt "90" ]; then
       break
   fi
   # compute remaining events
--- a/bin/MadGraph5_aMCatNLO/runcmsgrid_LO.sh
+++ b/bin/MadGraph5_aMCatNLO/runcmsgrid_LO.sh
@@ -68,6 +68,9 @@ fi
 #generate events
 ./run.sh $nevt $rnum
 
+echo "run finished, produced number of events:"
+zgrep \<event process/events.lhe.gz |wc -l
+
 cd $LHEWORKDIR
 
 mv process/events.lhe.gz events_presys.lhe.gz
--- a/bin/MadGraph5_aMCatNLO/runcmsgrid_LO.sh
+++ b/bin/MadGraph5_aMCatNLO/runcmsgrid_LO.sh
@@ -69,51 +69,27 @@ fi
 # FORCE IT TO PRODUCE EXACTLY THE REQUIRED NUMBER OF EVENTS
 #########################################
 
-echo "produce backup tarball"
-cd $LHEWORKDIR
-tar cfJ gridpack.tar.xz ./
-cd process
-
 produced_lhe=0
 run_counter=0
 run_random_seed=$rnum
 remaining_event=$nevt
 
-echo "start the loop"
 while [ $produced_lhe -lt $nevt ]; do
-
+  
   let run_counter=run_counter+1 
   echo Running MG5_aMC for the $run_counter time
   remaining_event=$(($remaining_event - $produced_lhe))
-
-  if [  $run_counter -gt "1" ]; then
-      echo "refresing the working directory (i.e. remove mgbasedir, process and and untar gridpack again"
-      cd $LHEWORKDIR
-      rm -rf mgbasedir process
-      tar xf gridpack.tar.xz
-      cd process
-  fi
   
   #generate events
   echo run.sh $remaining_event $run_random_seed
-  ./run.sh $remaining_event $run_random_seed # TO REALLY PRODUCE EVENTS
+  ./run.sh $remaining_event $run_random_seed
   
   produced_lhe=$(($produced_lhe+`zgrep \<event events.lhe.gz | wc -l`))
-  mv events.lhe.gz $LHEWORKDIR/events_${run_counter}.lhe.gz
-  # cp $LHEWORKDIR/events_${run_counter}.lhe.gz $LHEWORKDIR/events.lhe.gz # TO DELETE
-  
+  mv events.lhe.gz events_${run_counter}.lhe.gz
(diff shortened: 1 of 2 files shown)


perrozzi:
@qliphy @bendavid 
I didn't dare modifying the NLO script because is a bit more involved.
Modification under testing, will report asap

franzoni:
hello @perrozzi 
the logic of incrementing +1 the random seed can do
PROVIDED 
it does "collide/interfere" with the logic used by wmagent to set seeds

Please follow this"
https://hypernews.cern.ch/HyperNews/CMS/get/wmDevelopment/787.html

bendavid:
This issue simply does not exist for NLO, since the part of the code which produces this behaviour is not used there.

bendavid:
Also I'm not sure the gridpack can safely be run multiple times without cleaning intermediate files.

Suggest to compare LHE files from the same random seed and number of events from the cleanly untarred gridpack Vs the reused one.

vlimant:
given that asking for N (N too large) might fail also on retry, I have a suggestions.
   Do not retry more then X times (X=10?)
OR
   Add the "n_i" events from each internal run in a final lhe file, until you reach N

vlimant:
commented without reading the code. looks like "Add the "n_i" events from each internal run in a final lhe file, until you reach N" is already what is done

bendavid:
Good news is that lhe files produced with the same random seed and requested number of events seem to be identical whether the gridpack is cleaned and untarred again or not.  So this part seems safe at least.  (And then the only remaining possible issue is the logic for the random seeds)

vlimant:
from earlier discussion, doing +1 or +=(any deterministic number) is fine

perrozzi:
@bendavid then I can probably revert the last commit (b2a14db) right?

bendavid:
Yes that's right.  (and better revert it, because repeated untarring can put high load on the worker node filesystem)

bendavid:
What's the purpose of limiting the number of events per iteration?

perrozzi:
fixed a stupid bug and added a limit to 5k events per iteration.
In a test I was able to produce 50K with a problematic gridpack

vlimant:
if that works, we could go back to 8 cores ...

perrozzi:
others options (also in combination) would be:
-  to limit the number of events per iterations only when  `$ncpu > 1`
- limit the number to `$nevt/$ncpu` instead of 5000
- limit the number only starting from the second iteration whenever is required
@bendavid do you have any suggestion/comment?

vlimant:
Why limiting and not just let the program produce the amount of events it can, and add those together ?

perrozzi:
maybe was not clear, I just rephrased. every option is referring to limit the number of events per iterations

perrozzi:
I think that is good to limit the number of events per iteration to reduce the overhead.
my understanding is that whenever you ask for 25K events, the program will desperately try to create them all, so if the seed is problematic you will need to wait a lot of time and get in output only few events.
I have a case for which I was repeatedly asking for 25K events per iteration and I was getting only few tens or hundreds of events each time in 2h, for all the consecutive seeds.
The same gridpack produced flawlessly 50K events in chunks of 5K, each iteration lasting a couple of minutes.

Another way of saying, always according to my understanding, is: when there is a problematic seed, to generate each event requires much more time so better to keep the number of events per seed low and run on many seeds to maximize the probability to get all the events on time.

vlimant:
Ok

vlimant:
for what concerns the seed collision. At the moment, there is no guarantee that two jobs are getting different seeds since they are populated at job runtime using ::populate

https://github.com/cms-sw/cmssw/blob/CMSSW_9_3_X/IOMC/RandomEngine/python/RandomServiceHelper.py#L200

With the implementation of +=1 on the seed, running some basic toy

1000 jobs
max seed number 900000000
flat prob of 10 retries
tested 10000 times
mean collision 0.0168
fraction of no collision 99.48 %

and 

100 jobs
max seed number 900000000
flat prob of 100 retries
tested 10000 times
mean collision 0.0166
fraction of no collision 99.95 %

 so it would be better to have less jobs doing more +=1 to reach higher stats (running long 8-core jobs) to limit the collisions.

With the schema of s = int(s*1.5)% 900000000

1000 jobs
max seed number 900000000
flat prob of 10 retries
tested 10000 times
mean collision 0.0424
fraction of no collision 98.50 %

 and 

100 jobs
max seed number 900000000
flat prob of 100 retries
tested 10000 times
mean collision 0.2443
fraction of no collision 98.90 %

perrozzi:
ok, so according to an example of failing workflows, the number of events per job should be around max 25K events. If we use 5K flat, this means ~5 resubmissions (assuming that the number of failures will become ~negligible). even with 1000 jobs / workflow the collision should be small. The number of events can be increased to 10K for the first iteration and 5K for the others, or any other combination like 10K chunks until less events are produced, then reduce to 5K...
comments?

one last thing I want to try is the following: to some extent, it is clear that this behavior depends on BOTH the number of events requested AND the seed. I want to check if the first few events are always the same with the same seed but different number of requested events

perrozzi:
likely bingo: the events **are** different, I just did a test with gridpack `/afs/cern.ch/user/y/ytakahas/public/forEXOMC/SingleLQ3ToTauB_5f_madgraph_LO_t-channel-M900_slc6_amd64_gcc481_CMSSW_7_1_30_tarball.tar.xz` 
launching as
`./runcmsgrid.sh 101 7233555 8`
`./runcmsgrid.sh 255 7233555 8`
results attached
[events_presys_255_7233555_8.lhe.txt](https://github.com/cms-sw/genproductions/files/1517335/events_presys_255_7233555_8.lhe.txt)
[events_presys_101_7233555_8.lhe.txt](https://github.com/cms-sw/genproductions/files/1517336/events_presys_101_7233555_8.lhe.txt)

vlimant:
wait. by not changing the random seed anymore, you will end up doing "run 5k original_seed" several times in a row, and get the same events over and over

bendavid:
Hi,
Running with the same random seed but different number of events IS expected to produce different events, since the splitting across subprocesses is different, but there could still be nasty statistical correlations, so definitely each run should use a different random seed.

perrozzi:
correct. I had an extendet chat with Olivier to understand better.

At last he said another thing (which I will not consider for the moment)

./bin/internal/gen_ximprove.py:727:    max_iter = 9
./bin/internal/gen_ximprove.py:907:    max_iter = 9
./bin/internal/gen_ximprove.py:1025:                    'maxiter': self.max_iter,
./bin/internal/gen_ximprove.py:1170:                    'maxiter': self.max_iter,
./bin/internal/gen_ximprove.py:1232:    max_iter = 6
./bin/internal/gen_ximprove.py:1257:            self.max_iter = 5
./bin/internal/gen_ximprove.py:1432:        elif step >= self.max_iter:
./bin/internal/gen_ximprove.py:1460:                                              conservative_factor=self.max_iter)
./bin/internal/gen_ximprove.py:1468:        elif step < self.max_iter:
./bin/internal/gen_ximprove.py:1469:            if step + 1 == self.max_iter:
./bin/internal/gen_ximprove.py:1475:
so in this file (inside the gridpack from 2.4.2) you can increase max_iter if you want/need
this is to increase the number of iteration inside MG5 to limit the need for you to use multiple seed
you might want to also change this one:
./Source/run_config.inc:21:      integer    max_iter                   !Maximum number of iterations
./Source/run_config.inc:22:      parameter (max_iter = 9)              !during refinement
allowing more iteration will reduce the number of times that MG fails to generate events

bendavid:
Ok, I'm still not convinced there is a benefit to having a maximum number of events per iteration (but I also don't think it does any harm)

perrozzi:
merge fix added, changes propagated to several mg5_amc gridpacks, PR ready to be merged.
At some point we will likely to propagate this to all the existing mg5_amc gridpacks (script is ~ready)

PR #1580, opened 2017-12-06
Added configs for WW semileptonic production

valsdav:
The configs work with submit_phantom script.

### DIFF
--- a/bin/Phantom/cards/production/13TeV/WWSemileptonic/mu_ewk+qcd.cfg
+++ b/bin/Phantom/cards/production/13TeV/WWSemileptonic/mu_ewk+qcd.cfg
@@ -1,0 +1,189 @@
+# ---- ----  ---- ----  ---- ----  ---- ----  ---- ----  ---- ----  ---- ----  ---- ---- #
+#                                                                                        #
+#           template cfg file for the usage of the phantom generator in CMS              #
+#                                                                                        #
+# see https://twiki.cern.ch/twiki/bin/view/CMS/PhantomGeneratorCMS for more details      #
+#                                                                                        #
+# ---- ----  ---- ----  ---- ----  ---- ----  ---- ----  ---- ----  ---- ----  ---- ---- #
+
+
+[general]
+# where the zipped version of the compiled phantom code can be found
+# and where the gridpacks will be created
+
+# CMSSW environment for the gridpack creation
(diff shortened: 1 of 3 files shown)


govoni:
these are OK for the private studies of interference they are intended for.
For any official productions, the CMSSW release -- and hence the architecture -- would have to match the one used in the production.

govoni:
+1

perrozzi:
@govoni is there a runcmsgrid.sh inside phantom's gridpacks? can you please point me to one of them?

govoni:
Ciao Luca, 
yes there is one. This is the unxipped version of the gridpack, for which I am asking an MCM text the SMP GEN contacts:

/afs/cern.ch/user/g/govoni/public/gridpackTesting_SS_CMSSW_930

Please note that these cfg files Davide is asking are for LHE-only studies for interference evaluation, which are committed here for bookkeeping in CMS only.

perrozzi:
it is advisable to have a scram_arch/cmssw environment settings like this
https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/runcmsgrid_powheg.sh#L29-L61
where `SCRAM_ARCH_VERSION_REPLACE` and `CMSSW_VERSION_REPLACE` are replaced with the ones used to create the gridpack in the gridpack creation script

govoni:
the right CMSSW version and scram architecture to be used for the LHE events production are already enforced in the script.

For example:

scram -a slc6_amd64_gcc630 project CMSSW CMSSW_9_3_0
cd CMSSW_9_3_0/src
eval `scram runtime -sh`
cd -

this comes from here: 
https://github.com/cms-sw/genproductions/blob/master/bin/Phantom/submit_phantom.py#L146

where the cmssw and architecture setup is hardcoded:

https://github.com/cms-sw/genproductions/blob/master/bin/Phantom/submit_phantom.py#L167

perrozzi:
a more flexible structure like the one suggested above is more advisable, just to avoid problems (for instance if 930 would be deprecated at some point)

govoni:
Ciao @perrozzi , it makes sense for the release, less for the architecture. Assuming that compatible pdf libraries are used, the release change is probably not a problem, while the architecture should stick to the one used for the gridpack generation. I'll prepare a modification of the script.
Can we move this discussion to the submit_phantom.py file, though? This PR is not related to that.

PR #1597, opened 2017-12-18
MC 2016 Wprime_VBF_WZ_inclu

IreneZoi:
I am adding script and example cards for Wprime_VBF_WZ_inclu for official 2016 production


.gitignore
```diff
@@ -0,0 +1,5 @@
+# Compiled source #
```
perrozzi: this file is not relevant

perrozzi:
we usually ask to provide one example mass point and then a script to produce the others to avoid overloading the repository, can you please update the PR?

perrozzi:
hi, you apparently added files but didn't remove the others

IreneZoi:
Sorry, I didn't understand what I should remove. I now have the script to produce all the mass points, the directory that is used by the script to produce the mass points and an example for a certain mass point. Is it this last one that I should remove? Thanks

perrozzi:
I believe you have, in this PR, all the files for all the mass points and not just a representative datacard and the script to produce the other points (in fact there are 150 files). is this correct?

PR #1603, opened 2017-12-20
adding relative error plot to the output

govoni:
for a better understanding of the chi2 trends, adding the relative uncertainty of the calculations to the pdf outputs in the verification phase.


govoni:
of course, already tested.

PR #1613, opened 2017-12-29
Fragment for HH->bbZZ->bb4L production

ndefilip:
Fragment for HH->bbZZ->bb4L production


ndefilip:
Fragment for HH->bbZZ->bb4L production

perrozzi:
why pythia only?
@OlivierBondu can you please have a look

OlivierBondu:
@perrozzi : this is as for the other HH requests: we have common gridpacks 'up to HH', and then it's up to each analysis to have the pythia decay fragment ready, which is this PR

The decay fragment look fine to me

PR #1617, opened 2018-01-10
updating st_FCNC datacards 2015 to 2017 settings

wajidalikhan:


bin/MadGraph5_aMCatNLO/Utilities/scripts/copyOldCardsTo2017.sh
```diff
@@ -9,7 +9,7 @@ if [ -z $1 ]; then
 fi
 
 base_folder=$(git rev-parse --show-toplevel)/bin/MadGraph5_aMCatNLO
-
+echo $base_folder
```
perrozzi: are these modifications needed in the central repository?!?

perrozzi:
can you please upload only one representative datacard of each type and a script to produce the others? this to reduce the number of files in the repository and avoid slow downs

bin/MadGraph5_aMCatNLO/Utilities/scripts/copyOldCardsTo2017.sh
wajidalikhan: Hi Luca, 
Or instead of committing to central repository we can add a disclaimer on tiwki to use the script with caution, the modifications are for personal use, I am still working on it. 

The old script has a problem that it does not change the pdf label to the new recommended one (306000) 

https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/Utilities/scripts/copyOldCardsTo2017.sh#L53-L57 

as a result it just takes in the old pdf 230000. This has been cross-checked by looking at the LHE file produced with the modified script and the already existing script. The modified script produces the correct modification in the *run_card*.

wajidalikhan: Ahh ugly!, There are other other TTbar files are also accidentally got committed I will close the PR and will commit only st_FCNC samples.

perrozzi: @kdlong can you please crosscheck?

wajidalikhan: @perrozzi 
for the st_FCNC there are 19 different processes each having different run_card, implementing the generate processes in the script will make the script less readable and more difficult to spot the possible errors and also one has quick cross-check of data cards using a web browser using the link given in the gen-fragment, otherwise one has to download the tarball and than to extract the input cards. 

It is good to have a script in case of samples with a particle of different spin with a variable mass points, which generates tons of files. What do you think? Does this sound convincing?

PR #1619, opened 2018-01-11
Add missing argument to submit_condor script

kdlong:


perrozzi:
what is this `ALL` argument?

kdlong:
Oops I totally missed this. There is now and option to say which steps to run, as CMSConnect splits the steps by diagram generation and integrate (with MadSpin intended to be separate step at some point). For normal condor submit we just run in one go. The new argument previously hadn't been added.

PR #1626, opened 2018-01-15
Added MCFM+JHUGen cards for SIG and BSI(1,10,25)

carolhungwt:


perrozzi:
I see a mechanism to prevent producing more events than needed, but not one to prevent producing less than needed, which caused issues in production. any comment?

carolhungwt:
in commit 84f43d9, a mechanism of dealing with under-production has also been added

PR #1632, opened 2018-01-17
WplusHJ, WminusHJ and ggHZ template cards in all modes

gourangakole:
Used the approved sample card from the respective directories to make a template card. And used the sameto get the final card with different higgs mass. 
"Width" is taken from [here](https://arxiv.org/pdf/1307.1347.pdf#page=284)


perrozzi:
I would suggest to move all the folder under a Higgs folder to improve the searchability. Also, it would be good to have the exact POWHEG process name somewhere in the folder names for better clarity (this has not always been done so far, we will need to ensure homogeneity

chayanit:
ciao @perrozzi,  we have already committed many Higgs processes under this folder. are you suggesting to reorganize everything from Higgs? can you clarify a bit? I recall we used to have that in 2016 but it was a bit messy that we have cards related to Higgs everywhere inside and outside subfolder named Higgs. btw, what do you mean by the exact POWHEG process? thanks :)

perrozzi:
I would like to have all the Higgs-related folders inside a main folder called Higgs, because otherwise you need to browse between other hundres of folders and processes...
Concerning the POWHEG process, I think at the moment is ok as it is, I get it back :-)

chayanit:
@perrozzi we made a new PR #1633 to include all the cards here as well. Could you close this PR and merge another one? thanks!

PR #1633, opened 2018-01-18
Reorganisation of the Higgs cards

ShevchenkoRostyslav:
Merge PR from gourangakole manually: #1632 . Reorganisation of the Powheg Higgs cards to a separate folder


perrozzi:
Hi, things look mostly ok.
I have a question though, since I only see new files: could it be the case that you didn't move some of the existing files and created additional copies? (I hope what I say is clear)

ShevchenkoRostyslav:
Hi @perrozzi ! You are right, I forgot to git rm the old folders. Thanks for pointing this out! I just committed the cleaning.

perrozzi:
you can always, directly use `git mv file`

PR #1634, opened 2018-01-19
Cards for cross-check production of Mono-Z spin0

AndreasAlbert:
These samples will be used to cross-check a private production of mono-Z spin0 samples (2016 data).
The private production is necessary because the already officially produced samples are faulty and we need the results urgently.


python/ThirteenTeV/MonoZ/dmsimp_spin0/MonoZ_DMSimp_Spin0_cff.py
```diff
@@ -0,0 +1,45 @@
+import FWCore.ParameterSet.Config as cms
+
+externalLHEProducer = cms.EDProducer("ExternalLHEProducer",
+    args = cms.vstring(''),
```
perrozzi: this is only a template, right?

AndreasAlbert: Yes exactly. It's the same hadronizer for both points.

PR #1635, opened 2018-01-22
modified gridpack script to avoid compilation problems in ggHH

agrohsje:
disabled unneeded pythia 8 part of ggHH makefile


perrozzi:
have you checked that it doesn't break up `gg_H` ?

agrohsje:
Good point. For the time being it doesn't, but to be sure there is never an issue, I split the if statement.

perrozzi:
sorry, I don't want to be too picky but leaving the code as it was and add another if only for gg_HH would be better

agrohsje:
Done.

PR #1637, opened 2018-01-22
scale uncertainties for fragments needed for H->ZZ

hroskes:
copied from
https://cms-pdmv.cern.ch/mcm/requests?range=TOP-RunIIFall17wmLHEGS-00033,TOP-RunIIFall17wmLHEGS-00065&page=0&shown=127

and deleted the old scale variations, since they're not needed anymore


efeyazgan:
"deleted the old scale variations, since they're not needed anymore" --> Which ones are your talking about? ME or PS scale variations? We need to have them both.

hroskes:
They had 
```
TimeShower:renormMultFac = 4.0    #or 2.5
SpaceShower:renormMultFac = 4.0   #or 2.5
```

I thought that was the same as the new ones, which have for example
`isr:muRfac=0.25, fsrConHi fsr:muRfac=0.25,isrConLo isr:muRfac=4.0,fsrConLo fsr:muRfac=4.0`

efeyazgan:
OK, so you mean you do not use the parton-shower-scale-varied hadronizer config files. I think it would be good to still keep them (until the parton shower weights are used by many more processes).

hroskes:
I can put them back if you want, but I'm a little confused about their purpose.  If you're going to run with those configurations, you may as well just run with the equivalent ones that include weights for all variations.. am I missing something?

efeyazgan:
No, I think it is fine. I wanted to make sure you did not change the ME variations.

hroskes:
Thanks!

PR #1643, opened 2018-01-24
WGGJets and ZGGJets for 2017

gourangakole:
Original card are taken from [1](https://github.com/cms-sw/genproductions/tree/pre2017/bin/MadGraph5_aMCatNLO/cards/production/13TeV/WGGJets_5f_LO_MLM), [2](https://github.com/cms-sw/genproductions/tree/pre2017/bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZGGJets_ZToHadOrNu_5f_LO_MLM) and [3](https://github.com/cms-sw/genproductions/tree/pre2017/bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZllGGJets_5f_LO_MLM)
then changed the pdf and applied cleaning.


efeyazgan:
You should set
T        = pdfwgt

gourangakole:
Thank you @efeyazgan . Could you check once please.

efeyazgan:
auto_ptj_mjj should be set to T (unless there is a specific reason for not doing it).

gourangakole:
Hello @efeyazgan .... updated that too. I was not sure that is recommended for 2017.

PR #1644, opened 2018-01-25
Adding Njet binned Z->ll, Z->nunu, W->ellnu cards 

adewit:
This adds 0/1/2-jet binned cards for Z->ll, Z->nunu and W->ellnu (aMC@NLO)

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/dyellell0j_5f_NLO_FXFX/dyellell0j_5f_NLO_FXFX_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/dyellell0j_5f_NLO_FXFX/dyellell0j_5f_NLO_FXFX_proc_card.dat
@@ -1,8 +1,6 @@
 set low_mem_multicore_nlo_generation True
 
 import model loop_sm-ckm_no_b_mass
-#switch to diagonal ckm matrix if needed for speed
-#import model loop_sm-no_b_mass
 
 define ell+ = e+ mu+ ta+
 define ell- = e- mu- ta-
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/dyellell0j_5f_NLO_FXFX/dyellell0j_5f_NLO_FXFX_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/dyellell0j_5f_NLO_FXFX/dyellell0j_5f_NLO_FXFX_proc_card.dat
@@ -1,0 +1,13 @@
+set low_mem_multicore_nlo_generation True
+
+import model loop_sm-ckm_no_b_mass
+#switch to diagonal ckm matrix if needed for speed
+#import model loop_sm-no_b_mass
+
+define ell+ = e+ mu+ ta+
+define ell- = e- mu- ta-
+
+generate p p > ell+ ell- [QCD] @0
+
+output dyellell0j_5f_NLO_FXFX -nojpeg
+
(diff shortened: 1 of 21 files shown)


bin/MadGraph5_aMCatNLO/cards/production/13TeV/dyellell0j_5f_NLO_FXFX/dyellell0j_5f_NLO_FXFX_proc_card.dat
```diff
@@ -0,0 +1,13 @@
+set low_mem_multicore_nlo_generation True
+
+import model loop_sm-ckm_no_b_mass
+#switch to diagonal ckm matrix if needed for speed
+#import model loop_sm-no_b_mass
```
efeyazgan: delete this commented input

efeyazgan:
Don't you want to add 
$DEFAULT_PDF_MEMBERS = reweight_PDF ! reweight to get PDF uncertainty ?

adewit:
I just made a quick new gridpack with $DEFAULT_PDF_MEMBERS = reweight_PDF instead of .true. = reweight_PDF and that seems to have completely turned the PDF uncertainty off! (with the setup in the current version of the cards at least we get the replicas for pdf306000). Where does the replacement of $DEFAULT_PDF_MEMBERS happen? On the master branch I see it here: https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/gridpack_generation.sh#L491, but it's missing from this script on the mg26x branch.

kdlong:
In 26x the PDF weights get added by a separate module. You can also add dynamic scale  variations, which is probably a good idea. One could even implement a m_{ll} scale pretty easily, I can point to an example if there is interest in implementing this.

It's run here: https://github.com/cms-sw/genproductions/blob/mg26x/bin/MadGraph5_aMCatNLO/runcmsgrid_LO.sh#L87

which is filled by https://github.com/cms-sw/genproductions/blob/mg26x/bin/MadGraph5_aMCatNLO/gridpack_generation.sh#L573-L574

If you ran the event generation and still see no weights, can you send me the gridpack?

efeyazgan:
@kdlong could you please add specific instructions for 26 in https://twiki.cern.ch/twiki/bin/viewauth/CMS/Moriond18MC#Checks_to_the_datacards_BEFORE_T

adewit:
I see, so the problem is just that I have store_rwgt_info set to False and that is why it's not running. Will fix this shortly. At that point, does the configuration of 'reweight_scale' and 'reweight_pdf' in the run card still matter at all? I guess I can copy the settings from https://github.com/cms-sw/genproductions/blob/mg26x/bin/MadGraph5_aMCatNLO/cards/examples/wplustest_5f_NLO/wplustest_5f_NLO_run_card.dat#L95-L103

kdlong:
At least one has to be set to true (easier for it to be scale). See https://answers.launchpad.net/mg5amcnlo/+question/654417

It's kind of annoying

adewit:
Ok, thanks. So I think it should be correct now but please let me know if there are other things that need to be addressed. I'll wait for this PR to be merged before reproducing the gridpacks.

perrozzi:
things look ok to me, @kdlong @efeyazgan ?

adewit:
I know everyone has lots of other things to do, but the 2-jet gridpacks can take several days to produce so it would be good to set them going as soon as possible. I do want to make sure there are no further issues with the cards before doing so, so it would be great if any further comments (or a merge if everything is ok) could follow soon!

perrozzi:
I think that things are ok as far as you make sure that the scale and pdf weights are there.
I would have a rethorical question concerning the mg5_amc version to use, whether it should be 260 or 242 given that we don't need the bias weight anymore and 260 in principle has not finished the validation.
I would suggest to move on with 260 as a pragmatic solution as you always used that so far.

kdlong:
I don't see a strong argument against 260. The cards looks good to me. The reweighting can be tweaked after the gridpack is made if it has to be, as long as the store_rwgt_info and reweight_scale are set to true (which they are).

PR #1655, opened 2018-02-02
TT with modified top width

jfernan2:
- TTbar configs with modified top width: 0.7 , 0.85, 1.15 and 1.3 times the nominal top width of 1.31GeV
- Fixes in configs for old Moriond17 Semileptonic hdampUP/Down

### DIFF
--- a/bin/Powheg/production/2017/13TeV/TT_hvq/TT_hdamp_widthx0p7_NNPDF31_NNLO_dilepton.input
+++ b/bin/Powheg/production/2017/13TeV/TT_hvq/TT_hdamp_widthx0p7_NNPDF31_NNLO_dilepton.input
@@ -5,30 +5,32 @@
                    ! If POWHEG is interfaced to a shower MC, refer to the shower MC
                    ! documentation to set its seed.
 
-!Heavy flavour production parameters
 
-numevts NEVENTS    ! number of events to be generated
-iseed   SEED       ! Start the random number generator with seed iseed 
+
+numevts NEVENTS
+iseed SEED
 ih1   1        ! hadron 1
 ih2   1        ! hadron 2
 
-! To be set only if using LHA pdfs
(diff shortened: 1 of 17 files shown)


bin/Powheg/production/2017/13TeV/TT_hvq/TT_hdamp_widthx0p7_NNPDF31_NNLO_dilepton.input
```diff
@@ -0,0 +1,89 @@
...
+lhans1 306000
+lhans2 306000
+! NNPDF31_nnlo_as_0118
+ebeam1 6500    ! energy of beam 1
+ebeam2 6500    ! energy of beam 2
+qmass  172.5     ! mass of heavy quark in GeV
+facscfact 1    ! factorization scale factor: mufact=muref*facscfact 
+renscfact 1    ! renormalization scale factor: muren=muref*renscfact 
+
+hdamp 237.8775
```
perrozzi: is hdamp re-fitted or is kept the same?

jfernan2: It is kept the same as nominal sample, the process and tuning is the same so no need to refit, just a top width change e.g.
https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/production/2017/13TeV/TT_hvq/TT_hdamp_NNPDF31_NNLO_dilepton.input#L26
Thanks

perrozzi:
I would suggest to have a representative card for one width and a script to create the others

jfernan2:
But now that the cards have been created, should I revert to your solution? We do not expect any additional width sample to be created/requested.

perrozzi:
ok, fine for me. @efeyazgan can you please crosscheck the cards?

PR #1656, opened 2018-02-04
Update systematics call to so weights aren't duplicated

kdlong:


perrozzi:
I assume this is validated and ready to merged, right?

kdlong:
Yep

PR #1658, opened 2018-02-05
Fix NLO reweight with precompilation

kdlong:
This is NOT yet ready to merge. Still testing but putting here for reference.

Run reweight at NLO with pre-compilation of all reweight modules to avoid compilation on the grid. Also clean up synatx for reweight and add some error checking.


kdlong:
This is working for me for the Higgs mass effects samples

@AndreasAlbert maybe someone in Exo can check with your NLO samples? @maierbenedikt was interested in the past I know.

@senka @mmozercern this could also be tested for the aTGC NLO samples

agrohsje:
Hi Kenneth 
https://github.com/kdlong/genproductions/blob/c4a0df22ae441671739748d37461a84405ca601c/bin/MadGraph5_aMCatNLO/gridpack_generation.sh#L542 and the next 2 lines should have no madevent as there is cd $WORKDIR/process/madevent in line 527. 
Cheers, Alexander.

agrohsje:
Hi Kenneth,
the pathes in reweight still needs to be adjusted like e.g.: 

#reweight if necessary                                                                                                                                                                                              
if [ -e ./madevent/Cards/reweight_card.dat ]; then
    echo "reweighting events"
    mv events.lhe.gz ./madevent/Events/GridRun_${rnum}/unweighted_events.lhe.gz
    cd madevent
    echo "0" |./bin/madevent --debug reweight GridRun_${rnum}
    cd ..
    mv $LHEWORKDIR/process/madevent/Events/GridRun_${rnum}/unweighted_events.lhe.gz $LHEWORKDIR/process/events.lhe.gz
fi

Cheers, Alexander.

agrohsje:
Hi @kdlong,
using gcc6 indeed avoids  the f2py/gfortran problems. However, there is one remaining issue and that is the --remove_wgts=all option in the systematics section. What was the reason to add it ? The command removes the model reweighting. 
Anyway, modulo this  and the above changes are sufficient to get LO reweighting running. 
Cheers, Alexander.

kdlong:
Thanks Alexander. The ```remove-wgts=all``` options is something they added in 2.6, but I didn't think about the fact that it's clearly not appropriate here. 

When the file is initially created, you have to have some weights store for the PDF/scale weight info to be scale (reweight_scale = True usually for us). The systematics module then adds these weights in addition, and this option was to remove such duplicates. However, clearly we need to be more careful with it when additional weights are added that shouldn't be removed. There are other options besides "all" we could use if necessary [1]. Looking at the file produced without this option, do you think one of these is appropriate?

[1] https://cp3.irmp.ucl.ac.be/projects/madgraph/wiki/Systematics

agrohsje:
Hi @kdlong , I don't see any duplicate entry. I have the model reweighting first and then it starts with the scale variations. Just dropping the option works for me and the output looks like I want it to look.

perrozzi:
sorry, just ro understand: this needs further polishing before being merged, right?

kdlong:
@perrozzi Yes it does

@agrohsje Can you send me your cards/LHE file? In general the scale weights are added during the first step and then added again at the systematics step, which is kind of annoying as it adds duplicate weights and screws up the weight ordering. Maybe this doesn't happen with the reweight step. I think it would be worth understand when this does and doesn't happen and then setting remove_wgts/keep_wgts appropriately.

agrohsje:
Hi @kdlong 
I copied the gridpack path to 
/afs/cern.ch/user/g/grohsjea/public/gluontop_rwgt
Cheers, Alexander.

perrozzi:
ciao, is this pr still relevant? if so, it has conflicts, and I guess it should be merged before to merge 26x into master

perrozzi:
ping. please solve conflict asap :)

kdlong:
Conflict should be resolved. Testing now. @agrohsje could you run a test job as well?

agrohsje:
Hi Kenneth, I am going to test your branch now. I will ping you once I am done.

agrohsje:
@kdlong my LO test worked perfectly fine. Did you test NLO reweighting? If you have cards at hand with reweighting I could run it as well.

kdlong:
@agrohsje I also ran LO. It worked, but I'm having a problem with MadSpin after the fact and trying to figure out if I made a mistake in the cards or if it's a bug in the script. Can you try at NLO? It would also be good if it uses MadSpin.

agrohsje:
@kdlong my LO test worked perfectly fine. Did you test NLO reweighting? If you have cards at hand with reweighting I could run it as well.

agrohsje:
Hi @kdlong , I tested NLO + Madspin and it is working well. I tested NLO+Madspin+Reweighting and I got an error. However, this looks like a MG5_aMC@NLO problem that I should report:
http://www.desy.de/~agrohsje/hidden/validate_2.6/events.lhe.gz_0.lhe_tag_1_debug.log 
What was the problem you encountered ?

agrohsje:
The key error is related to a process change definition. I assume this is because I am running on decayed events? 
I will do a PR for an additional patch that Olivier provided for correct tags asap and I will see how to fix the NLO problem.

perrozzi:
ping

kdlong:
The problem I had came from the directory structure being messed up, so that the MadSpin run information wasn't stored in the correct location. This should be fixed now. I'm testing again to make sure.

agrohsje:
@kdlong 
do you have an example that uses madspin + reweighting @ nlo  or should we follow up with olivier ? I wanted to ask him a few more things, so I could do that as well?

perrozzi:
ping

agrohsje:
So NLO + reweighting without madspin doesn't work for me either. I ended up with the "same" key error.
Do you have any news on this combination?

agrohsje:
I still had the Madspin issue for LO and fixed it in my version. I have problems reweighting the chunks of 2500 events. It is currently working for the first 2500 events, but then it silently crashes. I will see how to fix it. Another point is that the reweighting info is not kept in the header when using Madspin. Should be easy to fix. Moreover, I got a patch from Olivier so that the first weight has the correct name.

kdlong:
@agrohsje can you make a pull request to my branch with the changes you made?

PR #1659, opened 2018-02-05
Hig vbf phantom 2017

skeshri:
Hi @govoni 

Please may you check this PR. I have made template config files for 4l and 2lnu final states. Both want to use same selections only the final states are different.

If you would like to see gridpacks then I will share them with you to have a look. Please let me know.

Thanks
Sumit


govoni:
Dear Sumit, as a matter of fact there might be an issue in Phantom which authors are checking, so I would wait before submitting. I will meet them in two days from now, and will learn more then.

skeshri:
Dear @govoni 

Many thanks for this information. I will wait for update from you.
So my understanding is if the problem get solved you will upload new version of compile code. correct?

Thanks
Sumit

govoni:
correct

govoni:
Dear Sumit,

they are working on the fix. Is it getting late? Authors claim that the effect of the fix should be minor, so you might want to consider going forward if needed. In this case, I would recommend telling them for what analysis exactly the sample is meant, to make sure we're not doing the wrong thing.

pietro

skeshri:
Dear Pietro

Sorry I missed your email somehow.
I don't know if the fix has already done by now. 
We are required VBF off shell SM Higgs sample. If that fix is not required then I already have gridpacks for those samples.

Sumit

govoni:
Dear Sumit,

I added some questions for you in this PR.
Still, I would prefer to decouple the cfg files from the modification of the submission script.

pietro

skeshri:
Dear @govoni 

As I understood from your suggestion I have decoupled all the config files and uploaded them. 
I don't know If you meant the same?

Regards,
Sumit

govoni:
Dear @skeshri,

thanks a lot for your quick reply.
Here are my comments: 
1. GIT is a big mess in general as a tool ;)
2. what is this file?
bin/Phantom/cards/production/13TeV/HZZ_VBFoffshell_Phantom/Higgs_125_4lOr2l2v_OnlyB.cfg.tmpl
it has no channel, nor folder name. Why is this part of this pull request?
3. I could not find any datacards for the final states with two charged fermions and two neutrinos, all fo the same flavour (like: "e e_ ve ve_" for example), are you planning to add those? In this case, are you still planning to veto top quarks?
4. the selections are fine I guess, remembering that the mll at 4 GeV will mean a lot of inefficiency, that lepton et and pt are extremely correlated and one will have to remember to apply both in the analysis as well, and that other selections might anyhow sculpt the mll distribution.

Cheers,

pietro

skeshri:
Dear Pietro,

1) Yes, it is some times ;)

2)  These are the template files which I'm using as an input to a script  to produce gridpacks for all channels. But any way I forgot to remove them I will remove them as I have uploaded all the config files for each channel separately. 
3) This I discussed in my group at some point. All of the same flavours will be considered in interference part, as we can't explicitly tell phantom what Higgs will decay in either ZZ or WW? So same flavour would be part of interference. Please tell me if this consideration is wrong. 
I really don't quite understand this top quark veto in this point.

4) We would like to keep the same selections as they were in 2016.

Thanks,
Sumit

perrozzi:
so can the pr be merged or further work is needed?

govoni:
The top quark veto issue is the following. With the final state of 2 charged leptons and 2 neutrinos, also ttbar can be produced. Of course the two lepton pairs (l+l-  and vv respectively) will not resonate on the Z mass, so this contribution is probably not relevant, it's just for you to know.
It is correct that you cannot ask Phantom to choose between WW and ZZ. I did not understand the consideration on the interference, but this anyhow is not a show stopper for this particular PR. Let's discuss this via email if needed.

govoni:
+1

skeshri:
Please don't merge it. I will need to make some update before.
Give me some time. Will update soon and then merge.

Thanks for understanding.

govoni:
I will review those as well then :)

perrozzi:
what is the status here?

skeshri:
Hi @perrozzi,

We have checked some distribution by analyzing LHE files and compare them with earlier(2016) phantom samples. we found some issue (a min pt cut on jet was enabled). This has been disabled. So all the gridpacks are redone. still some of them yet to finish. 
Also we are understanding more about what we are getting. 
I will update very soon! Thanks for patience.

Best,
Sumit

skeshri:
Dear @perrozzi and @govoni ,

Sorry for delaying this. We were trying to understand distributions in 4l and 2l2nu channel. It took some time to digest the distribution comparisons between 2l2nu and 4l final states. Since we are making request for phantom 2l2nu sample for the very first time, so we are extra careful for this request. Now we understand and satisfied with the results. We also debugged the problem that was coming due to pt cut of 10 GeV on jets. Now there is no cut on jet pT. May you please merge this request now? 
Thanks for the patience!

Best Regards,
Sumit

govoni:
Hello @skeshri, do you have a reference to the two issues, namely the understanding of the distributions in the comparison between 4l and 2l2v, as well as the jet pt problem? I don't see a big issue in putting a minimal cut on the jets pT, while I can see possible reasons of strange results in the calculations (vicinity to the beam) and waste of events (10 GeV is an extremely low cut for jets production).
Best regards.

skeshri:
Hi @govoni 
Sorry for not be specific.  This Jet pt cut is not the problem for Signal only case but it is problematic for Background only. So I'm attaching link [1] here for the very initial study. Some follow up after the [1] can be found in [2]. Please let me know if you have any doubts from slides.

Regards,
Sumit

[1] https://indico.cern.ch/event/718326/contributions/2957693/attachments/1627217/2591795/FollowUp_updated.pdf
[2] https://indico.cern.ch/event/718599/contributions/2975188/attachments/1635361/2608871/FollowUp_3_Phantom_sample.pdf

govoni:
Hello @skeshri ,

thanks a lot for the details. I am a bit surprised by some of your plots, like the ones in page 14: if you do not put the pt > 10 GeV cut, you get the red line, right? Then, I am surprised that the red and blue do not overlap in the central region: the red should always be higher than the red.
Another example is page 16: how comes that the presence of a cut at 10 GeV affects the plots up to 400 GeV?
How are you scaling those histograms?

pietro

skeshri:
Hi @govoni ,
Thanks for your comments and questions.
Yes red line belongs to no pt > 10 GeV cut on jets. And about your question for eta distribution i don't understand what would you mean by "the red should always be higher than the red"  But since this distribution is normalized to area that is why this distribution looking like this in slide 14. I also made same distribution for 30k events each, but unnormalized attached. 

for your question from slide 16 : it may be again the effect of normalization but I also checked the un-normalized distribution and I see the same effect, but I'm not sure why? 

leadjet_eta_
boson_pt_

govoni:
Hello @skeshri ,

thanks a lot for the plots. What do you mean by "un-normalised"? What does the integral of the first distribution you are plotting correspond to? Is it a cross-section, or is it just what you get when you plot a given number of MC events (the same number for red and blue)?

pietro

skeshri:
Hello @govoni 

un-nomalised means I didn't divide the distribution by its area. Just to compare shapes of two different distributions I normalized both distributions to unity. And yes  It is just what I got when I plot for 30k generated events for both nlue and red.

perrozzi:
shall this be merged?

govoni:
Hello @skeshri, what are the two cross-sections, with and without the pT cut on the jets? I am asking, since you will end up with quite a different offline efficiency there. Sorry, I am not getting the full picture maybe.

skeshri:
Hi @govoni 
For background only in 4e final state :
total cross section with no pt cut = 3.06289642142999969E-004  +/-  7.27421157729138349E-007 pb
and with pt cut = 2.19147415644999939E-004  +/-  2.01289695919686857E-007 pb

There is some difference in between these two cross sections.

Regards,
Sumit

govoni:
Hello @skeshri,

thanks for the numbers. You will have to take into account that the statistical power of your sample will be reduced with respect to the past, due to the missing pT cut.
Anyways, the presence of a eta cut on the jets makes the calculation safe from the point of view of divergences close to the beamline, so I am OK with the merge.

govoni:
+1

skeshri:
Hi @govoni 

Thanks for your help and to give green light for this PR. 
But in past this pt cut was also missing. When I tried to compare distributions from the events generated by me this year with the distributions from 2016 VBF phantom sample, there I realized due to this missing pt cut I got this difference. By default this pt cut is present in the template r.in card this year but was not in earlier production[1].


[1] https://github.com/cms-sw/genproductions/blob/master/bin/Phantom/cards/production/13TeV/r_ew126_scale4_4l.in.TEMPLATE_step1_onlyB_mjjJets_13TeV#L338

PR #1661, opened 2018-02-07
Displaced SUSY for Fall17 part2

jalimena:
This is to update the EXO Displaced SUSY pythia8 fragments to use tune CP5. In addition, a few mass and lifetime points have been added, and a few removed. These fragments will be used for the Run 2, Fall17, part2 production.


efeyazgan:
Could you please leave one example file for one mass and lifetime point and instead add a script producing other mass/lifetime points? thanks.

jalimena:
hi @efeyazgan , please let me know if my latest commit is what you had in mind

efeyazgan:
Yes @jalimena. Thanks.

PR #1662, opened 2018-02-08
WIP: Cluster handling via MG plugin

AndreasAlbert:
**DO NOT MERGE YET**

**Goal:**
Move our modified versions of CondorCluster and LSFCluster into a madgraph plugin "CMS_CLUSTER".
Use plugin by setting cluster_type to "cms_condor" or "cms_lsf".

**Implementation**
I simply copied the code in cluster.py after applying all patches to the plugin and renaming the LSF and Condor Cluster classes to a CMS-specific name. I removed the corresponding patches.

**Testing**
Ran wplustest_4f_LO on lxbatch and CMSConnect. Both work fine, but that is no guarantee that all of the code is still functional.

@kdlong @khurtado thoughts?


kdlong:
Awesome! I will try it out shortly. 

@bendavid will also be interested in following this. The goal is to have this as a first step in moving towards different condor behavior with lxplus/ CMSConnect/Tier2 so we can implement the -spool option without using this option when it's not necessary.

bendavid:
Yes, this is great!  One thing we'll have to keep track of in the future, is to do a diff on cluster.py when updating madgraph versions to check if there are any changes to the standard lsf and condor classes which can/should be ported into our custom versions.

Of course even better would be to somehow make the code modular such that our custom versions could inherit from the standard ones with minimal changes, but not sure how feasible that would be without major reorganization of code.

AndreasAlbert:
@bendavid Good point regarding modularization. Skimming the `CondorCluster` code, it seems that we  leave the `remove` and `submit` functions untouched, so these can be inherited out-of-the-box, which would be a start.

AndreasAlbert:
Updated to add trivial spool support.
Spool usage now activated by using `cluster_type = cms_condor_spool`.
For completed jobs, `condor_retrieve_data` is called automatically. NB: No error handling implemented, yet. As pointed out by @pmandrik, one needs to take into account that jobs go into HOLD state while data is being transferred.

First test at lxplus worked when submitting `wplustest_4f_LO` from `/tmp/user`

@kdlong @bendavid @pmandrik: thoughts?

kdlong:
Excellent! Can you try with a heavier process?

bin/MadGraph5_aMCatNLO/submit_condor_gridpack_generation.sh
```diff
@@ -3,7 +3,7 @@ source Utilities/source_condor.sh
 
 name=$1
 carddir=$2
-workqueue="condor"
+workqueue="condor_spool"
```
kdlong: I think cms_condor should still be the default, just force spool ```if [[ $HOSTNAME == *"lxplus"* ]]``` for example

bendavid:
We do want this to eventually work submitting from non-lxplus machines at CERN.

kdlong:
True. You could add a more sophisticated check, or you can always edit it manually if you do/do not want to spool. But in cases where it's not needed, the overhead of spooling the input/output is a waste.

AndreasAlbert:
I did some more testing with DY+jets. 0 jets, 0+1 jets and 2 jets work out fine. Starting at 0+1+2+3 jets, I run into trouble because output file retrieval fails for a subset of jobs, which the code cannot recover from:

`ClusterId 61471 with HoldReason: Error from slot1_8@b6545958d8.cern.ch: STARTER at 188.185.56.123 failed to send file(s) to <188.185.71.142:9618>: error reading from /pool/condor/dir_28324/G4: (errno 2) No such file or directory; SHADOW failed to receive file(s) from <188.185.56.123:39273>`

I am investigating why this happens / what to do about it.

pmandrik:
Dear Andreas,
I was able to reproduce this error. The "Executable" of the failing jobs is "work/processtmp/SubProcesses/survey.sh" with "Arguments" in you case "0 3 4" corresponding to G3 and G4 working folders. The "output" and "error" condor logs do not contain anything for me, so I run survey.sh locally and successfully created both folders without errors. After that I submitted just a single survey.sh script modified in order to print a lot of debug information as job with the same options as in MadGraph to condor and receive the same error due to failing of receiving the files, but the output of survey.sh looked as if it was interrupted during the execution. Finally, after increasing the requested queue time, memory and disk space of the job it successfully finished in condor cluster. So, looks like a JobFlavor="testmatch" option may be a solution.

PS. I'm also think it will be usefull to put a multiple_try decorator before the madgrap compile function because due to the unknown reason sometimes the fortran compiler has a Segmentation Violation and gridpack generation failed at this step.

kdlong:
Thanks @pmandrik, did you look at making the suggested changes?

AndreasAlbert:
@kdlong Yes, @pmandrik made a PR to my fork with an updated code. I have successfully used this version of the code to be able to produce a `dyellell01234j_5f_LO_MLM` gridpack. Only downside at this moment is that condor seems to throw a lot of warnings, but those may be related to temporary issues on the scheduler side.

kdlong:
Excellent! Can you share some example warnings? @khurtado would probably have insight into weather they are a concern or not (I can also ask condor experts at Wisconsin)

AndreasAlbert:
Let me quote @pmandrik from the [original PR](https://github.com/AndreasAlbert/genproductions/pull/1)

> condor_transfer_data:
>condor_transfer_data:
>condor_transfer_data:
>INFO: Idle: 0, Running: 52, Completed: 2405 [ 13h 5m ]
>condor_transfer_data: Can't find address for schedd bigbird14.cern.ch
>condor_transfer_data: Can't find address for schedd bigbird14.cern.ch
>condor_transfer_data: Can't find address for schedd bigbird14.cern.ch

khurtado:
@AndreasAlbert @kdlong 
I think those messages are an indication that the condor master at `bigbird14.cern.ch` crashed at some point or got overloaded an unresponsive to the point the `condor_transfer_data` request failed. Many servers are configured to have condor automatically restarted when they crash though. It's hard to know what exactly happens without the SchedLog/MasterLog.

I'm looking into the lxbatch twiki and there are some observations regarding using the spool :
https://twiki.cern.ch/twiki/bin/view/ABPComputing/LxBatch

```
However the -spool option may overload the htcondor server if the amount of data transiting is too larger (>20 MB)
```

perrozzi:
what is the status here?

pmandrik:
I noticed that the finished jobs do not removed from condor_q even after condor_transfer_data call.
Because of this we are retrieving output multiple times for the same job and maybe load the server.
I'm proposed a patch when I added a self.fetched_jobs list in order to store ids of finished and fetched jobs and skip them after. I'm also still have a "condor_transfer_data:" stderr (condor_transfer_data return code 1 in this case) but the number of this errors is much much less after patch.

Do you think it will be usefull to create ticket in order to try to understand the meaning of "condor_transfer_data:" printout?

kdlong:
@pmandrik I think we should understand why the jobs aren't removed from condor_q and if this is intended behavior.

The jobs are never removed, or are only removed after the data is fully transferred, which can result in multiple transfer requests?

kdlong:
Also as for the issue with the server error, I think it's ok if we aren't able to run hugely complex processes because of limitations in the data transfer. As long as most processes can be run we can continue to recommend CMSConnect for the most complex.

bendavid:
The volume of data to be transferred can likely be substantially reduced if we can make sure the pdf sets are accessed directly from cvmfs rather than transferred to the working node.

kdlong:
@bendavid this should be the case. I got the impression that it's when transferring data back (I believe the necessary files to be transferred are specified) in which case it's mostly a function of how many subprocesses that are simultaneously trying to transfer data. You could perhaps deal with this by making a limit on the number of calls to condor_transfer_data made per time/finishd jobs.

pmandrik:
Dear @kdlong ,

>The jobs are never removed, or are only removed after the data is fully transferred, which can result in multiple transfer requests?

looks like the jobs will be removed from queue automatically after a time: "An exception is if the job is submitted using condor_submit -spool. For this case, the default expression causes the job to be kept in the queue for 10 days after completion." [0]

Also, I believe, in the plugin we do not spawn a multiple process but just call condor_transfer_data  once and wait the return code before next call.

[0] http://research.cs.wisc.edu/htcondor/manual/v7.6/10_Appendix_A.html

bendavid:
Is there some additional option which could be passed to modify this behaviour in the -spool case?

pmandrik:
>Is there some additional option which could be passed to modify this behaviour in the -spool case?
I added this option into the patch [0] after tests. I'm also would like to note that we still should save fetched jobs ids because there are some delay between finish of the data retrieving and status changes in the queue.

[0] https://github.com/AndreasAlbert/genproductions/pull/2

PR #1664, opened 2018-02-09
stub for script to patch LO gridpacks

perrozzi:


kdlong:
Maybe we should create a new directory called "hacks" and put it there instead of in scripts :)

perrozzi:
+1 :-)

PR #1665, opened 2018-02-09
Uplaod 2016 Final WWjj_SS_dim8 EFT cards that were never pushed 

xjanssen:
Doing some bookeeping in pre2017 directory as we will need these cards for 2017 MC production and the PhD student which is leaving never pushed them in git (the wrong one with QCD=99 are in the pre2017 tag)


bin/MadGraph5_aMCatNLO/cards/production/pre2017/13TeV/VBS/WWjj_SS_dim8/WWjj_SS_dim8_ewk_customizecards.dat
```diff
@@ -0,0 +1,20 @@
...
+set param_card anoinputs 3 0.0
+set param_card anoinputs 4 0.0
+set param_card anoinputs 5 0.0
+set param_card anoinputs 6 0.0
+set param_card anoinputs 7 0.0
+set param_card anoinputs 8 0.0
+set param_card anoinputs 9 0.0
+set param_card anoinputs 10 0.0
+set param_card anoinputs 11 0.0
+set param_card anoinputs 12 2.000000e-12
```
efeyazgan: why not 0.0?

efeyazgan:
You have multiple definitions of leptons. Please clean up.

bin/MadGraph5_aMCatNLO/cards/production/pre2017/13TeV/VBS/WWjj_SS_dim8/WWjj_SS_dim8_ewk_customizecards.dat
xjanssen: because this was the optimal point to start from in term of statistic I think. Then the SM is recovered as one of the LHE weight here:
https://github.com/xjanssen/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/pre2017/13TeV/VBS/WWjj_SS_dim8/WWjj_SS_dim8_ewk_reweight_card.dat#L75

xjanssen:
Ah yes, well I removed the extra lepton definition but I think it does no matter as the sample was containing taus. I just re-extracted these cards from the gridpack as everything was missing

kdlong:
Just so it's clear, the point of this pull request is to document the cards used in the current analyses, to be shared with a comparison with ATLAS. The proper cleanup can come when they are ported for 2017 production :)

efeyazgan:
Thanks for the clarification (that saves time).

PR #1673, opened 2018-02-14
MonoPhoton DMsimp

yiiyama:
Cards for Monophoton DMsimp samples with photon pT > 130 GeV cut at the gen level.
Most of the updated files are in cards/production/2017/13TeV/MonoPhoton_DMsimp but there is also a hadronizer fragment. The fragment is a generic FxFx configuration with +1jet and xqcut = 10 GeV, but somehow does not exist in the current master branch of cms-sw.


yiiyama:
OK, Andreas told me that I should submit a template + example + script instead. Closing the PR.

PR #1674, opened 2018-02-14
Cards for monophoton DMsimp samples

yiiyama:
We would like to remake the DMsimp (NLO QCD) monophoton samples because the existing version has a cut on the mediator pT, which makes the photon pT not necessarily high and thus makes our acceptance low (primary discriminating variable in the analysis is photon pT).

Production request is here:
https://docs.google.com/spreadsheets/d/1SHuXn9vgIq1jOr9ixVhqcWXC7a8ZNxCQvj1n9iXlkDI/edit?usp=sharing

The PR contains a directory MG5/cards/production/2017/MonoPhoton_DMsimp which has the card templates and a script to generate the cards, along with one example subdirectory. Also, I am PRing a generic hadronizer fragment for +1jet FxFx (xqcut 10 GeV) which for some reason does not exist in the current master.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/MonoPhoton_DMsimp/Vector_Mphi-1000_Mchi-1_gSM-0p25_gDM-1p0_pta130/Vector_MonoPhoton_NLO_Mphi-1000_Mchi-1_gSM-0p25_gDM-1p0_pta130_customizecards.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/MonoPhoton_DMsimp/Vector_Mphi-1000_Mchi-1_gSM-0p25_gDM-1p0_pta130/Vector_MonoPhoton_NLO_Mphi-1000_Mchi-1_gSM-0p25_gDM-1p0_pta130_customizecards.dat
@@ -1,0 +1,17 @@
+set param_card mass  55 1000
+set param_card mass  52 1
+set param_card dminputs 2 1
+set param_card dminputs 3 0
+set param_card dminputs 4 0.25
+set param_card dminputs 5 0.25
+set param_card dminputs 6 0.25
+set param_card dminputs 7 0.25
+set param_card dminputs 8 0.25
+set param_card dminputs 9 0.25
+set param_card dminputs 10 0
+set param_card dminputs 11 0
+set param_card dminputs 12 0
+set param_card dminputs 13 0
+set param_card dminputs 14 0
+set param_card dminputs 15 0
+set param_card decay 55 auto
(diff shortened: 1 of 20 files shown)


perrozzi:
why CUETP8M1 ? You should use CP5

yiiyama:
We want this for a 2016 analysis, so all our background MC are done with CUETP8M1. Also, in particular this request is a replacement for these datasets [1] which had an unfortunate gen-level cut that made them unusable. Would you say we should use CP5 nevertheless?

[1] https://cmsweb.cern.ch/das/request?view=list&limit=50&instance=prod%2Fglobal&input=dataset%3D%2FDarkMatter_MonoPhoton_NLO_Vector_Mx-*_Mv-*_gDM1_gQ0p25_13TeV-madgraph%2FRunIISummer16MiniAODv2-PUMoriond17_80X_mcRun2_asymptotic_2016_TrancheIV_v6-v1%2FMINIAODSIM

perrozzi:
please then use the `pre2017` folder instead of `2017`

yiiyama:
But the UE tune is defined in the Hadronizer fragment, and there's no `pre2017` directory for those, no?

perrozzi:
(meant the mg5_amc cards)

yiiyama:
The MG5 cards are compatible with the latest gridpack production script, and the gridpacks are made with the recent master HEAD of cms-sw genproductions. So I believe the MG5 cards should be where they are now. We do plan to request samples with CP5 for the 2018 analysis, using the same gridpacks.

perrozzi:
pre2017 and 2017 productions have a change in both central PDF and Pythia tune.
I would personally discourage a sample featuring 2017 PDF and pre2017 tune, that would be confusing.
Better to have 2 copies of the datacards one in pre2017 and the other in 2017 folders with consistent settings...

perrozzi:
I agree to put all the copies of the cards in genproductions.
I would discourage the use of a single gridpack when at analysis level the default will be a PDF different than the central one. if you really want to do that, please at least make sure that you specify the PDF name in the primary dataset name in McM to (try to) minimize the probability of things going wrong...

yiiyama:
OK, I made the copy of the cards in pre2017 + the CP5 version of the hadronizer. As I said, we would like to use the gridpack made with 2017 cards and the CUETP8M1 hadronizer for the immediate production request. We'll follow your suggestion about making the PDF choice clear in the dataset name.

mtrovato:
@yiiyama : shouldn't you use 2017 cards and CP5 tuning as in semi-invisible Higgs decay PR (https://github.com/cms-sw/genproductions/pull/1678)? I think the same comment from Efe should apply here

yiiyama:
@mtrovato We want to be able to compare with the previous DMsimp sample (link above) and with the LO samples, which are all done with CUETP8M1. But if it is too much burden to use CUETP8M1 for this request, we are fine moving to CP5.

mtrovato:
@yiiyama : thanks for explaining. From our side it's not too much burden, as long either CUETP8M1 or CP5 is chosen

perrozzi:
from the PR point of view things look. please be careful when preparing the requests (and possibly use consistent pdf and tune...)

PR #1684, opened 2018-02-19
HJJ HWW Powheg input card

d4space:
identical to the https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/production/2017/13TeV/Higgs/HJJ_NNPDF31_13TeV/HJJ_NNPDF31_13TeV_M125.input.
To be used for hWW.


efeyazgan:
Then, why do you need to add this additional card?

d4space:
The reason is to make clear what hww is using.
And for the possible parameter changes like mass, we need to set up a directory.

Thanks,
SangEun

PR #1687, opened 2018-02-22
Compilation of VH_ew

covarell:
Minimal changes to make VH work:

- use of no fixed line length in f90
- patch HW_ew to have basic routines outside analysis package
- add missing libraries

IMPORTANT: These are only changes to make VH(J)_ew compile. Producing events seems complicated (need to produce gridpack with one option, run with another option and save results as weights). 

It appears similar to NNLOPS procedure: scripts/weight files can be adapted to deal with this procedure but we have to see with Oleari if potential pitfalls are there, similar to the NNLOPS case (e.g. meaningless weights if too few events per job are run).

### DIFF
--- a/bin/Powheg/patches/hwew.patch
+++ b/bin/Powheg/patches/hwew.patch
@@ -1,0 +1,207 @@
+--- scalupveto_HW.f	2018-02-22 16:32:52.000000001 +0100
++++ scalupveto_HW.f.new	2018-02-22 16:34:17.000000001 +0100
+@@ -784,3 +784,100 @@
+       endif
+              
+       end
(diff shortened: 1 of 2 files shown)


covarell:
Ahem... Any comment on the event generation issue?

perrozzi:
Ciao Roberto, I have read your comments and I would suggest a "modular approach" i.e. if this PR makes it at least to compile, then we can test it and try to produce events with it. Once also this works, we can try to make sure that we have the results right :-)
Does it make sense to you?

covarell:
Alright, I will do such test.

PR #1689, opened 2018-02-24
cards for resonant and non-resonant HH

acarvalh:


vischia:
Dear GEN,

do we have to wait for the PR to be approved to move on?
If so, could you please approve it?

Thanks in advance!
Cheers,
Pietro

perrozzi:
would it be possible to upload only one representative set of cards for the parameter scan, and a script to create the others? we are trying to keep the number of files low in the repoitory to avoid slow down

acarvalh:
I did that for the resonant case, for the nonresonant it would involve a
bit more of scripting work - is it absolutelly necessary?
The easiest way to reduce the file load is I leave only the customize cards
(remove extramodels/run of all but one representative)

2018-03-03 10:07 GMT+01:00 perrozzi <notifications@github.com>:

perrozzi:
are the others datacards except customize cards the same for all?

acarvalh:
Yes

vischia:
@perrozzi, then would it be OK to remove those and have a script that copies the same files to all the other filenames?

perrozzi:
yes that would be sufficient, thanks for your understanding

acarvalh:
Hi Pietro,
I had just removed the duplicated extramodels and run_cards
thanks
Xanad

2018-03-04 23:02 GMT+01:00 perrozzi <notifications@github.com>:

PR #1692, opened 2018-02-28
cards for TT FCNC 13 and 14 TeV

zenaiev:
cards for 16 TT FCNC samples containing t->Z decays, for both 13 and 14 TeV, in total 2*16 = 32 directories each containing 6 files


perrozzi:
we kindly ask to upload only one representative set of datacards for a parameter scan, uploading a script to produce the others. can you do that?

zenaiev:
Yes, adding such a script which I used to produce all cards, together with the input representative cards. In fact madspin cards are different for different models, so still there is a number of required representative cards, but other 4 out of 6 cards are identical -> not duplicating them now.
Also I add simple README to explain what needs to be done.

efeyazgan:
Please remove unnecessary comments (e.g "go go go") and unused definitions (e.g. in some cards leptons are defined but not used).

zenaiev:
Done.

Issue #1670, opened 2018-02-12
[mg26x] MadSpin crashes when decaying events after gridpack generated succesfully


cmantill:
I produced gridpacks for WJetsToQQ process at NLO usng M.G. 26x and these cards:

https://github.com/cmantill/genproductions/tree/mg26x/bin/MadGraph5_aMCatNLO/cards/production/13TeV/WJetsToQQ_5f_NLO_FXFX_bias
https://github.com/cmantill/genproductions/tree/mg26x/bin/MadGraph5_aMCatNLO/cards/production/13TeV/WJetsToQQ_5f_NLO_FXFX

both gridpacks where generated succesfully (although they took some time to generate), e.g. see the log here:

http://cmantill.web.cern.ch/cmantill/gridpack_generationWJetsbias.log

However, when I untar the gridpack and test it in a CMSSW env (9_3_0), I obtain the following error:
INFO: 
   --------------------------------------------------------------
      Summary:
      Process p p > V [QCD] @0 ; p p > V j [QCD] @1 ; p p > V j j [QCD] @2
      Run at p-p collider (6500.0 + 6500.0 GeV)
      Number of events generated: 1000
      Total cross section, incl. bias (DO NOT USE): 3.302e+14 +- 1.2e+12 pb
   --------------------------------------------------------------
 
INFO: The /local-scratch/cmantil1/CMSSW_7_1_30/src/process/Events/cmsgrid/events.lhe.gz file has been generated.
 
INFO: Events generated 
reweight -from_cards
decay_events -from_cards
INFO: Running MadSpin 
INFO: This functionality allows for the decay of resonances 
INFO: in a .lhe file, keeping track of the spin correlation effets. 
INFO: BE AWARE OF THE CURRENT LIMITATIONS: 
INFO:   (1) Only a succession of 2 body decay are currently allowed 
************************************************************
*                                                          *
*           W E L C O M E  to  M A D S P I N               *
*                                                          *
************************************************************
INFO: Extracting the banner ... 
INFO: process: p p > V   
INFO: options:  
INFO: detected model: loop_sm-ckm_no_b_mass. Loading... 
set seed 1023747
set seed 1010835
set seed 1016767
set seed 1025432
set ms_dir ./madspingrid
set Nevents_for_max_weigth 250 # number of events for the estimate of the max. weight
set max_weight_ps_point 400  # number of PS to estimate the maximum for each event
set BW_cut 40                # cut on how far the particle can be off-shell
[...]

PR #1697, opened 2018-03-02
[mg26x] Add patch to prevent madspin crash.

AndreasAlbert:
This patch will be part of MG2.6.2:
https://bazaar.launchpad.net/~mg5core1/mg5amcnlo/2.6.2/revision/305

It solves the problem discussed here:
https://github.com/cms-sw/genproductions/issues/1670


perrozzi:
@kdlong @qliphy @agrohsje please crosscheck

agrohsje:
Hi @perrozzi ,
the patch works fine, is consistent with the official one, gridpacks with madspin get properly produced. 
I guess that's sufficient as a cross check, right?
Cheers, Alexander.

PR #1566, opened 2017-11-29
2017 5TeV cards

gkrintir:
These are the powheg & madgraph5_amc@nlo cards for 5.02 TeV production.

All powheg cards successfully tested in CMSSW_9_3_0.

All madgraph5_amc@nlo cards tested to reproduce the results from the cards used in 2015 5.02 TeV production (within unc.).

Additional cards may come later (e.g. follow up on WZ&ZZ production currently known to be buggy).

PR #1706, opened 2018-03-06
t-channel with Vtd=Vts=0.7

ckmackay:
With more sensible comments


efeyazgan:
One final question: are the parameters assume the new values discussed in  https://github.com/cms-sw/genproductions/pull/1566 and https://indico.cern.ch/event/688821/

ckmackay:
These are the non standard values for testing Vtd=Vts= 0.707 for either decay or production. I have left them the same as were asked for by Orso et al previously.

PR #1708, opened 2018-03-06
Adding Extended Weak Isospin model for 14 TeV

mpresill:


mpresill:
I added the .mdl files and the headers of each LHE generated which contain info on the batch used for the generation (approximately from line 527 to 647).

Here I also add an example of the batch files used:

Model: MajN_2
Model changed: False
Gauge: Feynman
Process: p,p->e,NEs
Decay: NEs->e,jet,jet
Composite: p=u,u~,d,d~,s,s~,c,c~,b,b~
Composite: jet=u,u~,d,d~,s,s~,c,c~,t,t~,b,b~
Composite: e=e+,e-
pdf1: LHA:NNPDF30_lo_as_0130:0:1
pdf2: LHA:NNPDF30_lo_as_0130:0:1
p1: 7000
p2: 7000

Parameter: MUU=9000
Parameter: lc=9000
alphaQ: MUU
Number of events (per run step): 100000
Filename: NAME
NTuple: False
Cleanup: True
Parallelization method: local
Max number of cpus: 16
sleep time: 3
nice level : 19
nSess_1: 5
nCalls_1: 10000
nSess_2: 5
nCalls_2: 10000

Unfortunately I cannot give just one sript file to generate all the parameters since this is a 2-dimensional scan (not doable in CH), and also the step-size is not "regular".
One can just run the batch file above several times just substituting the lines
Parameter: MUU=
Parameter: lc=
with the suitable choice of parameters.
Here I've used those following pts:
lc=6000, MUU=6000
lc=9000, MUU=3000
lc=9000, MUU=6000
lc=9000, MUU=9000
lc=12000,MUU=500
lc=12000,MUU=3000
lc=12000,MUU=6000
lc=12000,MUU=9000
lc=15000,MUU=500
lc=15000,MUU=3000
lc=15000,MUU=6000
lc=18000,MUU=500
lc=18000,MUU=3000
lc=21000,MUU=500

perrozzi:
can you remind me how many files would be, if you would upload them all?

perrozzi:
you mean on top of the 20 already in the pr?

perrozzi:
up to you. if you think that is much less error prone to have these 14 files in genproductions, please add them. otherwise we merge the pr as it is

mpresill:
I think Headers are enough, so you can proceed with these. Thanks

PR #1709, opened 2018-03-08
tZq_ll_herwigpp and tzq_nunu cards

ckmackay:


efeyazgan:
This line  
PYTHIA8   = parton_shower 
should be changed to herwig!

PR #1710, opened 2018-03-08
m4l cut for POWHEG ZZ

hroskes:
We're hoping to use these samples in the HZZ group for around 125 GeV and for the very high mass offshell tail.  We don't get enough statistics in these regions from the full mass spectrum ZZ samples, especially when we apply jet categorization.  In addition, some B2G analyses would like to use dedicated generation at low mass, 10-40 GeV.

For the changes to the POWHEG code, I asked the POWHEG authors for help, and Giulia Zanderighi agrees with this implementation.

I generated a small sample using the 100-160 GeV input card and got very good agreement with the full sample (extracting the LHE information from /ZZTo4L_13TeV_powheg_pythia8/RunIIFall17MiniAOD-94X_mc2017_realistic_v10-v2/MINIAODSIM).  Both samples are normalized separately using the cross sections from POWHEG, and both the shapes and normalizations agree.

Please let me know if you need any more information.

(more plots: [plots.tar.gz](https://github.com/cms-sw/genproductions/files/1794530/plots.tar.gz))
m4l
phi1


perrozzi:
I wonder if we should have this as default for ZZ or create a switch to activate it.
@kdlong @agrohsje @qliphy ?

kdlong:
To me it looks like you're already provided flags that can be set for the m4l mass cut. I assume if nothing is specified they are left off, which seems like the ideal setup to me.

perrozzi:
sorry I had completely overlooked, you are right :-)

hroskes:
Thanks!  Just to confirm, yes, if it's not found in the input file POWHEG sets it to some default hardcoded value, I think it's -1d38 or something like that.  If m4lmin is -1d38, we don't need to do anything special since m4l is always bigger than that.  If m4lmax is -1d38, this gets caught by https://github.com/cms-sw/genproductions/blob/03bc648f9ffacc39568c47d54138752fd9570c0a/bin/Powheg/patches/zz_m4lcut.patch#L38

In this one I don't have an m4lmax, and it works fine: https://github.com/cms-sw/genproductions/blob/03bc648f9ffacc39568c47d54138752fd9570c0a/bin/Powheg/production/2017/13TeV/ZZ/ZZ_4L_800%2BGeV_NNPDF31_13TeV.input

I also use the usem4lcut flag, which avoids the extra computations of adding the 4 momenta and taking the mass when there's no cut.

PR #1720, opened 2018-03-16
Cards for WZjj + custom user patch support

kdlong:
2017 cards for WZjj EW production. In order to have a custom dynamic scale choice, I've also added patches to the systematics module. Sorry for mixing the two together but I needed something to test it on in any case. I don't think we want to strongly advertise this feature, but if someone is going to customize MG themselves we would rather they do it in a transparent way.


perrozzi:
in principle it looks ok, it has been tested and is transparent to others unless they put .patch files in the directory, right?

kdlong:
Oops I missed this. Yes, it should be used only very rarely and only have an effect if a .patch file is added to the cards directory. I don't think we have to advertise this, but we can mention it if people are already trying to do something customized.

PR #1722, opened 2018-03-16
Fix bug introduced in LO runcmsgrid

kdlong:
Fix bug introduced in LO runcmsgrid for 26x


perrozzi:
validated and ready to merge?

kdlong:
Yes. The bug with runcmsgrid_LO.sh was a pretty blatant one which needs to be fixed.

PR #1723, opened 2018-03-16
Updates for semi-invisible Higgs fragments

yiiyama:
- We want to run a Summer15GS + Summer16DR request with these fragments, but CP5 tune cannot be imported in 71X. Hard-coding the configuration for the time being. Will remove the hard-coded part once we finish the 2016 analysis.
- @smrenna suggested using PID 53 (instead of 300022 which I randomly picked) for the dark photon. This is the CMS convention for a non-interacting vector.

### DIFF
--- a/python/ThirteenTeV/Higgs/Hadronizer_TuneCP5_13TeV_powhegEmissionVeto_1p_HToGDarkG_M-125_LHE_pythia8_cff.py
+++ b/python/ThirteenTeV/Higgs/Hadronizer_TuneCP5_13TeV_powhegEmissionVeto_1p_HToGDarkG_M-125_LHE_pythia8_cff.py
@@ -1,31 +1,6 @@
 import FWCore.ParameterSet.Config as cms
 from Configuration.Generator.Pythia8CommonSettings_cfi import *
-try:
-    from Configuration.Generator.MCTunes2017.PythiaCP5Settings_cfi import *
-except ImportError:
-    pythia8CP5SettingsBlock = cms.PSet(
-        pythia8CP5Settings = cms.vstring(
-        'Tune:pp 14',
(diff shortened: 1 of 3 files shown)


efeyazgan:
Sorry what do you mean by "CP5 tune cannot be imported in 71X"? 
It is there, e.g. https://github.com/cms-sw/cmssw/blob/CMSSW_7_1_31/Configuration/Generator/python/MCTunes2017/PythiaCP5Settings_cfi.py

yiiyama:
I meant it isn't available in 7_1_30, which is the default release used by the gridpack generation script (and so the release I use to do GEN studies with). I suppose new production requests can be made with 7_1_31 then? Dropped the part that was hard-coding the CP5 configuration.

efeyazgan:
RunIISummer15GS campaigns uses CMSSW_7_1_31 and RunIISummer15wmLHEGS uses CMSSW_7_1_32 (that also has CP5) for sometime now anyway.

yiiyama:
OK, didn't know. Thanks.

perrozzi:
so the PR can be merged, right?

smrenna:
I don't care so much what is the particular ID used for the DM particle (though it would be nice to use the codes that have been provided at the community's request).   However, the main point I was making was that particle for gen jets should know to not use that particular ID (53, whatever).   

This was my original comment:
--begin quote--
53 is spin-1, which is why I suggested that one.   but you could use any.

Based on this:
http://cmslxr.fnal.gov/source/RecoJets/Configuration/python/GenJetParticles_cff.py

I would suggest to use 1000022 for now.    But then, you see how someone could come
back later and think this has something to do with SUSY.  Oh, well.
--end quote--

PR #1730, opened 2018-03-20
single top tW mass variation cards

ckmackay:


efeyazgan:
I forgot whether this was discussed but I have two questions (to topmodgen - @jfernan2  in general as well) 
-- You still keep hdamp=mt even for 2017 production with CP5 tune?
-- You change hdamp when you do mass variations. Was this always like this and the suggested way to do it?

ckmackay:
This is what was done previously, so I repeated it for this year, I hope that was OK?

PR #1736, opened 2018-03-26
add LQ3 cards for HL-LHC studies

jalimena:
This PR adds LQ3 cards, which will be used for HL-LHC studies.


jalimena:
hi, can this PR be reviewed soon? thanks!

PR #1742, opened 2018-03-28
Dalitz ggH

gourangakole:
Dalitz ggH with HToMuMuG and ElElG decay.
Example cards are taken from [here](https://github.com/cms-sw/genproductions/tree/pre2017/bin/MadGraph5_aMCatNLO/cards/production/13TeV/higgs/dalitz/ggH120_012j_NLO_FXFX_HtoMuMuGamma)


gourangakole:
Removed individual mass points and added a script.
Could you please check @efeyazgan and @perrozzi

PR #1744, opened 2018-03-29
[mg26x] Port LO fix to runcmsgrid from master

AndreasAlbert:
**Changes**
1. Port original workaround from https://github.com/cms-sw/genproductions/pull/1564
2. Port fixup from https://github.com/cms-sw/genproductions/pull/1716
3. Small change to use different run directory name (`GridRun_PostProc_${rnum}`) for systematics call, and manually create that directory. 

**Testing**
Successfully ran 100k events with `wplustest_4f_LO` gridpack.


AndreasAlbert:
@kdlong @agrohsje

kdlong:
I checked this by generating a new gridpack and by patching and testing all the MG 2.6.0 gridpacks I'd already generated and see things working as expected. I think it's ready to merge. Thanks Andreas!

PR #1749, opened 2018-04-03
[mg24x] Increase number of events for combine_events.f

AndreasAlbert:
Patch to solve the problem discussed [here](https://hypernews.cern.ch/HyperNews/CMS/get/generators/3814/1/1/1/1/1/1/1/1.html)

Successfully tested by me and @wverbeke 

@kdlong @agrohsje: Do we still care about patching master/24x or should I just port the patch to the 26x branch and leave master as it is?


agrohsje:
As long as it exists I would suggest to patch it as well. Just to be sure nobody picks it up and runs into the problem.

AndreasAlbert:
Then this PR should be merged.

PR #1753, opened 2018-04-05
Forked master

d4space:
call parameter change for "production/2017/13TeV/Higgs/VBF_H_WW_NNPDF31_13TeV/VBF_H_WW_NNPDF31_13TeV_template.input"


d4space:
bin/JHUGen/install.py:

The original file had a problem generating several packs simultaneously, over-writing each directory name to "runcmsgrid.sh" file.
Which means only one pack generated at a time.
The new version can generate packs simultaneously. And it is tested.

Issue #1307, opened 2017-09-11
[2017 MC v2] checking POWHEG compilation and weights


perrozzi:
* Currently **POWHEG** set to CMSSW_9_1_1 (cfr [twiki](https://twiki.cern.ch/twiki/bin/viewauth/CMS/PowhegBOXPrecompiled#Download_all_the_necessary_scrip) )
      * Does it have all the needed PDFs?
      * Probably better to move to CMSSW_9_2_11 ? 

~

   * Currently **MG5_aMC@NLO** set to CMSSW_7_1_28 (cfr [generation script](https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/gridpack_generation.sh#L47) )
      * Likely OK to stay with that unless newer version require an update

perrozzi:
**POWHEG** instructions moved to CMSSW_9_2_11

perrozzi:
need to wait for the latest and greatest 92x and 71x to include all the required PDFs
https://github.com/cms-sw/cmsdist/pull/3419/files
https://github.com/cms-sw/cmsdist/pull/3417

perrozzi:
David suggests to move to 93x instead of 92x
https://github.com/cms-sw/cmsdist/pull/3415#issuecomment-329801298
This means to move from gcc530 to gcc630.
We need to test POWHEG in this configuration...
Unless anybody has ideas/arguments to stick to 92x

perrozzi:
Currently testing **compilation** for **all** processes using
CMSSW_9_2_12 (gcc530)
CMSSW_9_3_0_pre5 (gcc630)

**FAILING** so far:
weakinos (gcc630 and gcc530)
Wbbj (gcc630 and gcc530)
Wbb_dec (gcc630)

perrozzi:
**POWHEG** instructions moved to CMSSW_9_3_0

agrohsje:
@perrozzi 
Hi Luca, I just  tested gg_H using  
https://raw.githubusercontent.com/cms-sw/genproductions/master/bin/Powheg/examples/gg_H_quark-mass-effects_withJHUGen_NNPDF30_13TeV/JHUGen.input 
and 
https://raw.githubusercontent.com/cms-sw/genproductions/master/bin/Powheg/examples/gg_H_quark-mass-effects_withJHUGen_NNPDF30_13TeV/gg_H_quark-mass-effects_NNPDF30_13TeV.input 
[...]

PR #1755, opened 2018-04-10
switched to Powheg rev 3511 

agrohsje:
supposed to speed up generation of ggHH
physics should be unaffected

### DIFF
--- a/bin/Powheg/run_pwg.py
+++ b/bin/Powheg/run_pwg.py
@@ -454,6 +454,7 @@ if [ "$process" = "ggHH" ]; then
    sed -i -e "/LIBPYTHIA8/s|^|#|g" Makefile
    sed -i -e "s|LIBHEPMC=|# LIBHEPMC=|g" Makefile
    sed -i -e "/main-PYTHIA8-lhef:/s|^|#|g" Makefile
+   sed -i -e "s|LIBS+=-L:\\\$|LIBS+=-L\\\$|g" Makefile
 fi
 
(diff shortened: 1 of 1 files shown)


agrohsje:
I found a bug in the ggHH Makefile that the authors released. I will fix it. Than merge the PR yet!

agrohsje:
The PR should be ready to go. Please merge! Xanda will produce a new gridpack. We saw indeed significant runtime performance. Pinging the authors we also got additional info on settings so that replacing the executable is not enough and new presampling is needed.

yuanchao:
A quick test on CMSSW_9_3_0 with test_source_compilation.sh, several channels including dijet, directphoton, disquark, weakinos... failed to be compiled. Please see /afs/cern.ch/user/y/yuanchao/scratch/CMSSW_9_3_0/src/source/LSFJOB_160407743/STDOUT.

agrohsje:
Sorry for not checking. That was my mistake. I should have done so. I was not aware these Makefiles were touched. 
I don't have permissions to check your stdout. . 
Did all of them compile before? Do you know in which release it was tested? We can compare the Makefiles to that.

agrohsje:
The same problems were already discussed here: 
https://github.com/cms-sw/genproductions/issues/1307
Did anybody solve them?

perrozzi:
> Did anybody solve them?
not that I know.

What is the status as of now? Which is the last svn rev? and which processes don't compile?

PR #1756, opened 2018-04-11
Hadronizer with qCut=20 nJetMax=3and4

alsusy:
New Hadronizer for W/ZJetsQQ HT-binned samples.   qCut = 20 and nJetMax = 3 and 4 were applied.


efeyazgan:
This is for a pre-2017 analysis, right? (since the old tune is used)

alsusy:
Yes,  right. it is for 2016 data analysis.

PR #1757, opened 2018-04-11
fix when we get NaN in MCFM

hroskes:
Sometimes events with 4 lepton interference have NaN somewhere in the phase space.  When the event generation finds this NaN, MCFM crashes because there's an assertion in LHAPDF that q^2 > 0, and NaN is not > 0.

This is a patch to run MCFM multiple times and merge the LHE file together, dropping the events that are only partially written with no </event>.  Because MCFM doesn't always produce the exact number of events requested, the script selects random ones to drop.  (This was done already, but now it looks through multiple files.)

The implementation is mostly copied from the Madgraph one, with some adaptations.  I've run some local tests and it works well.

### DIFF
--- a/bin/MCFM/adjlheevent.py
+++ b/bin/MCFM/adjlheevent.py
@@ -46,10 +46,12 @@ import argparse
 parser = argparse.ArgumentParser()
 parser.add_argument("requestednumofevents", type=int)
 parser.add_argument("foundnumofevents", type=int)
+parser.add_argument("seed", type=int)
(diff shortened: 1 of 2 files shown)


perrozzi:
this is to fix the errors recently observed in production, right?

hroskes:
Yes that's right

perrozzi:
has this been tested?

hroskes:
I tested it by using a script to patch existing gridpacks, and it works great.  We haven't tested one produced from scratch yet.

PR #1759, opened 2018-04-13
Boost include no longer needed in 2.6.1

kdlong:
Simplify LHAPDF 6.2 patch since only a small fraction is still necessary. Unfortunately it is still partially broken in the official version and requires a small patch.


perrozzi:
tested and ready to be merged?

kdlong:
Yes

PR #1761, opened 2018-04-17
HctoWA_AtoMuMu cards for 2016 samples

jbhyun:


perrozzi:
are these datacards really meant to be used for 2016 conditions?
can you please upload only one reference set and a script to produce the others, in order to limit the number of files?

jbhyun:
Yes, for 2016 condition, and I simplified the cards to unit files and a script to produce other datacards.

PR #1765, opened 2018-04-23
adding 4T and 4VTo4L channels

acarvalh:


acarvalh:
This is for 2017. I confess that I did not paid attention for pythia tunning, I just copied what it is used on the rest of the cards of the folder. Those cards are more an storage for the decay modes, but I agree that it shall be better to have pythia tunning up to date.

Please confirm if the bellow is what should be used

        parameterSets = cms.vstring('pythia8CommonSettings',
                                    'pythia8CP5Settings',
                                    'pythia8PowhegEmissionVetoSettings',
                                    'processParameters'
                                    )

An I update it on all the cards of this folder.

efeyazgan:
This looks OK but you need imports as well. You can take a look at the link below for an example:
https://cms-pdmv.cern.ch/mcm/public/restapi/requests/get_fragment/HIG-RunIIFall17wmLHEGS-00531/0

acarvalh:
@efeyazgan following your example, should I add 'pythia8PowhegEmissionVetoSettings', to all powheg cards?

acarvalh:
Updated in all cards (not only the new decay modes)

acarvalh:
Is it all ok now here?

efeyazgan:
Hi, I see several things to be changed:
- This is for powheg or madgraph? Assuming it is powheg, please change the names of the files accordingly. 

- I see CUETP8M1 appearing, e.g. in https://github.com/acarvalh/genproductions/blob/582ebd4d37fbaba710143b9818c9a3812e4582d1/python/ThirteenTeV/Higgs/HH/ResonanceDecayFilter_example_HHTo2B2G_madgraph_pythia8_cff.py#L27 

- Before I sent you a fragment as example but taking the following as example would have been easier: https://github.com/cms-sw/genproductions/blob/master/python/ThirteenTeV/Hadronizer/Hadronizer_TuneCP5_13TeV_powhegEmissionVeto_1p_HToTT_0P_M-125_LHE_pythia8_cff.py

acarvalh:
It is 1/2 powheg and 1/2 madgraph. 
I am sorry for having missed this last spot, now there is no CUETP8M1 anywhere else!

acarvalh:
the ones named madgraph should be madgraph,
the example you sent is for powheg,  is this the correct setting to
madgraph
https://github.com/cms-sw/genproductions/blob/master/python/ThirteenTeV/Hadronizer/Hadronizer_TuneCP5_13TeV_aMCatNLO_FXFX_5f_max2j_LHE_pythia8_cff.py#L5
?
( instead of this
https://github.com/acarvalh/genproductions/blob/a21229d35076c6c397afe01129da05717b36f2f3/python/ThirteenTeV/Higgs/HH/ResonanceDecayFilter_example_HHTo2B2G_madgraph_pythia8_cff.py#L17
?)

2018-05-03 17:48 GMT+00:00 Efe Yazgan <notifications@github.com>:

efeyazgan:
Yes, please make the names of the files and the contents consistent.

For the hadronizer for madgraph if you use fxfx yes, the one you sent should be OK (if it's what you want).

acarvalh:
So that is not ok, I do not use matching...

What should I add for plain LO madgraph, please?

thanks

2018-05-04 8:44 GMT+00:00 Efe Yazgan <notifications@github.com>:

perrozzi:
as suggested by Efe, could you please change the names of the fragments to contain the tune, for example
`ResonanceDecayFilter_example_HHTo2B2G_madgraph_pythia8_cff.py`
to
`ResonanceDecayFilter_example_HHTo2B2G_madgraph_pythia8_CP5_cff.py`

bin/Powheg/run_pwg.py
```diff
@@ -20,17 +20,17 @@
 
 
 def runCommand(command, printIt = False, doIt = 1, TESTING = 0) :
-    if TESTING :
```
perrozzi: this file only has commit noise, I am afraid...
cfr https://stackoverflow.com/questions/21822650/disable-git-eol-conversions

acarvalh:
I restored the previous commit with 
git checkout a5a3a0313e6cfd14b339070bab4ad026f54d242a -- bin/Powheg/run_pwg.py

PR #1766, opened 2018-04-23
Hadronizer_TuneCP5 qCut=20, nJetMax=3 and 4

alsusy:
Hadronizer for madgraph with qCut=20 and nJetMax = 3 and 4.  To be used for W/ZJetsQQ 2017 samples.

### DIFF
--- a/python/ThirteenTeV/Hadronizer/Hadronizer_TuneCP5_13TeV_MLM_5f_max3j_qCut20_LHE_pythia8_cff.py
+++ b/python/ThirteenTeV/Hadronizer/Hadronizer_TuneCP5_13TeV_MLM_5f_max3j_qCut20_LHE_pythia8_cff.py
@@ -1,0 +1,34 @@
+import FWCore.ParameterSet.Config as cms
+
+from Configuration.Generator.Pythia8CommonSettings_cfi import *
+from Configuration.Generator.MCTunes2017.PythiaCP5Settings_cfi import *
+
+generator = cms.EDFilter("Pythia8HadronizerFilter",
+    maxEventsToPrint = cms.untracked.int32(1),
+    pythiaPylistVerbosity = cms.untracked.int32(1),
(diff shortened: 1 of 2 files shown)


perrozzi:
ciao, is there any specific reason to go to qcut=20 GeV instead of using the default 19 GeV?

alsusy:
20GeV was used in their previous analysis. The analyser tested 20 GeV and exo MC team agreed to keep 20 GeV.

perrozzi:
Sorry but I don't understand. The CMS-wide default for qcut is 19 GeV since Run1.
Why the previous sample (which one?) was done at 20 GeV?
I expect 19 or 20 not to change anything........
Can you please clarify? Thanks

AndreasAlbert:
Hi @perrozzi, what does "CMS-wide default" mean? For which processes and settings?

@alsusy is referring to [this previous request](https://cms-pdmv.cern.ch/mcm/requests?prepid=HIG-RunIISummer15GS-01266&page=0&shown=1151), which used qcut=20 for this process.

perrozzi:
For DY+jets I would say that most of the samples used 19 GeV, cfr https://github.com/cms-sw/genproductions/blob/master/python/ThirteenTeV/DYJetsToLL_M-50_HTbinned_TuneCUETP8M1_13TeV-madgraphMLM-pythia8_cff.py. 
If the previous VToQQsample was produced with 20 for some reason, ok. 
But I am asking myself if is worth having both 19 and 20 in the repository. Naively I would prefer to keep only the 19 to prevent confusion and edit manually the request where 20 is needed. And in any case, as said, 19 or 20 will not change much...
Would that make sense to you?

perrozzi:
(just to be 200% clear: if you want to use 20 GeV use 20 GeV in the request, for instance to be 100% compatible with 2016 settings, but I prefer not to have multiple qcut cfg in the repository)

PR #1769, opened 2018-04-26
[mg261] Remove loop in runcmsgrid_LO

AndreasAlbert:
Reversal of the commits from #1744

@kdlong @agrohsje @Saptaparna


AndreasAlbert:
(Can be merged as far as I am concerned. I verified that it works with a simple gridpack.)

PR #1771, opened 2018-04-27
add --stream to xmllint

perrozzi:
while investigating the problems with TOP-PhaseIISummer17wmLHEGENOnly-00084
I made a try and observed something I had not completely understood when only looking at the log.
The sequence
```
rwl_file 'pwg-rwl.dat'
rwl_format_rwgt 1

 finished computing weights ..

./runcmsgrid.sh: line 155:   591 Killed                  xmllint --noout ${file}_final.lhe > /dev/null 2>&1
xmllint integrity check failed on pwgevents.lhe

```
corresponds to the case in which the lhe file is correctly created (and valid too) but xmllint fails.
I checked and the lhe size for such large number of events is huge, 1GB,
and xmllint has memory problem. I googled it and found this thread
https://stackoverflow.com/a/32992580
by adding --stream things seem to work.
```
-bash-4.1$ xmllint --stream --noout cmsgrid_final.lhe
-bash-4.1$ xmllint --noout cmsgrid_final.lhe
Killed
```
I also checked that removing the last line I still get an exception, i.e. xmllint is working with the --stream option
```
-bash-4.1$ xmllint --stream --noout cmsgrid_final.lhe
cmsgrid_final.lhe:30607342: parser error : Extra content at the end of the document
</event>
        ^
cmsgrid_final.lhe : failed to parse
```


agrohsje:
the final error is that in the lhe of the campaign or because you opened and accidentally  modified it?

perrozzi:
the final error comes from the fact that I removed the last line **on purpose** to make a check

PR #1774, opened 2018-05-01
changes in mcfm script mdata config

carolhungwt:


qliphy:
@carolhungwt  ping.  Can you give some descriptions or do you have some TWIKI on MCFM gridpack?

PR #1775, opened 2018-05-02
Cards to the VLF analysis

casfisica:
Cards of the VLF  model, new analysis for the yellow report.
@jotadram6


perrozzi:
the PR contains also card for FFllUpTo3j_TauAndMuonDecays are they needed?

casfisica:
Hola 

Yes they are, in fact these are the important files for the generation, the other (VLF_...) are paramcards for variations of the model.
Thanks.
 
@jotadram6

casfisica:
The extra paramcards are the model interesting points, we want to produce.

PR #1777, opened 2018-05-03
add patch for mg26x without the mkdir aMCNLO command...

perrozzi:
when multi-run has been added as patch...


perrozzi:
plus script from efe to simply check for patched gridpacks (added comments to make output easy to understand)

PR #1778, opened 2018-05-04
adding cards for V'->V(qq)H(2V->2L2nu) 

cgalloni:
Cards for B2G diboson analyses with heavy resonances deacying into dibosons, HVT signals : Z'->Z(qq)H(VV->2L2nu)  and   W'->W(qq)H(VV->2L2nu)  for 2016 production. 
@aspiezia @zucchett


perrozzi:
can you please only upload one representative mass point and a script to produce all the others?
Thanks

PR #1779, opened 2018-05-04
Cards for st_tWll_5f_LO

ckmackay:
These cards have not been used for a while so I'm not sure about the end of the run_card, are the parameters of the systematics study needed? 
Can T = pdfwgt go anywhere?
Thanks


ckmackay:
Also in the proc_card there is a /h in the generate and add process lines which seems to cause problems, can I just remove this?

perrozzi:
Hi, the cards look ok to me, which problem seems the /h to cause?

ckmackay:
It failed at this mine and when I removed the /h the gridpack built with no problems, perhaps it was just coincidence. Are the /h needed, I couldn't find anything about them in the documentation?

ckmackay:
Updated run and madspin cards and can produce a gridpack now

PR #1785, opened 2018-05-08
1% and 5% cards: continuation of PR#1776

dmajumder:


perrozzi:
Hi, we usually ask to upload only representative cards and a script to produce all the parameter scan.
How many more cards you want to upload?
Is it ok to upload just the script to produce them?

dmajumder:
@perrozzi These should be the final set of cards. 
I don't have a script, it's just cp and sed commands to change the relevant lines in the cards. 
I'd rather we have all the cards here because it is not a huge number--20 samples combining mass and widths.

qliphy:
@dmajumder  You can refer to the script here: https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/VectorLQ/ProduceOtherMassPoints.sh

efeyazgan:
Hi @dmajumder, would you do the suggested update or shall we close the PR? thanks.

dmajumder:
@efeyazgan I didn't get time to update the PR with the scripts. However for archival purposes for these cards i think this PR should be accepted.

efeyazgan:
@dmajumder could you please keep only the minimum number of cards? Instead of a script you can write which parameters are used for the other cards in a readme. thanks.

dmajumder:
@efeyazgan i have updated the PR.

PR #1790, opened 2018-05-11
merge mg26x to master

perrozzi:


perrozzi:
@kdlong @agrohsje @AndreasAlbert @khurtado 
I waited a couple of days more before to merge mg26x into master to avoid mess during the long weekend, but now I don't understand anymore the diff between the 2 branches (I am afraid I merged a couple of PR that scrumbled a bit some of the files).
Can I kindly ask to have another look and fix the conficts? I swear after that the branch mg26x will be merged into master without further delay...

AndreasAlbert:
@perrozzi The conflict arises because`genproductions/bin/MadGraph5_aMCatNLO/Utilities/scripts` is a symbolic link in mg26x pointing to `genproductions/Utilities/`, while it is a proper folder in master.

So the question is what convention do we want.

AndreasAlbert:
Some of the additional files seem to have changed in my original version of the merge.
E.g. `/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Higgs/ggzh01j_5f_LO_MLM_M_ZToNuNu/ggzh01j_5f_LO_MLM_M_ZToNuNu_proc_card.dat` was present in `master`, but did not exist in `mg26x`. I think git somehow decided to delete that card in my merge. Now, when merging into master, it restores it. This should be fine, the question would be why the initial git merge did not just keep the card.

AndreasAlbert:
Why do we keep cards and scripts in the same repository? All of this would be much easier without the cards.

kdlong:
Agreed. We've discussed before. The obvious solution is to have a git submodule for the cards. I don't think there's any technical barrier to this other than implementing it and trying to do it in a way that introduces little additional confusion to the procedure of uploading cards and producing gridpacks. It just got pushed to the backburner with many other things, but now that things have cooled off a bit it's a good time to try again.

kdlong:
And for the symlink, I think this approach is preferred since the idea is to have the same scripts accessible to set PDFs for both MadGraph and POWHEG

AndreasAlbert:
Let's discuss that at #1793

perrozzi:
Ciao, I would prefer to separate the issues.
the most urgent and most important thing is to solve the conflicts.
concerning the convention on
" The conflict arises becausegenproductions/bin/MadGraph5_aMCatNLO/Utilities/scripts is a symbolic link in mg26x pointing to genproductions/Utilities/, while it is a proper folder in master."
I let @kdlong comment

AndreasAlbert:
@perrozzi @kdlong #1749 resolves the conflict by adopting the symbolic link approach. So you can merge #1749 into mg26x and then we can have a final look in this PR before merging -> master.

AndreasAlbert:
By the way: we did not separate cards for 24x and 26x in any way. This is not optimal, but there is no easy solution except for going through and sorting them.

agrohsje:
you can have the bias specification in 2.6.0 which wouldn't work in 2.4

perrozzi:
yes, but we are upgrading 2.4.2 branch to 2.6.0 and not the contrary, so should be fine ;)

AndreasAlbert:
yeah, I'm on it

AndreasAlbert:
I successfully ran `wplustest_4f_LO`  and `wplustest_4f_NLO`  gridpacks both locally and on cmsconnect. Checked that gridpacks can be executed and give consistent cross-sections between both modes.

gridpacks and output here: /afs/cern.ch/work/a/aalbert/public/2018-05-16_mg26x/gridpacks

AndreasAlbert:
@agrohsje @kdlong @perrozzi What about #1658? as @agrohsje pointed out to me, if we do not have that PR, we break the reweight_card feature in master. Is it ready to be merged?

perrozzi:
what about https://github.com/cms-sw/genproductions/pull/1801 ?

AndreasAlbert:
fine by me

PR #1795, opened 2018-05-16
14TeV Z'>ttbar cards

emanueleusai:
This PR adds 14TeV cards for Z'>ttbar. Needed for LHE campaign for upgrade studies.


emanueleusai:
Could you review this, please?
Thank you very much!

efeyazgan:
Could you please keep representative cards (e.g. for a single mass point) and if possible add a script to produce the other cards? thanks

emanueleusai:
Thank you for reviewing the cards.
If I understand correctly your request, these should be both in the commit already.
The reference cards are here:
https://github.com/emanueleusai/genproductions/tree/zp2j/bin/MadGraph5_aMCatNLO/cards/production/2017/14TeV/ZPrimeToTTJets_012j_LO_MLM/reference
while the script to generate the cards is here:
https://github.com/emanueleusai/genproductions/blob/zp2j/bin/MadGraph5_aMCatNLO/cards/production/2017/14TeV/ZPrimeToTTJets_012j_LO_MLM/generateCards.py

efeyazgan:
Yes, but would it be possible to decrease the number of files? i.e. keep only representative cards from which you can produce different ones (e.g. corresponding to different mass points).

emanueleusai:
Done.

efeyazgan:
Two more things:
- Could you please check https://twiki.cern.ch/twiki/bin/viewauth/CMS/Moriond18MC#Checks_to_the_datacards_BEFORE_T (e.g. remove jet definitions)
- Is there a particular reason that the PR is made for mg26 but not to master?

emanueleusai:
Done. This production requires MG 26+ because the large-width resonances cannot be produced with the older versions. This is a known problem, see https://hypernews.cern.ch/HyperNews/CMS/get/generators/3923/1.html

efeyazgan:
Thanks. Could you please remove commented lines like:
"# MASSES = [500, 750, 1000, 1250, 1500, 2000, 2500, 2750, 3000, 3250, 3500, 3750, 4000]
# WIDTHS = [0.01, 0.10, 0.30]
#format  [mass, [widths] ]" ?

PR #1802, opened 2018-05-24
ZprimetoNN cards

sihyunjeon:
Hi, this is Zprime to pair of heavy neutrino gen cards

Thanks in advance,

Sihyun.


efeyazgan:
Is this for mg26 for a specific reason?

sihyunjeon:
Not specifically but I had a chat about this with andreas 2,3 months ago and we thought it would be okay to use latest version.
I also checked our LHE files comparing v242 and v260 and there was no any odd differences between those two versions.

sihyunjeon:
@efeyazgan  Hi Efe, is there any problem for merging these cards?

efeyazgan:
No, I could look at the rest only now. Merging.

PR #1817, opened 2018-06-26
Update 0012 patch 

khurtado:
This should workaround a problem exiting ajob wrappers in the middle when multiple refine templates are copied and pasted into it.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/patches/0012-propagate-madevent-exitcode-properly.patch
+++ b/bin/MadGraph5_aMCatNLO/patches/0012-propagate-madevent-exitcode-properly.patch
@@ -1,5 +1,5 @@
 diff --git a/Template/LO/SubProcesses/refine.sh b/Template/LO/SubProcesses/refine.sh
-index b1a7314..7cd7a75 100644
+index 8276a2c..38aa774 100644
 --- a/Template/LO/SubProcesses/refine.sh
 +++ b/Template/LO/SubProcesses/refine.sh
 @@ -12,6 +12,10 @@ if [ -n "$SRT_LD_LIBRARY_PATH_SCRAMRT" ]; then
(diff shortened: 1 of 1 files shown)


qliphy:
Thanks @khurtado  
Would you please also update the master branch?
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/patches/0018-propagate-madevent-exitcode-properly.patch

khurtado:
Hi @qliphy I can, but apparently this is only needed in Madgraph v261 (according to Junho's tests, earlier versions (v240, v260) produced one ajob1 wrapper per integration channel, and the content was written from a fortran version that didn't use refine.sh as a template but wrote every single line  independently instead).

qliphy:
Thanks @khurtado   
Good to know that 242/260 is safe on this.  Would you please provide more details or links?  I found e.g. "Template/NLO/SubProcesses/ajob_template" are the same for 261 and 25X,  and can't find information related to what you mentioned from https://launchpad.net/mg5amcnlo/2.0/2.6.x

qliphy:
@khurtado :   About this "the content was written from a fortran version that didn't use refine.sh ",  I tried with MG242 but found ajob* wrapper is copied from refine.sh.

But anyway, as it has been tested by Junho, I think it is ready to be merged.  Just to keep an eye on possible effect of this PR on other versions of MG.

khurtado:
@qliphy Sorry for not replying back before. MG261 uses a python script that can use templates as a base. This uses `refine.sh` here: https://github.com/khurtado/MadGraph5_aMC-NLO/blob/cms-2.6.1/madgraph/madevent/gen_ximprove.py#L1103

260 and 242 on the other hand, seem to use the fortran code, which doesn't take any template but write the lines to a file on its own (lines are almost exactly like refine.sh, but any patch applied to refine.sh won't propagate to this fortran script due to the hardcoded write lines), this can be found here:
https://github.com/khurtado/MadGraph5_aMC-NLO/blob/cms-2.6.0/Template/LO/Source/gen_ximprove.f#L539-L777

As a result, 242 and before doesn't have any of the patches applied to `refine.sh`, so the `exit`  line at the end of the script here is not an issue. On top of that, according to Junho, 260 (and 242 too probably) create their own ajob wrapper per integration channel, however 261 can use a single ajob wrapper for more than 1 integration channel, the template `refine.sh` is copied and pasted more than once in the ajob wrapper, so the `exit` line in the patch before would make the wrapper exit before finishing all integration channels. This PR just enclose that exit in a conditional (exit with an error code only when there was an error in madevent)

qliphy:
@khurtado   Thanks a lot for the details.  Although I found in 260 there is also the python script as in 261:
https://github.com/khurtado/MadGraph5_aMC-NLO/blob/cms-2.6.0/madgraph/madevent/gen_ximprove.py

khurtado:
Yes, it just seems it wasn’t used until 261. The Fortran version is still
in 261 but now unused.

oliviermattelaer:
Hi,

Indeed since a while, the gridpack mode was still using the fortran version of the refine while the normal mode was using the python version. With 2.6.1 and the request to have gridpack running on read-only mode, they are now also using the python version instead of the fortran one.

In principle, no code are using the Fortran part anymore and it should be save to fully remove it. By security I have not done that yet.

Now concerning the PR, this seems to fix an issue created by some CMS patch and therefore I can not really comment. If you need something specific about this, do not hesitate to contact me (on launchpad if possible)

Cheers,

Olivier

PR #1818, opened 2018-06-28
MSSMD cards

weishi10141993:
Adding Higgs Exo MC experts @vischia @gourangakole @Claudio @Amandip

PR #1819, opened 2018-07-01
Updated MSSMD cards 

weishi10141993:
Old request was #1818. 
We'd like to require samples for 2017 data.
The pdf setting is updated in the run card.  Also removed parameters for SysCalc in the run card.

Again, adding Higgs Exo MC experts @vischia @gourangakole @amandip. The gridpacks are updated with the new cards at the same directory: /afs/cern.ch/work/w/wshi/public


weishi10141993:
@qliphy 
The cards are further cleaned. And the PDF is set to 306000 for 2017 MC in mg26 branch.

crossblaim:
Hi, @claudio here, I don't have any relation with this project, please stop mentioning me. Thanks.

PR #1820, opened 2018-07-02
Use rpath compiling option for GCC libraries

covarell:


covarell:
@smuzaffar @fabiocos please check that I have done a meaningful selection of library paths

covarell:
@smuzaffar @fabiocos do you mind having a look, please?

bin/Powheg/run_pwg.py
```diff
@@ -531,6 +532,13 @@ def runGetSource(parstage, xgrid, folderName, powInputName, process, noPdfCheck,
   
 echo "ANALYSIS=none " >> tmpfile
 
+# Add libraries now
+NEWRPATH1=`ls /cvmfs/cms.cern.ch/${SCRAM_ARCH}/external/gcc/*/* | grep "/lib64" | head -n 1`
+NEWRPATH1=${NEWRPATH1%?}
+NEWRPATH2=`ls /cvmfs/cms.cern.ch/${SCRAM_ARCH}/external/zlib-x86_64/*/* | grep "/lib" | head -n 1`
+NEWRPATH2=${NEWRPATH2%?}
+echo "RPATHLIBS= -Wl,-rpath,${NEWRPATH1} -L${NEWRPATH1} -lgfortran -lstdc++ -Wl,-rpath,${NEWRPATH2} -L${NEWRPATH2} -lz" >> tmpfile
+
```
smuzaffar: @covarell , why not add lhapdf libaray path in RPATHLIBS?

covarell: This is already the default in all POWHEG Makefiles, so no need to change

fabiocos:
@covarelli what is the plan? Reproduce all the gridpacks practically in use? Or just prepare for the new ones? This change should make unnecessary to use old CMSSW environments in the external LHE producer, so no need to modify the existing fragments, am I correct?

Furthermore, I am not sure whether this modification affects also JHUgen, that is a use case that should work as well

covarell:
At the moment the plan is to just remake the 4-5 gridpacks used in validation tests, and then prepare for new ones. 

Indeed, we just need to modify the fragments to specify the new gridpack location, nothing else.

JHUGen is contained in one of the validation gridpacks (this is done on purpose, exactly to spot such problems): we will immediately check if this is affected or not from IB tests.

fabiocos:
@covarell thank you, the important thing is that, besides fixing the validation in the short term, we need ensure that any production need can be managed with a slc6_amd64_gcc700 version of CMSSW_10_2_X.

covarell:
Hello, why was this very old commit not yet merged???

PR #1821, opened 2018-07-04
monoZ+LLP fragements

chadfreer:
Fragments to extend MonoZ samples to include long-lived particle decays.


qliphy:
Could you please keep representative fragment and add a script to produce the other files?

Moreover, are you requesting 2016 or 2017 sample?  For 2017, Pythia8 CP5 Tune instead of CUEP8M1 should be used here.

chadfreer:
I have created a python script which creates all of the subdirectories and python fragments. Would it be better to create a new pull request or would adding the script to the current branch suffice?

qliphy:
You can update this PR by adding the script and removing unnecessary fragments.

chadfreer:
I have deleted the unnecessary fragments and left only a script to create the fragments and a README . Please let me know if there is anything else that should be done for this PR.

PR #1824, opened 2018-07-11
fix 4V decays

acarvalh:
I also removed from the 4T cards an unnecessary statement that would make it crash on specific CMSSW releases


qliphy:
All look good to me. But maybe you don't need to have the following lines?:

https://github.com/acarvalh/genproductions/blob/b4472fa2d48adaf010bd01fa15e707b97ebcea6b/python/ThirteenTeV/Higgs/HH/ResonanceDecayFilter_example_HHTo4V_powheg_pythia8_CP5_cff.py#L29-L30

https://github.com/acarvalh/genproductions/blob/b4472fa2d48adaf010bd01fa15e707b97ebcea6b/python/ThirteenTeV/Higgs/HH/ResonanceDecayFilter_example_HHTo4V_madgraph_pythia8_CP5_cff.py#L27-L28

PR #1830, opened 2018-07-17
Rename the ggh012j_HPt-250ToInf_5f_NLO_FXFX_MH125 and fix the Hpt con…

ShevchenkoRostyslav:
Fix the name convention, remove the mass parameter and fix the wrong Hpt requirement.


ShevchenkoRostyslav:
Hi @qliphy . l.402 is just a comment, however, I have removed it in my new commit. I decided to completely remove the customizecards.dat file as it is not really needed.

PR #1831, opened 2018-07-18
update_gridpacks_local_mg26x.py

qliphy:
Update the patch inside the python file


efeyazgan:
Just to note here as well: we can also use the simplified version that has this correction as well:
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/Utilities/scripts/update_gridpacks_local_mg26x.py

PR #1832, opened 2018-07-23
Card for monoZprime pencil jet analysis for 2016 and 2017

varuns23:


qliphy:
Please don't upload all fragments. Just an example one together with the script to make others.  You can check other uploads for reference.

Moreover, please refer to
e.g.  remove the banner in the proc_card

PR #1835, opened 2018-07-26
WGamma_Monophoton_NLO_cards

gomber:
Hi All, 

I have created the cards for the Wgamma NLO sample, once I have comments/green light from your, I will do the same for ZNunugamma and Zllgamma samples too. I just want to make sure these cards looks fine, before I start committing other ones. 

Regards
Bhawna


AndreasAlbert:
The VG samples need MG 2.6, right? In that case, please add them to the mg26x branch rather than master

gomber:
Indeed Andreas, Ok i didn't relazie there is mg26X branch too! 
Will send the PR there then..

PR #1836, opened 2018-07-26
Wgamma NLO Cards in mg26X branch

gomber:
Hi All, 

As Andreas mentioned that I should create the PR in mg6X for the NLO samples. So here are the cards for the Wgamma Monophoton NLO samples. Please let me know if there are comments/questions. I will then upload the cards for Znunugamma/Zllgamma too.


qliphy:
Note like in master branch for 242 [1],  DEFAULT_PDF_SETS doesn't work for 26x.  

Please also check 
https://twiki.cern.ch/twiki/bin/view/CMS/Moriond18MC#PDF_settings_in_mg26x_Caveat_bei

[1] 
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/gridpack_generation.sh#L491

gomber:
Thanks @qliphy .
I have updated the card adding the following line as you suggested 👍 
 True = store_rwgt_info ! Store info for reweighting in LHE file

Please let me know if there is anything else I should change.

Regards
Bhawna

gomber:
Yeah i followed the example there. Updated it.

gomber:
@qliphy please let me know if the above files are fine, then I will upload other cards too.

qliphy:
@gomber  It looks ok to me, but please remove  https://github.com/gomber/genproductions/blob/ac1a7cbfbcd88b6761518f31a8d202fe61554298/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wgamma_Monophoton_Ptg130_NLO_fxfx/Wgamma_Monophoton_Ptg130_NLO_run_card.dat~

gomber:
@qliphy thanks, I will remove the ~ files and actually will start nw PR which is baed on mg26x version instead of master which is the case for this PR. And in that PR I will include Znunugamma, Zllgamma cards too.

Issue #1834, opened 2018-07-23
PDF/scale weights have systematic offset at NLO in MG 2.6.0


kdlong:
As reported in [1] some time ago, there seems to be a systematic offset in the central weight and scale weights when using NLO MG. This was reported to the authors but not followed up on. The issue is in standalone MG and seems to persist in the current version (2.6.2). The issue effects the few NLO samples which have already been produced with 2.6.0 for 2017, e.g. [2]. We should understand the source of this and have a solution before switching master to 26x.

[1] https://answers.launchpad.net/mg5amcnlo/+question/660414
[2] /ZGToLLG_01J_5f_TuneCUETP8M1_13TeV-amcatnloFXFX-pythia8/RunIISummer16MiniAODv2-PUMoriond17_80X_mcRun2_asymptotic_2016_TrancheIV_v6-v1/MINIAODSIM

qliphy:
Olivier provide a patch and we should include them in 26X branch.

Fortunately, it seems we still can use those already produced  samples to evaluate the uncertainties: "indeed this is an exact (up to numerical accuracy) factor of two.  So you can correct this issue easily via post-processing."

=== modified file 'madgraph/various/systematics.py'
--- madgraph/various/systematics.py 2018-03-25 08:30:56 +0000
+++ madgraph/various/systematics.py 2018-07-24 12:34:44 +0000
@@ -885,6 +885,7 @@
             for onewgt in cevent.wgts:
                 if not __debug__ and (dyn== -1 and Dmur==1 and Dmuf==1 and pdf==self.orig_pdf):
                     wgt += onewgt.ref_wgt
+ continue

                 if dyn == -1:
                     mur2 = onewgt.scales2[1]

kdlong:
Fixed by #1837

PR #1837, opened 2018-07-30
Add patch to fix systematics problem

kdlong:
Add patch provided by authors to fix issue #1834

Tested and working, should be ready to be merged. While testing, however, I noticed that there appears to be an issue with the patch applied to combine_events.f (the patch fails). @AndreasAlbert did you look at this?


AndreasAlbert:
@kdlong My bad. I ported this patch from 24x, but did not realize that it failed. In 24x, there were multiple copies of combine_events.f flying around the MG subdirectories, but in 26x, the MG authors cleaned it up. Additionally, they increased the hardcoded defaults in 26x, too, so I think we can scrap this patch for now.

AndreasAlbert:
PR here #1838

PR #1838, opened 2018-07-30
Remove unnecessary patch

AndreasAlbert:
The patch was necessary for 24x, but is obsolete in 26x since the MG authors implemented it themselves. If we ever run into a merging failure related to the maximum number of events read again, we can resurrect the patch trivially.


AndreasAlbert:
Prompted by discussion in #1837

PR #1839, opened 2018-07-30
Add tH_b 2017 cards

CAdloff:
Please check that only the tH_b will be added  in 
genproductions/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/
It is first time I use GitHub so I am not sure I have done things right (I see my comment "add tH_b" in previous branches... on my own GitHub account)


CAdloff:
I have followed Aniello slides to generated 2017 cards 
https://indico.cern.ch/event/682477/contributions/2796891/attachments/1563315/2462321/AnielloSpiezia_GenContact.pdf
So the cards where generated from 2016 cards using 
./Utilities/scripts/copyOldCardsTo2017.sh tH_b/
I have add 600, 650 and 700 mass points by just modifying 
set param_card mass 8000001 700 
in the  _customisecard and the output name in __proc_card.
Aniello has not mentioned I should do something with the width (this cards are for narrow width).

qliphy:
I see.  But these new cards are quite similar.  I would suggest you to just keep cards for one mass/width point.

CAdloff:
is it ok like this?
In the affirmative could you approve the cards? and the pull request?

PR #1848, opened 2018-08-08
modified cards for 2017 production of BulkGraviton VBF WW

IreneZoi:
In the previous version of the cards the extramodel one was not present and I think it is needed to avoid crashes during the gridpack production.
Also I've added a modification in the proc_card needed because of the removal of p and j definitions, as suggested in https://hypernews.cern.ch/HyperNews/CMS/get/generators/4014/1/1.html
and set pdfwgt to True in the run card


qliphy:
This branch has conflicts (with https://github.com/cms-sw/genproductions/pull/1849 which has already been merged) that must be resolved:  

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/exo_diboson/Spin-2/BulkGraviton_VBF_ZZ_inclu/BulkGraviton_VBF_ZZ_inclu_narrow_M/BulkGraviton_VBF_ZZ_inclu_narrow_M_extramodels.dat

PR #1852, opened 2018-08-08
Fix #1842

### LINKED ISSUE #1842: [CMSConnect] MG seg faults
Hi,

I keep getting seg faults for one of 400 MG subprocesses of my gridpack (full log at [1], subprocess log showing the seg fault at [2]). I do not get this error if I run in local mode, so I think it may be related to CMSConnect, but I have no good idea how to debug it seeing that there is so little information available in the logs.

@khurtado do you have any ideas for this one?

[1] https://gist.github.com/AndreasAlbert/b1105a66e52c792130b2d7d1fe47a340
[2] https://gist.github.com/AndreasAlbert/ae6659c7646a7531bb4066145e023a5d

khurtado:
This should fix #1842 by fixing the singularirity environment when `SRT_LD_LIBRARY_PATH_SCRAMRT` is not present but `SRT_LD_LIBRARY_PATH_SCRAMRTDEL` is.

Needs to be tested (@AndreasAlbert )


agrohsje:
Hi @khurtado @AndreasAlbert , I am trying to clean up our PR's. Can one of you comment if this is still needed? Should I merge. Please let me know. If so, is this also needed in UL2019 branch?

AndreasAlbert:
hi, as described in #1842, this fix resolved the problem I had at the time. Given that it's been so long, I do not have an immediate idea whether this issue still exists / whether people are still encountering it

agrohsje:
Thanks Andreas. @khurtado can you comment more. Should we still merge or just ignore for now?

khurtado:
@agrohsje I do not know if this is still an issue for others (it seems it was an unusual problem to begin with), but merging should be harmless. Since the patch for UL2019 is the same, applying it to that branch should also be harmless. That said, you might want to apply the PR and run some basic random workflow just to confirm it doesn't affect the basic functionality.

agrohsje:
Hi @khurtado,
these patches will conflict, or : 
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/patches/0009-fix-singularity-ld-library-path.patch ?

khurtado:
They shouldn't, as this patch (0020) assumes (0009) has been applied. So, it doesn't patch the original file, but the modified version by 0009. It does the same as (0016) does after (0009) in UL2019.

agrohsje:
When I try locally, I end up with: 
Reversed (or previously applied) patch detected!  Assume -R? [n] 
Apply anyway? [n] 
Skipping patch.
So all modifications in this file are skipped for me.

agrohsje:
Close this for now as patch was skipped when testing. Happy to re-open if needed.

PR #1858, opened 2018-08-10
Mg261

khurtado:
This fixes the `libquadmath` dependency problem stated in #1842 for branch mg261


khurtado:
@AndreasAlbert Same here, but for mg261 branch.

PR #1859, opened 2018-08-12
uploaded Scalar LQ cards for 17 and 18 production

gitytakahas:
Added pair and single leptoquark production processes, with undecayed LQ. 
For t-channel, LQ is decayed.


qliphy:
Usually we don't recommend to set large bwcutoff as "15000 = bwcutoff ! (M+/-bwcutoff*Gamma)"
Please set it to more standard value of "15".

gitytakahas:
Thanks Qiang; I changed it.

PR #1860, opened 2018-08-13
Adding cards for Zdark analysis

rosedj1:
@hroskes 
Heshy, 

These are the cards for the UF analysis as discussed with Lucien. There are cards for the following Zdark mass points: 

[1, 2, 3, 4, 7, 10, 15, 20, 25, 30, 35] GeV

Each is in their corresponding HAHMcards_MZD*/ dir. If there are any issues please let me know. Thanks!

- Jake


hroskes:
Actually I'm not the one who reviews these, that's up to the GEN conveners.

One issue that we know of is the PDF, which is set to cteq6l1 instead of an NNPDF version.  Also, the directory structure should be different.  Let's finalize all the details between us and Lucien by email and then you can push more commits.

qliphy:
Please remove the model file, which is :  https://github.com/cms-sw/genproductions/blob/f0b1bf4019d22be7bcbc35aad71f8cf3d725b2d2/bin/MadGraph5_aMCatNLO/UF_cards/HAHMcards_MZD10/HAHM_variablesw_v3_UFO.tar.gz

And please follow the instructions:

rosedj1:
My apologies. I'll follow those instructions. Do I remove ALL commented lines from, say, the run_card.dat?

qliphy:
As mentioned in the twiki, you can " remove the banner in the proc_card "

For run_card.dat, it is not needed to do so.

rosedj1:
Understood. The banner was the first thing I nixed, as was given in the TWiki. I just wasn't sure if the following comment pertained to all commented lines in, say, the run_card: "cleanup commented/unuseful lines". Per your suggestion, I simply won't touch the run_card. 

I'll fulfill the criteria in the TWiki, delete this PR, and then make a fresh PR. Is this OK?

hroskes:
You can also just push more commits to this branch and it fixes this PR.  (and you can edit the title too.)  Or if it's going to take longer, you can close this PR and then reopen it when you're ready.

qliphy:
> I just wasn't sure if the following comment pertained to all commented lines in, say, the run_card: >"cleanup commented/unuseful lines". 

For run_card.dat, you can keep the banner, but please clean the commented/unuseful lines.

qliphy:
Please don't upload all fragments. Just an example one together with the script (e.g. [1]) to make others.

[1] https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/VectorLQ/ProduceOtherMassPoints.sh

qliphy:
For PDF setting, please follow here:  https://twiki.cern.ch/twiki/bin/view/CMS/Moriond18MC#PDF_Choice_for_2017_production

hroskes:
Hi @qliphy and @efeyazgan,

About the PDF set, I understand from Lucien and @rosedj1 that the analysis is targeting a combination of 2016 and 2017 data.  Is the recommendation to use different PDFs for both years even in this case?  The results will be not entirely consistent.

It's most likely a small effect, but we ran into major issues in HIG-18-002.  In the end we had to use the alternate weight for NNPDF3.0 NLO for the 2017 samples because we saw inconsistencies between the two MC samples.  In that case it was more complicated because (a) we needed a consistent parameterization of the offshell tail of H->ZZ->4l and (b) we were looking at VBF and VH kinematics.  Neither of those applies to this analysis, but there will still be an unphysical cross section difference between the 2016 and 2017 samples that will be reflected in the analysis results.  Even if it's covered by systematics, it's better to calculate them the proper way, with the variations of that PDF set, rather than picking two "random" (with respect to each other) PDFs and averaging them.

I'm just helping with the sample production and I'm not really involved in the analysis, so maybe @rosedj1 can point the other analyzers here so they can comment.

Heshy

qliphy:
Thanks for the information.  Indeed we recently had a similar problem reported from Exotica Group [1].  And this usually happens in a new corner of phase space at high energy.  

@hroskes  @rosedj1  Would it be possible for any of you give an presentation at GEN meeting (every Monday 1 or 2pm)?  

[1] https://indico.cern.ch/event/746826/contributions/3094599/attachments/1696355/2737409/20180731_MCI_Zbmonoh_keynote9.pdf

rosedj1:
Yes, I can give a presentation at GEN meeting soon, but most of my group is away at the moment, for one reason or another. In the meantime, I have pointed the other analyzers to this PR, will be cleaning up the cards, and discussing the PDF sets with my group. Thanks for your time.

qliphy:
@rosedj1 Thanks. We have put you in the agenda of Sep 3rd GEN meeting:
https://indico.cern.ch/event/746827/

Please fell free to update the title.  We will contact you again before the meeting.

qliphy:
Should be this link:  https://indico.cern.ch/event/746818/

rosedj1:
@hroskes I've added the cards for the 263000 and 306000 and removed the useless lines as Qiang and the TWiki require. When you make the official request for samples, Lucien wanted me to say to make sure that we get 2016+2017, and if possible, 2018 would be nice. Thank you very much.
- Jake

hroskes:
Hi Jake, do you think you could get rid of [this folder](https://github.com/cms-sw/genproductions/tree/54696ae9146d6e5b4a0ab049f1d3b77caa88152b/bin/MadGraph5_aMCatNLO/UF_cards) and add a script to make the other mass points in [this folder](https://github.com/cms-sw/genproductions/tree/54696ae9146d6e5b4a0ab049f1d3b77caa88152b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Higgs/HToZZdTo4L_M125_MZd20_eps1e-2_13TeV_madgraph_pythia8), as [Qiang](https://github.com/cms-sw/genproductions/pull/1860#issuecomment-413445390) suggested earlier?

Also, the ones with 263000 should be in bin/MadGraph5_aMCatNLO/cards/production/**pre**2017/13TeV/Higgs/HToZZdTo4L_M125_MZd20_eps1e-2_13TeV_madgraph_pythia8/

rosedj1:
Hello, 

I've added the script, makecards.sh, which generates other cards for the user-specified mass points in their respective directories. Then, mkgridpack.sh, can be applied to the directory to generate the desired tarballs.  

Also, to be consistent, for the 263000 cards, I put them in: 
bin/MadGraph5_aMCatNLO/cards/production/pre2017/13TeV/**Higgs**/HToZZdTo4L_M125_MZd20_eps1e-2_13TeV_madgraph_pythia8/
instead of:
bin/MadGraph5_aMCatNLO/cards/production/pre2017/13TeV/**higgs**/HToZZdTo4L_M125_MZd20_eps1e-2_13TeV_madgraph_pythia8/
which was the preexisting directory. 

Finally, I deleted [the folder](https://github.com/rosedj1/genproductions/tree/UF_cards/bin/MadGraph5_aMCatNLO/UF_cards) from the command line and then made more commits, so hopefully these changes will be reflected on GitHub. Perhaps I need to manually delete it on GitHub instead. Not sure. 

Please let me know if everything is satisfactory. Thanks.

- Jake

qliphy:
There are many cards which are similar.  As you have already the scripts,  you can just keep some example ones and remove all the others.

qliphy:
For example, this one is not needed: 
https://github.com/cms-sw/genproductions/blob/e04d889d14e489c00c5333ae17cc670f3efafd8a/bin/MadGraph5_aMCatNLO/UF_cards/HAHMcards_MZD10/HAHM_variablesw_v3_run_card.dat

rosedj1:
@qliphy UF_cards directory deleted.

qliphy:
From your process definition, you should set ickkw=0
https://github.com/cms-sw/genproductions/pull/1860/files#diff-af380194b27520d4212c87d1d36576f0R62
and
xqcut to 0
https://github.com/cms-sw/genproductions/pull/1860/files#diff-afb7c1208ee7357f50367a83abe1c937R197

qliphy:
@rosedj1  Thanks. All look good to me.

PR #1828, opened 2018-07-13
ALP model cards

weishi10141993:
Similar to #1819,  they'll be used for central MC production for run 2017, except these cards are for an axion-like particle model.
Adding Higgs Exo MC experts @vischia @gourangakole @amandip
Once these cards are merged, I'll create gridpacks for the ALP central MC request.

PR #1861, opened 2018-08-13
Merge cards to MG5 v2.4.2 for central production

weishi10141993:
Hi Qiang,

Could you merge these cards for two signal models?

As I discussed with @gourangakole today, and also considering the current status of mg26x branch, I decide to regenerate gridpacks under mg5 v2.4.2, which is the current master branch for central MC production.

These cards were merged before in mg26x branch, they're basically the same, except I modified the pdf to $DEFAULT_PDF_SETS, here are the old PRs:
https://github.com/cms-sw/genproductions/pull/1819
https://github.com/cms-sw/genproductions/pull/1828
@vischia @amandip

Cheers,
Wei


weishi10141993:
Thanks Qiang. Do you need to revert the old PRs in mg26x branch in case there are conflicts? 

Wei

qliphy:
I think it is fine not to revert old PRs, as you may use mg26x in the next round.

weishi10141993:
Sounds good. Thanks!

PR #1865, opened 2018-08-24
better printing for calculateXSectionAndFilterEfficiency.sh

hroskes:
(don't lose newlines)


qliphy:
Thanks @hroskes   
If possible, would you please also update the 26x branch?:  https://github.com/cms-sw/genproductions/tree/mg26x

PR #1867, opened 2018-08-30
Updated lhapdf ID and med cut at 170 GeV for bbDM with 2HDM+a model

tiwariPC:
Dear all,

As according to the discussions about the genMET cut and the use of particular lhapdf id, we have updated the data cards with genMET cut at 170 GeV and lhapdf ID 305000 (as suggested by MC&I convenors).
Please review them.

Thanks and Regards,


efeyazgan:
It looks OK to me. @qliphy could you please take a look as well? thanks.

qliphy:
Can you remind us why you would like to go for 305000, i.e. NLO PDF, instead of the default NNLO one?

tiwariPC:
Hi Qiang Li,

This has been suggested by Andreas, as the NNPDF 3.1 NNLO PDF set gives strange results in extreme regions of the phase space. Hence we just switch from the "default" PDF to NNPDF 3.1 NLO, which does not seem to be affected and also could be used consistently for 2016 and 2017. Since we require samples for 2016 and 2017

Regards

efeyazgan:
Do you have any slides/documents about this? thanks

tiwariPC:
Hi Efe,
Please find the slides here 
https://indico.cern.ch/event/746826/contributions/3094599/attachments/1696355/2737409/20180731_MCI_Zbmonoh_keynote9.pdf

efeyazgan:
Thanks. So, you want use NNPDF3.1 NLO in the matrix element (ME). That's fine however then do you want to use a consistent PDF set in the parton shower  (PS) or not in the actual production. There are two possibilities, you use:

1) ME uses NNPDF3.1 NLO and PS uses CP3 (CP3 is a tune derived using NNPDF3.1 NLO)
2) ME uses NNPDF3.1 NLO and PS uses CP5 (CP5 is a tune derived using NNPDF3.1 NNLO)

Since this is an exotic sample both should be fine but it would be good if you decide in the group for what to do. If you'd like, we can have another iteration in one of the upcoming meetings.

PR #1871, opened 2018-09-10
adding POWHEG Wgamma

AndrewLevin:
I am adding the hadronizer and the powheg.input cards for the POWHEG Wgamma process.


qliphy:
@covarell @agrohsje @yuanchao   Can you have a look?

yuanchao:
Sure.

qliphy:
@covarell @agrohsje @yuanchao   

Is there any way to have all decay channels in a same request?

 vdecaymode 2  ! (1:electronic decay, 2: muonic decay, 3: tauonic decay)

efeyazgan:
Isn't it vdecaymode 123 ?

yuanchao:
I just checked that gridpack tarballs are created well for both Wpgamma and Wmgamma. As to remove the externalLHEProducer part, is it duplicated in the hadronizer script?

AndrewLevin:
vdecaymode 123 is not implemented for POWHEG Wgamma unfortunately

The externalLHEProducer is not included for other processes that I looked at e.g.

https://github.com/cms-sw/genproductions/blob/master/python/ThirteenTeV/HToSS/NLO_HToSSTobbbb_MH125_MS55_ctauS10_13TeV.py

I am not sure if we want the path to the gridpack hardcoded here.

agrohsje:
Yes. vdecaymode is a single number 1 or 2 or 3, there is no smart logic behind to have common setup. At least not when I checked a while ago. 
The setup is using NNPDF30_nlo_as_0118 ? Is this on purpose? I assume the Powheg settings were validated against a theory paper or used that way before? 
The shower fragment shouldn't be called CUETP8M1. What is the reason not to use the P8 internal emission veto but the CMS one? Is there a setting needed that is not part of the official version? 
Why shouldn't be there a common fragment? You don't run lhe and gen-sim separately?

yuanchao:
I also tested with LHE production of 1000 events for both processes. They are done without problems.

AndrewLevin:
I used NNPDF30_nlo_as_0118 because I thought this is the correct PDF set for use with 2016 data, but I am not sure. All of the POWHEG settings, other than the pdfset are taken from the powheg.input provided by the POWHEG authors with their release.

CUETP8M1 is changed to CP5 now.

The CMS emission veto is used because we need to veto electromagnetic radiation in addition to QCD radiation, since we are generating a process with a hard photon.

We are not planning to run lhe and gen-sim separately. However, we are planning to use the same fragment for all six charge flavor combinations. Should I produce the six gridpacks, then update with six shower fragments with the gridpack paths hardcoded?

agrohsje:
As you combine it with the CP5 tune I assumed it would be 2017 or 2018. @efeyazgan and @qliphy  can comment more on the combination of the NLO PDF and the proper tune in the context of 2016 data. 

The fact that you specified nevents and iseed is just because this is an example card but not what you used for gridpack generation? ndns1 should be removed as you use lhans. 

Why do you need these settings ?
pythiatune 320       ! (default 320, Perugia tune) tune for pythia6 
py8tune  5           ! (default 5) tune for pythia8 (14 -> Monash2013 tune)
I assume that was just for the example from the Powheg authors but now the parton shower is fully from P8.2 ? 

Was nfinalmode == 2 tested in the context of the Wgamma sample? What is the reason not to use 0 ?
Why do you add  'POWHEG:nFinal = 2' if you use EV1? I am puzzled that you run both routines.

Yes, six fragments seems the way to go if you generate it in one step.

qliphy:
It is for SM measurement and should be more for 2016.   And then the Tune should be set to CUETP8M1 instead of CP5.   The PDF can be NNPDF30_nlo_as_0118 corresponding to CUETP8M1.

AndrewLevin:
The entire pythia settings section from the powheg.input cards is removed now, since it was meant for use with the POWHEG authors hadronization code, not our CMS hadronization code which relies on a separate configuration file.

Also, I removed the hadronizer from this PR. Once this PR is approved, I will produce the 6 gridpacks and make a separate PR for the hadronizer. Please approve this PR independently.

AndrewLevin:
I removed the ndns1 and ndns2 lines and modified the seed and nevents lines. Please let me know if there are any other issues with the powheg.input cards.

agrohsje:
i have no further comments for now. thanks for all the fixes.

efeyazgan:
Could you please remove the commented input lines as well? thanks.

AndrewLevin:
The commented lines are removed now.

PR #1872, opened 2018-09-10
Adding MG5 cards for ggH+1j (@NLO) for cp-even, cp-odd and mixed-cp Higgs

danielwinterbottom:
Hi,

Can you please merge these 3 sets of cards to produce cp-even, cp-odd and cp-mixed H+1j events at NLO? Please let me know if any corrections are needed.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/GluGluToHToTauTauPlusOneJet_M125_13TeV_amcatnloFXFX_pythia8/GluGluToHToTauTauPlusOneJet_M125_13TeV_amcatnloFXFX_pythia8_extramodels.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/GluGluToHToTauTauPlusOneJet_M125_13TeV_amcatnloFXFX_pythia8/GluGluToHToTauTauPlusOneJet_M125_13TeV_amcatnloFXFX_pythia8_extramodels.dat
@@ -1,0 +1,1 @@
+HC_NLO_X0_UFO-v1.3.zip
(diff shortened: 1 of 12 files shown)


qliphy:
Please follow here:

Moreover, do you want to produce a NLO QCD or FxFx sample?  From your process card, it seems to be the NLO QCD one, but then why do you set ickkw=3 in your run card?

danielwinterbottom:
OK I have followed the twiki instructions and removed duplicated lines/the proc_card banner, except for this instruction "remove p and j definitions " - I kept the lines defining p and j as we need to use 5f scheme for p and j definition. If I removed these lines would the default definition use 4f scheme or is it ok to delete these lines also?

Yes the reason for the ickkw=3 is that we use FxFx merged H+0j H+1j and H+2j events [1] but we require additional statistics for the H+1j events which will later be stitched together with the inclusive H+0,1,2 jet sample. The H+1j events I think need to be generated with exactly the same settings as the inclusive sample including the ickkw=3 setting (this was done already to produce additional H+2j statistics and this is how it was done here anyway)

[1] https://cms-pdmv.cern.ch/mcm/requests?page=0&dataset_name=GluGluTo*HToTauTau_M125_13TeV_amcatnloFXFX_pythia8_tauola&prepid=*LHE*&shown=127

qliphy:
Thanks for the explanation on "ickkw=3". Looks clear to me now.

On the other hand,  "import model HC_NLO_X0_UFO-heft " seems to already imply 5 flavour as I can see from the screen output:  "Pass the definition of 'j' and 'p' to 5 flavour scheme."
So I think you can remove "remove p and j definitions" to follow the model default

danielwinterbottom:
OK I have removed the p and j definitions

PR #1874, opened 2018-09-15
add 100 NNPDF LO variations

syuvivida:
For Exotica/B2G MC production, we would like to use NNPDF 3.1 LO as the default PDF. In addition, we also would like to keep NNPDF 3.0 LO for comparison with previous 2015/2016 analyses. However, the 100 variation of these PDF sets are not saved by default. Therefore, runcmsgrid_LO.sh is modified.


qliphy:
Thanks @syuvivida   
However, 100 additional weights maybe a bit heavier, moreover those LO PDFs are used mainly by some analyses.  As listed in [our recent recommendation ](https://twiki.cern.ch/twiki/bin/viewauth/CMS/PhysicsComparisonAndGeneratorTunes#Recommendations_on_the_usage_of),   if LO PDF weights to be added, that can be done by hand.

PR #1879, opened 2018-09-20
add POWHEG Wgamma EmissionVeto1 hadronizers

AndrewLevin:
These POWHEG Wgamma hadronizers are associated with the Wgamma powheg.input cards in the merged pull request here: https://github.com/cms-sw/genproductions/pull/1871

As mentioned there, I am using the EmissionVeto1 plugin instead of the standard POWHEG emission veto plugin in order to properly veto QED radiation in addition to QCD radiation. In general, I have looked at the pythia code pythia8F77.cc provided by the POWHEG authors (I have placed a copy of this file here: /afs/cern.ch/user/a/amlevin/public/pythia8F77.cc), and tried to configure the EmissionVeto1 hadronizer the same way. I am not exactly sure whether EV1_QEDvetoMode should be 1 or 2, so please let me know if you have any further understand of that.

I have checked that these hadronizers run and will produce events, and I done a low-statistics gen-level comparison with the MadGraph_aMC@NLO Wgamma sample, which looks reasonable.


agrohsje:
Please fix the tune in parameterSets, Why nfinalmode 2? As written I added this for bbbar4l. Do you need it?

AndrewLevin:
The tune in parameterSets is fixed now.

I used nfinalmode 2 to avoid the crash at https://github.com/cms-sw/cmssw/blob/master/GeneratorInterface/Pythia8Interface/plugins/EmissionVetoHook1.cc#L312 which occurs at a rate of around 1 per 20000 events generated in this sample.

agrohsje:
But does this logic: 
https://github.com/cms-sw/cmssw/blob/master/GeneratorInterface/Pythia8Interface/plugins/EmissionVetoHook1.cc#L279
gives you the same result for the other 19999 events as nfinal=2 ? 
What are the diagrams in your case that cause problems ? For bbbar4l this is only because of the Z->ll,g->bbbar contribution.

qliphy:
@AndrewLevin   Can you explain a bit more your choice of the following?
                         EV1_pThardMode = cms.int32(1),
                          EV1_pTempMode = cms.int32(0), 
I can't fine them in pythia8F77.cc you mentioned.

AndrewLevin:
@qliphy, in the file pythia8F77.cc, they are just called pThardMode and pTempMode

qliphy:
@AndrewLevin   Yes, I noticed that.  But I can't find the recommended values for pThardMode and pTempMode.  There seems to be just some options.

AndrewLevin:
It seems they are both set to 0 by default. For example:

`pythia.readString("POWHEG:pThard = 0");`

qliphy:
Thanks.  There are some descriptions in pythia8F77.cc,  not sure whether the default value is better or not.  

Maybe you can consult with the author with all your settings and ask for their opinion.

AndrewLevin:
@agrohsje, here is an example of an event that causes the crash:

`--------  PYTHIA Event Listing  (complete event)  ---------------------------------------------------------------------------------
 
    no         id  name            status     mothers   daughters     colours      p_x        p_y        p_z         e          m 
     0         90  (system)           -11     0     0     0     0     0     0      0.000      0.000      0.000  13000.000  13000.000
     1       2212  (p+)               -12     0     0     3     0     0     0      0.000      0.000   6500.000   6500.000      0.938
     2       2212  (p+)               -12     0     0     4     0     0     0      0.000      0.000  -6500.000   6500.000      0.938
     3          2  (u)                -21     1     0     5     0   501     0      0.000      0.000    341.558    341.558      0.000
     4         -1  (dbar)             -21     2     0     5     0     0   501      0.000      0.000     -4.491      4.491      0.000
     5         24  W+                  22     3     4     0     0     0     0     -0.000     -0.000    337.067    346.049     78.329
                                   Charge sum:  1.000           Momentum sum:     -0.000     -0.000    337.067    346.049     78.329

 --------  End PYTHIA Event Listing  -----------------------------------------------------------------------------------------------`

In this event, count = 1 and nFinal = 2.

AndrewLevin:
I have set nFinal = 2 and nFinalMode = 0 now. As mentioned above, this will cause a small rate (less than 0.1%  events) of failures in the hadronization, but I think this is acceptable. 

The POWHEG authors use this logic in their hadronization code:

```
    // Extra check that we have the correct final state                                                                                                    
    if (count != nFinal && count != nFinal + 1 && count != nFinal - 1) {
      cout << "Error: wrong number of final state particles in event: count= " << count << endl;
      exit(1);
    }

    // Flag if POWHEG radiation present and index                                                                                                          
    bool isEmt = (count == nFinal) ? false : true;
    int  iEmt  = (isEmt) ? e.size() - 1 : -1;
```

I think the POWHEG authors code may not be correct. It seems to me it should be `bool isEmt = (count == nFinal+1) ? true: false;`. I have asked the POWHEG authors about this, but they do not always reply quickly, or ever. If the POWHEG authors confirm this, I can make a new CMSSW pull request to add this logic to CMSSW, probably with a new nFinalMode. But since it affects less than 0.1% of events, I think we can just go ahead for now with the current fragment and code.

agrohsje:
Thanks for the info. For regular cases the Powheg snippet is ok. If count corresponds to nfinal we deal with the underlying born and isemt is false. In the case of these samples however, the gamma can also not appear in the event  record, as you show above. If there is no additional radiation on top, ncount is just 1. That's why they added this extension  count != nFinal - 1 which is normally not present. So for this case, indeed isemt seems not properly defined. Is that what you asked them? Whom did you ask? I would suggest to implement this logic in EV1 correctly and use that. I agree that 0.1% is tiny and doesn't matter, but why not doing it consistent and properly. It shouldn't be much work?

AndrewLevin:
I asked these people:

To: "Mauro Chiesa"<mauro.chiesa@physik.uni-wuerzburg.de>, "Guido Montagna" <guido.montagna@pv.infn.it>, "Paolo Nason" <paolo.nason@mib.infn.it>, Oreste Nicrosini <oreste.nicrosini@pv.infn.it>, "fulvio.piccinini@pv.infn.it" <fulvio.piccinini@pv.infn.it>

What I asked them is this:

> Hi POWHEG authors,
> 
> I have another question relating to your Wgamma process. I found this logic in your pythia8F77.cc file:
> 
>     // Extra check that we have the correct final state                                                                                                    
>     if (count != nFinal && count != nFinal + 1 && count != nFinal - 1) {
>       cout << "Error: wrong number of final state particles in event: count= " << count << endl;
>       exit(1);
>     }
> 
>     // Flag if POWHEG radiation present and index                                                                                                          
>     bool isEmt = (count == nFinal) ? false : true;
>     int  iEmt  = (isEmt) ? e.size() - 1 : -1;
> 
> If count == nFinal -1, this means the event is a "W" event with no jet and no photon, so in this case isEmt should be false, right? So, to me it seems that the correct logic is:
> 
>     // Flag if POWHEG radiation present and index                                                                                                          
>     bool isEmt = (count == nFinal + 1) ? true : false;
>     int  iEmt  = (isEmt) ? e.size() - 1 : -1;
> 
> Andrew

I asked them yesterday, so I guess we can wait several more days to see if they reply.

agrohsje:
Shouldn't one distinguish between the two case? I.e. for the W case 1 + 1 = 2 == nfinal is the radiation case.

AndrewLevin:
The e-mail thread from the POWHEG authors I think resolved the issue: their logic in doVetoMPIStep regarding the isEmt flag is not correct, but given that pThardMode = 0, the variables that are set in that function do not have any effect outside that function, and that function returns false.

I will go forward with making a CMSSW pull request for EmissionVeto1 which will add nFinalMode = 3 in cmssw which will do the same thing as nFinalMode = 2 except that 

`count != nFinal && count != nFinal + 1` --> `count != nFinal && count != nFinal + 1 && count != nFinal - 1`

in order to avoid the crash.

agrohsje:
Agreed. However, you should check if nfinalmode == 3 and pthardmode !=0 is chosen and throw an exception with a proper error, so that nobody is using the "buggy code". You can add some sentences in the code in case we want to extend it later so that we still remember. thanks !

qliphy:
"The e-mail thread from the POWHEG authors I think resolved  the issue: their logic in doVetoMPIStep regarding the isEmt flag is not  correct, but given that pThardMode = 0, the variables that are set in  that function do not have any effect outside that function, and that  function always returns false."

But does that mean isEmt was wrongly set by the authors?  Will this affect results in their paper?

AndrewLevin:
@qliphy, I forwarded you then e-mail thread. I think the answers to your question are yes, isEmt is wrongly set by the authors and no, this does not affect the results in their paper.

qliphy:
Thanks @AndrewLevin 

Write down here Mauro Chiesa's reply for reference 
">>> The logic is not correct, however we do not use the function dovetoMPI:
>>> it is called, but it always returns false (=event not vetoed). We don't
>>> think that the MPI should be vetoed because of the powheg radiation.
....
>>> in the interface pThardMode = 0, so pThard is always set to scalup."

AndrewLevin:
Here is the CMSSW pull request: https://github.com/cms-sw/cmssw/pull/24763

PR #1881, opened 2018-09-20
VA card with mg261 for 2017, in which photon eta asymmetry is fixed

menglu21:
as there is a photon eta asymmetry for VA FXFX sample in Madgraph 2.4, new gridpacks are generated with MG2.6.1 for 2016 already. These cards are corresponding to 2017 sample, generated with MG2.6.1


qliphy:
As mentioned by https://indico.cern.ch/event/754500/contributions/3143876/attachments/1717820/2771965/SMP_VV_WGamma_18Sep2018.pdf#page=10&zoom=auto,-99,549

It might be better to remove this cut:
30=mll_sf

menglu21:
@qliphy ok, i'll re-generate the gridpack for ZA

PR #1883, opened 2018-09-23
reduce the mll cut in the ZGamma sample from 30 to 4 GeV

AndrewLevin:
In order to estimate the background in the WGamma analysis, we need to reduce the mll_sf cut from 30 to 4 in the ZGamma sample, such that events where an FSR photon takes most of the pt of its mother lepton are included. I have checked that this increases the cross-section of the sample from 75.13 to 76.98 pb, before the matching efficiency.


qliphy:
Would you please upload new cards instead of overwriting here? Otherwise people will get confused about  the cuts applied to current ZG sample.

AndrewLevin:
Yes, I have undone the change to the original card and uploaded new cards now in a new directory now.

PR #1887, opened 2018-09-25
Cards and fragment for SIDM model

phylsix:
Requesting 2017 MC for SIDM (Self-Interacting Dark Matter) model. Thanks! <weinan.si@cern.ch>


qliphy:
The cards are based on MG26X.  However, this version is still being verified yet. The master branch of genproductions is still with MG242.   So please change your cards to be consistent with 242.

qliphy:
and push to master branch instead of mg26x

PR #1890, opened 2018-09-27
Cards for MonoZ(ll) Pseudoscalar 2HDM

AndreasAlbert:
The cards are to be used for 2016+2017+2018.

I added a set of cards for a representative mass point, as well as code to create the cards for all points of interest.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/MonoZLL/Pseudoscalar2HDM/create_2hdm_cards.py
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/MonoZLL/Pseudoscalar2HDM/create_2hdm_cards.py
@@ -1,0 +1,94 @@
+#!/usr/bin/env python
+#-*- coding:utf-8 -*-
+import os
+import sys
+import fnmatch
+import logging
+import shutil
+from lib.card_lib import *
+from lib.TwoHDM_lib import *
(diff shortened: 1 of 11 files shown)


qliphy:
What is the PDF choice for 2016+2017+2018?  Seems all to be NNLO?  

https://github.com/cms-sw/genproductions/blob/7ed0231155ce590ce6f30a183c9bccc7685a8e96/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/MonoZLL/Pseudoscalar2HDM/templates/pseudoscalar_2hdm_5f/run_card.dat#L51-L53

You may check with recent updated recommendation:

qliphy:
Your process is NLO, but the run_card seems for LO?

AndreasAlbert:
Hi @qliphy, thanks for checking the cards.

Regarding PDF: I have checked the fraction of negative events with the NNLO PDF for all the relevant parameter space. The highest rate of negative events I have observed is around 10^-4, so I think it's fine.

Regarding LO run_card: Since this is a loop-induced process, MG wants a LO run card. If you try running with an NLO card, it crashes (I tried). There was a quick exchange about this on the [hypernews](https://hypernews.cern.ch/HyperNews/CMS/get/generators/3588.html) last year.

Regarding cut_decays: I am not using any decay syntax, so this setting has no effect.

PR #1891, opened 2018-09-27
Patch to fix makefile for collier in mg260

lviliani:
The patch fixes the problem related to loop induced processes with the collier library discussed in this thread:
https://hypernews.cern.ch/HyperNews/CMS/get/generators/4099.html
The solution is the one proposed here:
https://answers.launchpad.net/mg5amcnlo/+question/488071
This problem is there only for mg260, mg261 already contains the same fix by the authors.


qliphy:
Thanks!  Please also update at hn with this information.

PR #1894, opened 2018-09-28
change the lepton pt cut to 0 in the ZA low mll sample

AndrewLevin:
Following up on https://github.com/cms-sw/genproductions/pull/1883, the lepton pt should also be reduced in addition to the mll cut. When the lepton pt cut is changed from 15 GeV to 0, the cross-section before matching increases from 76.98 pb to 146.6 pb


qliphy:
Please note for ZA FxFx,  MG242 is buggy,  you should use MG26X.
https://hypernews.cern.ch/HyperNews/CMS/get/prep-ops/4041.html

PR #1895, opened 2018-10-02
TT_CEDM

sam7k9621:
I have already followed the further cleanup suggestions
Please let me know if there still are somethings I need to modify

Also, I met some access issue with uploading UFO file, I am contacting with Sasha and Wajid, they will help me to upload them

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TT_CEDM/ProduceOtherdtG.sh
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TT_CEDM/ProduceOtherdtG.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 firstdtG=1
-lastdtG=1
+lastdtG=5
 echo "You are creating the cards for the TT_CEDM_dtG process, starting from dtG=$firstdtG GeV to dtG=$lastdtG in steps of 1."
 
 index=0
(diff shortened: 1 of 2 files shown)


qliphy:
Your run_card.dat seems to be based wrongly on LO card instead of NLO one.  Please refer to 

https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/dyellell012j_5f_NLO_FXFX/dyellell012j_5f_NLO_FXFX_run_card.dat


Output directory name should match your proc card name:
https://github.com/cms-sw/genproductions/blob/840ed34d9d413f8e7176e8e10a3faaf3d1151bf1/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TT_CEDM/TTbar_LO_dtG0/TTbar_LO_dtG0_proc_card.dat#L4


And there is no need to update dyellell012j_5f_NLO_FXFX pro card, as the model already implies 5f:

sam7k9621:
Thanks for your response

However, I want to request for  LO samples, are there any particular reasons that I should request for NLO ? 

Here's the reference paper
https://journals.aps.org/prd/pdf/10.1103/PhysRevD.93.014020

qliphy:
From your proc card, I saw 
      generate p p > t t~ [QCD] @0 
this is for NLO process.

PR #1903, opened 2018-10-10
Z'-> ttbar signal samples for Fall17 production

karavdin:


qliphy:
No need to upload so many cards.  You can just keep cards for one benchmark points, together with the script to make other cards.

Can you also explain a bit more about multiple_try patch you would like to add?
https://github.com/cms-sw/genproductions/blob/898cdd736c5f316952dab52c8880d9619424b162/bin/MadGraph5_aMCatNLO/patches/0023-multiple-try-compilation.patch

karavdin:
Hi @qliphy, 
Thank you for quick feedback. Only benchmark points are left. This patch helps a lot to make gridpack production more smooth. I took it from suggestion on HN [1].
[1] https://hypernews.cern.ch/HyperNews/CMS/get/generators/3898/1.html

qliphy:
Just noticed you are committing to 26X branch.   Note the current default is 242 yet.  

26X still needs some time to be the default.

Moreover, your run_card seems to be based on 242 but not 26X.  Please have a check.

And the files are still too many.  This will slow down git clone quite much. Please only keep necessary files. Thanks!

karavdin:
We had issues to generate high mass points with 242 and were suggested switch to 26X version. To be consistent with generated with 2.6 v gridpacks (explicitly checked), I push cards to 26X branch.

qliphy:
But the run_card.dat files is based on 242 template not 26X.  Do you want to update run_card.dat to 26X?

karavdin:
Could you point me to 26X template, please? Do we miss anything critical in run_card? It took more than a month to produce gridpacks and I would like to avoid re-doing them unless its absolutely necessary.

karavdin:
Unfortunately I could not spot any critical difference in examples for 2.6 cards. However I can understand concern  to not have cards under 26X brunch if they don't use 26X functionality. Therefore I created another PR for master branch with cards tested with MG 2.4.2.  This PR is going to be closed.

PR #1904, opened 2018-10-10
Add MonoZ DMSimp cards for 2017/18.

AndreasAlbert:
Added one set of cards for Scalar, Pseudoscalar, Vector and Axial scenarios and the respective scripts to create all cards. 

The scalar and pseudoscalar scenarios use NNPDF 3.1 NNLO because they are gluon-initiated and are not affected by large fractions of negative events.

The vector and axial scenarios use NNPDF 3.1 NLO to avoid the weight issue. They will be run with tune CP3.


qliphy:
Is there any specific reason to go for 26X?

AndreasAlbert:
Not very specific, but already for 2016, the DM signals had been produced with MG2.5. Using 2.6 makes sure we do not down-grade :)

PR #1916, opened 2018-10-15
Hj minlo nnlops nnpdf31

Andrej-CMS:


covarell:
These look good to me

PR #1920, opened 2018-10-15
add Zto3photon

menglu21:
add Zto3photon for 2016 sample


qliphy:
Please remove these lines:  https://github.com/cms-sw/genproductions/blob/43d35a047843f012d8ae82d176e3546dbda8bf8c/bin/MadGraph5_aMCatNLO/cards/production/pre2017/13TeV/ZTo3G/ZTo3G_run_card.dat#L272-L283

qliphy:
It is not a MLM sample, so please set ickkw to 0 and xqcut to 0

PR #1930, opened 2018-10-19
[mg26x] Add patch to make MG5 more resilient against server errors

AndreasAlbert:
A better version of this hotfix will be included in the official 2.6.4
release, so we can remove this patch at that time.


AndreasAlbert:
@efeyazgan @qliphy This patch fixes a problem with MG trying to retrieve package info from a remote web server. At the moment, the web servers run without problems, so this patch is only a safety measure in case the servers will mis-function again in the future.

AndreasAlbert:
The other patch we discussed regarding the retrieval of the collier package tarball from hepforge is not necessary anymore. The problem was that the URL `http://collier.hepforge.org/collier-latest.tar.gz` was intermittently not working a while back. As of now, it seems to be working again, so we do not need to patch it.

PR #1939, opened 2018-10-25
Correct bug in request_fragment_check when checking Fall18 cmssw version

clacaputo:
Dear @efeyazgan @qliphy , I've found this error while testing with `request_fragment_check.py` some Fall18 requests that use **CMSSW_10_2_6**:

```
[ERROR] PS weights in config but CMSSW version is not 10_2_3 - please check!
[OK] Parton shower weight configuration probably OK in the fragment
```
I fix the script in order to print the error only if the CMSSW version in less than 10_2_3


efeyazgan:
Very good! Thanks a lot. Could you please fix the spelling of "preavious"?

clacaputo:
Hi @efeyazgan , done! Sorry about that.

efeyazgan:
thanks

PR #1942, opened 2018-10-26
Add additional partons

sam7k9621:
Add additional partons in proc_card.dat


qliphy:
You are not adding partoning but just repeat the process twice.

You may want to add
        add process p p > t t ~ j @1
        add process p p > t t ~  j j @2

but then you should also update your run_card for the MLM matching

sam7k9621:
I have modified the run_card for MLM matching and also the proc_card to add partons
Also, change the b mass into 0 for 5f

PR #1943, opened 2018-10-26
ggZH

qyguo:


qliphy:
@qyguo @hroskes   It would be good if any of you can come to GEN (not so urgent, possibly in late Nov or Dec.) to give an update on JHUGen and recent development.

PR #1946, opened 2018-10-31
H plus tau nu2017

slehti:
Updated Charged Higgs to tau nu MG5_aMCatNLO cards to 2017.


qliphy:
Please upload cards just for an example benchmark point together with a script to generate cards for other points. You can find examples in the genproduction cards area.

And do you need also to add decay for h- and t?
https://github.com/cms-sw/genproductions/blob/1efff76babc959dd117e7e936e7781c1ea6b283b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ChargedHiggs_heavy_M1000/XXX_madspin_card.dat#L9-L10

slehti:
Done.

Decays for H- and t added.

In 2016 production MadSpin did not work for samples mH+ > 500 GeV, assuming same here (taken into account in the heavy H+ script), letting Pythia decay samples with mH+ > 500 GeV.

PR #1954, opened 2018-11-06
Use condor batch system for making and running POWHEG gridpacks 

covarell:
First version of scripts for making gridpacks and running them in parallel.
- Added as separated files (not overriding existing ones), soon to become the default
- Additional user requests than basic "job flavour" (e.g. minimum memory, maximum disk space on the nodes etc.) can be specified in a file called additional.condorConf, to be placed in the main working directory
- Tested with full mode (f) and parallel mode (1239 + step 7 for HJ-NNLOPS)

### DIFF
--- a/bin/Powheg/run_lhe_condor.sh
+++ b/bin/Powheg/run_lhe_condor.sh
@@ -40,7 +40,7 @@ echo -e "echo \"Start of job on \" `date`" >> submitEvents.sh
 echo -e "source /cvmfs/cms.cern.ch/cmsset_default.sh" >> submitEvents.sh
 echo -e "tar -xzvf ${WORKDIR}/${tar}" >> submitEvents.sh
 echo -e "./runcmsgrid.sh ${nume} \$1 1" >> submitEvents.sh
-echo -e "mv cmsgrid_final.lhe ${WORKDIR}/cmsgrid_final\$1.lhe" >> submitEvents.sh
+echo -e "mv cmsgrid_final.lhe ${WORKDIR}/cmsgrid_final_\$1.lhe" >> submitEvents.sh
 chmod a+x submitEvents.sh 
 
 cat << EOF >> events.condorConf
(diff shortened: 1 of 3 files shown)


covarell:
@yuanchao

yuanchao:
Sorry for the late reply. With a quick look it seems ok. But I tried a simple example on lxplus, there's nothing returned after the condor job finished. LSF jobs on lxplus run in a temporal directory unless changed to AFS in the script. It's probably the same for condor jobs. Need to check the log for better understanding.

covarell:
Hi @yuanchao can you send me the command you tested?
Of course when I validated, I checked that something was returned :-)

yuanchao:
Ok. I'm new to HTCondor. It looks like I choose a too short job queue and the job got killed according to the log. I'm submitting another job for test.
'''
python ./run_pwg_condor.py -p f -i DMGG_NNPDF30_13TeV.input -m DMGG -f my_DMGG_c1 -q longlunch -d 1
'''
Will see how it works. Since this PR is rather isolated from other parts, we can accept it first and send issues if there's any later.

covarell:
Apparently the choice of the JobFlavour is not a trivial thing as we discussed with @kdlong. 
With "tomorrow" I could not get any job running after one day. With "microcentury" I could do most of the tasks but some ran out of time (e.g. step 2).
Efe suggested to use instead explicit requirements on estimated CPU and memory usage, which you can place in a text file called "additional.condorConf"

efeyazgan:
Hi @yuanchao, are there any news on this? thanks

covarell:
I think Yuan's suggestion is to merge the commit and then we progressively start updating the Twiki,
Do you have other validation results you would like to see before doing that?

covarell:
I have updated the Twiki (of course leaving all the LSF options in, because the PR is not approved yet and also LSF is still usable)
https://twiki.cern.ch/twiki/bin/view/CMS/PowhegBOXPrecompiled

PR #1956, opened 2018-11-07
update run cards with NNPDF31 LO & genelic hadronizer for LQ -> taub and topnu decay

gitytakahas:
Set the PDF to NNPDF31 LO (LHAID = 315200)

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ScalarLQ/NonRes_M500/NonRes_M500_customizecards.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ScalarLQ/NonRes_M500/NonRes_M500_customizecards.dat
@@ -1,4 +1,5 @@
 set param_card MASS 9000006 500
+set param_card DECAY 9000006 AUTO
 set param_card FRBLOCK 1 0.0
 set param_card FRBLOCK 2 0.0
 set param_card FRBLOCK 3 1.0
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ScalarLQ/NonRes_M500/NonRes_M500_run_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ScalarLQ/NonRes_M500/NonRes_M500_run_card.dat
@@ -49,7 +49,8 @@
 # PDF CHOICE: this automatically fixes also alpha_s and its evol.    *
 #*********************************************************************
  'lhapdf'    = pdlabel     ! PDF set                                  
- $DEFAULT_PDF_SETS = lhaid
+ 315200 = lhaid
+# $DEFAULT_PDF_SETS = lhaid
  $DEFAULT_PDF_MEMBERS = reweight_PDF      ! if pdlabel=lhapdf, this is the lhapdf number
 #*********************************************************************
 # Renormalization and factorization scales                           *
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ScalarLQ/Pair_M500/Pair_M500_customizecards.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ScalarLQ/Pair_M500/Pair_M500_customizecards.dat
@@ -1,4 +1,5 @@
 set param_card MASS 9000006 500
+set param_card DECAY 9000006 AUTO
 set param_card FRBLOCK 1 0.0
 set param_card FRBLOCK 2 0.0
 set param_card FRBLOCK 3 1.0
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ScalarLQ/Pair_M500/Pair_M500_run_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ScalarLQ/Pair_M500/Pair_M500_run_card.dat
@@ -49,8 +49,11 @@
 # PDF CHOICE: this automatically fixes also alpha_s and its evol.    *
 #*********************************************************************
  'lhapdf'    = pdlabel     ! PDF set                                  
- $DEFAULT_PDF_SETS = lhaid
+# $DEFAULT_PDF_SETS = lhaid
+ 315200 = lhaid
  $DEFAULT_PDF_MEMBERS = reweight_PDF      ! if pdlabel=lhapdf, this is the lhapdf number
+
+
 #*********************************************************************
 # Renormalization and factorization scales                           *
 #*********************************************************************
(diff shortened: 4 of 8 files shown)


qliphy:
Looks good.   But please notice the following written in[ the GEN twiki](https://twiki.cern.ch/twiki/bin/view/CMS/PhysicsComparisonAndGeneratorTunes#Recommendations%20on%20the%20usage%20of):
"And please also add NNPDF3.1LO weights if needed (ref) "

gitytakahas:
Thanks for merging Qiang. What do you mean by "add NNPDF3.1LO weights if needed (ref)" ? 
Isn't it sufficient to have, in run card, 

>  $DEFAULT_PDF_MEMBERS = reweight_PDF      ! if pdlabel=lhapdf, this is the lhapdf number

I may misunderstood?

qliphy:
For LO process,  as you can see from below,  "$DEFAULT_PDF_MEMBERS" doesn't play any role

https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/gridpack_generation.sh#L503-L512

Instead you the PDF reweight is defined here
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/runcmsgrid_LO.sh#L159

gitytakahas:
Hi Qiang,

Thanks a lot for the explanation. I see.
It seems that NNPDF31_lo_as_0130 is already defined in runcmsgrid_LO.sh. So I guess I don't need any action here (i.e. it will save PDF weights). Could you confirm if that is the case ?

Thanks,
Yuta

qliphy:
Hi Yuta,   you may need to change "NNPDF31_lo_as_0130.LHgrid 1" to "NNPDF31_lo_as_0130.LHgrid",  i.e. remove the "1" to include the PDF uncertainties.

gitytakahas:
Hi Qiang, 

I just tested run_cmsgrid_LO.sh (without any modification) and found that it produces PDF weights as you can see here. 

/afs/cern.ch/user/y/ytakahas/public/forQiang/cmsgrid_final.lhe

So, I think it is fine as it is now. Do you agree? 

Thanks,
Yuta

qliphy:
@gitytakahas   I looked into the lhe file, but could not find the weights from NNPDF31_lo_as_0130 (which has 101 variations including the central one).

BTW:  I found a strange thing. your process is 5-flavor one including b in the proton definitions, but the reweight PDF picked seems to be 4-flavor.  Can you have a check?

https://github.com/cms-sw/genproductions/blob/cdfa5a18099992af82ae162738eeae2eb9881a91/bin/MadGraph5_aMCatNLO/runcmsgrid_LO.sh#L203

gitytakahas:
Hi Qiang, 

Thanks for the check. I produced another LHE file with following change in runcmsgrid_LO.sh

> NNPDF31_lo_as_0130.LHgrid 1 ->  NNPDF31_lo_as_0130.LHgrid

/afs/cern.ch/user/y/ytakahas/public/forQiang/cmsgrid_final_v2.lhe

Could you please check, again, if this is OK ? 

I have no idea why it picks 4FS. Is this related to the fact that I had 

> NNPDF31_lo_as_0130.LHgrid 1 

before? 

Thanks,
Yuta

qliphy:
@gitytakahas     Hi Yuta, it still seems 4FS is picked.  Are you using the most updated genproductions?

  <weightgroup name="NNPDF31_nnlo_hessian_pdfas" combine="hessian">...
  <weightgroup name="CT10nlo_nf4" combine="hessian">...
  <weightgroup name="CT14nnlo_NF4" combine="hessian">...
   ....

I just make a gridpack using your card, and it picks the right 5FS PDF
 /eos/cms/store/user/qili/NonRes_M500_slc6_amd64_gcc481_CMSSW_7_1_30_tarball.tar.xz

  <weightgroup name="NNPDF31_nnlo_hessian_pdfas" combine="hessian">...
  <weightgroup name="NNPDF31_nnlo_as_0108" combine="hessian">...
  <weightgroup name="NNPDF31_nnlo_as_0110" combine="hessian">...
  <weightgroup name="NNPDF31_nnlo_as_0112" combine="hessian">...

gitytakahas:
Hi Qiang, 

I made sure to use the latest version. Could you please check again?

/afs/cern.ch/user/y/ytakahas/public/forQiang/cmsgrid_final_v3.lhe

Thanks again,
Yuta

gitytakahas:
it seems to me that, it still picks 4 FS ...

qliphy:
Yes, still 4FS PDFs.

Anyway, you can check with the gridpack produced by me:
 /eos/cms/store/user/qili/NonRes_M500_slc6_amd64_gcc481_CMSSW_7_1_30_tarball.tar.xz

I just do as following: 
git clone git@github.com:cms-sw/genproductions.git genproductions
and then use your cards
https://github.com/gitytakahas/genproductions/tree/6d67c800e73c98a572a7dea1e50e532e0a5853c7/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ScalarLQ/NonRes_M500

gitytakahas:
hmmm, I did the same.
Only difference would the different production. Can you please try with Pair_500 cards?

https://github.com/gitytakahas/genproductions/tree/6d67c800e73c98a572a7dea1e50e532e0a5853c7/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ScalarLQ/Pair_M500

qliphy:
With Pair_500 cards, you can find my results here:
/tmp/genproductions/bin/MadGraph5_aMCatNLO @ lxplus062

Everything looks normal, and 5FS is picked.

gitytakahas:
Thanks - to produce gridpacks, did you use girdpack_submission.sh or something else? 

Best,
Yuta

qliphy:
Just simply gridpack_generation.sh

qliphy:
You don't need to copy runcmsgrid_LO.sh when generating lhe.

There is a runcmsgrid.sh file in the gridpack, where the is5flavourscheme is set automatically and consistent with your process card. You can run runcmsgrid.sh to produce LHE.

gitytakahas:
Yes, ok, we solved the problem then. 
However, to get the PDF weights, I think we need to set 

NNPDF31_lo_as_0130.LHgrid 1 -> NNPDF31_lo_as_0130.LHgrid

in runcmsgrid_LO.sh, right? Should I make PR for that, because when gridpack is made, runcmsgrid.sh is copied from runcmsgrid_LO.sh. 

Please let me know,
Yuta

qliphy:
@gitytakahas   We discussed this before  among GEN group and decided not to include NNPDF31lo weights as default.  You can find a recent thread here:
https://github.com/cms-sw/genproductions/pull/1874

gitytakahas:
ok I see. then, I can just modify my local code (runcmsgrid_LO.sh) and create gridpacks. I then hand this to MC convener for the official production. Is this OK ? 

Thanks,
Yuta

qliphy:
That sounds good to me.

Best, 
QIang

PR #1964, opened 2018-11-15
switch to LO PDF set

karavdin:


qliphy:
Please further read the [GEN recommendations](https://twiki.cern.ch/twiki/bin/view/CMS/PhysicsComparisonAndGeneratorTunes#Recommendations%20on%20the%20usage%20of):  (1) you may use pythia8 CP2 Tune instead of CP5; (2)  And please also add NNPDF3.1LO weights if needed

PR #1968, opened 2018-11-19
add crads for VBS ZZjj

menglu21:
cards for three process, EWK+QCD, EWK and QCD


qliphy:
Can you make sure you don't want to include tau for ZZJJ processes?
              define l+ = e+ mu+
              define l- = e- mu-

menglu21:
> Can you make sure you don't want to include tau for ZZJJ processes?
> define l+ = e+ mu+
> define l- = e- mu-

i followed the cards for 2016 gridpack production, and also some elaboration on AN-17-002, 'Leptonic decays of the Z bosons are considered, ZZ → 4l, where l = e, μ.

PR #1970, opened 2018-11-20
monoHiggs CP2 setting included

fmanteca:


efeyazgan:
Hi, an alternative may be to make new fragments with CP2 and keep CUEP8M1 ones in case someone wants to use them in the future.

fmanteca:
Hi, I kept the CUEP8M1 files and added the new _CP2 ones

PR #1975, opened 2018-11-20
add cards for LNuJJ_EWK_SM_5f_LO

menglu21:
add cards for LNuJJ_EWK_SM_5f_LO


qliphy:
Please follow https://twiki.cern.ch/twiki/bin/view/CMS/Moriond18MC#Checks%20to%20the%20datacards%20BEFORE%20T
for example,  remove the banner in the proc_card 

For WGG_5f_NLO, it seems to be a FxFx request according to the proc card.  If so, please update the card name from NLO to FxFx, and change ickkw to 3 in run card.

And do you also want to do similar for ZGG?  i.e.  to make it FxFx but not NLO...

menglu21:
> Please follow https://twiki.cern.ch/twiki/bin/view/CMS/Moriond18MC#Checks%20to%20the%20datacards%20BEFORE%20T
> for example, remove the banner in the proc_card
> 
> For WGG_5f_NLO, it seems to be a FxFx request according to the proc card. If so, please update the card name from NLO to FxFx, and change ickkw to 3 in run card.
> 
> And do you also want to do similar for ZGG? i.e. to make it FxFx but not NLO...

ok, i have modified these cards.

PR #1980, opened 2018-11-21
Add cards for 2017 charged higgs to cb decay sample. 2HDM.

diracyoon:
Add cards for 2017 charged higgs to cb decay sample. 2HDM.


diracyoon:
Thank you very letting me know that. Actually, model has been changed slightly.

qliphy:
But still you can make a script like this one: 
https://github.com/slehti/genproductions/blob/6f2694044d190131a122a56f7c92a293b32c2ac7/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ChargedHiggs_taunu/ChargedHiggs_taunu_HeavyMass_NLO_cards.sh 

Then you just need to upload cards for one parameter point

diracyoon:
I understood. I've added sh file to generate cards for various mass points. I've removed unneccessary directories.

PR #1984, opened 2018-11-22
update HH bbVV 2l2nu high mass fragment

covarell:
@clacaputo @acarvalh
Following our discussion with Ilya Kravchenko on 2019 MC, I update the fragment to be with 2l2nu final states only.


covarell:
Please hold on because there seems to be a problem with this request

agrohsje:
Hi @covarell , should I close this PR or was it ok? I try to clean up a little bit.

covarell:
This sample is no longer needed, honestly I do not remember the problem rafter 2 years :-)

agrohsje:
Thanks @covarell. Let me close and clean up a little bit.

PR #1988, opened 2018-11-24
Generic hadronizers for CP2 and CP3

DryRun:
Similarly to https://github.com/cms-sw/genproductions/pull/1901, adds copies of the standard MG LO hadronizer (with PS weights) for CP2 and CP3.


DryRun:
Sorry, I just noticed that the files with and without "PSWeights" in the name are identical. I'll close this PR and figure out what we want to do about that.

PR #1992, opened 2018-11-27
cards for TT FCNC top-photon

zenaiev:


qliphy:
Please notice that we just changed master branch of genproductions from mg242 to mg26X

So please close this one and make a PR to the mg242legacy branch, if it is mg242 you want to use.

PR #1996, opened 2018-11-28
cards for TT FCNC top-photon

zenaiev:


zenaiev:
Hi Efe, this PR is into cms-sw:mg242legacy.

PR #1997, opened 2018-11-28
fix reweighting in NLO gridpack run script

AndrewLevin:
Currently, the reweighting is done twice when the NLO gridpack script is run. The line `echo "reweight=OFF" >> runscript.dat` has no effect because it is after the first `done`. When `reweight=OFF` is applied correctly, `reweight_card.dat` is moved to `.reweight_card.dat`, so the if statement `if [ -f ./Cards/reweight_card.dat ] ;then` needed to be changed.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/runcmsgrid_NLO.sh
+++ b/bin/MadGraph5_aMCatNLO/runcmsgrid_NLO.sh
@@ -77,6 +77,11 @@ if [ -f ./Cards/madspin_card.dat ] ;then
   domadspin=1
 fi
 
+doreweighting=0
+if [ -f ./Cards/reweight_card.dat ]; then
+    doreweighting=1
+fi
+
 runname=cmsgrid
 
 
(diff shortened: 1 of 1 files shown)


qliphy:
Thanks @AndrewLevin 
@soarnsoar @kdlong @agrohsje  Can you have a look?

agrohsje:
Dear @qliphy , @AndrewLevin,
I can confirm this with a quick standalone test. However, I would prefer testing at the beginning if there is a reweighting_card.dat and  if so, set an new variable (doreweighting which is initially false) to true. Everyhting following is then only based on this variable, including copying .reweighting_card.dat to reweighting_card.dat. The idea behind my suggesting is to be independent from the logic the MG5 authors are using. 
Cheers, Alexander.

AndrewLevin:
@agrohsje, I added a doreweighting flag.

agrohsje:
Thanks @AndrewLevin !

PR #1999, opened 2018-11-28
cards for TT FCNC top-photon pre2017

zenaiev:


qliphy:
You are using MadGraph embedded old PDF, please update to newer one with LHAPDF:
 
https://github.com/cms-sw/genproductions/blob/24d33436b4014a2d78eb3d62b6cd7a3e9caeb450/bin/MadGraph5_aMCatNLO/cards/production/pre2017/13TeV/TT_FCNC/TT_FCNC-T2AJ_aTleptonic_kappa_act_LO/TT_FCNC-T2AJ_aTleptonic_kappa_act_LO_run_card.dat#L51-L52

qliphy:
Also there are too many cards.  Many of them are similar, for example proc_card, can you try to reduce the file numbers?

zenaiev:
Hi Qiang, the old PDF is used on purpose, since we are producing MC samples for 2016 and need different gridpacks. The corresponding cards for 2017 and 2018 were just committed here https://github.com/cms-sw/genproductions/tree/mg242legacy/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TT_FCNC

zenaiev:
We wanted to use the same settings as for other 2016 FCNC requests, e.g. this one [1] done using these cards [2], with NNPDF30 NLO PDF and  CUET8PM1.
As for the cards, I could create a diff file comparing 2017/2018 and 2016 cards and commit it here with a short readme. Would it make sense?
[1] https://cms-pdmv.cern.ch/mcm/requests?prepid=TOP-RunIISummer15wmLHEGS-00154
[2] https://github.com/cms-sw/genproductions/tree/pre2017/bin/MadGraph5_aMCatNLO/cards/production/13TeV/TT_FCNC/madgraph_card

qliphy:
But then you are need to change "nn23lo1 = pdlabel" to "lhapdf = pdlabel ! PDF set", otherwise nn23lo1 will be used.

>As for the cards, I could create a diff file comparing 2017/2018 and 2016 cards and commit it here >with a short readme. Would it make sense?

Yes, this sounds good.

zenaiev:
Hi Qiang, thanks for the follow up. Indeed the reference 2016 run cards have two different entries for the PDFs: "nn23lo1 = pdlabel" and "230000 = lhaid", and the former is the relevant one, while I was confused by the two PDF entries and thought that the relevant one is "230000 = lhaid", i.e. NNPDF30 NLO. To be honest I do not know why the 2016 FCNC gridpacks were produced with nn23lo1, but I thought it is a good idea just to keep all TT FCNC 2016 samples consistent. Also I suppose these analyses (1) are not really sensitive to the PDFs, (2) there are more PDF weights such that the analysers can switch to another more modern PDF sets.
I am adding Kate @ckmackay and Emyr @EmyrClement to this discussion, maybe they want to comment.
Once we will settle this issue, I will update the cards by committing only the diff file w.r.t 2017/18 ones.

zenaiev:
Dear All, can we revive this pull request, please?

qliphy:
This is LO request, so PDF and Pythia8 Tune should not matter quite much.   However, I leave this for Top or TOPMODGEN to decide. 

Adding also @efeyazgan

ckmackay:
Dear All, perhaps we should use the correct PDF, even if this is different from the other FCNC requests in 2016, unless there was a specific reason to use this. I will quickly check with Reza to see if he remembers any reason for this.

efeyazgan:
Thanks for checking. I agree .

gvonsem:
Hi all, I made a new PR #2051 based on Sasha's cards but with the PDF changed from NNPDF2.0 to NNPDF3.0. Feel free to comment there, and/or let me know if something is missing.

PR #2004, opened 2018-11-29
2016 cards for C-Enriched background samples

panwarlsweet:
-Adding cards for C-enriched background samples for VHtocc analysis

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/pre2017/13TeV/DYCJetsToLL_M-50_Zpt-100to200_5f_LO_MLM/DYCJetsToLL_M-50_Zpt-100to200_5f_LO_MLM_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/pre2017/13TeV/DYCJetsToLL_M-50_Zpt-100to200_5f_LO_MLM/DYCJetsToLL_M-50_Zpt-100to200_5f_LO_MLM_proc_card.dat
@@ -1,0 +1,45 @@
+#************************************************************
+#*                     MadGraph5_aMC@NLO                    *
+#*                                                          *
+#*                *                       *                 *
+#*                  *        * *        *                   *
+#*                    * * * * 5 * * * *                     *
+#*                  *        * *        *                   *
+#*                *                       *                 *
+#*                                                          *
+#*                                                          *
+#*         VERSION 2.3.3                 2015-10-25         *
+#*                                                          *
+#*    The MadGraph5_aMC@NLO Development Team - Find us at   *
+#*    https://server06.fynu.ucl.ac.be/projects/madgraph     *
+#*                                                          *
+#************************************************************
+#*                                                          *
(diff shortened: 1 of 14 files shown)


qliphy:
Please check more https://twiki.cern.ch/twiki/bin/view/CMS/Moriond18MC#Checks_to_the_datacards_BEFORE_T

For example:  remove the banner in the proc_card 


For  w+ production,  please change e.g.  "generate p p > w+ cquark, w+ > ell+ vl / h @0"  to "generate p p > ell+ vl cquark  / h @0 ".
and set "bwcutoff" to smaller value like 15


And:    these lines should be removed: 

+# Parameter of the systematics study

panwarlsweet:
I have edited other things but for this
"For w+ production, please change e.g. "generate p p > w+ cquark, w+ > ell+ vl / h @0" to "generate p p > ell+ vl cquark / h @0 ". "

I see that for B enriched cards also it is defined like this "generate p p > w+ bquark, w+ > ell+ vl / h @0"

and in this card also they decay w later "https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/WJets_HT_LO_MLM/WJetsToLNu_HT-100to200/WJetsToLNu_HT-100to200_proc_card.dat"

Does it not matter it I don't write w is generate process?

qliphy:
There have been some similar discussions previously  in GEN group, and the recommended way is to have the so-called direct production, i.e.   "generate p p > ell+ vl cquark / h @0 ".

panwarlsweet:
Thanks for the comments. Then we will modify it accordingly and make the PR again. For now, I am closing the PR.

mastrolorenzo:
@qliphy just to my knowledge: I am assumiong that is there the exact same thing writing "generate p p > ell+ vl cquark / h @0" instead of the process: "generate p p > w+ cquark, w+ > ell+ vl / h @0", can you confirm this?

qliphy:
Some discussion can be found here https://www.dropbox.com/s/nr8td3k5x3k8h3h/genval_CompareVersion_sept.pdf?dl=0

You can also have a check with WJets gridpack setting

mastrolorenzo:
@qliphy many thanks for the link to the slides. Indeed, I see (slide 8) that there is a quite substantial changes in the spectrum of W-pt for momenta < 60 GeV. Since you said that this is the new recommendation I am assuming that this difference is understood. If this is the case, we will proceed in updating the cards as you suggested.

qliphy:
A more updated report is here https://indico.cern.ch/event/571835/contributions/2313850/attachments/1343093/2023590/genval_CompareVersion_sept.pdf#search=Lorenzo Russo

Anyway, the current gridpacks for WJets etc are made with the so called direct production method.

panwarlsweet:
ok, so it should be changed only for WJets samples, not ZJets?

qliphy:
Yes, for Z should also be similar.

PR #2015, opened 2018-12-07
Example Config for MG5 MLM with Herwig7

Andrej-CMS:
Dear all,
I added an example config on how to do MLM with MG5 and Herwig (requested by Jet-MET). This config is for DY+4jets.
I left the comments from the authors and added some myself, so the contacts know what is the meaning of the setting.  I hope this helps to widen the use of H7.
Kind regards,
Andrej


Andrej-CMS:
Hi @qliphy ,
Thank you very much! I answered to the HN thread

efeyazgan:
@Andrej-CMS thanks a lot!

PR #2017, opened 2018-12-09
Disappearing tracks cascade

aehart:
These are the files for the disappearing tracks cascade request in EXO, where long-lived charginos are strongly produced via an intermediate gluino.


qliphy:
Note you may also want to make changes to mg242legacy branch

aehart:
Admittedly, we kind of forgot about this PR, but we now have somebody actively working on our MC request, and I think it is ready to be merged.

And I don't think the mg242legacy branch is relevant for us, since we use Pythia8. Unless I misunderstand something, I think we just want these files on the master branch.

PR #2021, opened 2018-12-11
adding cards for ZZTo4B

patrickbryant:
Successfully generated a gridpack from these cards and the final state branching ratio was as expected. Based on the ZZTo2L2Q cards used for the VH analysis.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ZZTo4B01j_5f_NLO_FXFX/ZZTo4B01j_5f_NLO_FXFX_madspin_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ZZTo4B01j_5f_NLO_FXFX/ZZTo4B01j_5f_NLO_FXFX_madspin_card.dat
@@ -1,0 +1,22 @@
+#************************************************************
+#*                        MadSpin                           *
+#*                                                          *
+#*    P. Artoisenet, R. Frederix, R. Rietkerk, O. Mattelaer * 
+#*                                                          *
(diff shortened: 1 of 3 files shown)


patrickbryant:
The instructions on this twiki are out of date and not functional: https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideSubgroupMC#1_1_Using_wmLHE_campaigns

Until someone tells me to do something else, I will assume it is ok to keep the gridpack on my afs public

qliphy:
Usually there is no need to upload hadronizer python file ZZTo4B01j_5f_NLO_FXFX.py.  Instead one can add by hand in the McM request fragment, and change the gridpack path to the cvmfs one.

Please also noted the master branch has recently been changed to mg26x
https://twiki.cern.ch/twiki/bin/view/CMS/MG26x
You can still use the previous mg242 version with the mg242legacy branch.

But as you are now committing to the master branch with mg26x, please notice the following doesn't work now and you should set explicitly PDF for which you can refer to above twiki.

 $DEFAULT_PDF_SETS    = lhaid     ! if pdlabel=lhapdf, this is the lhapdf number
 $DEFAULT_PDF_MEMBERS  = reweight_PDF     ! reweight to get PDF uncertainty

patrickbryant:
So I should make a merge request with the mg242legacy branch and not include the fragment file?

patrickbryant:
I gather from your comment that the other option is to modify the run_card to match the pdf specifications of https://twiki.cern.ch/twiki/pub/CMS/MG26x/dyellell012j_5f_NLO_FXFX_run_card.dat

I have done this now and am trying to generate the gridpack again.

PR #2022, opened 2018-12-11
WV cards

sregnard:
Cards for 2017 and 2018 production of high-mass diboson resonances (GBulk->WlepWhad, W'->WlepZhad, and W'->Wlephinc), with LO NNPDF.


qliphy:
Please don't upload all the 180 cards,  instead you can upload example cards together with a script to change to other model parameters. You can easily find examples from genproductions area.

For LO process, you can remove this line
https://github.com/cms-sw/genproductions/blob/d00b2fc168cf11b0213a5750f79b65132dcd0eb7/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/exo_diboson/Spin-1/Wprime_WZ_WlepZhad/Wprime_WZ_WlepZhad_narrow_M1000/Wprime_WZ_WlepZhad_narrow_M1000_run_card.dat#L53

PR #2029, opened 2018-12-13
Update cards for 2017 (and 2018) production; uses LHAPDF ID 306000

jheikkil:


qliphy:
Please remove the banner in the proc card,  set bwcutoff to smaller value as 15 in run_card.dat.  

Moreover,  master branch is now with mg260.  If you want to use mg242, you can commit to the mg242legacy branch.

qliphy:
As for the version of MG,  you can check 

By default,  genproductions is using madgraph version 260, and if you want to use the old version of 242, you should check out with
git clone git@github.com:cms-sw/genproductions.git genproductions -b mg242legacy
There should not be much difference in Physics for your case.

As for bwcutoff,  you can refer to 
https://hypernews.cern.ch/HyperNews/CMS/get/prep-ops/5701.html
and
https://indico.cern.ch/event/560833/contributions/2268189/attachments/1321495/1982803/genval_result_ago.pdf#search=bwcutoff%20lorenzo

This was discussed and agreed in CMS GEN group.  But still some old gridpacks were produced with large bwcutoff.  The effect in your case may be small, but it is better to switch back to small value.

jheikkil:
How do I know which version of MG I should use?
Would it possible for you to update the twiki with clear instructions what needs to be changed in the cards? Currently it is rather challenging to follow.

qliphy:
Both are indeed ok.  But if you fell more comfortable, you can take the mg242legacy branch as we already mentioned in hn:
"If you still want to use v242, then you will need to make your pull requests to the mg242legacy branch(
git clone git@github.com:cms-sw/genproductions.git genproductions -b mg242legacy)"

jheikkil:
Thank you!

Recently we had to redo gridpacks for few samples in 2016 production. Now I would like to confirm that I used 242 correctly for those too. Can I see it somehow from the gridpacks?

qliphy:
Yes, we are trying to improve.  We have a preliminary twiki for 26X
https://twiki.cern.ch/twiki/bin/view/CMS/Mg26x

jheikkil:
Thank you! I still couldn't understand how to check if the weights were correctly made in the 2016 gridpacks. Could help me to understand if some line in this log file reveals if it was correctly done? 
[gridpack_generation_220_2016.log](https://github.com/cms-sw/genproductions/files/2676726/gridpack_generation_220_2016.log)

jheikkil:
After you have taken a look at the log file I sent above, could you also confirm that this is correct now: https://github.com/cms-sw/genproductions/pull/2030 ?
Thanks!

qliphy:
From the log, you are using   MG  VERSION 2.4.2.

Maybe you can close this now, switch to #2030

PR #2030, opened 2018-12-13
Update cards, 2017 setup, version 242

jheikkil:


qliphy:
These cards look good now, except one point:  for PDF setting, please follow
https://twiki.cern.ch/twiki/bin/view/CMS/Moriond18MC#PDF_Choice_for_2017_production_f

the logic can be read here
https://github.com/cms-sw/genproductions/blob/mg242legacy/bin/MadGraph5_aMCatNLO/gridpack_generation.sh#L492

jheikkil:
Previously, I tried to use the three lines 
  'lhapdf' = pdlabel ! PDF set
 $DEFAULT_PDF_SETS = lhaid

these never worked for me. My MC contact Yiwen Wen told me I could instead use the lines I had for the PDF setting. Do I really need to use these instead?

qliphy:
It should work, if you take the most recent genproductions mg242legacy branch.

But well, your process is LO, so indeed it doesn't matter as you can see from here:
https://github.com/cms-sw/genproductions/blob/mg242legacy/bin/MadGraph5_aMCatNLO/gridpack_generation.sh#L504
and PDF weights are added anyway in
https://github.com/cms-sw/genproductions/blob/mg242legacy/bin/MadGraph5_aMCatNLO/runcmsgrid_LO.sh#L159

For NLO, it will be different.

jheikkil:
Hi again, I have tried one more time, and for some weird reason the automatic PDF setting worked. 
I have created a new PR for cards: https://github.com/cms-sw/genproductions/pull/2031

PR #2031, opened 2018-12-14
Use the automatic PDF setting

jheikkil:


qliphy:
Looks good to me.  But please set bwcutoff to 15.

jheikkil:
Thanks for pointing that out to me! I guess AtoZh_M340 was the only one left with 15000? Could you confirm, please?

qliphy:
Looks good now. Thanks!

jheikkil:
Thanks for all the help! For future reference, could the twiki page be clarified with all this information?

qliphy:
Yes, of course.  The basic message is already in the hn announcement

We will make the twiki more clear.

PR #2033, opened 2018-12-15
2016 an 2017 ggH MINLO NNLOPS cards with hadronizer file

panwarlsweet:
- Adding 2016 and 2017 ggH_MINLO_cards
-Adding 2017 fragment file for ggH_MINLO


qliphy:
adding POWHEG experts @covarell @yuanchao

covarell:
1) POWHEG cards look ok. However, the MINLO card for 2017 is already present: https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/production/2017/13TeV/Higgs/HJ_MiNLO_NNLOPS_NNPDF31_13TeV/HJ_MiNLO_NNLOPS_NNPDF31_13TeV.input
So you just need to add the "pre2017" one (maybe changing the name as above?)

2) In the hadronizer, why is the emission veto deactivated? It should be active with nFinal = 2.

panwarlsweet:
Could you please let me know if everything  is fine with Fragment file so that it could be merged since it is required to submit the Moriond19 samples.

covarell:
They are fine. In the POWHEG card there are instrcutions that will be just ignored in HJ (like "fakevirt 1") but this will not change the result.

panwarlsweet:
Thanks for the confirmation. I would like to request to merge this PR.

PR #2037, opened 2018-12-17
tW mass variations

ckmackay:


qliphy:
adding @covarell @yuanchao

ckmackay:
Only change from Fall17 is hdamp values have been updated to 1.379*mtop in line with ttbar.

PR #2039, opened 2018-12-19
added ttbarDM gridpack submission script, two sets of templates for (…

agrohsje:
…pseudo)scalar mediators, as well as one specific example


qliphy:
Please remove "store_rwgt_info" which is not needed in mg242

agrohsje:
Hi Qiang,
I removed both blocks. Btw, should I add an example shower fragment? Who will handle replacing the gridpack path with the proper location in the fragment? The EXO MC contacts?  
Cheers, Alexander.

PR #2043, opened 2019-01-03
pre 2017 simulation

ksiehl:
simulation for cp odd couplings 2016 data


qliphy:
Please set store_rwgt_info to True

In proc card, you set 4 flavor:
however, in run_card, you take 5flavor PDF as 292200

Moreover, please just upload one set of cards for an example case

ksiehl:
Okay.  I'll just close the pull request and start over.

ksiehl:
Actually, I checked, and the 5flavor pdf was in the wz section, define j = j b b~ and define p = p b b~
The 4flavor ww section has PDF as 292000
Should I just open a new pull request with one from each, ww and wz, or do a separate pull request for each?

qliphy:
You can make a new PR with one from each.

PR #2045, opened 2019-01-06
low mass end ww

ksiehl:
Semileptonic channel, ww decays.


qliphy:
Do you really need this ptl cut?  Have you checked it will not bring any bias?
             48.0  = ptl     ! Min lepton transverse momentum

qliphy:
I don't see any update.  Or do you mean "48.0 = ptl" is fine for you?

ksiehl:
You should see it now. I changed 48 to 20.

qliphy:
Thanks for the information.  It is better to update the proc card to keep consistency.

PR #2051, opened 2019-01-09
cards for TT FCNC top-photon pre2017, with NNPDF3.0 PDF

gvonsem:
The top-photon FCNC cards in this PR are the same as in PR #1999, but with the PDF in the run cards changed from NNPDF2.0 to NNPDF3.0.


qliphy:
Please delete following lines in run_card files:
https://github.com/cms-sw/genproductions/blob/a8f2e3f1058a73b02bb70462ad921e5c2fc059c1/bin/MadGraph5_aMCatNLO/cards/production/pre2017/13TeV/TT_FCNC/TT_FCNC-aT2AJ_Tleptonic_kappa_act_LO/TT_FCNC-aT2AJ_Tleptonic_kappa_act_LO_run_card.dat#L272-L283

BTW:   It seems to me that [2017 cards ](https://github.com/cms-sw/genproductions/tree/mg242legacy/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TT_FCNC)  miss this line  "True	=  use_syst"

gvonsem:
The run cards are updated with the suggested blocks removed. 

Regarding the comment on the 2017 cards: these cards are from an earlier merged PR with corresponding samples already submitted. To be sure, let me add TOPMODGEN experts (@ckmackay and @EmyrClement) to this discussion to see if they have any opinion on that missing line in the [2017 cards](https://github.com/cms-sw/genproductions/tree/mg242legacy/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TT_FCNC).

ckmackay:
Yes the line is missing in 2017 cards, this will be added and the gridpacks remade. Apologies it's taken so long to respond to this! It was only looking into a query that we found this again.

Is there a reason why the lhe header info does not give the weight IDs with these cards? Are we missing something to ensure this information is included? Are the weights always in a standard order? If so what is that, as the requester needs the PDF IDS and can only see that there are 1080 weights, but not what they are for. Thanks for any help with this

qliphy:
@ckmackay   Thanks for pointing out this. It seems to be an issue with madspin which remove the lhe header.  I have checked other relevant gridpack [1] with madspin and found similar behaviour.  

One can also easily confirm this by commenting out the madspin step in runcmsgrid.sh, after which the lhe header appears.  In this way, you can get the PDF IDs.

But of course GEN should try to fix this and check whether mg26X has the same issue.   


[1] /cvmfs/cms.cern.ch/phys_generator/gridpacks/2017/13TeV/madgraph/V5_2.4.2/tW_tW_x53x53_madspin_narrow/tW_tW_x53x53_madspin_narrow_RH_M900_slc6_amd64_gcc481_CMSSW_7_1_30_tarball_runmode0_TEST.tar.xz

qliphy:
FYI: Alexander made a PR  to add wgt header back removed by madspin:  https://github.com/cms-sw/genproductions/pull/2370

PR #2052, opened 2019-01-09
2017 cards for VHbb samples

tiwariPC:
Dear all,
Please find the cards for 2017 cards for VHbb samples


qliphy:
For WBJets, you may also want to exclude top quark in the process definition.

Moreover,  "T  = cut_decays" will cut on W decay products, so depending on what you want, you may set " F  = cut_decays "  or write process as "generate p p > ell+ vl  bquark / t h"

qliphy:
To exclude top quark, you can do as following:

generate p p > ell+ vl bquark / h @0
->
generate p p > ell+ vl bquark / t  h @0

tiwariPC:
Does this have to be done for DY and Zbjets as well?

qliphy:
Only for WBJets, as  W+b can come from top

PR #2060, opened 2019-01-21
Adding heavyNeutrino cards

UAEDF-tomc:
All cards for the heavyNeutrino requests (trilepton) + including the generation script to produce them using the template cards


UAEDF-tomc:
@qliphy done

PR #2069, opened 2019-01-31
DYNNLOPS

intrepid42:
Enhance Wj and Zj with NNLO/MINLO weights, https://arxiv.org/abs/1407.2940


efeyazgan:
@covarell @yuanchao could you please take a look at this PR? thanks.

covarell:
Hi Efe, @intrepid42 and I already discussed general improvements to the new HTCondor workflow coming with these PR and they are fine to me. Indeed another pair of eyes looking at them (@yuanchao ) would help. 

My only comment is that probably new python routines that are only used for DYNNLO should be clearly marked as such (for example run_minlo appears like something that must be used for every MINLO process, while it is not).

yuanchao:
I am currently out of town this week. Will do during the coming weekends.

kdlong:
@covarell what are the improvements to HTCondor coming in with this PR? The main changes I see seem specific to this process. For example, what is the idea behind the "runInBatchDir" feature? Just that the script should be executed from a different initial directory? It still relies on a shared file system, right? What's the reason for not transferring output files, because it's assumed that they are produced on the sharedfs? 

On this note, the condor support is currently relies heavily on the shared file system and is also pretty CERN specific. I have been working on running at Wisconsin without relying on afs, but it will take a decent bit of reorganizing the code to integrate this in a straightforward way.

yuanchao:
@intrepid42  I quickly went through the PR and have some general comments:
* as a new stage id is added, please prepare the corresponding info in twiki and program "help" message
* saw some "fix me" comments and re-written in later commits, please make sure those commented lines and "fix me" are not in the final commit.
* Looks to me that those condor related patches are only specific for this process which is fine with me. Just would be better that they can be run out side of lxplus. (so that I can test locally)
* Please also commit the patch file to central or ask us to copy them to generator group web page. To me, it's not a good idea to point to some personal directory.

intrepid42:
Hi, for Condor, I made those fixes to prevent files from different processes overwriting each other. The transfer of the output files should be handled by the .sh script, otherwise they will all end up in whatever `initialdir`, instead of being sorted into dirs per jobs, or per parameter (scale) point.

I will still make all the "cleanup" changes you suggested.
The nnlo patches had to change a few times because that code had a few bugs, I will send them to you to include in the central powheg tarball.

covarell:
This now looks good to me @yuanchao ?

yuanchao:
> This now looks good to me @yuanchao ?

Sorry for the late reply. Yes, just that could the patch file be d/l from generator website?
(previously I've helped copied the patch files)

covarell:
ping

covarell:
The patch files are present in the generator website, can we proceed?

intrepid42:
Hi Roberto, I changed that already: https://github.com/cms-sw/genproductions/pull/2069/commits/389c04a8389b5355057bfe3787ee9cc2eef81571

But we will also need to make additional changes to the make_rwl.py (to reduce number of weights for the process), so we can wait for that before merging.

covarell:
Ok thanks.

I found a few other processes where running stage 2 and/or 3 in the initialdir in fact screws up the data files, so I would like to make runInBatchDir the default.

intrepid42:
Ok, feel free to merge this as soon as you like, I can just create a new pull request for smaller DYNNLOPS-specific additions

covarell:
Ok! @efeyazgan @qliphy

qliphy:
Thanks! Merged now.

PR #2074, opened 2019-02-04
DisplacedSUSYStopToLD

wadud001:
I'm a new EXO Gen/MC contact. This is work on a request.


efeyazgan:
@wadud001 why do you need to commit an empty txt file?

wadud001:
This was a test. Will submit another request once done.

Thanks.

PR #2077, opened 2019-02-05
Adding AJJ_EWK_SM_5f_NLO cards

UAEDF-tomc:
Adding the cards for electroweak (VBF) photon+2j production at NLO

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/AJJ_EWK_SM_5f_NLO/AJJ_EWK_SM_5f_NLO_FKS_params.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/AJJ_EWK_SM_5f_NLO/AJJ_EWK_SM_5f_NLO_FKS_params.dat
@@ -1,0 +1,75 @@
+! ==========================================================================
+! This file sets the different technical parameters intrinsic to the
+! FKS program and which controls the behaviour of the code at run
+! time.  The common user should not edit this file and only experts
+! should venture editing these parameters.
+! ==========================================================================
+!
+! ==========================================================================
+! Arbitrary numerical parameters used in the FKS formalism
+! ==========================================================================
+!
+! To be implemented by the FKS authors
+!
+! ==========================================================================
+! Parameters controlling the tests based on the IR poles comparison
+! ==========================================================================
+!
+! This threshold sets the limiting value for the comparison of the
+! relative difference of the IR pole from the OLP and the one computed
+! internaly by MadFKS.  The value below is used for the first PS
+! points to assess the sanity of the computation. A value ten times
+! smaller will be used for the systematic check of IR poles at
+! runtime.
+! Notice that the systematic comparison of IR poles at run time is
+! only performed when the Monte-Carlo over helicity configurations
+! method is not used.  Set this value to '-1.0d0' if you want the
+! check to always pass.
+#IRPoleCheckThreshold
+-1.0d0
+! Default :: 1.0d-5
+!
+! ==========================================================================
+! OLP (virtuals) behavior at run time
+! ==========================================================================
+!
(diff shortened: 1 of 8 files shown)


qliphy:
Please change to 
               True	= store_rwgt_info 
               PYTHIA8	= parton_shower 

And  are you sure you want a large Radius cut?
                0.4	= r0gamma ! Radius of isolation code

Also note if it is for 2017/2018, the default PDF should be NNLO (303600), to match with the Pythia8 Tune CP5:
                292000 = lhaid

qliphy:
And why do you need a  FKS_params.dat?

UAEDF-tomc:
> And why do you need a FKS_params.dat?
The FKS_params.dat is modified to turn off the #IRPoleCheckThreshold (set to -1)

UAEDF-tomc:
> Also note if it is for 2017/2018, the default PDF should be NNLO (303600), to match with the Pythia8 Tune CP5:
> 292000 = lhaid

I assumed that we could still use
 _$DEFAULT_PDF_SETS = lhaid
 $DEFAULT_PDF_MEMBERS  = reweight_PDF_
as we did before. Since when did this recommendation change? Then we should update the twiki: https://twiki.cern.ch/twiki/bin/viewauth/CMS/QuickGuideMadGraph5aMCatNLO#PDF_Choice_for_2017_production

qliphy:
Yes, you can still use for 2017/2018 requests
    $DEFAULT_PDF_SETS = lhaid 
This will set NNLO PDF for you.


>The FKS_params.dat is modified to turn off the #IRPoleCheckThreshold (set to -1)

Can you give some more explanation or ref link?  And have you got green light from the authors?

UAEDF-tomc:
> > The FKS_params.dat is modified to turn off the #IRPoleCheckThreshold (set to -1)
> 
> Can you give some more explanation or ref link? And have you got green light from the authors?
I don't have the explanation myself, though it should have been discussed privately between @azzurrip, @pfs and aMC@NLO authors (Marco Zaro)

UAEDF-tomc:
> And are you sure you want a large Radius cut?
> 0.4 = r0gamma ! Radius of isolation code

Yes, there was also a draj=0.4 cut at LO, so we want to stay with this for NLO too

qliphy:
Thanks for the info.   I have one last question on  0.4= r0gamma

This cut affect the additional jet emission in NLO.   For jet close to photon within the distance of r0gamma,  the smooth isolation will be applied.  Probably, a smaller value of r0gamma may be better, as the additional jet will be more included inclusively.  

In other similar process, 0.05 was used, e.g.:  
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/pre2017/13TeV/ZATo2LA01j_5f_NLO_FXFX/ZATo2LA01j_5f_NLO_FXFX_run_card.dat

UAEDF-tomc:
@azzurrip Could you give your opinion on this r0gamma cut?

> Thanks for the info. I have one last question on 0.4= r0gamma
> 
> This cut affect the additional jet emission in NLO. For jet close to photon within the distance of r0gamma, the smooth isolation will be applied. Probably, a smaller value of r0gamma may be better, as the additional jet will be more included inclusively.
> 
> In other similar process, 0.05 was used, e.g.:
> https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/pre2017/13TeV/ZATo2LA01j_5f_NLO_FXFX/ZATo2LA01j_5f_NLO_FXFX_run_card.dat

azzurrip:
I see what you mean.  
Ok yes let's have a 0.05 cut so to cover better all NLO effects.

UAEDF-tomc:
Ok, I have updated this in the cards.
@qliphy  I have addressed all comments of your review now

PR #2079, opened 2019-02-06
shepa configs for 2017/2018 production

Andrej-CMS:


efeyazgan:
@GurpreetSinghChahal could you please take allok?

GurpreetSinghChahal:
Hi @Andrej-CMS, thanks! Here are a few questions about ttH run card. I'll send comment about ttbb after cross-checking with ATLAS run cards
1. Even though the tune wouldn't matter much for this process but why are not we using default pdf (NNPDF30_nnlo_as_0118) that was used in default sherpa tune? 
2. The Njets/Ljets are defined in the run card but not used. Currently this CKKW* is not merging anything because there is no additional jet in the final state. Btw, how many additional jets are needed at LO or NLO for this sample?  
3. What is the target analysis? Since the detailed decay modes are not defined so H will just decay according to the branching ratios so a few analysis may suffer due to less statistics in some decay mode.

Andrej-CMS:
Dear @GurpreetSinghChahal 

1. I was not aware that the NNLO pdf was used for the tune. If you confirm that this is indeed the case, then I will reproduce the gridpack
2. This process doesn't work with additional jets and crashes during compilation if one adds additional jets. This card is on purpose without additional jets
3. The cards are on purpose inclusive in decay, so the resulting sample can be used for all ttH analyses.

efeyazgan:
@GurpreetSinghChahal could you please confirm that this is OK to merge now? thanks,

Andrej-CMS:
ping

agrohsje:
Settings out-dated. Will close for now.

PR #2093, opened 2019-02-12
Madgraph cards for H(125)-> 2h_D -> 4Z_D -> 8 ell

cecilecaillol:
Commit for both 2016 and 2017 cards


efeyazgan:
Hi, could you please keep only representative cards for one parameter sets and if possible a script to produce the other cards. thanks.

cecilecaillol:
Hi, I have kept only one mass point, all the rest is deleted now.

efeyazgan:
Thanks. Could you also get rid of commented/unusued lines such as 
#set group_subprocesses Auto
#set ignore_six_quark_processes False
#set loop_optimized_output True
#set loop_color_flows False
#set gauge unitary
#set complex_mass_scheme False
#set max_npoint_for_channel 0
thanks.

cecilecaillol:
Done

PR #2105, opened 2019-02-20
Update Hadronizer_TuneCUETP8M1_13TeV_MLM_5f_max2j_qCut60_LHE_pythia8_…

smdogra:
…cff.py

 To take into account  that for  2018 fragment should include PS weights w.r.t 2017 production.


qliphy:
For 2017 and 2018, the default Pythia8 Tune is CP5.   While CUETP8M1 is for 2016.
So this fragment indeed is not needed to be updated with PS weights w.r.t. 2017 production.

PR #2108, opened 2019-02-25
updated cards for exo_diboson/Spin-1/Wprime_Wh_Whadhbb and exo_diboson/Spin-1/Zprime_Zh_Zhadhbb for 2017 MC signal production

abenecke:


qliphy:
For Zprime card,   you define p as 4-flavor, 

probably you want also to include b and b~?

abenecke:
We do the same for the Wprime card.
And I did that to be consistent with the other Diboson signal sample.
Is that a problem?

qliphy:
Because for Zprime, there can be contributions from b b~ > Zprime.

I am asking this because there was some similar discussion before
https://hypernews.cern.ch/HyperNews/CMS/get/generators/4078/2.html

Usually the effect from b and b~ is small, but you may want to have a check, just in case.

abenecke:
I have added the contributions from b b~ > Zprime to the ZPrime cards

PR #2111, opened 2019-02-26
Cards and fragment for iDM model

afrankenthal:
Hello, these cards and fragment are for requesting 2018 MC for the iDM (inelastic dark matter) model. Thank you!


qliphy:
ickkw should be 1 instead of 0,  as you are producing MLM matching samples. 

Please also remove banners in the proc cards.

afrankenthal:
Hello, could you clarify something? I was under the impression that icckw = 0 wouldn't matter since in my Pythia configuration fragment I am turning off the option to read the matching parameters from the LHEF header. The option "JetMatching:setMad = off" is documented here: 

http://home.thep.lu.se/~torbjorn/pythia81html/JetMatching.html

I then set the matching to MLM via the option "'JetMatching:jetAlgorithm = 2" here:

https://github.com/afrankenthal/genproductions/blob/9775a405b2350d95cd58ab5cf4893784f99d62f4/python/ThirteenTeV/iDM/iDM_DarkPhotonToMuMu_ctau-XXX_TuneCP5_13TeV_MLM_4f_max2j_pythia8_cff.py#L36

However, if that is not sufficient then I am happy to change also icckw in the MG card. Thanks!

afrankenthal:
I'm sorry, the option to use MLM matching is actually "JetMatching:scheme = 1", and not "JetMatching:jetAlgorithm = 2" as I previously stated.

qliphy:
ickkw is more than triggering the pythia matching, as you can see here:
https://hypernews.cern.ch/HyperNews/CMS/get/generators/3967/1/1.html

afrankenthal:
Thank you for the clarification! I have changed icckw to 1 and xqcut to 20 (with my Pythia qcut at 30), and I'm validating it right now to check that the cross section doesn't change too much. I'll push the commit when it's done.

PR #2113, opened 2019-02-27
increase ncalls for low mass POWHEG ggH_quark-mass-effects

hroskes:
and require this in request_fragment_check.py

As I say in the comments there, it's probably ok for < 150 GeV, but we thought it was ok for < 300 GeV and this turned out to be wrong.  I'd prefer to be safe for future gridpacks.  However, the problem is only known to happen around 170 - 230 GeV.

Again this only applies to POWHEG ggH_quark-mass-effects.  As far as I know no other POWHEG processes are affected - at any rate VBF_H, HZJ, HWJ, and ttH seem to be fine for the mass ranges we generate in HZZ.  Gridpacks that don't use POWHEG definitely don't have this problem.

https://its.cern.ch/jira/browse/CMSCOMPPR-4735
https://its.cern.ch/jira/browse/CMSCOMPPR-4874

### DIFF
--- a/bin/MCFM/run_mcfm_AC.py
+++ b/bin/MCFM/run_mcfm_AC.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python
-from optparse import OptionParser
 from argparse import ArgumentParser
-import os, sys, re
+import os, re, subprocess, sys, textwrap
 
 def checkallfiles():
 	if not os.path.exists('./runcmsgrid_template.sh'):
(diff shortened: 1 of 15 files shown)


qliphy:
Thanks a lot!  Adding @Andrej-CMS @covarell @yuanchao 

As this involves also checking script, I will let @efeyazgan  have a further look

hroskes:
I tested it on
```
python request_fragment_check.py --prepid HIG-RunIIFall18wmLHEGS-00069 --bypass_status
python request_fragment_check.py --prepid HIG-RunIIFall17wmLHEGS-03148 --bypass_status
```
The first gives [OK], the second gives [WARNING]s, as expected (that will change once I get a new gridpack but probably not for a few days at least).

efeyazgan:
looks good. thanks a lot.

efeyazgan:
I was too quick. It doesn't work with the "--ticket" option anymore. Could you please fix it? thanks.

efeyazgan:
In fact, it doesn't work for --prepid either. 

File "request_fragment_check.py", line 284
    matches = {name: re.search(r"^"+name+" *([0-9]+)", content, flags=re.MULTILINE) for name in desiredvalues}

hroskes:
Will fix in the next 15-20 minutes

hroskes:
@efeyazgan ok, I see the issue, the default python on lxplus is very old and doesn't allow this syntax.  I'm making a new PR.

PR #2116, opened 2019-02-28
add Qstar -> QW and QZ sample cards with CP2 tune

clelange:
These are the Pythia cards for the Q* samples to be used in the Run-2 legacy B2G diboson analysis. With respect to the already existing ones, they have the new tune CP2 to avoid issues with negative PDF weights. Cross sections are set to 1 consistently.

FYI @ahinzmann

### DIFF
--- a/python/ThirteenTeV/QstarToQW/QstarToQW_M_1000_TuneCP2_13TeV_pythia8_cfi.py
+++ b/python/ThirteenTeV/QstarToQW/QstarToQW_M_1000_TuneCP2_13TeV_pythia8_cfi.py
@@ -1,0 +1,33 @@
+import FWCore.ParameterSet.Config as cms
+
+from Configuration.Generator.Pythia8CommonSettings_cfi import *
+from Configuration.Generator.MCTunes2017.PythiaCP2Settings_cfi import *
+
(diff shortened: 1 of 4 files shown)


efeyazgan:
Note that there is a plan to use NNPDF31_NNLO_as_0118_MC set which is positive definite and can be used with CP5. We are validating this set now.

efeyazgan:
And in any case, please do not put different files for each parameter of the same process. You could have a few representative ones and a script to provide the other files with different parameters.

clelange:
Hi @efeyazgan 

OK, I've added the scripts that I used to create the configs and only added the 1 TeV mass points for both cases. I also force-pushed to remove the other files from history.

We'd be happy to use the NNPDF31_NNLO_as_0118_MC set and can go to CP5 in this case, it just depends on the timescale, since these samples should enter production soon. Do you think we can ask for production with the new PDF set already now (and then use CP5)?

clelange:
Hi @efeyazgan - can this be merged, please? Physics coordination would like us to have the analysis out by summer, and missing signal samples are a real blocker. Using CP2 seems to be the safer choice at this point.

PR #2117, opened 2019-03-01
no spin correlations for ttbar inclusive

GiuliaNegro:


efeyazgan:
Looks correct but @agrohsje , could you please double check? thanks.

agrohsje:
Hi Giulia, hi Efe, what is the reason to produce an inclusive no spin corr sample? 
As a comment: I checked the tarball powhegboxV2_rev3633_date20190225_testggHH.tar.gz and it has the proper nospincorr option. So keep fingers crossed. :-)

agrohsje:
Sorry one more question. It is on purpose NNPDF 3.0 ?

GiuliaNegro:
Hi Alexander, a no spin corr sample was requested by Georgios Bakas. It has NNPDF 3.1.  @EmyrClement and @ckmackay can confirm it.

agrohsje:
Me culpa 306000 is NNPDF31_nnlo_hessian_pdfas! About the final state, maybe worth checking if he really needs all-hadronic?

GiuliaNegro:
At the moment we are requesting all possible final states (not only fully hadronic), but I will check with Georgios.

agrohsje:
I know that he wants to do some boosted analysis, but not sure what exact final state. In case all-hadronic would not be involved that could gain us some statistics and save resources, while keeping dilepton might be interesting for the 18-006 follow-up.

GiuliaNegro:
Georgios confirmed that he needs all-hadronic final states (see https://hypernews.cern.ch/HyperNews/CMS/get/top-modeling-and-generators/282/1/1/1/1/1/2.html), so we can proceed with the request.

PR #2120, opened 2019-03-04
Run request checking script without cookie

justinasr:
Run request checking script without SSO cookie and return number of errors as exit code. This pull request should be squashed before merging because it has multiple commits which are not important.

Also, I added 
```
res = [res]
```
line, so changes made by me would be as small as possible as I did not want to re-indent half of the file.


efeyazgan:
Thanks. I've run a few tickets and requests and at least nothing seems to be broken. There are still changes needed to move some warnings to errors and vice versa.

PR #2121, opened 2019-03-05
monoH gridpacks production using condor batch and gz=0.2

fmanteca:


qliphy:
Why do you need to upload  gridpack_generation.sh?  Is it different from the one in the master branch of genproductions?
https://github.com/cms-sw/genproductions/blob/611f3675fd4bbeed1337fde8a06224935707e512/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/monoHiggs/Zp2HDM/Zprime_Zh_Zinv/gridpack_generation.sh

qliphy:
Moreover, we also provide  condor submission script,  can you just use this? 
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/submit_condor_gridpack_generation.sh

fmanteca:
Hello,

Up to now we were using Madgraph5 2.4 to produce the gridpacks, that was the main reason for using the old gridpack_generation.sh. After a discussion among us, we agree to use MG5 2.6, so I will remove the gridpack_generation.sh from the merge request. Thank you

On the second comment, is there some documentation for running your submit_condor_gridpack_generation.sh available? I have been trying to submit the several jobs in parallel, but the job output is displayed in my screen and when I stop the display then the job is removed from condor.

qliphy:
You can follow GEN twiki:  https://twiki.cern.ch/twiki/bin/view/CMS/QuickGuideMadGraph5aMCatNLO

fmanteca:
OK thanks. I have already updated the scripts in the new commit

qliphy:
>Up to now we were using Madgraph5 2.4 to produce the gridpacks, that was the main reason for >using the old gridpack_generation.sh. After a discussion among us, we agree to use MG5 2.6

We have kept a branch (https://github.com/cms-sw/genproductions/tree/mg242legacy) with MG5_2.4.2
So you can also commit cards there.

However, MG5_2.6.0 is newer and it is better if you can use this verion. 


> so I will remove the gridpack_generation.sh from the merge request. 

Please remove not needed file, e.g. submit_gridpack.sub,  submit_gridpack_condor.sh and gridpack_generation.sh etc.

fmanteca:
Ok, I have already removed them. (last commit updated cfg)

PR #2133, opened 2019-03-09
delete /tmp/USER/gridpack/eos/* to save space

qliphy:
Otherwise for a check with many requests, there will be no space left and lead to exit


qliphy:
@efeyazgan   Can you please have a look at this?

PR #2141, opened 2019-03-14
HNL with tau couplings (2016+2017)

cecilecaillol:
Heavy neutral leptons with couplings to taus
2016 and 2017 gridpacks


qliphy:
Why do you add "SM_HeavyN_FilesWithUFO.tgz"  in the extramodels.dat cards?

As note [here](https://twiki.cern.ch/twiki/bin/viewauth/CMS/Mg26xTests),  In mg26x, we recommend to set "True = store_rwgt_info" in run_card.dat for NLO process, which is necessary for the "systematics" module afterwards.

cecilecaillol:
Your suggestions have been implemented. SM_HeavyN_FilesWithUFO.tgz was not necessary.

qliphy:
It seems you still miss  "True = store_rwgt_info"  in HeavyNeutrino_trilepton_M-1000_V-0.01_tau_NLO_run_card.dat

cecilecaillol:
I thought it was for 2017 datacards only. Now it is also added in the 2016 version.

qliphy:
"True = store_rwgt_info" is required for MG26X version (in master branch of genproductions).

PR #2146, opened 2019-03-18
use the provided powheg.input card for parallel stages 2 and 3

AndrewLevin:
Currently, the powheg.input card provided with "-i" is ignored for parallel stages 2 and 3. For example, in these cases, it would be ignored:

```
# step 2
   python ./run_pwg_condor.py -p 2 -i powheg_Zj.input -m Zj -f my_Zj -q longlunch -j 10
# step 3
   python ./run_pwg_condor.py -p 3 -i powheg_Zj.input -m Zj -f my_Zj -q longlunch -j 10 
```

The powheg.input card may need to be changed between steps 1 and 2 and 3, e.g. the parameters `nubound` and `ncall1`.

### DIFF
--- a/bin/Powheg/run_pwg_condor.py
+++ b/bin/Powheg/run_pwg_condor.py
@@ -1692,6 +1692,7 @@ if __name__ == "__main__":
         os.system('cd '+rootfolder+';bash '+scriptName)
 
     else                    :
+        os.system('cp -p '+args.inputTemplate+' '+args.folderName+'/powheg.input')
         runEvents(args.parstage, args.folderName,
                   args.eosFolder + '/' + EOSfolder, njobs, powInputName,
                   jobtag, args.prcName)


qliphy:
adding @covarell @yuanchao @agrohsje

covarell:
Indeed, once the powheg.input card is provided at step 0, it is manipulated to modify the parallelstage only inside the workdir and it is not copied every time.

We are going to make some changes for ggHH and to improve file handling, we can include this feature too.

covarell:
However, I would be surprised if it works with just your one-line change because the initial powheg.input should contain the NEVENTS and SEED strings (if they are not changed to sensible numbers inside the runEvents routine, the program should crash no?)

AndrewLevin:
I am not sure if it works for all purposes, but for gridpack generation in 3 parallel stages, I think it does work. Since `manyseeds` is set to `1`, I think the `seed` is irrelevant. Also, the number of events is overwritten here

https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/run_pwg_condor.py#L997

so it should be set correctly when the gridpack is run (as long as provided powheg.input has `numevts` not commented).

yuanchao:
Well, NEVENTS and SEED are taken from comment arguments directly. If new input card is assigned, these two options need to be present too. Then the rest is just copy the corresponding code for different steps.

AndrewLevin:
@yuanchao, yes, for gridpack generation in 3 parallel stages, I think `iseed` and `numevts` only need to be present in the card used in the last step, since that is the card that is put in the gridpack, but for other uses of this script, they would need to be present in other steps

covarell:
Indeed in practice iseed and numevts are not used in the intermediate parallel stages. My worry was only that if they are read as strings not numbers, even if ignored, the programs may crash anyway. 

But I did not test this: if you have and there are no problems, then I approve this change.

covarell:
Superseded by https://github.com/cms-sw/genproductions/pull/2167

PR #2152, opened 2019-03-22
HToZdZdTo4l cards and template for various mass points

rosedj1:
Included is the template for making ZdZd cards and the already-prepared cards for mass points.


qliphy:
Do you really need to upload so many cards?  Most of the run_card.dat files indeed are the same

rosedj1:
> Do you really need to upload so many cards? Most of the run_card.dat files indeed are the same

Updated.

qliphy:
For LO process with mg26X (similar as for mg242), you should also add in run_card.dat: " True  = use_syst      ! Enable systematics studies"

rosedj1:
> For LO process with mg26X (similar as for mg242), you should also add in run_card.dat: " True = use_syst ! Enable systematics studies"

Added.

qliphy:
"True     = use_syst" is fine.

PR #2153, opened 2019-03-24
Remove warning for datasets with MCFM in the name

hroskes:


efeyazgan:
I think you also need to make the following change:
            if ind == 3: --> if ind >= 3
in line https://github.com/cms-sw/genproductions/blob/32f097d84517f364edf7e86630ce5646b12b835e/bin/utils/request_fragment_check.py#L342 
thanks

hroskes:
@efeyazgan I don't think so.  It says break, not continue, so once it gets there for ind == 3 it won't continue for ind == 4

efeyazgan:
OK, I see what you wanted to do now.

PR #2159, opened 2019-03-26
updated model and updated PDF

tiwariPC:
Dear all,
We have updated the model and the PDF which we will use in the sample production.
Please have a look and merged this PR.
Thanks a lot,
Praveen

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/bbDM_2HDMa/Template/extramodels.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/bbDM_2HDMa/Template/extramodels.dat
@@ -1,1 +1,1 @@
-2HDMPZ2II_v1.tgz
+Pseudoscalar_2HDM.tgz
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/bbDM_2HDMa/Template/proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/bbDM_2HDMa/Template/proc_card.dat
@@ -1,4 +1,4 @@
-import model Pseudoscalar_2HDM
+import model Pseudoscalar_2HDM-bbMET_5FS
 
 define p = g u c d s u~ c~ d~ s~ b b~
 define j = p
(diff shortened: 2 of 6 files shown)


qliphy:
Why do you need this specific PDF with ID 13100?

tiwariPC:
This pdf has been suggested by Katerina Lipka, and Exo MC contacts, MCI and MET+X convenors have given the green light the for this pdf hence we have used this PDF

qliphy:
Ok, thanks.  To leave a record for future reference, can you give a bit more details or provide a link to refs?

tiwariPC:
For the details, please have a look here :
https://indico.cern.ch/event/800504/contributions/3326573/attachments/1800774/2937275/bbDM_StatusUpdate_22022019.pdf#page=3
For the pdf selection, the discussion was over mail.

qliphy:
But "13100 CT14nlo" may brings inconsistency with Pythia8 Tunes.
If you don't want NNLO PDF, then why not just use NNPDF3.1LO with CP2 Tune?  

tiwariPC:
We will use the following combination for our sample production:
ME (NNPDF3.1 NLO) + PS (CP3)
this combination has also been suggested in the twiki no?

qliphy:
Yes,  that is fine.  But then why setting 13100 (CT14nlo) in run_card.dat?

tiwariPC:
Dear Qiang Li,
Thanks a lot for clarification.
We will use NNPDF3.1 NLO as the central pdf. I checked here: 
https://lhapdf.hepforge.org/pdfsets
I see that for NNPDF3.1 NLO the pdf ID is 305000.
But this PDF id I do not see in the following list:
https://github.com/cms-sw/genproductions/blob/master/MetaData/pdflist_5f_2017.dat
Our main motive is that we want to save  NNPDF3.1 NLO as central pdf and also save 13100 (CT14nlo) for systematics.
Could you please let me know what is the way to do that and how to check if the systematics are saved in the LHE file?
Thanks a lot,

qliphy:
But there is  "305800 NNPDF31_nlo_hessian_pdfas 103"  which you can use.

To check PDF information, you can refer to: 
https://twiki.cern.ch/twiki/bin/viewauth/CMS/HowToPDF

tiwariPC:
Dear Qiang,
I have updated the cards, please have a look.
Thanks a lot

PR #2167, opened 2019-03-27
Modifications to support the ggHH process (mtdep=3)

covarell:
Also implementing two suggestions:
- Accepting https://github.com/cms-sw/genproductions/pull/2146 and superceding it  @AndrewLevin 
- Setting runInBatchDir as the default for parallel stages, as it is completely tested now @intrepid42

### DIFF
--- a/bin/Powheg/run_pwg_condor.py
+++ b/bin/Powheg/run_pwg_condor.py
@@ -926,7 +926,7 @@ chmod 755 runcmsgrid.sh
 def runEvents(parstage, folderName, EOSfolder, njobs, powInputName, jobtag, process) :
     print 'run : submitting jobs'
   
-    sedcommand = 'sed -i "s/NEVENTS/10/ ; s/SEED/3/ ; s/parallelstage.*/parallelstage ' + parstage + '/ ; s/xgriditeration.*/xgriditeration 1/" '+folderName+'/powheg.input'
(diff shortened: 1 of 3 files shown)


efeyazgan:
@covarell  this is ready to be merged, right? thanks.

covarell:
Well, yes, as usual I would first let @yuanchao and other people in the thread comment (especially because I merged together several updates)

yuanchao:
@covarell  Just had a quick view. Could you clarify the followings:
* powhegboxV2_rev3633_date20190225_testggHH.tar.gz <--- this is only for test?
* is 10 events good enough (not degrading the grid)
* linking obj-gfortran is not needed as directly copied?

covarell:
The tarball is indeed a "test" in the sense that it contains code not publicly available yet in the POWHEG-BOX. It was provided by theorists to CMS for testing it.

I put a random number for stage 2 and 3 because I think this is not used. Is it used? If so, we need to increase it.

I do not know about obj-gfortran, this is not a change I did.

covarell:
In the last commit I moved to the default nEvents (2000)

About obj-gfortran, bb4l and the like are removed from powhegjhugen run card, since these processes do not support JHUGen

yuanchao:
Ok. Then I'm sure if it's a good idea to put the test package in the official script.
As to the rest, it's fine with me. Thank you!

covarell:
Done! I think this can be merged now.

PR #2169, opened 2019-03-29
ChargedHiggs_taunu 2017/13TeV: added missing parameter store_rwgt_info

slehti:
Added parameter store_rwgt_info into the H+ ->taunu run files missing it, recommended for MG526.


qliphy:
store_rwgt_info is needed only for NLO but not LO.  
We have add this in our twiki:  https://twiki.cern.ch/twiki/bin/view/CMS/Fall18Repack

slehti:
Do I remove the store_rwgt_info from the LO cards, or does it matter? What do you prefer?

qliphy:
Yes, please remove store_rwgt_info from LO cards. Thanks!

PR #2176, opened 2019-04-04
Adding hadronizers for VBF photon

yiiyama:
We are going to request background samples for a VBF photon production cross section measurement. The samples will be aMC@NLO GJets with FxFx merging (which is technically not valid for GJets final state but could be usable at high photon pT; we have been encouraged to give it a try). We will apply a phase space cut with dijet mass because otherwise the cross sections are enormous.
There is already a EDFilter that applies a dijet mass cut using GenJets, but since the module cleans the jets only against leptons, we have to create a photon-vetoed jet collection first to feed to the module.
The fragment has been validated with the GJets gridpack that we are going to use in the production.

I should point out that the ak4GenJetsNoNuNoGamma collection produced by this fragment stayed in the event (the standard event content commands seem to keep the collection). If that can cause issues downstream in the production, I can add an explicit drop statement at the end of event content configurations.

### DIFF
--- a/python/ThirteenTeV/GJetsVBF/GJets_Mjj-400_TuneCP5_13TeV_aMCatNLO_FXFX_5f_max2j_qCut45_LHE_pythia8_cff.py
+++ b/python/ThirteenTeV/GJetsVBF/GJets_Mjj-400_TuneCP5_13TeV_aMCatNLO_FXFX_5f_max2j_qCut45_LHE_pythia8_cff.py
@@ -1,0 +1,88 @@
+import FWCore.ParameterSet.Config as cms
+
+from Configuration.Generator.Pythia8CommonSettings_cfi import *
+from Configuration.Generator.MCTunes2017.PythiaCP5Settings_cfi import *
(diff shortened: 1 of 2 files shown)


yiiyama:
@menglu21 PR is made :)

qliphy:
Adding @agrohsje @kdlong  @lviliani @covarell

qliphy:
Adding @agrohsje @kdlong  @lviliani @covarell

yiiyama:
Hi @qliphy @agrohsje @kdlong @lviliani @covarell, is there any problem with this PR?

kdlong:
@yiiyama I see no technical issue, assuming you've checked that the distributions aren't biased by the selection. In principle you can filter on whatever you want as long as you effectively below the phase space you're interested in.

Do you really gain anything from the FxFx merging? I would think that after the high mjj cut you're fully covered by the 2j contribution and there's not much point in merging in the 0,1j.

yiiyama:
Hi @kdlong sorry for a late response. Yes, the filter efficiency for the 1j sample is extremely low at ~2%, but since it's not zero, we'd need these events too.

PR #2180, opened 2019-04-05
Adding Inclusive Leptonic Decays, Renaming one File

atishelmanch:
Hello,

I'm making this pull request in order to add the inclusive fully leptonic and semi leptonic decays. Process-wise, this is:

WWgg->lnulnugg
WWgg->qqlnugg

where 'l' can be an electron or muon, and q can be 'd', 'u', 's', 'c'.
I've also renamed a previously pulled fragment to fix a typo (there are not 21 muons in that file's final state). Does this look ok?

Thanks,
Abe Tishelman-Charny

CC: @acarvalh


efeyazgan:
Are the corresponding run and proc cards already in github?

atishelmanch:
Hello,

Yes, they are here: 

https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/pre2017/13TeV/exo_diboson/Spin-0/Radion_hh_narrow/Radion_hh_narrow_M900

Thanks,
Abe Tishelman-Charny

PR #2183, opened 2019-04-08
2016 vbf zprime zh

IreneZoi:
These cards were never done before, or at least I could not find them. They are a modification of https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/exo_diboson/Spin-1/Zprime_Zh_Zhadhbb for the VBF production mode for 2016 MC


qliphy:
You define 5f for "p" = u c d s b u~ c~ d~ s~ b~ ,  but not for "j"
In run_card.dat, you apply 263400, i.e. the 4f PDF set.    Can you make it consistent?

PR #2186, opened 2019-04-08
configure seed for powheg stages 2 and 3

hroskes:
- seeds for the 10 jobs start from the one provided on the command line
- fix some headers to use the right python

if @covarell agrees


covarell:
Can you please be more specific? 
- Why is the current pwgseeds.dat file (numbers from 1 to njobs) not correct? AFAIK it is sufficient that the numbers are different from each other.
- What is "the right python"? If you type "python run_pwg.py" rather than "./run_pwg.py" (as it is advertized in the Twiki) doesn't it select the python version coming from cmsenv?
@yuanchao

yuanchao:
Indeed as we advertized in twiki that "python run_pwg.py", it should not be a problem. Though I think it's ok to point to the python specified in the environment.
As to the seeds, I think we should stick to the current pwgseeds.dat.

hroskes:
Hi, is there any obstacle to merging?

qliphy:
@covarell @yuanchao    It seems ready to be merged, right?

covarell:
Ok for me

PR #2190, opened 2019-04-10
data cards for 2016 bbDM_2HDMa sample

tiwariPC:
Dear all,
This Pull Request is for 2016 sample production for the bbDM analysis with 2HDM+a model.
2017  datacards are already there in the following PR:
https://github.com/cms-sw/genproductions/pull/2159

Please have a look and merge the PR
Thanks a lot,
Praveen


qliphy:
Are there any differences from the already merged 2017 cards?

tiwariPC:
only thing is that I have made these data cards using madgraph 242

qliphy:
Indeed you can use the same MG260 gridpacks for both 2016 and 2017/2018 productions.  If so there is no need to upload these cards made for MG242.

tiwariPC:
Okay thanks a lot, I shall close this PR

PR #2193, opened 2019-04-10
Datacards for samples to be produed in Fall18wmLHEGS campaign

grasph:
Datacards for samples to be produed in Fall18wmLHEGS campaign for the HGG group.


efeyazgan:
Hi @grasph instead of 52 files, could you please add only a few representative cards and if possible a script to generate the other cards? thanks.

grasph:
The production of the samples is already submitted and changes in datacards were minor (limited to new recommendations). I close the pull request.

PR #2195, opened 2019-04-12
First commit for leptoquark madgraph production, decaying to b+ele or b+mu

DavidMorse:
First commit for 2016-2018 leptoquark madgraph production, decaying to b+ele or b+mu

Single and pair production

Separate cards for 2016 and for 2017-2018:

2016 uses default PDF and Tune to be in coordination with bg samples.

2017-2018 uses NNPDF_31_lo_as0130 and TuneCP2, to avoid problems seen at high resonance mass with negative pdf weights, as recommended by EXO MC contacts


DavidMorse:
@DryRun @jotadram6 @sethcooper

DryRun:
Hi @qliphy , could this PR be merged? Thanks!

qliphy:
In general look good to me.  
Is there any special reason to use fixed scales instead of the default dynamical one?

     True = fixed_ren_scale  ! if .true. use fixed ren scale
     True = fixed_fac_scale  ! if .true. use fixed fac scale
     300  = scale            ! fixed ren scale
     300  = dsqrt_q2fact1    ! fixed fact scale for pdf1
     300 = dsqrt_q2fact2 ! fixed fact scale for pdf2

DavidMorse:
The renormalization and factorization scales are set to the value of the leptoquark resonance mass. This is the suggestion of the theorists who built the model. It is also done this way to calculate the production cross section. This has been done this way historically for leptoquark searches for many years.

PR #2196, opened 2019-04-15
Vector LQ with ME (NNPDF3.1 LO) + PS (CP2)

IHEP-CMS:
Vector LQ (pair and single production) previously produced with NNPDF3.1 NNLO are now produced with NNPDF3.1 LO in order to remove the dependence from negative weights.


efeyazgan:
@aspiezia, I assumed you know that the recommendation for ultra legacy is to use the no-negative-weight version of NNPDF3.1 NNLO (i.e. NNPDF31_nnlo_as_0118_MC) with CP5.

PR #2200, opened 2019-04-18
Update fragment tune and name for ttbar Pyquen PbPb 2018

gkrintir:


qliphy:
Can you add the tune to the fragment's name?

gkrintir:
thanks for spotting @qliphy 

it was Z2star, i mistakenly committed the wrong file

the tune name is now also reflected to the cfg filename

PR #2202, opened 2019-04-19
updated WPrime cards for 2018 MC production

abenecke:
added extramodel card
changed pdfwgt from T to F 
increased mass range


qliphy:
Please check updated twiki: https://twiki.cern.ch/twiki/bin/view/CMS/Fall18Repack

pdfwgt should set to be False

abenecke:
Sorry for the confusion, I meant changed pdfwgt from T to F.
in the run card it is set:
 F = pdfwgt
Is that what you meant?

qliphy:
Ok, then that should be fine.

PR #2216, opened 2019-05-08
TauDecays included

wajidalikhan:


efeyazgan:
@wajidalikhan , could you please include only a few representative cards and a script to produce the others? thanks.

wajidalikhan:
Hi @efeyazgan, I remember there was a reason for not using a script to produce the data cards for single top FCNC samples in past. However, we do have a script to produce the data for TT FCNC process. So that is why we I have not done it in that way.

wajidalikhan:
Hi @efeyazgan, I asked Kate about it and it looks that there was no script to do ST ones and it would be not useful to write a script for these ST FCNC samples since we have a variety couplings with which samples are produced. Its not just a single data card file which which scans for the mass points and rest remains the same. Let me know if this makes sense to you.

PR #2239, opened 2019-05-27
adding new ttgamma samples

llechner:
adding the cards for the new nominal and pT binned ttgamma signal sample cards for hadronic, singleleptonic and dileptonic decays


qliphy:
please remove MG5_aMC banner if present 
https://twiki.cern.ch/twiki/bin/view/CMS/Fall18Repack

llechner:
done

PR #2242, opened 2019-05-28
Updates to remove MLM matching and add the taus in 4B cards

georgia14:


qliphy:
These are too many similar cards for different parameters.  Would you please just upload some benchmark cards together with a script to make other cards?   You can find such examples in some recently merged PRs.

georgia14:
Hi @qliphy 
Ok, I have updated the PR . I have replaced with template directory cards and a script to generate all a0 mass points.
Thanks, Georgia

georgia14:
Could a MC expert please review if " pythia8PSweightsSettings " are needed in the Pythia fragment I added ?
Thanks, Georgia

qliphy:
parton shower weights (pythia8PSweightsSettings) are recommended to add for 2018 samples.

https://twiki.cern.ch/twiki/bin/view/CMS/Fall18MC


On the other hand,  usually shower fragment is not supposed to upload to genproductions area, which is usually for keeping generator cards.    Shower fragment now is directly put in McM request by MC contacts.

georgia14:
Ok, thanks . @clacaputo suggested me to add the Pythia fragment under genproductions. Let me know if you want me to remove.

qliphy:
Ok, you can still keep it here for reference.

Your cards look fine to me. I will let others to comment before merging it later today or tomorrow.

PR #2244, opened 2019-05-29
Fix patch 0006 to restore the LHE header

lviliani:
@kdlong @covarell @agrohsje @efeyazgan @qliphy 

This is to fix the issue discussed in this thread:
https://hypernews.cern.ch/HyperNews/CMS/get/generators/4324.html

The patch was originally meant to add the central weight in the scale variation weight group of the LHE header, but the results was a messed up header.

Note that this is not the default in MG, i.e. the systematics module does not write the central weight in the scale variation group (checked with MG standalone), but it is consistent with what CMS has done so far.

The fix has been tested with the wplustest_4f_LO process.
OLD patch: http://lviliani.web.cern.ch/lviliani/test_patch_0006/cmsgrid_final_oldpatch.lhe
NEW patch: http://lviliani.web.cern.ch/lviliani/test_patch_0006/cmsgrid_final_newpatch.lhe

**NB**: While checking this I also noticed that pdfs for which we don't store variations, e.g.:
https://github.com/cms-sw/genproductions/blob/mg261/MetaData/pdflist_4f_2017.dat#L4-L6
get written in the LHE header outside weight groups (see the LHE files above). This also happens in MG standalone but seems not consistent with what we did in MG242.
I'm not sure whether this can be a problem or not.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/patches/0006-add-central-weights.patch
+++ b/bin/MadGraph5_aMCatNLO/patches/0006-add-central-weights.patch
@@ -1,6 +1,6 @@
---- a/madgraph/various/systematics.py	2017-05-01 21:02:10.402600552 +0300
-+++ b/madgraph/various/systematics.py	2017-05-01 21:01:44.759601146 +0300
-@@ -228,7 +228,7 @@
+--- a/madgraph/various/systematics.py	2019-05-29 10:48:35.000000001 +0200
++++ b/madgraph/various/systematics.py	2019-05-28 16:02:38.000000001 +0200
+@@ -314,7 +314,7 @@
          else:
              lowest_id = self.get_id()        
  
@@ -9,7 +9,7 @@
          all_cross = [0 for i in range(len(self.args))]
          
          self.input.parsing = False
-@@ -261,7 +261,7 @@
+@@ -348,7 +348,7 @@
                  print event
                  raise Exception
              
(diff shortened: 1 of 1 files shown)


qliphy:
Is this also needed for the master branch (with mg260) and mg261_UL branch?

lviliani:
> Thanks! Is this also needed for the master branch (with mg260) and mg261_UL branch?

I think so, let me do a quick test with the master branch to be sure.

lviliani:
I confirm it is needed for master branch as well.
And I guess also for mg261_UL, given that it is forked from mg261.

qliphy:
Thanks @lviliani   
Then after this get merged, please also make PRs to master and mg261_UL branches.

kdlong:
I think we should seriously consider if we want to implement this. In principle it's not a problem to not store the central scale variation, and we've already abandoned being consistent across samples.

lviliani:
I agree with @kdlong , we need to discuss this option as well.

Let me add that in any case we need to fix the issue with the LHE header, either removing the patch completely (thus using the MG default and not storing the central weight), or using this fix.

kdlong:
Right. I also agree with you that we should address the issue with the weight group labeling. This is the type of change that I would think the authors would be willing to integrate, which makes it easier to maintain.

agrohsje:
Fully agreed. If that can be centrally fixed, that would be useful and ideal. 
@kdlong Why would you prefer not having the central value as first one? 
@qliphy What are the differences between UL branch and none-UL branch of 261. Just the PDF's? If so, wouldn't it be easier to handle this separately then 2 branches? What is the purpose of 26x branch? It is 2.6.0 as master. Is it still needed? Just to be sure we don't get lost and consistently maintain all branches.

qliphy:
@agrohsje  The difference between UL and non-UL 261 branches  is indeed PDF and environment settings (Positive PDFs currently are only in SLC built 10_6_0+).

26X is historical and not needed but probably we can keep it for a moment.

efeyazgan:
Hi,
Can we come to a conclusion on this soon (by the end of this week)?  For me the best option seems to not  use this for UL and document how to deal with it very clearly. We lose consistency with the past samples but at least UL will have less problems.

agrohsje:
@kdlong Hi Kenneth, can you comment on why you want to drop the first weight? I would prefer keeping it for now until the general weight issues are settled. But maybe there is something I am missing.

kdlong:
@agrohsje I'm not saying I am opposed to it, just that I'm concerned with the number of patches we have and the issues it causes if one isn't updated properly in the future, as in this case. So maybe we should ask whether this is really worth adding.

agrohsje:
Thanks @kdlong for the feedback. That is indeed true. My only concern is that it could be useful to keep this for debugging purposes. But I don't know. I think we should just take a decision. @qliphy @efeyazgan @covarell @lviliani do you have any preference?

lviliani:
Hi, I don't have a strong opinion.
But if we decide to drop this patch we have to make sure that all analyzers are aware of the changes in the LHE header wrt what we had before.

qliphy:
I second Lorenzo

efeyazgan:
same here

agrohsje:
Ok. So if nobody has a too strong opinion I would suggest to fix the patch for now in all relevant branches and keep it a little more. Meanwhile maybe @lviliani can check with Olivier the issue with info outside the weight group above? Once @kdlong updates the weight objects, we can reconsider as by then we have a more complete picture.

efeyazgan:
@lviliani is this ready to be merged? We need to start the UL production soon and this seems to be the last missing piece.

lviliani:
I tested this patch also with more complicated processes (FxFx + madspin) and it works as expected. So yes, it is ready to be merged.

lviliani:
Should I make a PR also for other branches? master, mg261_UL, UL2019?

efeyazgan:
Yes please. Thanks.

lviliani:
I guess there are server problems now and I can't make a new PR... I will try again later.

PR #2245, opened 2019-05-29
Lljj cards + cream02 support

tomcornelis:
Addition of the LLJJ_105-160_EWK_SM_5f_LO cards
Now doing this pull request I realize there's also another older commit in my own branch which restores the cream02 support (which apparently was dropped in the move from v242 to the new master branch).
Given that it is currently the only pbs cluster type those changes are not affecting the other ways of producing a gridpack, so I don't see a problem having this in the central repository too.


qliphy:
Please remove these lines:  https://github.com/cms-sw/genproductions/blob/c1faef53499e7a9c861db9c3f570d79956289d51/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/LLJJ_105-160_EWK_SM_5f_LO/LLJJ_105-160_EWK_SM_5f_LO_run_card.dat#L273-L284

qliphy:
About cream02 support,  I am adding @kdlong @AndreasAlbert  @lviliani  @covarell  @efeyazgan @agrohsje for more comments.

agrohsje:
In principle I don't mind adding this as it is orthogonal. However, as we provide several options to prepare gridpacks that we support as much as we can, I think adding something that we don't consider actively is not ideal. I would prefer dropping this in central scripts. If we decide to support the option I am fine with keeping it.

agrohsje:
Will close this PR. @tomcornelis  please ping me in case this is not ok.

PR #2252, opened 2019-06-05
Update XtoAAto4b_X1000A50_proc_card.dat

osherson:

PR #2254, opened 2019-06-05
Update XtoAAto4b_X1000A50_run_card.dat

osherson:


qliphy:
There is a conflict with https://github.com/cms-sw/genproductions/pull/2252 

Please make several commits together in one PR.

Moreover, as this is a loop induced process, the run_card.dat setting should be more like LO but not NLO.   Please follow this example [1] to update your card.

[1] https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/examples/ggzh01j_5f_LO_MLM/ggzh01j_5f_LO_MLM_run_card.dat

agrohsje:
@osherson Is this still needed or can we close it?

osherson:
Yes, thank you,

- Marc

PR #2257, opened 2019-06-07
correct xsec in gridpack mode

qliphy:
https://bugs.launchpad.net/mg5amcnlo/+bug/1830567


kdlong:
While clearly we'd rather have this information correct rather than being 0 all the time, Olivier's suggestion to properly use the cross section written in the <init> block does seem reasonable. But more than anything I think it'd be great to drop the dependency on the merge.pl script. Could we move to a workflow that did the merge by reading multiple LHE files in cmsRun?

PR #2260, opened 2019-06-10
last changes to run card

osherson:
Still not able to compile with error:

Command "launch" interrupted with error:
KeyError : 'y0'
Please report this bug to developers

           More information is found in 'MS_debug'.

           Please attach this file to your report.

"y0" is the name of the new particle in our model.


osherson:
Actually looking further, I also see this error:

Command "generate_events pilotrun" interrupted with error:
Exception : 0.4  =  dparameter5  can not be mapped to an integer

However, the 0.4 was copied from the example you asked me to follow, so I'm nor sure how that's failing.

osherson:
Actually, I also noticed:
Command "generate_events pilotrun" interrupted with error:
Exception : 0.4  =  dparameter5  can not be mapped to an integer
But 0.4 is copied from the example you provided. Is there something wrong with my MG setup?

qliphy:
Can you add you girdpack generation log somewhere?
Moreover, where is the model file you used, i.e. Phi2_Simp.tar.gz ?

osherson:
Hi Qiang,

It is here:
however, I just noticed that there is a mis-match between the name in my
card and the name of the model in that directory. Not sure when that was
introduced but I will try running with the correct name.
I'll send the log from that trial.
- Marc

osherson:
Actually in my version I had already made this change. I've attached the
log.
- Marc

qliphy:
I can reproduce your error.  It seems to be due to run_card.dat.
I works well for me If changing to use  https://raw.githubusercontent.com/cms-sw/genproductions/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/DYJets_HT_LO_MLM/DYJets_HT_mll50/DYJets_HT-incl/DYJets_HT-incl_run_card.dat     (updated xqcut to 100 in addition as in your card)

osherson:
Hi! Thanks! That does indeed seem to work, except that I got this warning:
CRITICAL: Branching ratio larger than one for 54
Does that matter? How can I check that the gridpack "works"?
- Marc

qliphy:
This warning means the actual decay width of particle with PID=54 is larger than what you set in param_card.dat

You may set a more correct value or try set  "DECAY 54 AUTO"

Please also refer to e.g.  https://answers.launchpad.net/mg5amcnlo/+question/663487

osherson:
Hi, we have updated to using the customizecards, and are looking to push this to master. Please let us know if we're missing anything.

PR #2261, opened 2019-06-12
Update for LQToBMu and LQToBEle example cards

DavidMorse:
Update for LQToBMu and LQToBEle example cards, using only NNPDF31_LO PDF and CMSSW_9_3_8 for production

@DryRun @jotadram6 @efeyazgan


efeyazgan:
Why a particular CMSSW version but not 9_3_X?

DavidMorse:
I just mean that I made the gridpacks with 9_3_8, as is the default in the gridpack creation script from the Twiki. Let me know if that doesn't answer the question

efeyazgan:
That's OK.

Issue #2237, opened 2019-05-27
MG261:  Cluster handling via MG plugin


qliphy:
I would like to come back to the discussions in
https://github.com/cms-sw/genproductions/issues/1869
i.e.   to port https://github.com/cms-sw/genproductions/pull/1662 into mg261 branch

However, I got this error message after adding #1662 into mg261 branch, while running with a simple example with ./submit_condor_gridpack_generation.sh  wplustest_4f_LO cards/examples/wplustest_4f_LO/:

"root: cms_condor is not recognized as a supported cluster format."

@AndreasAlbert @kdlong @agrohsje   @soarnsoar   Can you have a look?

This is related to 
https://hypernews.cern.ch/HyperNews/CMS/get/generators/4340.html

qliphy:
As Andreas correctly pointed out,  the error message is thrown around line 3120 of madgraph/interface/common_run_interface.py, i.e. in MG5, not in CMS scripts.

After further tracking down, it seems at Line 1873 [1] in madgraph/various/misc.py,  plugin_path is empty and this seems to trigger the problem.

qliphy:
Workaround provided by #2262 and #2263

PR #2262, opened 2019-06-13
Mg261 Condor Support

qliphy:
https://github.com/cms-sw/genproductions/issues/2237

### DIFF
--- a/bin/MadGraph5_aMCatNLO/PLUGIN/CMS_CLUSTER/__init__.py
+++ b/bin/MadGraph5_aMCatNLO/PLUGIN/CMS_CLUSTER/__init__.py
@@ -1,0 +1,751 @@
+## import the required files
+# example: import maddm_interface as maddm_interface # local file
+#          import madgraph.various.cluster as cluster #MG5 distribution file
+
+from madgraph.various.cluster import *
+import time
+
+#local file
+
+################################################################################
+# Copyright (c) 2009 The MadGraph5_aMC@NLO Development team and Contributors             
+#
+# This file is a part of the MadGraph5_aMC@NLO project, an application which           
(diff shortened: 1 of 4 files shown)


soarnsoar:
Hi Qiang. I found that this PLUGIN/CMS_CLUSTER is not copied to mgbasedir when using submit_cmsconnect_gridpack_generation.sh

I don't know what causes this problem yet.

qliphy:
@soarnsoar   Can you give me a log for this?

qliphy:
@soarnsoar  It seems with cms_connect we still have the issue "root: cms_condor is not recognized as a supported cluster format."  A quick workaround is to change in gridpack_generation.sh from "cms_condor" to "condor" in several places.   In the meantime, we have a further check for cms_condor support .

soarnsoar:
@qliphy , 
You can find the logfile here :

/afs/cern.ch/work/j/jhchoi/public/logfiles/mg261_dy0j_lo_cmscluster_Cp/


But I don't understand why this line 

https://github.com/cms-sw/genproductions/blob/0f4c6a90737f2424c3805494057b98343f9bbbdb/bin/MadGraph5_aMCatNLO/gridpack_generation.sh#L121

doesn't work when running submit_cmsconnect_gridpack_generation.sh.

When I ran the gridpack gen with "gridpack_generation.sh"at cms connect, the "CMS_CLUSTER" directory was copied properly. But when with submit_cmsconnect_gridpack_generation.sh, it wasn't.


This is what I observed. I haven't investigaed it further yet sorry. But please note that this symptom is only for "submit_cmsconnect_gridpack_generation.sh" script.

qliphy:
@soarnsoar    Should be solved with #2284 and #2285 
Would you please have a test? Thanks!

PR #2265, opened 2019-06-13
fixed syntax in critical madspin patch (0015) to make sure all decay …

agrohsje:
…me's are properly generated

### DIFF
--- a/bin/MadGraph5_aMCatNLO/patches/0015-fix-critical-madspin-bug.patch
+++ b/bin/MadGraph5_aMCatNLO/patches/0015-fix-critical-madspin-bug.patch
@@ -1,7 +1,7 @@
 === modified file 'MadSpin/decay.py'
---- ./MadSpin/decay.py	2017-12-02 01:16:01 +0000
-+++ ./MadSpin/decay.py	2018-01-11 10:40:04 +0000
-@@ -2746,37 +2746,8 @@
+--- ./MadSpin/decay.py	2019-06-14 00:17:19 +0200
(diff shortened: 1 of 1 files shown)


agrohsje:
Hi @ckmackay . This PR should fix the madspin problem in 2.6.1. I will try to redo your gridpacks:
https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TGGJets_leptonDecays_4f_ckm_LO
and let you know.

qliphy:
@agrohsje   Thanks a lots!  Let's wait for your test result before merging this.   Please also make a same PR to mg261_UL branch.

agrohsje:
@ckmackay @qliphy @efeyazgan I tested this PR with Kates gridpacks and everything is working well. There were some patch complaints which I just fixed to avoid confusing output. 
One remark. I noticed that using these cards https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TGGJets_leptonDecays_4f_ckm_LO the pilot run fails to generate enough events, or better say, not the required number of events. This is because of singularities in the setup, e.g. missing dR cuts. At the same time, some dR cuts are already 0.4 i.e. as on detector level. Is that intended? 
Anyway, these points are independent from the actual problem, my PR. Should I merge?

qliphy:
Thanks @agrohsje   Can you also update mg261_UL branch accordingly?

PR #2266, opened 2019-06-13
Cards for NMSSM resonant production of Higgs + scalar (X -> YH -> bb bb)

l-cadamuro:
This PR contains the cards to generate the NMSSM signal X -> YH -> bb bb.
The free parameters of the model are mX and mY.
Folders containing the cards for each of the (mX, mY) pair are generated with the script generate_grid.py (the grid currently listed in the script is for test only, more points are generated for the central request).
The script simply replaces the mX and mY values in the cards contained in the Template folder.


qliphy:
Please remove these lines: https://github.com/cms-sw/genproductions/blob/7bca911805bd22a9f4641951a099457014785f13/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_bbbb/Template/run_card.dat#L269-L281
as we mentioned here: https://twiki.cern.ch/twiki/bin/view/CMS/Fall18Repack

l-cadamuro:
Thanks, this is now removed with the last commit.

PR #2273, opened 2019-06-20
modify the permission of runcmsgrid_LO.sh to executable

soarnsoar:
Hi. The runcmsgrid_LO.sh is not executable now. I changed the permission.


qliphy:
thanks!  Can you also commit to UL2019 branch?

PR #2276, opened 2019-06-20
UL2019 for powheg

yuanchao:
Hi,
I take the latest run_pwg.py and the corresponding scripts from master branch. The make_rwl.py script is updated to have the new NNPDF31 0118 mc_hessian sets available.

### DIFF
--- a/bin/Powheg/make_rwl.py
+++ b/bin/Powheg/make_rwl.py
@@ -1,0 +1,170 @@
+#!/usr/bin/env python
+
+'''
+Script for POWHEG weighting calculation
+By Yuan CHAO  08/06/2016
+'''
+
+import sys
+import os
+
+if len(sys.argv) < 3:
+    print """\
+ERROR: Please specify if the Flavor scheme for which you want to define the weights is 5F (1) or 4F (0), and the central PDF
+Example of usage for 5F:  python make_rwl.py 1 306000
+Example of usage for 4F:  python make_rwl.py 0 320900
+"""
+    sys.exit(1)
+
+is5FlavorScheme = str(sys.argv[1])
+CentralPDF = str(sys.argv[2])
+forDYNNLOPS = str(sys.argv[3]) if len(sys.argv) > 3 else 0
+
+# is5FlavorScheme = True
+# CentralPDF = 306000
+# processes4Flavor = ['ST_tch_4f', 'bbH', 'Wbb_dec', 'Wbbj',]
+
+# if processName in processes4Flavor:
+  # is5FlavorScheme = False
+  # CentralPDF = 320900
+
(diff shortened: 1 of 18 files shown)


qliphy:
thanks @yuanchao  !
@agrohsje @covarell   Can you have a double check before merging this PR?

agrohsje:
@yuanchao @qliphy I just checked the request. While it looks generally good, I have a few follow-ups to make sure we don't miss anything. 
1.) These patches are different from the one in master. Is there a reason not to adjust them? 
./bin/Powheg/patches/pdfweights.patch
./bin/Powheg/patches/pwhg_lhepdf.patch
./bin/Powheg/runcmsgrid_powhegjhugen.sh
./bin/Powheg/runcmsgrid_powheg.sh
2.) Should we add JHUGen: runcmsgrid_template_VBFoffshell.sh as well, to make sure MC contacts have it available? 
3.) I assume we don't want to modify Powheg/Jhugen from the UL branch but do it via master only? If so the following missing files should be no problem. 
/bin/JHUGen/install_DM.py install.py
/bin/Powheg/source/ -> don't do it in ul
/bin/JHUGen/runcmsgrid_template_DM.sh
/bin/JHUGen/runcmsgrid_template.sh
4.) The examples for Powheg differ compared to master. I don't know if MC contacts are considering them or not? 
5.) In UL we have no subdirs in production. Do we care about having  16/17/18 or not. I assume having none is fine. Same holds for /bin/JHUGen/cards
6.) I would recommend to remove 
bin/Powheg/runcmsgrid_powheg_hdampgrid.sh
to avoid mistakes. 
7.) About PDFs. Should we harmonize the order of weights between mg5 and powheg for UL?
What is the idea behind the variation splitting in hessian and replicas in pdf variations? I cannot connect the name with the given PDFs. What was the feedback from Juan on 4F pos. version prospects? I thought he send an email about that, but I cannot find it.

qliphy:
Thanks @agrohsje 

UL2019 branch is cloned from mg261_UL, while mg261_UL is based on mg261 branch.  However, the powheg scripts in mg261 branch is already quite decoupled and different from the master.

So we should first update UL2019 branch to have the Powheg scripts matched with master branch.  

What do you think?

agrohsje:
HI @qliphy . In fact this is what I did. I checked the Powheg scripts from Yuan and compared to the master. That's how I realized that the branch would miss some important updates where I think the patches are the most relevant.

qliphy:
@agrohsje @yuanchao 
Please check https://github.com/cms-sw/genproductions/pull/2283
which basically update powheg scripts to master branch

Note "UL2019" branch was suggested at last week, as the name "mg261_UL" is confusing and seems to only for madgraph.

If @2283 is merged, @yuanchao  please commit same changes as here to UL2019 then.  Thanks!

qliphy:
#2283 is merged now.
@yuanchao   Would you please commit changes to UL2019?  Thanks!

qliphy:
You can open a new PR.

PR #2280, opened 2019-06-21
MonojetLQ cards and model added

alpakpinar:
This pull request is for Monojet leptoquark model, which was discussed in MC&I meeting on last Monday, June 17.


AndreasAlbert:
(Note: This PR contains the cards for one parameter point + template cards and scripts to generate all other points)

PR #2284, opened 2019-06-24
Add Plugin for CMSConnect for MG261

qliphy:
Add Plugin for CMSConnect for MG261


qliphy:
See https://github.com/cms-sw/genproductions/pull/2262

PR #2285, opened 2019-06-24
add Plugin for cmsconnect

qliphy:
add Plugin for cmsconnect


qliphy:
https://github.com/cms-sw/genproductions/pull/2262

PR #2288, opened 2019-06-24
update PDF sets for UL2019

yuanchao:
Adding 325300, 325500, and 30600 PDF sets.

### DIFF
--- a/bin/Powheg/make_rwl.py
+++ b/bin/Powheg/make_rwl.py
@@ -68,6 +68,8 @@ if int(is5FlavorScheme) == 1:
             # weight id, LHAPDF id, name, replicas to be written
             "PDF_variation1 , hessian" :
             [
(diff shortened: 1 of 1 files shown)


qliphy:
Thanks @yuanchao     325300, 325500, and 30600 PDF sets are currently only available with CMSSW_10_6_X.  Probably we should make this clear at some place, e.g. https://twiki.cern.ch/twiki/bin/viewauth/CMS/PowhegBOXPrecompiled

agrohsje:
Hi @qliphy 306000 is available since long. Only  325300, 325500 were added, no?

qliphy:
@agrohsje   Yes, but there seems to be re-ordering of 306000 here.

agrohsje:
@qliphy I was referring to the TWiki warning. There only  325300, 325500 should be mentioned, no?

qliphy:
@agrohsje   Yes, right.  In addition, we should also mention the CMSSW release for UL2019 should be 10_6_X

PR #2289, opened 2019-06-25
DYNNLOPS updates, with parallelstage fixes

intrepid42:


covarell:
May also fix the problems observed in:
https://hypernews.cern.ch/HyperNews/CMS/get/generators/4375.html
Veryfing now.

agrohsje:
@covarell Yesterday, I tested the gridpack production for melu and it works if I fix the input card. I didn't touch the code. So that shouldn't be related. I will reply asap.

covarell:
Thanks. I can confirm this.

covarell:
So we can merge this? The rest is DYNNLOPS-specific so I assume it is validated by @intrepid42 ? Or you prefer to wait and commit more on top?

qliphy:
@efeyazgan @agrohsje   This looks good to me.  You can go ahead and merge this.

intrepid42:
Hi, I quickly checked that https://github.com/cms-sw/genproductions/pull/2289/commits/b813d75db35995491779face50da06925c2c83a1 indeed did not break anything, so this should work for everyone

covarell:
We have a complaint from @lviliani about this.
"it seems that folderName is needed here, otherwise this breaks the run_pwg_parallel_condor.py script 
https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/run_pwg_condor.py#L60-L62 "
Can you please check?

intrepid42:
Puh, what command exactly is he using and how is it failing? I know that parallelstage was broken before when we set `runInBatchDir = True` and that is fixed here.

Looking at the code, `execname = folderName + '/run_' + tag` might now be broken when `runInBatchDir = False` because we set `f.write('initialdir              = ' + rootfolder + '/' + folderName + '\n')` afterwards :s
But I may also be mistaken and all the paths need to be defined from `rootfolder`, so it would be good to get hands at the error message, and the condorConf, please!

lviliani:
Hi, I noticed the problem when trying to make a gridpack for WWJ NNLOPS.
The command I used is:
`python ./run_pwg_parallel_condor.py -i examples/V2/WWJ_NNLOPS_13TeV/WWJ_NNLOPS_13TeV.input -m WWJ -f my_WWJ -q workday -j 10`
I don't have the old log files now, but I fixed the problem in my local script by adding `folderName` here:
https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/run_pwg_condor.py#L61-L63
or at least this change seems to work because jobs are running now (but maybe there's another reason hidden somewhere).
Btw the problem was that the .out, .err, and .log files got written in `rootfolder` instead of in `folderName` (that was the path written in the condorConf).

lviliani:
But I didn't try with this PR, so maybe you fixed it already in a different way.

intrepid42:
> maybe you fixed it already in a different way

Yes, indeed... https://github.com/cms-sw/genproductions/pull/2289/commits/5b8c8862031c904b7c679b21976e971bedc23c47

PR #2295, opened 2019-07-03
fix wrong tarball and max weights

covarell:
Fix wrong tarball (should be deleted from AFS area) and augment number of maximum weights: at the moment there are 1284 in UL2019.


covarell:
Urgently needed for UL2019 as reported in:
https://hypernews.cern.ch/HyperNews/CMS/get/generators/4374/1.html
and followups.

PR #2299, opened 2019-07-06
udapted patch 0004 for mg5.265

agrohsje:


agrohsje:
This PR is to adjust  patch 0004 for 5.265 and to add a new patch 0003 which avoids an error message in reweighting which is tagged as crucial but a print out bug in MG5_aMC@NLO. The printout is removed in 5.266, so patch 0003 can be removed in future branches.

qliphy:
Thanks, looks good to be merged.

PR #2300, opened 2019-07-08
not enough events check

efeyazgan:


efeyazgan:
@agrohsje please take a look at the new warning message.

bin/utils/request_fragment_check.py
```diff
@@ -479,6 +480,17 @@ def root_requests_from_ticket(ticket_prepid, include_docs=False):
                             dn = dn + "-amcatnlo"
                         if matching_c == 3:
                             dn = dn + "-amcatnloFXFX"
+                gp_log_loc = my_path+'/'+pi+'/gridpack_generation.log'
+                if mg_gp is True or amcnlo_gp is True and os.path.isfile(gp_log_loc) is True:
+                    not_enough_events.append(os.popen('grep \"saving rejects to\" '+gp_log_loc).read())
```
agrohsje: This error message appears when a patch didn't work. In principle we should fix such problems before releasing a version, but still I see it from time to time in gridpacks, so we should make ourself aware and avoid such problems. So this would be a separate thing and independent from not enough events in pilotrun.
For Madspin you might extend the warning and specify that this problem affects Madspin.

efeyazgan: Now they are split but what shall we suggest the contacts for the above case? To re-make the gridpack? thanks.

agrohsje: Hi @efeyazgan , I would call this failing patch or so. It is totally unrelated to not having enough events. 
If a gridpack has problems with applying patches, the MC contact should understand if the patch is relevant for this gridpack (i.e. a madspin patch will be irrelevant  when deay is done in the process card) and inform us immediately so that we can fix it.

PR #2306, opened 2019-07-12
Port https://github.com/cms-sw/genproductions/pull/2289 to UL2019

covarell:
Critical fix by @intrepid42 ported to UL branch (the full PR is ported, so DYNNLOPS specific changes are also in, which I think is good...)

Markus, the PR as is will create a discrepancy w.r.t. default PDFs in DYNNLOPS and default (pos-def) PDFs for other samples, please check whether this is correct.


covarell:
Sorry I cannot get the title nice :-)

PR #2314, opened 2019-07-19
JHUgen updates and fixes for VBFoffshel production

skyriacoCMS:


qliphy:
Adding @hroskes

hroskes:
Hi @qliphy, I was working with Savvas on putting these gridpacks together, and as far as we can tell they're correct.  @qyguo is preparing the requests.

skyriacoCMS:
@qliphy Thank you

PR #2319, opened 2019-07-22
NMSSM_XYH_bbbb for Run2016

fravera:
Added the MG242 datacard for NMSSM_XYH_bbbb 
datacards are generated using generate_grid.py which edits datacard templates in the Template folder


qliphy:
Some comments as below:

1. Why do you upload .gitignore?
2. It is a LO non matching process, so icckw should be 0.
3. Please remove https://github.com/cms-sw/genproductions/blob/9e055e5a4008a7f37b8696b6688dd9dd96d29f6c/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_bbbb/Template/run_card.dat#L272-L285

fravera:
I uploaded the requested changes, I also matched some of the parameters to be the same of the folder in the Master branch for the 2017-2018 productions.
I noticed that there are some flags that are present in 2016 and not in 2017-2018 and viceversa:
in 2016 but not in 2017-2018
1 = highestmult
1 = ktscheme
False = pdfwgt

in 2017-2018 but not in 2016
None = bias_module
{} = bias_parameters
average = event_norm 
1,2,3,4,5,6,21,1000001,1000002,1000003,1000004,1000005,1000006,1000021,2000001,2000002,2000003,2000004,2000005,2000006 = pdgs_for_merging_cut
-1 = ptlund
Is it fine?

I also changed  292000 = lhaid as suggested by the MC contacts

Many thanks,
Fabio and @l-cadamuro

qliphy:
The difference you mentioned indeed are not 2016 vs 2017/2018, but between MG242 and MG26X,  
As you are committing to mg242leagcy branch, you don't need to add those lines from MG26X, e.g.  bias_module, bias_parameters...

PR #2324, opened 2019-07-23
Cards for WJetsToQQ_HT-1200toInf for Run3

ph2632:


qliphy:
Please set bwcutoff  to a smaller number like 25
pdfwgt should be F:  https://twiki.cern.ch/twiki/bin/view/CMS/Moriond18MC

PR #2325, opened 2019-07-25
 Cards for gridpack 2017/2018 production of Top 3/2 BSM model 

karavdin:


qliphy:
T = cut_decays ! Cut decay products 

Do you really want to cut on decay products?

karavdin:
@qliphy, there are cuts on min jet and photon pt, which I would like to keep for consistency with 2016 gridpacks.

PR #2329, opened 2019-07-30
Add special TT4L samples for HIG-19-001 and HIG-19-009

usarica:
These are special samples needed for HIG-19-001 and HIG-19-009 as they proceed to approval/approval reprise:
- proc_cards are essentially the same except for the generate (or add process lines)
- run_cards and customizecards are identical.
- Making this PR myself after discussion with HZZ conveners and the HZZ MC contact.
- The preference is to request these samples for 2016-2018 with the same (NNPDF3.0) configurations in order to be consistent with the rest of the analysis elements, but I don't know exactly how to do that.

Please let me know if anything needs to be changed, and I will be happy to make the commit.

cc: @qyguo

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TT2LX_4LFilter_5f_LO/TT2L2Nu_4LFilter_5f_LO_customizecards.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TT2LX_4LFilter_5f_LO/TT2L2Nu_4LFilter_5f_LO_customizecards.dat
@@ -11,7 +11,7 @@ set param_card decay 24 2.085
 set run_card use_syst True
 #set run_card pdlabel lhapdf
 set run_card lhaid 263000
-set run_card bwcutoff 15000
+set run_card bwcutoff 15
 set run_card ickkw 1
 set run_card ktscheme 1
 set run_card asrwgtflavor 5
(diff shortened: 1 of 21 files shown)


usarica:
Also adding @hroskes who provided the instructions and is the HIG-19-009 editor

qliphy:
Please remove these lines: https://github.com/cms-sw/genproductions/blob/eb8e8d4dd5fe8eba4e7796395112dd78c1cc4a09/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TT2LX_4LFilter_5f_LO/TT2L2Nu_4LFilter_5f_LO_run_card.dat#L274-L280

And set bwcutoff to a smaller value as 15. 


It is possible to use a custom PDF set (e.g. for 2016 production), please refer more to
https://twiki.cern.ch/twiki/bin/view/CMS/Moriond18MC#PDF_Choice_for_2017_production_f

usarica:
@qliphy Do these changes satisfy your request, or should I explicitly change the run_cards for the pdf id explicitly?

qliphy:
I found bwcutoff is set very large in the customizecards files:
 set run_card bwcutoff 15000

It would be nice if you can move those run_card settings from customizecards.dat to run_card.dat

usarica:
Oops, you are right, I fixed it now.
The run_card setting and customizecards should be the same (at least now) except for the lhapdf id.

qliphy:
ickkw should be 0 instead of 1, as your process is not a mlm matching one but just LO.

I would suggest you to move all the run settings to run_card.dat instead of customizecards

usarica:
Ok, I think I migrated all of them. Do you think that is ok now?

PR #2330, opened 2019-07-30
WWZTo4L, tau included

menglu21:
input cards for WWZ/WZZ/ZZZ to 4L, the previous sample not include tau, added now.


qliphy:
It seems you miss to add "True = store_rwgt_info"

menglu21:
> It seems you miss to add "True = store_rwgt_info"

added; the processes have been modified to at most one additional jet, so ickkw is changed to 3

PR #2333, opened 2019-08-01
UL17_18_16 gen settings consistency check

efeyazgan:
This should only be merged when all HTCondor nodes get a cookie! (hopefully early next week)


qliphy:
https://github.com/cms-sw/genproductions/blob/3fce7e5306d1819560ae26997a6214c1cb1cf49e/bin/utils/request_fragment_check.py#L435

This seems to be buggy, and different from master branch

qliphy:
Should be 
https://github.com/cms-sw/genproductions/blob/master/bin/utils/request_fragment_check.py#L395

efeyazgan:
@qliphy, thanks. I've updated the PR but this is still not ready to be merged.

qliphy:
Removed the check on 16 core with low filter eff,  after the discussion with PPD:  
https://indico.cern.ch/event/841756/

qliphy:
Justinas updated on Aug 15th that "cern-get-sso-cookie seems to be available in HTCondor"

PR #2335, opened 2019-08-03
fix f2py compiling issue under SL7

qliphy:
https://hypernews.cern.ch/HyperNews/CMS/get/generators/4411/1/2/1.html


qliphy:
Under SL7,  there are different versions of f2py (used for MG reweighting to make I/O fast),  the default one seems to be f2py3.6 which triggers python3, while MG26X is not ready with python3 yet.

Currently a quick and simple hack:  by setting f2py to f2py2.7

There might be other more elegant way.  Comments are welcome.

qliphy:
@grasph @lviliani @covarell @efeyazgan @agrohsje

efeyazgan:
This solution looks good. thanks.

grasph:
Perfect. Thanks Qian.

Philippe.

PR #2346, opened 2019-08-14
added datacards for 2b2tau final state

amandipde:


qliphy:
Output directory name should be consistent with the proc card name:

https://github.com/cms-sw/genproductions/blob/460b5b754be4e10da8ba58bd53b0d5ead280836f/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/bbtautau_final_state/Wh01_M125_Toa01a01_Tobbtautau/Wh01_M125_Toa01a01_Tobbtautau_proc_card.dat#L15

amandipde:
Hi,
This directory I am using as temporary to create dir. on the basis of different mass points where the name are compatible. 
`https://github.com/amandipde/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/bbtautau_final_state/createCards_Wh01_M125_Toa01a01_Tobbtautau_2017.sh#L8`

qliphy:
ok then. Thanks!

PR #2352, opened 2019-08-19
updated cards  for 2016 wprime wz whad zhad

IreneZoi:
with extra-models cards and script to produce the mass points.
Old cards were in:
https://github.com/IreneZoi/genproductions/tree/mg240/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson/Spin-1/Wprime_WZ_WhadZhad


qliphy:
You set 5f in proc card:   define q = u c d s b u~ c~ d~ s~ b~
while in run card, you have 4f PDF, 263400,
it is better to make them consistent.

IreneZoi:
I had the same also in the already merged PR #2349 & #2350 
Should I update it there too?

qliphy:
Yes, please.  It was overlooked for the other two PRs.

PR #2359, opened 2019-08-26
DY(mumu)gamma cards in madgraphLO  with dimuon mass 20-110GeV

afanfani:


qliphy:
Please remove the banner in the proc_card.dat. Otherwise the cards look good.

afanfani:
the banned has been removed from DYGTo2MuG_mll_20-110_pta10_LO proc cards

PR #2362, opened 2019-08-28
Make NLO EW same-sign WW vector-boson scattering available in CMS

covarell:
From https://arxiv.org/abs/1906.01863 (Pellen et al.).

Only tested in SLC7, other tests ongoing.  Please have a look and comment but wait a moment for merging.

@kdlong @arapyan @lviliani


govoni:
Ciao @covarell, this is very timely! 
I have a curiosity: how did you set the usual poweheg parameters, i.e. ncall1, itmx1, ncall2, itmx2, fold*, etc.?

covarell:
We are not at that stage yet, I am checking that the process can indeed generate events.
At the moment you can see we take the default in the package that is set to very low values (ncall=1000), not OK for real runs.

covarell:
Sorry forgot to include @yuanchao

covarell:
This should be OK now for merging, it also contains a couple of minor fixes as requested in . https://hypernews.cern.ch/HyperNews/CMS/get/generators/4447.html. 

We received the suggestions in the last commit on the POWHEG parameter setting from M. Pellen, which I am testing now. Technically now it should be fine.

arapyan:
Hi Roberto, if I remember correctly Matthieu mentioned that the code was only producing the emu final state. He said the other final states (ee,mm) will be included soon. Are these already included in this setup you are using?

covarell:
Yes, it is included. See settings in the input card:
https://github.com/cms-sw/genproductions/pull/2362/files#diff-d1466e525517182c586386d253a588aaR3

PR #2372, opened 2019-08-30
Fix false hold job warning

grasph:
When generating gridpack with the cms_condor queue I'm getting a warning  "WARNING: Failed condor job ...". I believe it's a false warning due to a bug in PLUGIN/CMS_CLUSTER/__init__.py. See proposed changes Line 363-364. Other changes are improvement of warning messages.


qliphy:
Thanks.  looks good.

Adding @AndreasAlbert  who wrote the plugin for a double check.

qliphy:
Thanks!  Can you also make a PR to master and mg265 branches?

PR #2374, opened 2019-08-30
adding Suu_Diquark model to mg265

evahal:
Adding Suu_Diquark model for dijet and paired dijet analysis.  mg265 should be used.


qliphy:
These lines are not needed:
https://github.com/cms-sw/genproductions/blob/7e9923bf711a879fcbe11be1aa070bc76c581a67/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Suu_Diquark/Suu_Diquark_S8000_chi1900/Suu_Diquark_S8000_chi1900_run_card.dat#L278-L281

as the module systematics will be called later:
https://github.com/cms-sw/genproductions/blob/mg265/bin/MadGraph5_aMCatNLO/runcmsgrid_LO.sh#L171-L175

qliphy:
Param card is not mandatory

Instead you can consider a customize card
https://twiki.cern.ch/twiki/bin/view/CMS/QuickGuideMadGraph5aMCatNLO#Preparation_of_the_cards_Mostly

PR #2376, opened 2019-08-31
new d and MD points

yguler:
ADD model generator fragment for EXO with new d and MD points.


qliphy:
Please only add 1-2 example fragments, as the others are quite similar.

efeyazgan:
Hi @yguler 
You shouldl not use PDF:pSet = LHAPDF6:NNPDF31_lo_as_0130 with tune CP5.  
If you want to use CP5, then you should use NNPDF31_NNLO_as_0118
or if you want to use NNPDF31_lo_as_0130, you should use CP1 or CP2.

efeyazgan:
@yguler thanks for fixing. 
I leave the rest to @qliphy and @agrohsje.

qliphy:
This PR seems to mix others.  Can you make it clean?

agrohsje:
As there was no follow-up and the PR pushed >800 files I will close.

PR #2380, opened 2019-09-02
TT_T2ZJ_aTleptonic_kappa_zct_LO cards for 2016

ckmackay:
TT_T2ZJ_aTleptonic_kappa_zct_LO cards for 2016
Set 
F = pdfweight
is that correct for 2016?


qliphy:
pdfwgt is better to set to false.

Please just add one signel line " F = pdfwgt"  instead of two different lines on "pdfwgt"

ckmackay:
Done, thanks

PR #2396, opened 2019-09-17
remake inclusive LO(MLM matching) DY gridpack for herwig7 showering

menglu21:


qliphy:
Is there any difference from other cards? e.g. 
https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/DYJets_HT_LO_MLM/DYJets_HT_mll50

menglu21:
> Is there any difference from other cards? e.g.
> https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/DYJets_HT_LO_MLM/DYJets_HT_mll50

the only difference is set pdfset=263000 explicitly and remove '$DEFAULT_PDF_MEMBERS = reweight_PDF'

qliphy:
But why do you want to change to 263000 (NNPDF30_lo_as_0130)  instead of the default PDF for 2017/2018?  Note for Herwig7 CH3 Tune,  the default PDF is NNPDF3.1 NNLO

menglu21:
> But why do you want to change to 263000 (NNPDF30_lo_as_0130) instead of the default PDF for 2017/2018? Note for Herwig7 CH3 Tune, the default PDF is NNPDF3.1 NNLO

sorry for forgetting to mention this is for 2016, for 2017/2018, i will use the current setting.

bendavid:
Why can't the existing gridpacks already used for madgraph+pythia be used with Herwig?

menglu21:
> Why can't the existing gridpacks already used for madgraph+pythia be used with Herwig?
when i use the existing gridpack for local run, it returns error:
'grep: /tmp/melu/gridpacks//SMP-RunIISummer15wmLHEGS-00338/mgbasedir/Template/LO/SubProcesses/refine.sh: No such file or directory
Traceback (most recent call last):
  File "request_fragment_check.py", line 852, in <module>
    MGpatch.append(int(os.popen('grep -c _CONDOR_SCRATCH_DIR '+my_path+'/'+pi+'/'+'mgbasedir/Template/LO/SubProcesses/refine.sh').read()))
ValueError: invalid literal for int() with base 10: ''
GEN Request Checking Script returned exit code 1 which means there are 1 errors'

since the gridpack was generated using MG2.2.2(2016 sample), i'm not sure if it contains all the necessary file that the request check script required.

qliphy:
Yes, gridpack with MG2.2.2 is too old.

However,  for Herwig7, as mentioned above, one should always use the CH1 or CH3 tunes  (you can check GEN-17-001).  So by default, one should use NNPDF3.1 NNLO.   This is not the same as for Pythia8, for which we have CUETP8M1 before CP5.

And also note, Herwig7 is not yet available in 7_1_X.   @Andrej-CMS  may comment more here.

Andrej-CMS:
@qliphy @menglu21 
As Qiang said correctly, Herwig7 is not available in CMSSW_7_1_X. The reason is that it doesn't compile in gcc_481. I have notified the authors and they are having a look, but it's unclear how long this will take or whether they will make Herwig7 available for such an old gcc version.

menglu21:
ok, thanks for noticing the authors. i discussed with the analyzers, they said they will go with 2017/18, so no need to produce the 2016 at present.

PR #2414, opened 2019-10-01
update Powheg part to be same as in master branch

qliphy:
rm -r .../Powheg
cp -r  fromMasterbranch/Powheg .
git add -A * 
git commit -m "update Powheg part to be same as in master branch"
git push origin mg265


qliphy:
@covarell @agrohsje   Would you please have a check?
This PR is to update Powheg part in mg265 branch to be the same as in master branch.

(We plan to switch mg265 to be the new master soon)

covarell:
It looks good to me!
I wonder if we should take the chance to remove completely run_pwg.py and run_pwg_parallel.py, which depend on LSF queues, and are anyway very outdated. AFAIK there is no CMS site left which still supports that batch system.

qliphy:
Ok, I will let this to be decided later.
I will merge this PR now.   And you can then update the PDF as suggested by Alexander yesterday.

PR #2415, opened 2019-10-01
update QCD ZG

menglu21:
the previous gridpack was produced using mg261, it was found that maxflavor=5 was overwritten by maxflavor=4. remade the gridpack in mg260, the runcard under directory 'process' seems correct now.


qliphy:
@menglu21   Can you explain a bit more why maxflavor gets overwritten in mg261?  and not in mg260?

menglu21:
> @menglu21 Can you explain a bit more why maxflavor gets overwritten in mg261? and not in mg260?

unfortunately, i can't reproduce the buggy gridpack, i tried with mg261 and the new gridpack is fine.

agrohsje:
Will close due to too many conflicts and inactivity. Please make a new PR if needed. Thanks so much!

PR #2416, opened 2019-10-01
Fix default PDFs for POWHEG in UL2019 branch

covarell:


qliphy:
Thanks @covarell   
Would you please also update the master and mg265 branch if necessary?

covarell:
I do not know if it is necessary in fact...
If we are keeping three branches exactly up-to-date with each other (including small details like default PDFs) then why do we need three branches?  :-)

qliphy:
@covarell  Right.  Buf or mg265 branch we will also use it for UL production.  So it is better to update PDF there similarly.  Can you take care of this?  Thanks!

PR #2417, opened 2019-10-01
update Powheg part to be same in mg265 as in UL2019 branch

covarell:
The PR is here: it is not very good, because there were updates in master that were not propagated to UL2019, so we are going back on a few aspects on mg265 (which is the same as master now) with this PR.

Again (also in view of the massive script update that is coming up from @klong & co.) maintaining three branches in parallel is definitely not possible. Feel free to accept/reject this commit...


qliphy:
Thanks!  As mg265 will be the new master, I think it is better to keep it similar as current master, while just update PDF as done in https://github.com/cms-sw/genproductions/pull/2416/files

qliphy:
Done vis  this commit:   https://github.com/cms-sw/genproductions/commit/f46d2a98123e38c4ae88c640bd2752f13be61825
(Sorry I directly merged it.  Please have a check if I miss anything there)

covarell:
Thanks @qliphy, the commit you point to seems ok to me (I see two deleted files that seem to have nothing to do with POWHEG, on those I cannot comment but it seems they were uploaded by mistake)

PR #2419, opened 2019-10-01
Check OS vs SCRAM_ARCH in Madgraph gridpack generation

DryRun:
A number of analyzers have produced nonfunctional gridpacks due to logging into lxplus7 and making a gridpack with an SLC6 SCRAM_ARCH. This PR adds a simple check to avoid this problem.


agrohsje:
Thanks David. Do you mind modifying gridpack generation in the UL2019 branch accordingly?

DryRun:
Sure, no problem.

PR #2420, opened 2019-10-01
Check OS vs SCRAM_ARCH in Madgraph gridpack generation (UL2019 version)

DryRun:
A number of analyzers have produced nonfunctional gridpacks due to logging into lxplus7 and making a gridpack with an SLC6 SCRAM_ARCH. This PR adds a simple check to avoid this problem.

Port of #2419


agrohsje:
Thanks!

PR #2423, opened 2019-10-02
Add cards+bias module for Z'(qq) w/flat mass and pT for tagger training

DryRun:
Add cards+bias module for Z'(qq) w/flat mass and pT for tagger training.


qliphy:
Thanks.  It would be interesting to hear a report on this bias sample once it is done.

PR #2424, opened 2019-10-06
14TeV Zmm mass bins: m120-200 and m200-400

afanfani:
14TeV Zmm in mass bins: powheg cards for m120-200 and m200-400 needed for Run3

### DIFF
--- a/bin/Powheg/production/2017/14TeV/DYToLL_NNPDF31_14TeV_MllBinned/Z_mumu_NNPDF31_14TeV_M_120_200.input
+++ b/bin/Powheg/production/2017/14TeV/DYToLL_NNPDF31_14TeV_MllBinned/Z_mumu_NNPDF31_14TeV_M_120_200.input
@@ -37,6 +37,4 @@ iseed    SEED    ! initialize random number sequence
 mass_low 120       ! M Z > mass_low in GeV
 mass_high 200    ! M Z < mass_high in GeV
 
-pdfreweight 0       ! PDF reweighting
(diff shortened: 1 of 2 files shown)


qliphy:
@covarell @yuanchao @agrohsje  Can you have a look? Thanks!

agrohsje:
Looks good to me. ndns lines could be removed. For my understanding: Why are systematic weights turned off?

afanfani:
Concerning the systematic weights: I started from existing cards at 13TeV corresponding to samples currently used in MUO Pog (btw also in UL). I only changed energy and PDF. 
In the context of Run3 studies I believe it’s not relevant. 
As I said, similar cards are used to produce the tarballs in UL.

agrohsje:
Unless we are sure that nobody else will ever use these samples for physics I am fine, otherwise I would recommend including LHE weights. Just to be sure samples are useful for all potential analyzers.

afanfani:
Given the limited scope of Run3 I believe it’s fine without LHE weights. 
This pull request refers to Run3 so it can go.

For the UL it might be a different story. For UL I used existing gridpack and the requests were approved months ago. Some of the mass bins are already produced. Anyway we might agree on producing a new set of these samples with “improved” cards.

agrohsje:
Removed the obsolete lines about systematics. In the new logic they get anyway overwritten. The samples will contain systematic weights. We should remove the confusing lines in runcmsgrid.sh.
Anyway, all good, no need to worry, will merge and close.

PR #2430, opened 2019-10-09
DYJets_HT-40to70 mll50

Andrej-CMS:
Missing HT slice for DYJets_HT mll 50toinf


qliphy:
It is fine to add this missing HT slice card.  However, please note "pdfwgt" may be recommended to set to "False", at lease for UL productions.
https://hypernews.cern.ch/HyperNews/CMS/get/generators/4492/1/1/1.html

PR #2434, opened 2019-10-10
updated aQGC VVjj all-hadronic 2016 Madgraph cards

stalbrec:
Hello,

this updates all 2016 aQGC VV hadronic cards. 
However, there are still two problems. Maybe someone here can help me with these:

- when I'm trying to generate gridpacks with reweight_cards larger than 98304 bytes the [gridpack_generation.sh](https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/gridpack_generation.sh) script exits because of (i think) a inherted exit code from `cat` in the [gridpack_helpers.sh#L67](https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/Utilities/gridpack_helpers.sh#L67) script.
If i comment out the `set -o pipefail` in [gridpack_generation.sh#L703](https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/gridpack_generation.sh#L703) or just use `head` in gridpack_helper.sh instead of `cat` this part of the script runs fine.

- I found that two LHAPDF IDs (292000 and 293000) given on the [PDF@CMS Twiki](https://twiki.cern.ch/twiki/bin/viewauth/CMS/PDFatCMS) correspond to different sets than specified on the [LHAPDF website](https://lhapdf.hepforge.org/pdfsets).
For example:

|                **On the TWiki** |                        |
|---|---|
|                      292000 | NNPDF30_lo_as_0118     |
| **while on the lhapdf website** |                        |
|                      292000 | NNPDF30_nlo_nf_4_pdfas |
|                      262000 | NNPDF30_lo_as_0118     |

I also posted on hn-cms-generators@cern.ch about this last week (see [thread](https://hypernews.cern.ch/HyperNews/CMS/get/generators/4489.html)).

Do you have any suggestions, what I should to about these issues?

Thanks,
Steffen


agrohsje:
Hi Steffen,
I would suggest you to do a PR and modify gridpack_helpers. Using head should be sufficient.It should be one of the first commands. You might add head -20 or similar.
GEN is not really working on the PDF TWiki. I was surprised not to see an answer from PDF people.  You might ping Katerina Lipka or Klaus Rabbertz directly to fix. In general, LHAPDF should be right. You can open the corresponding pdf and check the *.info file to be 100% sure it is what you need.  
If you want to use a specific PDF as default, make sure that the PDF is also included in the variations!
 One final point. Can you try producing a gridpack, untar and do ./runcmsgrid.sh 10 1231 1 and check the cmsgrid_final.lhe having the proper weights and everything as you expect it. 
Cheers, Alexander.

stalbrec:
Hi Alexander,

thanks for the help!
As you maybe already saw, i made the PR with the changes to gridpack_helpers and it got merged.

I also wrote an mail to Katerina and Klaus, and Klaus confirmed, that the mismatch is a simple typo that has now been fixed.
My next question regarding this would be, if the mentions of 292000 on the GEN TWikis (mainly on [Moriond18MC](https://twiki.cern.ch/twiki/bin/view/CMS/Moriond18MC) and [QuickGuideMadGraph5aMCatNLO](https://twiki.cern.ch/twiki/bin/view/CMS/QuickGuideMadGraph5aMCatNLO) are aiming to "recommend" NNPDF30_nlo_nf_4_pdfas or 
NNPDF30_lo_as_0118 for 2016 samples (i.e. was it chosen based on the information from the PDFatCMS TWiki, which were wrong).

Another question I have relates to the flavourscheme: Do I have to set it based on the model I am using (i.e. 5f in this case)?

Regarding the weights in the cmsgrid_final.lhe: I ran it with a few different scenarios (different PDF sets + (no) reweighting) and it looks good to me so far.

Thanks,
Steffen

stalbrec:
Hi Alexander,

thank you very much, that was helpful! 
The bosons decay hadronically in madpsin, so they would lead to bjets. 

A new version of the SMEFT model just got copied to the web repo. 
The new version is in 5f scheme and adds 3 more SMEFT parameters (S2,T3 and T4).
So I will be updating the reweighting cards, the proton definition and the lhaid.

Also I'll have to add EWK and QCD VBS cards similar to the ones from [2017 VVjj-semileptonic](https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/VBS/VVjj_semileptonic). Should I move the aQGC cards and the EWK&QCD cards (once they are ready) in one common sub-directory of the VBS directory?

Cheers,
Steffen

agrohsje:
Any updates on this? A common VBS folder looks good to me. 
Cheers, Alexander.

stalbrec:
Hi Alexander,

sorry for the long silence. 
I've been comparing the SMEFT models in different flavour schemes. Since different flavour schemes result in same results across models (see [slides](https://indico.cern.ch/event/865716/contributions/3647670/attachments/1950928/3238659/VBS-hadronic_steffen.pdf#page=13)) and the newer model includes 10 operators less (and only one additional instead of 3 as stated by me before) we will stick with the old model.

I will update the branch as soon as possible (with correct PDF set and everything moved into common VBS folder).

stalbrec:
From my site, this is ready to get merged.

stalbrec:
ping @agrohsje

agrohsje:
Please check your madspin card: 
decay w- > j j
decay w- > j j
There are redundant lines. 
As you are mixing 4fs and 5fs for proton,jets. What is the reason to pick 4 flavor scheme for 
 asrwgtflavor ?

stalbrecsArchive:
Oh yes, I'll remove the redundant line.
When I changed asrwgtflavor I didn't have the jet definition in mind. I will change asrwgtflavor back to 5.

stalbrec:
Oh yes, I'll remove the redundant line.
When I changed asrwgtflavor I didn't have the jet definition in mind. I will change asrwgtflavor back to 5.

agrohsje:
+1 @qliphy

qliphy:
+1

stalbrec:
ping @agrohsje @qliphy

PR #2435, opened 2019-10-10
Change default SCRAM_ARCH to slc7_amd64_gcc630 (master)

DryRun:
I'm not sure whether this is appropriate, but this PR changes the default `scram_arch` to `slc7_amd64_gcc630` in gridpack_generation.sh. The recently added SCRAM_ARCH vs OS consistency check (https://github.com/cms-sw/genproductions/pull/2421) seems to be causing some confusion, because users don't know that the SCRAM_ARCH can be specified from the command-line argument, and end up with the default slc6 scram_arch on lxplus7. 

Alternatively, the consistency check could tell the user to login to lxplus6 instead.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
+++ b/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
@@ -614,13 +614,16 @@ if [ -n "$5" ]
   then
     scram_arch=${5}
   else
-    scram_arch=slc6_amd64_gcc630 #slc6_amd64_gcc481
+    scram_arch=slc7_amd64_gcc630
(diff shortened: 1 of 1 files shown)


agrohsje:
Hi David. I think there is a small problem with this PR. Non UL-requests can run on slc6 or slc7 machines. For that slc6 is the only choice as slc7 gridpacks will not run on slc6 machines. That's why we only added slc7 as default to the UL branch where we are sure they run on slc7. We should make this clear again in the next MCCM.

DryRun:
Hi Alexander - OK, thanks for the information! I'll close the PR.

PR #2437, opened 2019-10-15
ST tW DS cards for Fall17 and Fall18

ckmackay:


agrohsje:
hi @ckmackay , should we merge or wait for consistent settings? if we keep these settings. it might be worth adding a comment to the input cards pointing out the problem. we should make sure  nobody is using them without being unaware of the inconsistency.

ckmackay:
@agrohsje I added a comment to each file to explain the settings.

PR #2439, opened 2019-10-15
fix typo in PDF checking from cards, replacing pw_pdf with mg_pdf

afanfani:
fix typo for PDF checking in request_checking_script


efeyazgan:
thanks

PR #2432, opened 2019-10-10
update PDF setting in WJetsToLNu cards

harrypuuter:
This is an update of the run Cards for WJetsToLNu according to the recommended PDF settings mentioned in https://twiki.cern.ch/twiki/bin/viewauth/CMS/QuickGuideMadGraph5aMCatNLO#PDF_Choice_for_2017_production

PR #2442, opened 2019-10-15
WJets cards with pdfwgt set to F

aloeliger:
WJets inclusive and jet binned cards with pdfwgt set to F for use in UL2017 based on recent studies.


qliphy:
As discussed here: https://github.com/cms-sw/genproductions/pull/2432
The direct production is preferred.

aloeliger:
Something like that then? Is that the correct way to do handle direct production?

qliphy:
looks good, except this one:
https://github.com/cms-sw/genproductions/blob/909156a3ff68a640529cf76b5f8b4aa1b185ca23/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/WJetsToLNu_pdfwgt_F/W2JetsToLNu_13TeV-madgraphMLM-pythia8/W2JetsToLNu_13TeV-madgraphMLM-pythia8_proc_card.dat

aloeliger:
That should fix the typos I think.

PR #2447, opened 2019-10-18
update WJets and DYJets cards

wang0jin:
Reopen previous pull request with modification of the "auto_ptj_mjj" and removal of a few systematic weight lines.


agrohsje:
Dear @wang0jin @qliphy, the mll min, max cuts were dropped for DY in this PR. Why was this done? Were gridpacks produced with these updated cards?

PR #2448, opened 2019-10-21
Zmumu powheg mass bins: Update PDF for UL and sys weights

afanfani:
Zmumu powheg mass bins: Update PDF for UL2019 and sys weights


agrohsje:
Hi Ale,
thanks for the PR. Would you mind removing these 2 lines:
ndns1 131         ! pdf set for hadron 1 (mlm numbering)
ndns2 131 ! pdf set for hadron 2 (mlm numbering)
It will be overwritten by lhans but removing them would be cleaner. 
Thanks, Alexander.

afanfani:
I've removed the ndns lines from the cards.

PR #2456, opened 2019-10-28
PR from Jihun

Hooooon12:
separated out the hard coded functions into templates
changed optparse to argparse


covarell:
@qliphy @agrohsje this is the famous PR for restructuring the POWHEG scripts.

I will test it immediately for 2-3 processes (especially those that were added after the work from Jihun had already started, like NLO VBS W+W+)

qliphy:
@covarell @Hooooon12   Thanks a lot!

covarell:
I see two main problems, but could not investigate the cause yet:
1) When running with JHUGen.input in the main work directory, it gets removed and the JHUGen generator is not activated.
2) When running step 1 (-p 1 -x something) the files appear in the main work directory and not in the process directory.

covarell:
Hi, so a follow up here: 

1. is a non-problem. In fact someone made the JHUGen.input option configurable via -g and I did not know this.

2. The logic in Template/prepareJob_template.txt is not correct. In this script you cd to the POWHEG work directory, then pushd to the CMSSW_BASE and then popd (so back to the POWHEG work directory). This should not be done because all the output from the various parallel jobs will be written in the same POWHEG work directory overwriting technical files and making jobs fail (especially dangerous for steps 2 and 3).

The correct logic is cd to the POWHEG work directory, then check that cmsenv works OK, then cd back into the main root directory of the running node (outside POWHEG work directory). Here all the executables and .dat files are copied so the job can run independently of others: only at the end of the jobs everything should be copied back to workdir/folderName - this is already correct.

covarell:
Can you make these changes or should I fork the PR and work on it?

Hooooon12:
Hi, sorry for the late reply.
I've fixed the concerned part in prepareJob_template.sh, please confirm that.
(When I ran it personally it seemed to work well)
By the way, I noticed there were conflicts in the latest commit and fixed that in run_pwg_condor.py .

covarell:
Thanks, will test again

covarell:
More comments:
- Here: https://github.com/cms-sw/genproductions/blob/ca2c08ac29033966f9266f0f28a22c1b7281c65b/bin/Powheg/run_pwg_condor.py#L313
The string inputname is not defined, should be the same as the other routines - inputName = folderName + "/powheg.input" (probably a problem introduced in merging the conflict)
- Here: 
https://github.com/cms-sw/genproductions/blob/ca2c08ac29033966f9266f0f28a22c1b7281c65b/bin/Powheg/run_pwg_condor.py#L19
Can we move to the latest version? ( powhegboxV2_rev3691_date20191021.tar.gz  ) Then @intrepid42 can start modifying DYNNLOPS once this PR is merged.

agrohsje:
hi @covarell if the last question about version goes to us, I am fine with it.

covarell:
Thanks that you confirmed! :-) In fact it was a modification request for @Hooooon12

Hooooon12:
Hi, I fixed the conflict here:

https://github.com/Hooooon12/genproductions/blob/4252bd96aaee54165c87145c06a246aca7bf1252/bin/Powheg/run_pwg_condor.py#L306-L308

and updated the source file:

https://github.com/Hooooon12/genproductions/blob/4252bd96aaee54165c87145c06a246aca7bf1252/bin/Powheg/run_pwg_condor.py#L19

please confirm it once again, thanks!

covarell:
Thanks again Jihun, will give another try

covarell:
This is now OK to merge. I will need then to apply a simple patch to pwhg_main rev. 3691 because of a new syntax not compiling in gfortran7 but I can do once this is approved.

qliphy:
Thanks. Now merged.

PR #2459, opened 2019-10-31
st_tWlnuZ and st_tWqqZ cards for 2016

ckmackay:
2016 version of these cards:
https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/st_tWlnuZ_5f_LO
https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/st_tWqqZ_5f_LO
Only changes made are the PDF sets in the run_card.


qliphy:
ok, but please check PDF variations are also included for 263000.  
For example, you may need to change by hand this file:
https://github.com/cms-sw/genproductions/blob/master/MetaData/pdflist_5f_2017.dat

PR #2468, opened 2019-11-12
Add ZZ and ZH to 4b cards for 2016 PDFs

patrickbryant:
Adding MadGraph cards for ZZ to 4b and ZH to 4b Powheg .input file using recommended PDFs for 2016 MC samples.


agrohsje:
There are 0 files changed. Going to close now.

PR #2469, opened 2019-11-13
Setting pdfwgt=F in run card

Sumantifr:


Sumantifr:
This is for the run card for Wprimetotb directory.

ph2632:
So, this update implies that gridpacks should be re-produced (with "F=pdfwgt" configuration) right? 
Antonis, MC-contact

agrohsje:
Yes. Please.

Sumantifr:
We will do. We are also trying to address the comment on merging issue in the ticket [1]. 
We will provide new gridpacks after we take care of it.

[1] https://its.cern.ch/jira/browse/HIGHPRIOREQ-30

PR #2474, opened 2019-11-17
add cards for ZJets and GJets using sherpa, with EW correction

menglu21:


menglu21:
@GEN conveners, could you have a check on these cards and approve the cards if they are fine, thanks.

qliphy:
Adding @GurpreetSinghChahal

GurpreetSinghChahal:
@menglu21 Thank you very much!

Marek suggested to avoid pt binning samples, and encouraged to directly go for pt_0_to_inf by controlling the number of events by using Enhance_Observables, etc. in the run cards. After discussing with other Sherpa authors and ATLAS, I'd suggest to go for DYjetstoemu (180M) and DYjetstotaus (90M) separately. In ATLAS, they always produce e/mu and tau DYjets separately. Even though, we can enable tau spin correlations with SOFT_SPIN_CORRELATIONS=1 if we wangt to produce e/mu/tau together but considering the tau spin correlations could make the sherpack production slow. So let's produce them separately. 

You can take this card for DYjetstoemu and create a similar one for taus. Please let me know incase of any issue: https://gitlab.com/GurpreetChahal/sherpa-setups/blob/master/Run.dat_DY_01234jets_pt0_to_inf

menglu21:
> @menglu21 Thank you very much!
> 
> Marek suggested to avoid pt binning samples, and encouraged to directly go for pt_0_to_inf by controlling the number of events by using Enhance_Observables, etc. in the run cards. After discussing with other Sherpa authors and ATLAS, I'd suggest to go for DYjetstoeemumu (180M) and DYjetstotautau (90M) separately. In ATLAS, they always produce e/mu and tau DYjets separately. Even though, we can enable tau spin correlations with SOFT_SPIN_CORRELATIONS=1 if we want to produce e/mu/tau together but considering the tau spin correlations could make the sherpack production slow. So let's produce them separately.
> 
> You can take this card for DYjetstoeemumu and create a similar one for taus. Please let me know incase of any issue: https://gitlab.com/GurpreetChahal/sherpa-setups/blob/master/Run.dat_DY_01234jets_pt0_to_inf

Hi Gurpreet, thanks for the check, i have several question on the card setting:
1. what's the meaning of this line, '
`PP_RS_SCALE VAR{0.25*H_Tp2};`', i didn't find information on this line.
2. do i need to change the pdf to NNPDF30_nnlo_as_0118
3. 
`Enhance_Observable VAR{log10(PPerp(p[2]+p[3]))}|1.|3. {3,4,5,6,7,8};`, is {3,4,5,6,7,8} consistent with NJET=4? or should we use {3,4,5,6}

GurpreetSinghChahal:
Hi @menglu21 Thank you! 
1. incase of ≥ 2 jets at NLO, it's recommended to use this setup, which improves the unweighting efficiency 
2. Thanks for reminding. yeah, let's use Sherpa's default PDF= NNPDF30_nnlo_as_0118
3. Thanks, it's corrected to {3,4,5,6} but even if we're writing {...7,8}, it will not create any problem because we don't have that final state

menglu21:
@GurpreetSinghChahal thanks, i'll close this one and make a new PR

PR #2487, opened 2019-11-28
adding NMSSM bbgg final state cards for 2017

panwarlsweet:


panwarlsweet:
Hello @qliphy ,

(Started discussion here https://github.com/cms-sw/genproductions/pull/2563)

For these cards, I checked the effect of pdfwgt = true / false effect. I am attaching the pT-distribution of Y and Higgs bosons. From plots, it seems a small effect (by default it is true in submitted Run2 samples). Could you please have a look and suggest if they are fine enough to do the analysis? I am sorry I should have written an email but I thought it is good to have this discussion with PR.
image

image


Thanks,
Lata

qliphy:
Thanks for the checks.  
I notice your process is a LO one, but not a MLM matching one (i.e. several processes with different jet multiplicities),  if so, "pdfwgt" will not affect you.

panwarlsweet:
Yes, the process is LO. Thanks for your quick reply and confirmation.

PR #2494, opened 2019-12-04
Updating source_condor.sh for Python 3 compatibility

eshwen:
Updating the Python command used for checking existence of `htcondor` package for Python 3 users


agrohsje:
As MG5_aMC@NLO is not yet Python 3 compatible, what is the reason to do this PR in master?

eshwen:
The PR is mainly for future-proofing as Python 2's end of life is approaching. I chose `master` because it seems to still be maintained (while the `mg26x` branch I normally use has seen no updates in a while). I can target a different branch if required

agrohsje:
Will close this and leave it to the python3 development branch.

PR #2496, opened 2019-12-04
update sherpa Gjets card

menglu21:


GurpreetSinghChahal:
Hi @menglu21 Thanks a lot! The cards looks ok, but could you please remind me the source of isolation cut? Was it there in the previous MG5 samples?
FYI @rajat116

menglu21:
> Hi @menglu21 Thanks a lot! The cards looks ok, but could you please remind me the source of isolation cut? Was it there in the previous MG5 samples?
> FYI @rajat116

Just add offline talks here to close this thread, the isolation part is following setting from MG.

PR #2499, opened 2019-12-04
Fix for default params not being overwritten and for reweighting bug

danielwinterbottom:
1. fixing bug in gridpack script where specified params card is ignored and default parameters are always used
2. fixing bug in gridpack script that causes weights computed using reweighting feature to be removed
3. Adding LO weights in addition to NLO weights to cards used for ggH CP gridpacks


agrohsje:
hi daniel, what is the timescale for these samples?  
because 2. will not be sufficient. there is more work needed. there is already an open pr 2170 for this.

danielwinterbottom:
I have already produced the gridpacks for these samples, I just thought it was better to make a PR for the fixes so that it doesn't cause problems in the future. I can remove the changes made for 2 if there is already a open PR?

agrohsje:
ok. yes please do so. i see you don't use madspin, so you didn't encounter the other problems.

danielwinterbottom:
ok I have removed the reweighting fix for now

agrohsje:
thanks. one more comment. isn't the purpose of the customize card to adjust the default param card? any reason you cannot use that functionality. i assume this was introduced to avoid mistakes but @qliphy might comment more.

qliphy:
Yes, it is simpler to use customize card.

agrohsje:
Hi @danielwinterbottom ,
can you do the param_card modifications also to the UL2019 branch? 
Will merge with master,
Alexander.

danielwinterbottom:
Yes I will do this today

danielwinterbottom:
Ok sorry for the delay I added this and made a PR to UL2019 branch: https://github.com/cms-sw/genproductions/pull/2535

PR #2503, opened 2019-12-04
Charged Higgs to TB NLO

Jangbae:
Added charged higgs to TB NLO data cards for 14 different mass points.


qliphy:
As the cards are similar for different model parameters, would you please upload some example cards only ?

Issue #2508, opened 2019-12-06
Bug in run_pwg_condor.py


EmyrClement:
Hi,

Looking at these lines:

https://github.com/cms-sw/genproductions/blob/a4a93103faa51123763e0a9b387915a2ff33744c/bin/Powheg/run_pwg_condor.py#L956-L965

it appears a merge conflict wasn't resolved before merging into master?  The script does not run.

Thanks,
Emyr

agrohsje:
Fixed now. Thanks for reporting.
Cheers, Qiang and Alexander.

PR #2464, opened 2019-11-06
Suu Diquark

evahal:
Adding Suu Diquark files to master branch

PR #2509, opened 2019-12-07
Revert "Suu Diquark"

qliphy:
Reverts cms-sw/genproductions#2464

https://github.com/cms-sw/genproductions/issues/2508


DryRun:
Tagging @evahal

PR #2524, opened 2019-12-31
Added 2016 hadronizers

afrankenthal:
Hello, I'd like to add the 2016 hadronizers to iDM which use the CUEP8M1 tunes rather than the CP5 ones used in 2017+18. That is the only difference in the new hadronizers. Thank you!


qliphy:
Why not renaming 
iDM_DarkPhotonToMuMu_2016_ctau-XXX_TuneCP5_13TeV_MLM_4f_max2j_pythia8_cff.py 
as 
iDM_DarkPhotonToMuMu_2016_ctau-XXX_TuneCUEP8M1_13TeV_MLM_4f_max2j_pythia8_cff.py ?

afrankenthal:
Good catch! Just submitted a fix. Thanks!

qliphy:
Thanks. But please also update https://github.com/cms-sw/genproductions/blob/c38af68d5e0b8cb0002581f91b2e9cf51a02e3e8/python/ThirteenTeV/iDM/createFragments.py#L33

afrankenthal:
Whoops, my bad. It's fixed now.

qliphy:
And also this?  
template_2016 = 'iDM_DarkPhotonToMuMu_2016_ctau-XXX_TuneCP5_13TeV_MLM_4f_max2j_pythia8_cff.py'

afrankenthal:
Oof, yeah you're right. Sorry should've checked more carefully!

PR #2534, opened 2020-01-23
Update 2016 cards & add new 2017 cards for GGF narrow-width Radion, Wprime and BulkGrav ZV->llqq samples

nurfikri89:
This PR is to update cards and add new cards for GGF production samples of narrow-width Radion, Wprime and BulkGrav resonances decaying to ZZ or ZW which then exclusively decays to llqq. 

Card updates involve the migration of the 2016 production cards in the pre2017 tag ([Radion](https://github.com/cms-sw/genproductions/tree/pre2017/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson/Spin-0/Radion_ZZ_ZlepZhad) [Wprime](https://github.com/cms-sw/genproductions/tree/pre2017/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson/Spin-1/Wprime_WZ_WhadZlep), [BulkGrav](https://github.com/cms-sw/genproductions/tree/pre2017/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson/Spin-2/RSGraviton_ZZ_ZlepZhad)) and also extensions to the range of mass points such that it covers from 400 GeV to 8 TeV.

New cards are added for 2017 production with the same range of mass points.


nurfikri89:
ping! Are there any problems with our cards?

qliphy:
Look good to me.  @agrohsje  can you have a double check?

agrohsje:
Can you comment on your 4 vs 5 flavor scheme choice. Why do you need to exclude b's from the proton? Can't you treat them massless everywhere and use 
5 = maxjetflavor as well as 
5 = asrwgtflavor . 
Can  you confirm that the only difference between 2016 and 2017 run_cards is 
 'lhapdf'  = pdlabel     ! PDF set                                     
 263400    = lhaid       ! if pdlabel=lhapdf, this is the lhapdf number
versus the "generic" form. Which version of the code you want to use for 2016? Just asking to make sure you include PDF weights as well.

nurfikri89:
> Can you comment on your 4 vs 5 flavor scheme choice. Why do you need to exclude b's from the proton? Can't you treat them massless everywhere and use
> 5 = maxjetflavor as well as
> 5 = asrwgtflavor

The choice made is purely based on being consistent with the already available 2016 samples [here](https://github.com/cms-sw/genproductions/blob/pre2017/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson/Spin-0/Radion_ZZ_ZlepZhad/Radion_ZZ_ZlepZhad_narrow_M1000/Radion_ZZ_ZlepZhad_narrow_M1000_proc_card.dat#L6) and [here](https://github.com/cms-sw/genproductions/blob/pre2017/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson/Spin-0/Radion_ZZ_ZlepZhad/Radion_ZZ_ZlepZhad_narrow_M1000/Radion_ZZ_ZlepZhad_narrow_M1000_run_card.dat#L71). We'll gladly change to 5 flavour scheme and  maxjetflavor to 5 if that is the more appropriate choice.

> Can you confirm that the only difference between 2016 and 2017 run_cards is
> 'lhapdf' = pdlabel ! PDF set
> 263400 = lhaid ! if pdlabel=lhapdf, this is the lhapdf number
> versus the "generic" form. Which version of the code you want to use for 2016? Just asking to make sure you include PDF weights as well.

I can confirm the only difference between 2016 and 2017 run_cards is specifying the PDF set (again on the basis of [consistency](https://github.com/cms-sw/genproductions/blob/pre2017/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson/Spin-0/Radion_ZZ_ZlepZhad/Radion_ZZ_ZlepZhad_narrow_M1000/Radion_ZZ_ZlepZhad_narrow_M1000_run_card.dat#L52)) for 2016 cards. We intend to use version 2.6.5 for 2016.

nurfikri89:
Hi @qliphy @agrohsje, we can drop the p definition in our [proc_card](https://github.com/cms-sw/genproductions/blob/3c96f49caaa188e409e8883d4b656c5146c829c3/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/exo_diboson/Spin-0/Radion_ZZ_ZlepZhad_narrow/Radion_ZZ_ZlepZhad_narrow_M/Radion_ZZ_ZlepZhad_narrow_M_proc_card.dat#L6) and maxjetflavor to 5 if that is the recommended way to go. As I had mentioned above, our initial choice is purely based on being consistent with the already existing 2016 samples.

agrohsje:
Hi @nurfikri89, I hope there was some reason for the specifications in 2016. I have no recommendation. I just like to understand. The question also depends on the model.

clelange:
Hi, @nurfikri89 pointed me to this discussion.
Thanks for carefully reviewing these data cards, @agrohsje 
We've discussed in the past on hypernews: https://hypernews.cern.ch/HyperNews/CMS/get/generators/4078/2.html (based on GEN recommendations) and also for the diHiggs samples we decided to move to 5FS. Quoting from an older email that I wrote to the HH4b high mass team:

> In principle, sticking to the 4FS would be
> nice, but that would mean using NNPDF 3.0 LO PDFs while in the CP2 tune
> NNPDF 3.1 LO is used. For consistency, it's probably OK to use NNPDF 3.1 LO
> PDF (5 FS) with the corresponding CP2 PYTHIA tune. It would be good to check
> for each model we use if it supports 5FS, and for those that don't, check
> the effect, but that's quite some work... I would think that Eiko's study is
> somewhat generally applicable, and then 4FS vs 5FS should not have a large
> effect. We don't really have bottom quarks in the initial state for most
> processes. I studied 4 vs 5 FS for bbH in the past, and the overall
> kinematic effect wasn't large (and covered by the PDF uncertainties).

Bottom line was that we moved to 5FS to be consistent with the PDFs even though it is inconsistent with the corresponding 2016 simulation. The differences are, however, negligible.

agrohsje:
Thanks @clelange . @nurfikri89 please fix cards to have consistently 5FS in proc and run cards.

nurfikri89:
Thanks @clelange and @agrohsje ! Just to confirm should this fix be applied only for the 2017 cards or should be applied also for 2016 cards?

agrohsje:
You can leave the 2016 in the mixed scheme if this is just for documentation. But I would strongly recommend to have dedicated path label for these so that they don't get picked up in the future.

nurfikri89:
@agrohsje I've made the changes and have been committed to my branch

agrohsje:
+1 from my side for merging. @qliphy please go ahead.

PR #2535, opened 2020-01-28
adding params card fix for UL branch

danielwinterbottom:
Adding the fix to ensure the params card is not ignored when making gridpacks to the UL2019 branch, as requested in https://github.com/cms-sw/genproductions/pull/2499#issuecomment-577660862


agrohsje:
Thanks!

PR #2536, opened 2020-01-29
Re-adding Suu diquark files

DryRun:
Trying https://github.com/cms-sw/genproductions/pull/2464 again, which was reverted due to merge conflicts (https://github.com/cms-sw/genproductions/pull/2509). 

@evahal


agrohsje:
Hi @DryRun . Just to check this is all 4f and the default definition in the UFO?

DryRun:
Hi @agrohsje - 

Eva made some more changes that include using 5f instead of 4f, and committed them in https://github.com/cms-sw/genproductions/pull/2537. Sorry for the double PR!

- David

PR #2540, opened 2020-01-31
Perform input checks before execution

vargasa:
Relocating the input checks so to avoid any execution (and waiting times) without valid input

### DIFF
--- a/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
+++ b/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
@@ -588,10 +588,6 @@ name=${1}
 # name of the run
 carddir=${2}
 
-if [ -z ${carddir} ]; then
-    echo "Card directory not provided"
-fi
-
 # which queue
 queue=${3}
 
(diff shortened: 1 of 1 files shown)


agrohsje:
Thanks Andres. Wouldn't it make sense to also check that the name argument is provided : 
https://github.com/cms-sw/genproductions/blob/81a5ec1f3442224d6c6b009d6b47454808a40ff0/bin/MadGraph5_aMCatNLO/gridpack_generation.sh#L586  ?
Can you do the same PR to UL2019 please?

vargasa:
Hello @agrohsje, I also see `MG5_aMC_v2.6.7.tar.gz` is now available. Should it be changed too? current script version is `2.6.5`. Let me know

vargasa:
@agrohsje `name` check is already in place [here](https://github.com/cms-sw/genproductions/blob/3ca293360fd2d566888893d9075c488326298caf/bin/MadGraph5_aMCatNLO/gridpack_generation.sh#L661)

agrohsje:
Right. Maybe you can also move your check below 
"#catch unset variables" 
to make it cleaner. Yes, please also modify to 2.6.7 branch.  Thanks !

agrohsje:
Hi @vargasa . Sorry. I realize a typo in my comment above. I wanted to say:"Yes, please also modify the 2.6.7 branch."  We already have a development branch for 267. Don't modify the version in master. We need to make sure patches are still appropriate. Moreover, we run extensive physics validation before such a change. Rest looks good.

agrohsje:
@qliphy  +1 for merging.

PR #2548, opened 2020-02-07
Model: dark-Higgs WW 

calderona:
New MC model and cards for Dark-Higgs --> WW signal inside the EXO group.


agrohsje:
Looks good to me. Maybe one question: Should we really have the hadronizer here? @qliphy feel free to merge.

qliphy:
We got noticed that it is possible setting 'event_norm = average' (default is 'sum') for LO processes will cause issues in production.  Can you have a check by generating gridpacks with different event_norm setting and test locally?

calderona:
I have tested both settings: sum and average, and it works with both. If you recommend to move to the default 'event_norm =sum', we will go ahead with it

qliphy:
Thanks.  Please set event_norm as sum

PR #2555, opened 2020-02-17
Add cards for the leading NLO EW contribution to ttW production

pieterdavid:
These cards simulate pp > ttW+jet through EW diagrams (by excluding diagrams with more than one QCD coupling), which is the dominant NLO EW contribution (real diagrams) to ttW production; a pt > 50GeV cut on the leading jet is added to avoid the region where a full NLO treatment is necessary.
It should be safe to combine this with the existing TTWJetsToLNu samples; I first tried using the cards for that process, but ran into problems when generating a gridpack, so I used the run card from the MG 2.6.7 session where I checked the diagrams (and compared with one that was recently added, DarkHigs_WW - I changed the defaults where I knew they needed changing (maxjetflavor=5, pdf settings), but left the following: `average` instead of `sum` for event_norm (TTWJetsToLNu uses average), auto_pt_mjj (which should matter without xqcut, if I read the comment correctly), and the syscalc settings which look similar in a different format) - please let me know if there is a guideline on this, or a list of things to check to make sure this is all correct - thanks in advance!

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TTWJets/TTWJetToLNuEWJetPt50_5f/TTWJetToLNuEWJetPt50_5f_customizecards.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TTWJets/TTWJetToLNuEWJetPt50_5f/TTWJetToLNuEWJetPt50_5f_customizecards.dat
@@ -1,0 +1,4 @@
+#put card customizations here (change top and higgs mass for example)
+set param_card mass 6 172.5
+set param_card yukawa 6 172.5
+set param_card mass 25 125.0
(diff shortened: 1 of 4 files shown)


agrohsje:
Hi Pieter,

please remove this section: 
systematics = systematics_program ! none, systematics [python], SysCalc [depreceted, C++]
['--mur=0.5,1,2', '--muf=0.5,1,2', '--pdf=errorset'] = systematics_arguments ! see: https://cp3.irmp.ucl.ac.be/projects/madgraph/wiki/Systematics#Systematicspythonmodule
# Syscalc is deprecated but to see the associate options type'update syscalc'
4 = asrwgtflavor
should be 5. 
I would remove any cut you don't need, i.e. loosen dR_ll if they are not intended. 
Cheers, Alexander.

pieterdavid:
Thanks Alexander! The commits I just pushed should fix these issues (I don't think I need any eta or dR cuts at ME level, so I disabled all of them)

agrohsje:
Thanks so much. One final question. You use pt jet > 20 for all other jets. Seems high? What are your final jet cuts? I would suggest lowering the general cut to 10 and just keep the leading high, no? 
Otherwise +1 from my side. @qliphy feel free to approve depending on Pieters feedback.

pieterdavid:
Thanks Alexander, I reduced the (non-leading) jet pt cut to 10 to be safe indeed.

PR #2557, opened 2020-02-19
Add cards for ZprimeVBF samples ZprimeVBF_WW and ZprimeVBF_tau_tau

padeken:
Hi, we would like to have the signal for ZprimeVBF in the central production. The corresponding presentation is here:
https://indico.cern.ch/event/863944/


qliphy:
Please don't upload 800 files!
You can add a few of them together with a script to make others.

padeken:
Hi, ok I changed it. Must have missed the instruction for this. Is this better?

qliphy:
icckw should be 0 as this is a LO process.

And please follow more https://twiki.cern.ch/twiki/bin/view/CMS/Fall18Repack
e.g.  banner of proc_card.dat can be removed.

qliphy:
replace the .... = lhaid line by: 
$DEFAULT_PDF_SETS = lhaid
    replace the ... = reweight_PDF line by: 

padeken:
Hmm about the ickkw we do have explicit jets in the final state, shouldn't there be a (MLM) matching?

qliphy:
your process is VBF, and the two jets are EWK but not QCD ones, so no need for matching.

qliphy:
Please also remove these:
https://github.com/cms-sw/genproductions/blob/12278dea16d076e371722920cae7e28d2a4e9888/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ZprimeVBF_WW/run_card.template#L256-L268

Otherwise looks good to me.   @agrohsje

padeken:
Sorry is there anything still to be done for this PR?

qliphy:
+1 @agrohsje

agrohsje:
There so nothing you set up your own. It is all default, right? If so, suggest to remove: 
set group_subprocesses Auto
set loop_color_flows False
set gauge unitary
set complex_mass_scheme False
set max_npoint_for_channel 0
Why do you need the param_card.template? Can't you work with the customize_card to modify only the specific options?
I cannot find the model on https://cms-project-generators.web.cern.ch/cms-project-generators/ .
Please upload.

padeken:
Thanks I removed the unneeded setup, but for the customize_card, I am not sure this will work in this model since it is based on a mg4 model. If it is not too complicated I would like to keep it this way, were we validated the output.
For https://cms-project-generators.web.cern.ch/cms-project-generators/ I don't have access to write so I asked the model to be uploaded. It was uploaded to https://cms-project-generators.web.cern.ch/cms-project-generators/SSMZp/ssmzp_nonuniversal_v4.zip
(Note the subfolder).

agrohsje:
How do you plan to include the template param card in the gridpack_generation.sh. You could also test customize_card so that everybody can run your setup without special modifcations. And if indeed there is a problem, do it via the param_card. 
Can you comment more on the process line: 
generate p p > zp j j $$ z w+ w- zp / g a h , ( zp > w+ w- ) QCD=0
There is no impact from  $$ z w+ w- zp. Same for QCD=0 which acts on the decay only.

padeken:
Sorry I had to work on other stuff. Now I finally had time to debug something. So yes we can remove the QCD=0 and also the constraint. It removes basically some diagrams that we know are not VBF, but they end up in the same final state and the physics is solid.

But I did realize that I did a simplification that does not work for one parameter case we need a different model, because the main contributing diagrams change significantly.

Sorry I did not have time to convert it to the customize_card format. Since the model is v4 and the format of the settings change.

Coming back to this after some time I also did something not nice with git, sorry!

padeken:
Hi, is there any open issue?

agrohsje:
Dear @padeken 
I see that you removed now everything in the generate line. But not that 
generate p p > zp j j $$  w+ w-  is not the same as 
generate p p > zp j j . 
Can you check what you want.

padeken:
Hi, yes we want to separate the WW final states and the tau tau final states. This is what we want.

agrohsje:
No, no. I was talking about something else. Do you generate on purpose p p > zp j j ? This  will also include diagrams where the Z' is radiated from a W+ and then decays into a jet pair. You will reduce the contribution from this diagrams by your mass cut. Do you prefer keeping these diagrams or remove them by doing : 
generate p p > zp j j $$ w+ w-

padeken:
We tried to keep these diagrams, but the contribution is very small and mg has problems with the inclusive generation. So yes since the Z' will be much heavier than the W there is no need to keep this diagrams.

agrohsje:
Does that mean you will change the line to :
generate p p > zp j j $$ w+ w-
?

padeken:
Ahh no ok sorry I was confused. As it is with
generate p p > zp j j , ( zp > w+ w- )
and 
generate p p > zp j j  ,  zp > ta- ta+ 
it works and we would keep it this way.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ZprimeVBF_WW/run_card.template
```diff
@@ -0,0 +1,249 @@
...
+#
+#*******************************                                                 
+# Parton level cuts definition *
+#*******************************                                     
+#                                                                    
+#
+#*********************************************************************
+# BW cutoff (M+/-bwcutoff*Gamma) ! Define on/off-shell for "$" and decay  
+#*********************************************************************
+  15000  = bwcutoff      ! (M+/-bwcutoff*Gamma)
```
sihyunjeon, line 88: @agrohsje
Is it okay to set bwcutoff=15000 here? The resonance and decays are cracked into two parts for this process.
We had some issues with DY Zprime to tautau (p p > zp, zp > ta+ ta-) samples that were already produced (WRONGLY) for EXO and reviewing their cards. But there bwcutoff was set to 15000 and I saw ~20% increase in cross sections compared to using bwcutoff=15.

agrohsje: 20% is a lot. What was the mass and the width. The Zp should have a BW shape with a cutoff of width * bwcutoff. So if the particle is very narrow 15 or 15k will make some difference, because you cut hard in the BW. However, for the Z the cross section would only change by ~1%.

sihyunjeon: 3000 GeV mass point has ~80 GeV width so... In these cases bwcutoff=15 will be safe right?

Issue #2542, opened 2020-02-02
Making a 2016 gridpack


DryRun:
I was wondering how the logic for 2016 PDFs is supposed to work in the gridpack creation scripts. Particular these lines, https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/Utilities/gridpack_helpers.sh#L12-L23, which control the default PDFs used for 2016 vs. 2017. Should this line:

`elif grep -q -e "\$DEFAULT_PDF_SETS" -e "\$DEFAULT_PDF_MEMBERS" $CARDSDIR/${name}_run_card.dat; then`

be something like this instead:

`elif grep -q -e "\$DEFAULT_2016_PDF_SETS" -e "\$DEFAULT_PDF_MEMBERS" $CARDSDIR/${name}_run_card.dat; then` ? It seems like the first "elif" block will never be entered. 

Thanks,
David

agrohsje:
Hi @DryRun . I hope @kdlong or @qliphy can comment more on the logic. I must admit it is not clear to me how the current splitting is supposed to work.

qliphy:
Currently for 2017/2018, PDF can be handled automatically, if one set in run_card.dat the following as in GEN recommendation:
                   $DEFAULT_PDF_SETS = lhaid                                  [1]
                   $DEFAULT_PDF_MEMBERS = reweight_PDF           [2]

However, for 2016, one has to set it explicitly by hand as one may want a different central PDF.  
In this case, as @DryRun mentioned, as there is no [1] [2] set as above, 
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/Utilities/gridpack_helpers.sh#L18-L23
this will not leave the run_card.dat as default.
On the other hand, due to the following:
https://github.com/cms-sw/genproductions/blob/master/Utilities/scripts/getMG5_aMC_PDFInputs.py#L18
https://github.com/cms-sw/genproductions/blob/master/Utilities/python/PDFSetsChooserTools.py#L10-L16
the reweight PDF sets will just follow 2017.
In runcms

kdlong:
Right. This was on my todo list for so long it fell off. If someone would like to have it, it's quite straightforward to implement, I can provide input at least.

agrohsje:
@DryRun Would you be available to do it with the help of @kdlong ?

DryRun:
Hi @agrohsje - sure, I'd be happy to.

PR #2558, opened 2020-02-21
Update the cards for the run 2 VBF HH production

l-cadamuro:
This PR updates the cards for the VBF HH production.
The corresponding gridpacks have been reproduced to be compatible with the current version requirements.
All the samples are being requested to be remade with updated showering options and increased event statistics.

List of changes introduced :

- run and process cards added in all the subfolders (were just in 2 out of 6 subfolders previously)
- added a new sample ``VBF_HH_CV_1_C2V_0_C3_1_13TeV-madgraph`` that is needed for analysis and validation.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/VBF_HH/VBF_HH_CV_0_5_C2V_1_C3_1_13TeV-madgraph/VBF_HH_CV_0_5_C2V_1_C3_1_13TeV-madgraph_extramodels.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/VBF_HH/VBF_HH_CV_0_5_C2V_1_C3_1_13TeV-madgraph/VBF_HH_CV_0_5_C2V_1_C3_1_13TeV-madgraph_extramodels.dat
@@ -1,0 +1,1 @@
+SM_HEL_UFO_noLightYukawa_HH_VBF.tar.gz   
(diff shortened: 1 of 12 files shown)


qliphy:
+1 @agrohsje

l-cadamuro:
Thansk @qliphy, @agrohsje , let me know if this can be merged.

By the way, I assume that these are fine for the 2017/2018 campaigns (not UL), but which changes should be done to produce the same gridpack for the 2016 conditions?
I could not find anything specific on the [TWiki](https://twiki.cern.ch/twiki/bin/viewauth/CMS/QuickGuideMadGraph5aMCatNLO)

Of course if this configuration is OK for the 3 years we'll just use the same gridpacks for all.

Cheers,
Luca

qliphy:
For 2016, you need to change PDF if you want to use the old 2016 Pythia8 Tune.
https://twiki.cern.ch/twiki/bin/view/CMS/Moriond18MC

l-cadamuro:
Perfect, thank you very much!

I'll make a second version for 2016 then, changing the pdf to 292000.
Let me know if I should commit that one as well.

Btw, do I read correctly from your link that the ``$DEFAULT_PDF_MEMBERS  = reweight_PDF`` has to be removed for 2016?

Cheers,
Luca

qliphy:
> Btw, do I read correctly from your link that the `$DEFAULT_PDF_MEMBERS = reweight_PDF` has to be removed for 2016?

Yes.  You can also find some discussions here:
https://github.com/cms-sw/genproductions/issues/2542

l-cadamuro:
Thank you @qliphy @agrohsje  ! All the gridpacks are now ready for both conditions.

If all is OK can this be merged?
I think that the HIG MC contacts (@gourangakole @attikis) await for the merge to copy the gridpack to cvmfs, which is in turn needed to move forward with the requests.

Cheers,
Luca

qliphy:
+1 @agrohsje

PR #2559, opened 2020-02-21
Add cards for the ggH madgraphMLM production in high Higgs pt bins

wang0jin:
These cards are for samples used in the study of the running of Higgs masses. The analysis requires more statistics of the ggH process in higher Higgs pt bins. The samples are produced with a cut on Higgs pT to be larger than 120GeV or 270GeV. 

These are example cards of 125GeV Higgs mass point. Cards for other mass points are similar.


qliphy:
Please remove these lines:
https://github.com/cms-sw/genproductions/blob/3a06f0f8e1706ba6fd47028fdd4056f8d91cdbf1/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ggh01j_M125_HiggsPt120toInf/ggh01j_M125_HiggsPt120toInf_run_card.dat#L290-L294

https://github.com/cms-sw/genproductions/blob/3a06f0f8e1706ba6fd47028fdd4056f8d91cdbf1/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ggh01j_M125_HiggsPt270toInf/ggh01j_M125_HiggsPt270toInf_run_card.dat#L291-L294

Moreover, why do you need a pythia card?

wang0jin:
Hi Qiang,

Will the scale variation automatically included in systematic weights without setting parameters of the systematic module?

I think the pythia is used to custermize the decay the Higgs to two photons. They did not do it in Madgraph because they found in that case the Higgs pt restriction would be ignore.

Thank you.

Kind regards,
Jin

qliphy:
> Will the scale variation automatically included in systematic weights without setting parameters of the systematic module?

Yes, this is added later here:
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/runcmsgrid_LO.sh#L171-L174
 
> I think the pythia is used to custermize the decay the Higgs to two photons. They did not do it in Madgraph because they found in that case the Higgs pt restriction would be ignore.

For gridpack generation, do we need a pythia card?  I mean pythia is set later in the fragment in McM request.

wang0jin:
I see your points. Yes, you are right. I have removed these lines and pythia cards as you suggested. Thanks a lot.

Cheers,
Jin

qliphy:
+1 @agrohsje

agrohsje:
Hi @wang0jin,
I have a few comments on your cards. It is all 5 flavor scheme, no? Please change 
4 = asrwgtflavor . I would suggest to remove the first "37" lines of your process card. 
You cut on 20 GeV jet. This is no issue due to the high Higgs pT cut, right? 
Why did you change photon cuts? There are no photons involved.
Cheers, Alexander.

wang0jin:
@agrohsje Thanks a lot for the comments. The analyzer has updated the cards accordingly. The gridpacks and requests will also be updated once the pull request is merged.

Best regards,
Jin

agrohsje:
set default_unset_couplings 99
set group_subprocesses Auto
Hi Jin,

thanks for the modifications. Any reason not to remove this default stuff: 
set loop_color_flows False
set gauge unitary
set complex_mass_scheme False
set max_npoint_for_channel 0
import model sm
define l+ = e+ mu+
define l- = e- mu-
define vl = ve vm vt
define vl~ = ve~ vm~ vt~

Nothing of this was added by you? 
Can you confirm that the model is 5 FS by default? 
Where do you apply the Higgs pT cuts?

ashimroy:
Hi Alexander,
    There is no reason to keep the default staff. If you want I can remove if there is no problem in the production. 
What do you mean by the model is 5 FS by default. Are you talking about this 5 = asrwgtflavor? 
 The higgs pt cut is applied in run card. {25:120} = pt_min_pdg

Thanks,
Ashim

agrohsje:
I would suggest to remove everything in the proc_card that you don't need to add explicitly and that was there as defaut. This makes it easier to read. We can afterwards always get the additional info from the gripack itself. 
About 5FS: is the b massive or massless in the param_card of the model?

wang0jin:
Hi @agrohsje ,

I have check this model in madgraph and found that it is actually 4 flavour scheme. You can find these info attached below. 

Please check these cards and see whether they are good for merge. Thank you very much.

Best regards,
Jin



MG5_aMC>import model heft-full
INFO: Change particles name to pass to MG5 convention 
Kept definitions of multiparticles l- / j / vl / l+ / p / vl~ unchanged
Defined multiparticle all = g u c d s u~ c~ d~ s~ a ve vm vt e- mu- ve~ vm~ vt~ e+ mu+ t b t~ b~ z w+ h h1 w- ta- ta+
MG5_aMC>display multiparticles
Multiparticle labels:
all = g u c d s u~ c~ d~ s~ a ve vm vt e- mu- ve~ vm~ vt~ e+ mu+ t b t~ b~ z w+ h h1 w- ta- ta+
l- = e- mu-
j = g u c d s u~ c~ d~ s~
vl = ve vm vt
l+ = e+ mu+
p = g u c d s u~ c~ d~ s~
vl~ = ve~ vm~ vt~
MG5_aMC>display particles b
Particle b has the following properties:
{
    'name': 'b',
    'antiname': 'b~',
    'spin': 2,
    'color': 3,
    'charge': -0.33,
    'mass': 'mdl_MB',
    'width': 'ZERO',
    'pdg_code': 5,
    'line': 'straight',
    'propagator': '',
    'is_part': True,
    'self_antipart': False,
    'type': '',
    'counterterm': {}
}

agrohsje:
5 = asrwgtflavor     ! highest quark flavor for a_s reweight 
Needs to be 4 in this case.

wang0jin:
@agrohsje Yes, It was modified to 4 already. 

Thanks,
Jin

agrohsje:
Not yet. Search for asrwgtflavor here 
https://github.com/cms-sw/genproductions/pull/2559/files
and you will see it is still 5.

wang0jin:
Sorry, you are right. I thought you were talking about jet flavor option. Now it is corrected. Thank you!

wang0jin:
Hi Alexander,

Do you think these cards are good for merge? Is it OK to proceed with gridpack production.

Thank you.

Best regards,
jin

PR #2560, opened 2020-02-22
update Phantom script to fit CONDOR

menglu21:


qliphy:
Thanks.  It would be great if you can give a report on Phantom at the MEFG+Validation meeting on 9th March.

menglu21:
> Thanks. It would be great if you can give a report on Phantom at the MEFG+Validation meeting on 9th March.

Hi Qiang, 

Ok, i think i can give an update on this topic.

Cheers,
Meng

qliphy:
Thanks.  Added to the agenda: https://indico.cern.ch/event/873797/

agrohsje:
Hi @menglu21 . Should we still wait before merging? Can you quickly comment on your next steps? I assume your PR was triggered by the point Benedikt raised, or ?

menglu21:
> Hi @menglu21 . Should we still wait before merging? Can you quickly comment on your next steps? I assume your PR was triggered by the point Benedikt raised, or ?

Hi @agrohsje , it's due to the pdf issue. the difference seems not small, i think we may need to reproduce the sample. The sample size is small (WA: 700K, ZA: 300K), so i guess we can afford redoing the production in a short time.

PR #2561, opened 2020-02-24
Model: dark-Higgs WW->2l2nu

calderona:
New MC model and cards for Dark-Higgs --> WW --> 2l 2nu  signal inside the EXO group.


qliphy:
Please remove https://github.com/cms-sw/genproductions/pull/2561/files#diff-dcd8daf62bd005468b81e7cbec2e0402R279-R291
as suggested in GEN twiki: https://twiki.cern.ch/twiki/bin/view/CMS/Fall18Repack

agrohsje:
@calderona  @SenneVanPutte   Why are the p settings not consistent with :
https://github.com/cms-sw/genproductions/pull/2565/files 
Only the W decay should be different no? Please coordinate.   
Did you see that tauola was needed for the samples? That would be an interesting result. From the P8 manual I assumed this wouldn't be needed. Should also be consistent between the 2 PR's.

calderona:
Sorry, which p settings? there should be consitent for sure among the two. 
About Tauola, we had implement the tau decay in pythia. We are still waiting for the EXO MC contacts to have the ok on this procedure.

agrohsje:
The proton definition. Yours is the default one while the other request drops the gluon and has b's in there, which would correspond to 5FS. I assume the default in the model is 4FS as the original hunting the dark higgs paper had h -> bbbar. But in your case this is not relevant and the 5FS should give you a better estimate of the cross section (if you need that).

SenneVanPutte:
So if I understand correctly we should change p the same way that I changed j and Alicia should define j as I did?
Also should I include g in the j and p definition?

agrohsje:
If you want to run in the 5FS, you should set the b mass to 0 (and the yukawa to avoid hs couplings to your massless b) and do 
p = p b b~ 
j = p . 
There is no need to exclude gluons: MG5 will realize that gluon fusion doesn't matter here.

calderona:
Ok thanks, we go with the 4FS for both decay channels

calderona:
Hi, is there a problem with this PR? We need to have merged.

qliphy:
Why bwcutoff is set to so large a value as 200?  Usually we recommend it to be smaller as 15 or 25.

Otherwise it looks fine to me.  @agrohsje

calderona:
It was suggested by the EXO contacts to extend that range and not cut the W mass.

calderona:
I have just write to the them asking to back to 15 GeV.

agrohsje:
Hi @qliphy . I checked the impact on the cross section and didn't see any change beyond the statistical uncertainty. Maybe we could ask validation to redo some of the plots Lorenzo did with 2.6.5.

qliphy:
@agrohsje   Thanks.  I agree.  

Then please move forward.

agrohsje:
checked that bwcutoff has no impact on cross section

PR #2562, opened 2020-02-25
Completing set of shape benchmark cards being used for UL now

panwarlsweet:


panwarlsweet:
Hello @qliphy !

Could you please confirm that in the UL run cards the pdfwgt should be set to "True"? ( I am using mg265 branch for grid-packs)

I find this on this slide https://indico.cern.ch/event/746828/contributions/3087449/attachments/1712834/2761990/2018-09-10_GENnews.pdf#page=11&zoom=auto,-74,254

but Twiki [1] mention to set it false. Could you please clarify this contradiction?

Thanks,

Lata

[1] https://twiki.cern.ch/twiki/bin/view/CMS/Fall18Repack

qliphy:
@panwarlsweet   pdfwgt recommendation is updated to be set to "false"

Thanks for pointing this out:
We have updated the link on the twiki: https://twiki.cern.ch/twiki/bin/view/CMS/Fall18Repack

panwarlsweet:
Thanks for quick reply @qliphy 

I would like to ask one more question. If the samples which (grid-packs from MG_2.6.5)are already in production for Run-2 or UL have this setting "pdfwget=True", one should submit them again?

Since I find on slide [1] that it largely affects MG_2.4.2 is it fine if one uses the sample with the true setting in MG_2.6.5 grid-pack for the Run2/UL analysis (only if the samples are already in production)?

Kindly help me to understand this.

Thanks,

Lata

[1] https://indico.cern.ch/event/855110/contributions/3613939/attachments/1930820/3197907/2019-09-27_vht_265.pdf#page=9

qliphy:
pdfwgt is now recommended to set to false.   As for previously produced sample with "pdfwget=True", you can first check how large the effect/difference could be.

qliphy:
I notice your process is a LO one, but not a MLM matching one (i.e. several processes with different jet multiplicities), if so, "pdfwgt" will not affect you.

PR #2565, opened 2020-02-28
Cards for dark higgs WW->jjlnu

SenneVanPutte:
Cards for Dark-Higgs --> WW --> jjlnu signal inside the EXO group.


agrohsje:
checked that bwcutoff has no impact on cross section.

PR #2569, opened 2020-03-02
add ggZZ012j_MLM cards for VBS ZZ

colizz:


qliphy:
+1 @agrohsje

agrohsje:
Hi @colizz . How will qcut be choosen. I saw that xqcut is quite low. But just curious.

colizz:
Hi @agrohsje , thanks for your comments. 
We indeed had a study on xqcut in VBS ZZ analysis. We found that for loop induced events, ISR jets can be emitted directly from loop, resulting in generally softer jets. The below slide shows the cross-section and DJR plots with (xqcut, qCUT)=(5, 10) compared with conventional value (10, 20), indicating that the former seems a better choice for this process.
https://docs.google.com/presentation/d/1GHpPyWbML7fMkCkEl7XaDyDBkHRY73vuYr_XG0uZodo/

agrohsje:
With "bad" you are referring to the second plot in the upper row, with the red circle? Do you have more qcut plots for xqcut 10? The acceptance is very low with 5/15.

colizz:
Hi,
It's also because there is a hollow from the first plot in the upper row (error bar cannot compensate this), indicating the unsmooth stitching. We did not try more qCUT plots, but simply choose the recommended qCUT value based on xqcut. 
Indeed the acceptance is low, but we already produced enough LHE given this 8% matching rate (since the generation time is beyond McM limit we chose the pLHE campaign), so there's not so much to worry.

agrohsje:
Thanks for the info. Maybe that is something to follow-up if we want to go for gridpacks. Will merge.

colizz:
Many thanks!

PR #2571, opened 2020-03-09
Xmlchsub

efeyazgan:


efeyazgan:
@qliphy @agrohsje I did this in a hurry, could you please double check?

agrohsje:
@efeyazgan Thanks so much! I couldn't spot the line where the gridpack is copied back to eos. But maybe I just missed it?

efeyazgan:
@agrohsje I've updated it (moved the main operations to a function). Could you please take a look again? (BTW in the meanwhile, it looks like someone had modified the problematic gridpack for pdmvserv_task_TOP-RunIIFall17wmLHEGS-00391__v1_T_200212_212120_3782 to add the --stream option. )

agrohsje:
Thanks. Enrique started patching some by hand. Do you need to untar the old gridpack before modifying runcmsgrid.sh. I see that you tar czf later but I cannot find tar xzf .

efeyazgan:
Yes, untarring has been done for many other purposes: 
https://github.com/cms-sw/genproductions/blob/f195ccb78f181ef66dab7b62d280b7bfa5ef4208/bin/utils/request_fragment_check.py#L681

PR #2583, opened 2020-03-20
Added bJet config with pThat 80 and 100

Taburis:
For b Jet analysis, the b jet MC generation configuration added for minimum pThat 80 and 100 separately. The configuration is the same as the previous Pythia8 CP5 tune with pThat 15, but changed the low pThat cut from 15 to 80 and 100.


agrohsje:
You are changing 635 files.

Taburis:
Sorry committed to wrong repository.

PR #2588, opened 2020-03-27
updated the mediator ID

tiwariPC:
This PR includes an updated ID for the mediator mass.
Please have a look and merge it.
Thanks,
Praveen


qliphy:
Is 9900032 the recommended PID?

tiwariPC:
> Is 9900032 the recommended PID?

yes it is in the model file which we are using, it is only used in mass block

PR #2589, opened 2020-03-28
evtGen checks

jordan-martins:
Hi Efe, Alexander and Qiang,

I made some inclusions to the request_fragment_check to avoid mistakes when using EvtGen.

I hope that it is fine.

Best, Jordan.


agrohsje:
Thanks Jordan. Should we have some checks on the versions as discussed with Alberto.

jordan-martins:
what do you mean @agrohsje !? the dec and pdl files with the 2014 tag!?

agrohsje:
Thanks @jordan-martins .
+1 for merge.

PR #2294, opened 2019-06-29
aQGC cards for VVjj semileptonic VBS

asar0786:
These cards contains 9 samples required for VVjj semileptonic precess. They are

1. pp -> W+W- JJ  -> L+Nu QQ JJ
2. pp -> W+W- JJ ->  QQ L-Nu JJ
3. pp -> W+Z JJ ->  QQ L+L- JJ
4. pp -> W+Z JJ -> L+Nu QQ JJ
5 pp -> W+W+ JJ -> L+Nu QQ JJ
6. pp -> W-Z JJ -> L-Nu QQ JJ
7. pp -> W-Z JJ -> QQ L+L- JJ
8. pp -> W-W- JJ -> L-Nu QQ JJ
9. pp -> ZZ JJ -> L+L- QQ JJ

PR #1513, opened 2017-11-13
adding 2016 Madgraph cards for aQGC_VVjj_hadronic

stalbrec:
These Madgraph cards are based on these pre2017 tree cards [/production/13TeV/VBS/VVjj_semileptonic/aQGC](https://github.com/cms-sw/genproductions/tree/pre2017/bin/MadGraph5_aMCatNLO/cards/production/13TeV/VBS/VVjj_semileptonic/aQGC). They are changed to hadronic decay channel. (2016 PDF set)

PR #2592, opened 2020-04-07
updated aQGC VVjj all-hadronic 2017 Madgraph cards 

stalbrec:
Hi,

this updates the 2017 version of the aQGC VVjj all-hadronic Madgraph cards and moves them to common VBS directory.
The only difference to the 2016 cards (PR #2434) is the used default PDF set (`$DEFAULT_PDF_SETS  = lhaid`).

Best,
Steffen


agrohsje:
Looks good to me. One question which I missed before. I see this line was changed:
generate p p > w+ z j j QED=3 QCD=1 NP=1
Was this used before? Can you comment more.

stalbrec:
This line was changed since `.. QED=3 QCD=1 NP=1` does only include the aQGC diagrams. 
I used cards with this line in 2018 for private samples (#1513 ).
The updated line includes SM (+SM-aQGC Int.) and is also used for the semi-leptonic cards (see for example #2294 ).

agrohsje:
@qliphy +1

PR #2596, opened 2020-04-08
adding SM VVjj hadronic 2016&17 Madgraph cards

stalbrec:
Hi,

this adds madgraph cards for SM VBS hadronic for ~~2016&~~ 2017. 
~~The only difference in 2016 and 2017 is the choice of PDF set(s).~~

These cards include the following nine processes:
ZZ bb,   j_nob j_nob,  bb j_nob j_nob,  jj jj
ZW jj,  bb jj,  j_nob j_nob jj
ssWW  jj jj
osWW  jj jj

Here j_nob is defined without `b b~`. To achieve this I simply tried to redefine the multiparticle `j` in the madspin card. With this I was able to produce gridpacks, but when I ran `runcmsgrid.sh` madspin crashed. This does not happen when i introduce a new multiparticle `j_nob`.
Is this okay?


govoni:
Dear All,

may I ask what version of the EFT model is used here? I understand that the Eboli basis has gone through a sequence of updates:

http://feynrules.irmp.ucl.ac.be/wiki/AnomalousGaugeCoupling#no1

that include bugfixes and elimination of operators. During a recent meeting ATLAS noted that we're using in VBS SSWW + WZ a quite old basis, from the presence of the M6 operator.
It may be worth updating it to a newer one.

stalbrec:
Hi Pietro,

in the cards for the respective aQGC samples for 2016 (#2434) and 2017 (#2592) we indeed use the older model (last updated 2015 on [cms-project-generators.web.cern.ch](https://cms-project-generators.web.cern.ch/cms-project-generators/)). 
I had a quick look into the updated version in November (see [slides](https://indico.cern.ch/event/865716/contributions/3647670/attachments/1950928/3238659/VBS-hadronic_steffen.pdf#page=13)). This did not show any major differences in the resulting distributions I were looking into.
I did not know that the operators that were "missing" in the updated version with respect to the older one, were in fact eliminated. Do you have or could point me to some info on that?

Thanks and best regards,
Steffen

govoni:
Hello Steffen,

unfortunately not, I just reported what I've been told to by ATLAS people. It may be worth contacting the authors and ask for feedback in this situation. Are you already in contact with them?

pietro

stalbrec:
No, I'm not in contact with them yet. Should i ask them, or would you like to do that?

govoni:
I think you may want to do it, as you're doing the analysis, discussions with them may be useful in general. Feel free to put me in CC if you find it useful.

stalbrec:
I remove the pre2017 cards, since they would've not been used.

stalbrec:
ping

kdlong:
@agrohsje @qliphy is there any point in having a specific request for 2016 with an old PDF set? I would prefer to harmonize the gridpacks across years with the most modern PDF set and tune when producing 2016 samples for the first time.

qliphy:
> @agrohsje @qliphy is there any point in having a specific request for 2016 with an old PDF set? I would prefer to harmonize the gridpacks across years with the most modern PDF set and tune when producing 2016 samples for the first time.

Both options should be fine.
It is either inconsistency between signal and background,  or between 2016 and 2017/2018.

qliphy:
> ping

+1  @agrohsje

kdlong:
> > @agrohsje @qliphy is there any point in having a specific request for 2016 with an old PDF set? I would prefer to harmonize the gridpacks across years with the most modern PDF set and tune when producing 2016 samples for the first time.
> 
> Both options should be fine.
> It is either inconsistency between signal and background, or between 2016 and 2017/2018.

From my perspective inconsistency in the signal is the bigger problem, since we tend to have a single signal strength between years, so having a single signal template is preferable. Background differences are more easily absorbed in uncertainties.

govoni:
On the other hand, in each year the analysis result is extracted as the excess of signal over the background-only hypothesis, hence differences in model between signal and background within the same year may produce artefacts. Inconsistencies across the years in the pdfs choice, on the contrary, are similar to those you may expect between ATLAS and CMS for example, and once you know how to treat the correlations, the systematic uncertainties, which are estimated independently in each year, may be properly correlated.

kdlong:
But normally we do the fit simulatenously. So the background and signal is 2016+2017+2018, and there are naturally differences in experimental uncertainties. But ultimately we define a cross section as mu*sigma_theory, and we take for granted that sigma_theory is one number. In the case of different signals, it technically isn't. In the case of different backgrounds, as long as the uncertainties are properly estimated and correlated in each case, it's ok. 

There really shouldn't be artefacts greater than a few percent due to PDF and tune. In the case of backgrounds, this is rarely an issue, but for signal, its not something we typically account for, because the signal uncertainties should largely cancel out in the cross section measurement, and they are negligible in the significance calculation.

govoni:
I understand the fact that variations across pdf versions are hopefully smaller than pdf uncertainties. Concerning the combination, instead, one usually separates the three years in the fit: the result is the combination of three separate fits, not a single fit on lumped distributions. Because of this, it seems to me more relevant the consistency within a year, rather than across years.

kdlong:
One uses separate bins per year in the likelihood, but normally 1 single signal modifier (mu), which is the thing that is most important to have consistent.

govoni:
I am not so adamant about that: it's for sure important, yet it's the ranking between the two type of consistencies that I am not sure always has to be that way. The point is probably simply that one expects the effect to be small anyways.

agrohsje:
Sorry. The setup confuses me. Why treating the b as massless, using 5 FS, but then not including the b in the p defintion? 
import model sm-ckm_no_b_mass
generate p p > z w+ j j / h QED=4 QCD=0 @ 1
add process p p > z w- j j / h QED=4 QCD=0 @ 2

For my own analyses I am doing what Kenneth is suggesting: Being consistent in the signal mode.

stalbrec:
Hi, 
Yes - sorry about the confusion. Thanks for pointing this out. I made a mistake there. 
I also have to change the exclusion of `t t~` and `h` again.
Sorry again, for getting this PR so messy.

agrohsje:
Thanks. Will merge.

PR #2600, opened 2020-04-10
Cards for M-650

ph2632:
Cards copied an modified from M-700.


qliphy:
You don't need to upload these cards, as they are very similar as for M-700.

PR #2603, opened 2020-04-14
[Exotica Mono H cards] Python scripts and templates to produce input cards for the ZpBaryonic model

syuvivida:
This commit includes the instruction, python scripts, and template cards to produce madgraph input cards for the MonoH analysis, the ZpBaryonic model. The structure follows that in the Zp2HDM model. 
https://github.com/syuvivida/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/monoHiggs/Zp2HDM/Zprime_A0h_A0chichi

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/monoHiggs/ZpBaryonic/README.md
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/monoHiggs/ZpBaryonic/README.md
@@ -14,7 +14,7 @@ after executing python script, gridpacks will be generated in several minutes
 git clone git@github.com:cms-sw/genproductions.git
 # Produce cards
 cd genproductions/bin/MadGraph5_aMCatNLO/
-cp cards/production/2017/13TeV/monoHiggs/ZpBaryonic/gridpacks_ZpBaryonic.py
+cp cards/production/2017/13TeV/monoHiggs/ZpBaryonic/gridpacks_ZpBaryonic.py .
 cp cards/production/2017/13TeV/monoHiggs/ZpBaryonic/pdflist_4f_2017.dat ../../MetaData/.
 cp -r cards/production/2017/13TeV/monoHiggs/ZpBaryonic/cards/* cards/. 
 python gridpacks_ZpBaryonic.py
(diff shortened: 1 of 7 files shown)


qliphy:
+1 @agrohsje

agrohsje:
Maybe I am missing something. What is the reason to have this 5F PDF specified in the run_card: 315200 . And why adding it in the script: 315200 NNPDF31_lo_as_0130 101 ? 
My naive understanding is that you just use your default list and its systematics.

syuvivida:
From the proc card, you could see that this is a 4F LO process. For the 2015/2016 analysis, we have been using 4F NNPDF 3.0 LO PDFs. But in NNPDF 3.1, LO PDFs do not have 4F any more. Therefore, we could only use NNPDF 3.1 5F LO PDFs (315200) as our default. Since it's a 5F PDF set, I also need to add this addition line in pdf list_4f_2017.dat to include its PDF weight variation. 

One could also argue why we don't simply use NNPDF 3.1 NNLO or NLO 4F PDF sets. 
Note, this is one of the monoHiggs models that has large cross section change when using NLO/NNLO PDFs for this LO process (rather than LO PDFs). See [1][2]. The cross sections could drop to 50% of the old numbers when using NLO/NNLO PDFs at high Z' mass.

Therefore, in order to be consistent with our previous analysis, we were suggested by the GEN group to use NNPDF 3.1 LO PDFs as the default but keep the other PDFs in the list to study the systematic uncertainties. 

[1] https://indico.cern.ch/event/746826/contributions/3094599/attachments/1696355/2737409/20180731_MCI_Zbmonoh_keynote9.pdf
[2] 
Distribution comparison
https://github.com/syuvivida/darkmatter_figures/tree/master/monoH/ZpBaryonic/PDF_comparison

syuvivida:
BTW, if I understand correctly from this twiki instruction, since we are using NNPDF 3.1 LO PDFs, we should use CP2 tunes then?

agrohsje:
Didn't realize that NNPDF 3.1 LO has no 4 FS and totally forgot about the discussion we already had in GEN about that. Sorry. 
Yes. CP2 would be right.

Issue #1793, opened 2018-05-14
Separate cards and scripts repositories


AndreasAlbert:
**Proposal:**
put cards into repository A
put scripts into repository B

@kdlong suggested using git submodule so that we can keep the folder structure. Advantage: Can keep cards in genproduction subdirectory, so everything looks the same as it is now. Disadvantage: We have shower fragments and cards for different generators in different folders anyway, so we either have to use multiple submodules or change our directory structure to have all cards under a common root path.

Any downsides to this?

AndreasAlbert:
Step 1 would be to collect all cards in a central location in the repository. 
I'd propose to just create a new `genproduction/cards` directory, create subfolders in it for each of the generators and move the cards there. To make it maximally easy, I'd also move the python fragments folder into cards. A shell script that does this is [here](https://gist.github.com/AndreasAlbert/b31468b30d66aa0407f5c8a8e063d977) and the resulting folder structure would look like this:

```
.
├── bin
│   ├── FPMC
│   ├── GenValidation
│   ├── JHUGen
│   ├── MadGraph5_aMCatNLO
│   ├── MCFM
│   ├── Phantom
│   ├── Powheg
│   ├── Sherpa
│   ├── Starlight
│   └── utils
├── cards
│   ├── Alpgen
│   ├── BlackMax
│   ├── CalcHEP
│   ├── Charybdis
│   ├── CompHEP
│   ├── FPMC
│   ├── GenValidation
│   ├── JHUGen
│   ├── MadGraph4
│   ├── MadGraph5_aMCatNLO
│   ├── MCFM
│   ├── Phantom
│   ├── python
│   ├── QBH
│   └── Sherpa
└── test
    └── calculateXSectionAndFilterEfficiency
```
Step 2 is to pack up the new `cards` directory in a submodule.

AndreasAlbert:
Do we care about the commit history for the cards? If so, step 0 would be to run `git filter-branch` to extract the relevant part of the history.

perrozzi:
I would say yes we care about the card history

kdlong:
Do you have a branch where this is in progress already Andreas?

AndreasAlbert:
@kdlong Yes, my clone of genproductions is here: https://github.com/AndreasAlbert/genprod2,
the cards repo is here: https://github.com/AndreasAlbert/genprodCards

[...]

PR #2606, opened 2020-04-15
[EXO MonoHiggs hadronizer python] put back CP2 tunes for MC samples that use NNPDF 3.1 LO PDFs

syuvivida:
The hadronizers of CP2 tunes for the 5 Higgs decay modes were in the genproduction package originally but removed by another analyzer by accident (CP2 files were removed but CP3 were added for NLO PDFs). Now we just put these files back since we need to use these hadronizers for the production of ZpBaryonic MC samples.

### DIFF
--- a/python/ThirteenTeV/monoHiggs/pythia8_hadronizer_nomatching_HWWllnunu_cff_CP2.py
+++ b/python/ThirteenTeV/monoHiggs/pythia8_hadronizer_nomatching_HWWllnunu_cff_CP2.py
@@ -1,0 +1,31 @@
+import FWCore.ParameterSet.Config as cms
+from Configuration.Generator.Pythia8CommonSettings_cfi import *
+from Configuration.Generator.MCTunes2017.PythiaCP2Settings_cfi import *
+
+generator = cms.EDFilter("Pythia8HadronizerFilter",
+                         maxEventsToPrint = cms.untracked.int32(1),
+                         pythiaPylistVerbosity = cms.untracked.int32(1),
+                         filterEfficiency = cms.untracked.double(1.0),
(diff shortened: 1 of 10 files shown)


agrohsje:
If not strictly needed from your side, we usually don't add shower fragments to genproductions. The file can always be taken from e.g. mcm.

agrohsje:
Let me know if I can close.

syuvivida:
@agrohsje 
In addition to shower, the python file also includes the decays of Higgs boson into each specific channel; for each decay channel, we share the same gridpack production in which Higgs boson is not decayed. Our EXO MC contact also requires us to include the link to this python file in the google spreadsheet. Therefore, we do need to commit these configuration files.

agrohsje:
Do you maybe have a private folder you could use for that and link? I can merge but it is really unrelated to the default purpose of genproductions. We can also clarify with exo mc contacts.

DryRun:
Eiko pointed us to this discussion - OK, we can certainly stop adding these to genproductions. Our policy was for analyses with unique configurations to upload a single fragment + script for making the rest, for the sake of future reproducibility, but (a) this does seem redundant with MCM, which anyhow is a better reference, and (b) genproductions is huge already.

(Somewhat related: I guess https://github.com/cms-sw/genproductions/issues/1793 was abandoned?)

syuvivida:
Sorry, since it was a while ago and I am not surer if the same recipe applies now. My understanding is that we use 
***_CP2.py for 2017 MC 
and ***_CP2_PSweights_2018.py for 2018 MC, 
is this still correct?

agrohsje:
there is no problem to use ps weights in 2017 as well. we don't force people but it might be interesting to have.

syuvivida:
ok I see.

PR #2607, opened 2020-04-15
Updated Zp -> Gh cards.

jhakala:
This is an update to the cards for Hgamma signal samples. The parameters have been changed somewhat relative to the previously generated 2016 samples [1]. The model is identical to the earlier one, however. The earlier versions of the cards were submitted in a PR at [2].
[1] https://cms-pdmv.cern.ch/mcm/requests?tags=EXO-Zprime_Gh_hbb-JHakala&page=0&shown=127
[2] cms-sw#1319

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Zprime_Gh/Zprime_Gh_M1200/ZpHgamma_UFO-M1200_customizecards.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Zprime_Gh/Zprime_Gh_M1200/ZpHgamma_UFO-M1200_customizecards.dat
@@ -1,0 +1,2 @@
+set param_card MZp 1200
+set param_card WZp .1200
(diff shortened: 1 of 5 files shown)


agrohsje:
please remove these default settings: 
set group_subprocesses Auto
set low_mem_multicore_nlo_generation False
set loop_color_flows False
set gauge unitary
set complex_mass_scheme False
set max_npoint_for_channel 0
import model sm
define l+ = e+ mu+
define l- = e- mu-
define vl = ve vm vt
define vl~ = ve~ vm~ vt~

Is this 5FS or 4FS? maxjetflavor is 4? Confirm this is intended.
What is the reason not to go for additional partons? Wouldn't that give you a more realistic simulation of the process?

jhakala:
Thanks @agrohsje, I removed the default settings and changed maxjetflavor to 5. The reason was simply because the older cards had 4 = maxjetflavor (I am not sure why) but I think it would be good to set it to 5 instead as you suggest. Changes have been committed.

agrohsje:
thanks! what about additional jet and merging? 
cheers, alexander.

jhakala:
Hi Alexander, do you have any particular suggestions for those items (additional jets and/or merging)? I have kept these close to the original cards from the author (Bogdan) and it would be nice to stay consistent if possible/reasonable.

jhakala:
Just a ping on this request. Are things moving forward, or are there any further suggestions that should be dealt with?

agrohsje:
As you don't include additional jets, they will be generated by the parton shower, so the pT might be softer and cutting on jets may lead to some bias compared to your backgrounds. 
Having additional jets at ME level and including MLM jet merging, might give you a more realistic simulation in case you plan cutting on jets.

PR #2610, opened 2020-04-20
Updated genvalidation cards 20200420

menglu21:
Updated genvalidation cards 20200420, the main changes are the pdfs, and the process cards have been trimmed.


qliphy:
@menglu21   Thanks!  I notice you also uploaded several param_card.dat files, which may not be needed.  Otherwise it looks good to me.
And you also uploaded phantom files and other cards, are they related to genvalidation? or you want to upload them together.

@agrohsje

menglu21:
> @menglu21 Thanks! I notice you also uploaded several param_card.dat files, which may not be needed. Otherwise it looks good to me.
> And you also uploaded phantom files and other cards, are they related to genvalidation? or you want to upload them together.
> 
> @agrohsje

Hi Qiang,  I deleted the para cards. As for the phantom update, i think it would be better to merge it since it has been checked to be ok. And for the 5f cards, i think it's ok to merge them even if we don't need to generate new WA/ZA samples

qliphy:
+1  @agrohsje

menglu21:
Hi @agrohsje ,could you please comment on this thread, i'd like to update the script for validation using condor, but i don't want to accumulate too many things in the same commit here.

agrohsje:
1.) Do we want to keep this or change to false like we currently recommend: 
True	= pdfwgt for MLM 
2.) Would it make sense to take a 2.6.5 run_card as mg5 would generate in standalone mode and then adjust the relevant settings? Seems some options are missing. 
3.) Suggest to remove default lines like 
 set loop_color_flows False
set gauge unitary
set max_npoint_for_channel 0
set group_subprocesses Auto
set complex_mass_scheme False

menglu21:
Hi @agrohsje , these cards have been updated

PR #2612, opened 2020-04-21
Fix reading of stage2 grids

intrepid42:
Hi, I noticed during the ZjMiNNLO validation that these lines have an unwanted consequence when used with more complicated samples that require parallelstages.
They prevent powheg to read all the stage2 grid files during event generation, using only `pwggrid.dat = pwggrid-0001.dat` instead.

This leads to a poor estimate of the total cross section used for generation and a higher number of upper bound violations, that are also visible as non-unit weights when using `ubexcess_correct 1`.

Illustrating plot, compare with and without "stage2fix":
image

### DIFF
--- a/bin/Powheg/Templates/createTarBall_template.sh
+++ b/bin/Powheg/Templates/createTarBall_template.sh
@@ -13,6 +13,12 @@ rm -f $$WORKDIR/$$folderName'_'$$process'.tgz'
 
 cp -p $$WORKDIR/run_pwg.py $$WORKDIR/$$folderName
 
+if [ -e $WORKDIR/$folderName/pwg-0001-stat.dat ]; then
+  cp -p $WORKDIR/$folderName/pwg-0001-stat.dat $WORKDIR/$folderName/pwg-stat.dat
+fi
+if [ -e $WORKDIR/$folderName/pwg-st3-0001-stat.dat ]; then
+  cp -p $WORKDIR/$folderName/pwg-st3-0001-stat.dat $WORKDIR/$folderName/pwg-stat.dat
+fi
 
 FULLGRIDRM=`ls $${WORKDIR}/$${folderName} | grep fullgrid-rm`
 FULLGRIDBTL=`ls $${WORKDIR}/$${folderName} | grep fullgrid-btl`
--- a/bin/Powheg/Templates/createTarBall_template.sh
+++ b/bin/Powheg/Templates/createTarBall_template.sh
@@ -13,11 +13,11 @@ rm -f $$WORKDIR/$$folderName'_'$$process'.tgz'
 
 cp -p $$WORKDIR/run_pwg.py $$WORKDIR/$$folderName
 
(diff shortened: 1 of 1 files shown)


agrohsje:
Hi @intrepid42 @covarell ,
does one of you know why these lines were included? There is no reason to remove the index, or do I miss something? Powheg will just read in all of them. There is no external merging needed. 
There are more ID depending lines:
https://github.com/cms-sw/genproductions/blob/45f04fc3575df27d4398d4b9556c75a0e02c169d/bin/Powheg/Templates/createTarBall_template.sh#L22
https://github.com/cms-sw/genproductions/blob/45f04fc3575df27d4398d4b9556c75a0e02c169d/bin/Powheg/Templates/createTarBall_template.sh#L66
Cheers, Alexander.

covarell:
L16 is wrong, it should check whether pwggrid.dat exists, and if not, use pwggrid-0001.dat. 

This was introduced at the LSF times, because for some processes stage 3 was so long that it never made it to the merging, not even with the longest queues. In those cases, the partial results of job 1 was used. 

But indeed it should not be done in all cases. I am not sure when this bug was introduced...

intrepid42:
With most recent powheg it seems that no merging of the stage2 grids happens in any preparation stage.
Instead, all the `pwggrid-*.dat` files are read at the beginning of event production iff there is no `pwggrid.dat`.

covarell:
Ok, this I did not know. Then the PR is fine for me.

covarell:
Hm, I am confused now...
First of all, why the $$ is replaced by single-$? I thought that was needed to correctly parse the .sh into the main python script. Is it redundant instead? @Hooooon12 
Second, it seems to me we are replacing again pwg-stat with pwg-stat-0001 in any case. The if should be "if there is no pwg-stat, then use pwg-stat-0001".
Am I missing something?

intrepid42:
Sorry, I just copied this over from my current working directory which is still using the version of Powheg scripts from a few months ago. changed `$ -> $$`.

We can modify the logic but there should not be any `pwg-stat.dat` after `paralellstage` runs.

agrohsje:
Hi @intrepid42, sorry, I need to ask one more question. Can you clarify the pwg-stat.dat logic? Isn't the grid logic also true for the stat? I.e. what happens if you have multiple pwg-stat-000x.dat files?

intrepid42:
Hi Alex, the `pwg-stat.dat` is not read by Powheg but just by the our CMS scripts to write the cross section with higher precision into the LHE header.

* `pwg-0001-stat.dat` and `pwg-st2-0001-stat.dat` are identical files produced at the end of stage2. It is statistically independent from the other `pwg-st2-000*-stat.dat`
* `pwg-st3-0001-stat.dat` is produced at the beginning of stage3 and contains the combined cross section of all the stage2 jobs. All `pwg-st3-000*-stat.dat` are identical

So we need to copy the combined cross section from stage3 to `pwg-stat.dat` to write that into the LHE header

agrohsje:
Thanks Markus. Do you mind updating ul2019 as well?

PR #2613, opened 2020-04-21
Initial pull request for LQ_cmutau

Physicsjong1:


bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/LQ_cmutau/LQ_cmutau_proc_card.dat
```diff
@@ -0,0 +1,41 @@
...
+import model sm
+define p = g u c d s u~ c~ d~ s~
+define j = g u c d s u~ c~ d~ s~
+define l+ = e+ mu+
+define l- = e- mu-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+import model LeptoQ_UFO
+generate p p > t t~, t > b w+, t~ > c~ ta+ mu-
+add process p p > t t~, t > c ta- mu+, t~ > b~ w-
```
swertz, line 40: Since you don't decay the W's explicitly here and don't use MadSpin, it means they will have to be decayed by Pythia. This means the spin correlation information will not be present.
Is that expected?

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/LQ_cmutau/LQ_cmutau_run_card.dat
```diff
@@ -0,0 +1,282 @@
...
+  False     = gridpack  !True = setting up the grid pack
+  -1.0 = time_of_flight ! threshold (in mm) below which the invariant livetime is not written (-1 means not written)
+  3.0 = lhe_version       ! Change the way clustering information pass to shower.        
+  True = clusinfo         ! include clustering tag in output
+  average =  event_norm       ! average/sum. Normalization of the weight in the LHEF
+
+#*********************************************************************
+# Matching parameter (MLM only)
+#*********************************************************************
+ 1 = ickkw            ! 0 no matching, 1 MLM
```
swertz, line 72: Shouldn't this be 0 if you don't have any matching?

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/LQ_cmutau/LQ_cmutau_proc_card.dat
Physicsjong1: To reply to the major question, I am not well understanding the generation process. Is a customize_card is necessary? I am just thinking naively that the extramodel we provided will contain all the parameters needed for the process. I would like to know what a customize_card or definition of the signal parameters does.

Yes, it is what we expected. We use inclusively W decay samples and are not considering the use of spin correlation.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/LQ_cmutau/LQ_cmutau_run_card.dat
Physicsjong1: Oh, you are right... I missed that point. I will correct it since we don't do matching.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/LQ_cmutau/LQ_cmutau_proc_card.dat
swertz: > We use inclusively W decay samples and are not considering the use of spin correlation.

Ok, thanks for clarifying.

swertz:
> I am not well understanding the generation process. Is a customize_card is necessary? I am just thinking naively that the extramodel we provided will contain all the parameters needed for the process. I would like to know what a customize_card or definition of the signal parameters does.

If the parameters are fixed to what you need them in the model, I would assume it is OK...

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/LQ_cmutau/LQ_cmutau_run_card.dat
swertz: Ok! Don't you need to regenerate the gridpack then?

Physicsjong1: I have updated the gridpack and placed it at the cernbox. Thank you!

agrohsje:
1.) Please remove this stuff : 
#*                  *        * *        *                   *
#*                *                       *                 *
#*                                                          *
#*                                                          *
#*         VERSION 2.6.3.2               2018-06-22         *
#*                                                          *
#*    The MadGraph5_aMC@NLO Development Team - Find us at   *
#*    https://server06.fynu.ucl.ac.be/projects/madgraph     *
#*                                                          *
#************************************************************
#*                                                          *
#*               Command File for MadGraph5_aMC@NLO         *
#*                                                          *
#*     run as ./bin/mg5_aMC  filename                       *
#*                                                          *
#************************************************************
set default_unset_couplings 99
set group_subprocesses Auto
set loop_color_flows False
set gauge unitary
set complex_mass_scheme False
set max_npoint_for_channel 0
import model sm
define l+ = e+ mu+
define l- = e- mu-
define vl = ve vm vt
define vl~ = ve~ vm~ vt~

2.) Please use consistently 5 FS. You treat the b as massless, no ? 

3.) Please add the W decay. I think that would be good if others want to use. I fully agree with the original comment from Sebastian. 

4.) Please remove this 
#
#systematics = systematics_program ! none, systematics [python], SysCalc [depreceted, C++]
#['--mur=0.5,1,2', '--muf=0.5,1,2', '--pdf=errorset'] = systematics_arguments ! see: https://cp3.irmp.ucl.ac.be/projects/madgraph/wiki/Systematics#Systematicspythonmodule
# Syscalc is deprecated but to see the associate options type'update syscalc'

5.) Please add the last line to the block : 
  lhapdf = pdlabel ! PDF set
 $DEFAULT_PDF_SETS = lhaid

6.) The purpose of the customize card is to adjust default parameters of the param card on the fly. this could be ok the top mass, the lq mass or their width.  Don't you need that?

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/LQ_cmutau/LQ_cmutau_proc_card.dat
minerva1993: I thought I already pointed out W decay and its spin information in the internal meeting? As far as I remember, attaching madspin had some problems in this model. Can you make further tests?

minerva1993:
I have a question about additional partons.
As I mentioned some times ago, because this process is basically ttbar production, ttjj events have ~4% of the contribution (you can see the measurement from ttbb/ttjj paper with l+j as well as dilepton channel). Also, ttj events will have a non-negligible contribution (larger than ttjj, of course). So without additional partons in the matrix element level, you may lose signal events with higher jet multiplicity. (I'm not sure about s/b though.) Also, it would be fair to include additional partons since ttbar background sample is generated with additional parton [1] in realistic analysis. (Please correct me if something is wrong) Unless, I think the final state of signal process is similar to tW channel, rather than ttbar. 

But in this case you should turn on jet matching, and tune xqcut since LQ mass is ~ 1 TeV.

[1] I guess this is the line: https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/production/2017/13TeV/TT_hvq/TT_hdamp_NNPDF31_NNLO_ljets.input#L73
and associated to pythia as 'POWHEG:nFinal = 2': https://cms-pdmv.cern.ch/mcm/public/restapi/requests/get_fragment/TOP-RunIIFall18wmLHEGS-00001

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/LQ_cmutau/LQ_cmutau_proc_card.dat
swertz: Can't you include the W decay in the madgraph process without using madspin?

Physicsjong1: I can add the explicit W decay modes in the madgraph process.
What I want to ask is,
since there are two W decay modes, hadronic or leptonic, if we request MC samples, then are two samples produced, respectively?

minerva1993: Isn't it possible?
define allw = u c d s u~ c~ d~ s~ ve vm vt e- mu- ve~ vm~ vt~ e+ mu+ t b t~ b~ ta- ta+
generate p p > t t~, (t > w+ b, w+ > allw allw), (t~ > w- b~, w- > allw allw)

swertz: Just write `, w- > all all` and `, w+ > all all` to include everything in one sample.

Physicsjong1: Then I will add w decay to "all" in our proc card and update. Thank you :)

Physicsjong1: Then I will add w decay to "all" in our proc card and update. Thank you :)

minerva1993: > Just write `, w- > all all` and `, w+ > all all` to include everything in one sample.

I think this will create un-intended diagrams such as w+ -> w+ h. (I see some warning in MG5) Is it ok to proceed? (Even though its cross section would be very small) For sure, in madspin, branching rations will be recalculated...

Physicsjong1:
To answer to each question,
> 1.) Please remove this stuff : 
> ...
> import model sm
> define p = g u c d s u~ c~ d~ s~
> define j = g u c d s u~ c~ d~ s~
> define l+ = e+ mu+
> define l- = e- mu-
> define vl = ve vm vt
> define vl~ = ve~ vm~ vt~

Removed it and updated the proc card.

> 2.) Please use consistently 5 FS. You treat the b as massless, no ?

I have changed into 5 flavour scheme. And then, is the "p" defined with containing "b" and "b~"?

> 3.) Please add the W decay. I think that would be good if others want to use it. I fully agree with the original comment from Sebastian.

I added inclusive W decay on the madgraph card.

> 4.) Please remove this

Removed it and updated the run card.

> 5.) Please add the last line to the block :
> lhapdf = pdlabel ! PDF set
> $DEFAULT_PDF_SETS = lhaid
> $DEFAULT_PDF_MEMBERS = reweight_PDF

Added the line to the block.

> 6.) The purpose of the customize card is to adjust default parameters of the param card on the fly. this could be ok the top mass, the lq mass or their width. Don't you need that?

We don't have a plan to adjust the mass or width of LQ, so we follow the parameter settings of the model.

swertz:
Thanks for the update! @agrohsje do you have additional comments?

agrohsje:
What about the comment from @minerva1993 ? Is the signal good enough to compare with SM ttbar. Of course you get additional jets from the PS, but the question is,will you bias your selection if the jet modeling is very different between signal and e.g. SM ttbar background?

swertz:
Not sure about that... @minerva1993 points to the powheg card for the ttbar background, i.e. ttbar @ NLO without explicit extra partons (so only the first jet is described at ME level, at LO, all the others are from the PS). I guess if you don't explicitly use the extra jets in the analysis it should not have a large impact?

On the other hand, looking at [FCNC ttbar cards](https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TT_FCNC/refcards/madgraph_card/TT_no_top_decay_LO_proc_card.dat) there were up to 3 additional jets at ME level.

agrohsje:
It might not be strictly needed. It fully depends on how the analysis is designed. My proposal was mainly that it shouldn't be too much additional work will it might have some benefits. But really just a proposal. I am also fine with merging.

minerva1993:
Since the cross section of signal process is calculated in effective theory, it is 'fixed' once mass and width of leptoquark are determined. So if one has more partons in ME level, the acceptance will get better when applying jet selection. (This can be important because the theory cross section is very small, order of ~ 10^(-6).)

Physicsjong1:
Thank you for the proposal @agrohsje. Our analysis is to calculate the upper limit of the process and our team has discussed for the analysis design. Additional jets could increase the acceptance but it requires much work for parton - jet matching. So we decided to progress with not applying additional jets and this may affect the higher jet multiplicity region. We know the limitation of no additional jets, but, in the first step of searching leptoquark at c mu tau channel, we would like to focus more on analysis. However, if the requirement of additional jets is mandatory, we would do that.

swertz:
> However, if the requirement of additional jets is mandatory, we would do that.

@agrohsje can correct me if I'm wrong, but I don't think it is.

Is this good to go then?

Physicsjong1:
> Is this good to go then?

We are ready to go.

swertz:
@agrohsje @Physicsjong1 
After checking the actual gridpack, with the current cards we have some inconsistencies:
- 4FS/5FS: The card was updated to have `maxjetflavor=5` and `asrwgtflavor=5`, but:
   - the b quark mass in the model is not set to zero
   - the PDF used (`325500=NNPDF31_nnlo_as_0118_nf_4_mc_hessian`) is a 4FS one
   - the b/b~ quarks are not included in the proton (I see `define p = g u c d s u~ c~ d~ s~` and `define j = g u c d s u~ c~ d~ s~` in `proc_card_mg5.dat` in the gridpack). 

It should be consistently decided between 4FS and 5FS.
- The top quark mass in the model is 172.0 instead of 172.5 used centrally in CMS. Please update the model, or better yet, add a customize card (e.g. as in [here](https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/HeavyNeutrino_Dilepton_NLO/HeavyNeutrinoToDilepton_Tchannel_NLO/HeavyNeutrinoToEMu_Tchannel_M400_NLO/HeavyNeutrinoToEMu_Tchannel_M400_NLO_customizecards.dat)) to fix that value.

agrohsje:
@swertz @Physicsjong1 This is exactly what I asked above:
2.) Please use consistently 5 FS. You treat the b as massless, no ?
Please make another pr to fix. Was there already something submitted?

Physicsjong1:
For applying LQ_cmutau cards for gridpack, the cards at "process/madevent/Cards/" in the gridpack are not updated.
The LQ_cmutau cards were just in "InputCards" directory.
Should I modify the gridpack_generation.sh file?
Also, I can add the customize card to ignore b mass and change top mass.
And, is there a recommended PDF for 5FS?

agrohsje:
Which cards didn't get considered? If you start from genproductions and follow the TWiki [1] everything will work. 
If you run in 5FS and specify $DEFAULT_PDF_SETS = lhaid you will get the right one. 
If the model is per default 4FS, please add 
p = p b b~
j = p 
to your proc_card.dat . You can use cumstomize_card.dat to adjust the b and top mass. 

[1]  https://twiki.cern.ch/twiki/bin/view/CMS/QuickGuideMadGraph5aMCatNLO

Physicsjong1:
I found the discrepancy when running the ./generate_gridpack.sh.
All of the values are the same but the only lhaid and maxjetflavor are different.
Is there a possibility that the model changes them?
I cannot figure out where the values change.

agrohsje:
Hi @Physicsjong1 , the problem is that your model has no restriction card with 5FS. You cannot use the customize_card to adjust the b mass to switch from 4FS to 5FS as the process was already generated at that point.  You can either adjust the model and upload the new version and use this one, or switch for simplicity consistently to 4FS.

Physicsjong1:
We decided to go with 4FS with b mass.
I have created a [new pull request](https://github.com/cms-sw/genproductions/pull/2642) with changing the max flavours on the run card.

PR #2614, opened 2020-04-22
Updated genvalidation script using cmsconnect

menglu21:
update the genvalidation script using cmsconnect, the compare script is to be updated


agrohsje:
Hi Meng. What is the reason not to just update the old tools? Why you want a separate directory?

menglu21:
> Hi Meng. What is the reason not to just update the old tools? Why you want a separate directory?

Hi Alexander, the updated script use sandbox to submit the jobs, so the script has been changed a lot. It's certainly fine to merge it with the previous directory.

agrohsje:
Hi @menglu21 . Thanks so much. As this will not affect central production I will merge. We can always submit fixes if needed.

PR #2615, opened 2020-04-22
update aQGC VVjj hadronic 2017 madgraph cards

stalbrec:
This updates the used SMEFT UFO model, renames the processes in order to be consistent with two processes added here as well (ZNuNuZJJnoB and ZBBZJJnoB).

The new UFO model was discussed in this [HN](https://hypernews.cern.ch/HyperNews/CMS/get/generators/4671.html).


agrohsje:
Any reason you add these to the proc_card: 
set group_subprocesses Auto
set complex_mass_scheme False
All default, no. 
Otherwise +1 @qliphy .

stalbrec:
These are just remnants from when i was adopting the cards for VVjj semileptonic.
If they are the default, I can remove those lines if you like.

stalbrec:
Should I remove those lines?

agrohsje:
Right. If default please remove. Qiang can merge afterwards. 
+1 from my side @qliphy

stalbrec:
I checked - they are all default.
I removed the lines.

PR #2617, opened 2020-04-27
fix defaultPDF

Hooooon12:
Hi, this is a temporary PR to fix default pdf setting in runGetSource function, from old one to the UL setting.

### DIFF
--- a/bin/Powheg/run_pwg_condor.py
+++ b/bin/Powheg/run_pwg_condor.py
@@ -278,7 +278,7 @@ def runGetSource(parstage, xgrid, folderName, powInputName, process, noPdfCheck,
 
     fourFlavorProcesses = ["ST_tch_4f", "bbH", "Wbb_dec", "Wbbj", "WWJ"]
     template_dict["isFiveFlavor"] = int(process not in fourFlavorProcesses)
-    template_dict["defaultPDF"] = 306000 if template_dict["isFiveFlavor"] else 320900
+    template_dict["defaultPDF"] = 325300 if template_dict["isFiveFlavor"] else 325500
 
     DYNNLOPS = ["Zj", "Wj"]
     if process in DYNNLOPS:


agrohsje:
Dear @Hooooon12,
can you comment on why we want that change in master? The setting works in UL? 
Thanks, Alexander.

Hooooon12:
Hi Alexander,
sorry for the late reply. This commit is to fix the disagreement in default PDF setting
between the main code: 
https://github.com/cms-sw/genproductions/blob/0635692e97dbea26a9ae344e9df53eb2d84c514e/bin/Powheg/run_pwg_condor.py#L757-L760
and a running function:
https://github.com/cms-sw/genproductions/blob/0635692e97dbea26a9ae344e9df53eb2d84c514e/bin/Powheg/run_pwg_condor.py#L281
in the master branch.

I saw the same issue in several branches, but I did PR for master now as I wasn't sure which branches need this change.
Does it seem ok? or Do we need this in other branches?
@covarell 

Best, Jihun.

agrohsje:
@covarell @qliphy I forgot what we agreed on. Did we want 325300 and 325500 as default PDF in both master and ul2019 or just in ul2019? I thought we agreed to keep 306000 and 320900 as default for master.

qliphy:
@agrohsje    For MG in master, we have already changed the default PDF to 325300 and 325500:
https://github.com/cms-sw/genproductions/blob/master/MetaData/pdflist_5f_2017.dat#L2
https://github.com/cms-sw/genproductions/blob/master/MetaData/pdflist_4f_2017.dat#L2

agrohsje:
@qliphy my worry is mainly ttbar where powheg is heavily used. i am fine with moving as for mg5 but we should warn people in this weeks mccm. 
+1 for merging

qliphy:
" i am fine with moving as for mg5 but we should warn people in this weeks mccm."
Agreed.

PR #2618, opened 2020-04-27
bugfix in filename for running xml_check_and_patch

llechner:
@efeyazgan 
fixing the filename in the arguments of the xml check script


efeyazgan:
thanks

PR #2621, opened 2020-04-30
darkHiggs to WWjjlnu: proc datacard redefine multiparticle to be safe

SenneVanPutte:
The name of the lepton multiparticles was changed:
l => myl
vl => myvl
To be sure nothing preexisting is overwritten, carrying unwanted changes.

The gridpack's where created with these datacards.


agrohsje:
I don't see the need for this PR. Can you please comment where a problem should appear?

SenneVanPutte:
I don't believe there really was a problem with this.
I just experienced some issues when I was producing the tar balls, while debugging I created this change.
Then when things started working I forgot to change it back.  
So it is just for completeness, if they are needed in the next step (sample production).

agrohsje:
Not to cause confusion, I would prefer not to merge and close this PR. It is ok to have different cards in the gridpacks. That's no problem. The cards as they are now in genproduction are bug-free and can be picked-up next time as they are.Are you ok with this?

SenneVanPutte:
Ok that is fine by me!

PR #2623, opened 2020-05-01
add_time_of_flight

matt-komm:
so far this only adds time_of_flight to the runcmsgrid_LO.sh script; let me know the best place to fix the runcmsgrid_NLO.sh script as well


agrohsje:
Hi Matthias, thanks for NLO you can put it here: 
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/runcmsgrid_NLO.sh#L131
For NLO I suggest to move it here:
https://github.com/cms-sw/genproductions/blob/20fe01aad5f5cc46e6ce0890f3988c3075f1f00b/bin/MadGraph5_aMCatNLO/runcmsgrid_LO.sh#L171
When Madspin is invoked the file will have a different name. This is currently not reflected in your PR.

agrohsje:
Hi @matt-komm . Do you mind updating the PR so that we can merge? Thanks so much!

PR #2624, opened 2020-05-01
Add DMsimp spin0 cards for monojet

z4027163:
Add several mass points for spin-0 DM signal for monojet process.


qliphy:
Please only upload cards for an example point.

Moreover, please remove lines as:
https://github.com/cms-sw/genproductions/blob/69a9b7f466c9f00ee7b2dc057c8aa8af2541fe14/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/monojetv/dmsimp/DMSimp_monojet_NLO_Scalar/Scalar_MonoJ_NLO_Mphi-MX_Mchi-MY_gSM-1p0_gDM-1p0_run_card.dat#L270-L282

z4027163:
@qliphy I have made the change according to the requirements.

qliphy:
Your run_card.dat seems to be a LO one, please change it to NLO.

dr-stringfellow:
@qliphy, the process is loop-induced. Madgraph chooses the LO run card format automatically. We need both LO run card and hadronizer.

qliphy:
@dr-stringfellow  Ok, thanks.  But then the cards name containing "NLO" look confusing.

PR #2626, opened 2020-05-06
- Temporary workaround for condor_chirp not being available in CMS

### LINKED ISSUE #2625: CMS Connect CODEGEN transfer methods broken
Hello, 

We use 2 different transfer mechanisms in the CODEGEN generation via CMS Connect. The first uses XRootD to the OSG stash service. The second one relies on `condor_chirp`, which has been supported in CMS. 
The XRootD service has undergone major changes in OSG, so it is not working at present. Also, last week, CMS shipped a new implementation of singularity in the Global Pool, which broke `condor_chirp`, not only for CMS Connect, but also other areas like production. 

There will be a hotfix for `condor_chirp`, but in the meantime, I will be providing a quick workaround to still be able to create gridpacks by using the native condor transfer mechanism instead.

khurtado:
Fixes #2625


qliphy:
Thanks @khurtado !

PR #2629, opened 2020-05-07
fix content2 variable name

llechner:
bugfix: content -> content2 variable name


llechner:
@efeyazgan 
Hi Efe, this line creates problems when validating powheg+madspin samples due to a bug in the naming of the variable. Due to the wrong naming it basically copies the runcmsgrid.sh into the external_tarball directory, however the files are different if the external_tarball is from powheg.

PR #2630, opened 2020-05-08
add HH to WWWW fragment

AndrewLevin:
This adds a HH to WWWW fragment with all W boson decays turned on.


agrohsje:
usually we don't recommend having pythia fragments here . can you comment on why you want this pr?

AndrewLevin:
what do you mean? it is in the same place as the other ones

agrohsje:
i was talking about why having shower fragments at all in genproduction and not just in mcm. but i fully agree, you are not the first committing here. :-)  let me merge.

PR #2631, opened 2020-05-09
fix reweighting for lo and nlo when using madspin

agrohsje:
changed order of madspin and rwgt as done in pr2170, can be closed 
madspin removed header for me weights -> added back
systematic evaluated only once -> modified run_card in gridpack helpers
compared output w and w/o madspin, w and w/o rwgt, lo and nlo at lhe level: no problems observed


qliphy:
Great, thanks!

PR #2636, opened 2020-05-12
add hh to wwww cards

AndrewLevin:
Add mg5_aMC hh to wwww cards for 2016 and 2017/2018.

Copied from https://github.com/cms-sw/genproductions/tree/dac0b0a9a49aeaddc656529d275e5d426effce44/bin/MadGraph5_aMCatNLO/cards/production/13TeV/NonRes_hh/GF_HH_node_SM and changed the PDF.


qliphy:
Please remove banner from the proc cards, and remove lines as below in run_card
https://github.com/cms-sw/genproductions/blob/fec5a962e8a2e854363378ba0a9e0a9e83cae23c/bin/MadGraph5_aMCatNLO/cards/production/13TeV/HHToWWWW/HHToWWWW_run_card.dat#L269-L280

AndrewLevin:
done

qliphy:
+1  @agrohsje

agrohsje:
Please remove these default lines? 
set group_subprocesses Auto
set complex_mass_scheme False
import model sm
define l+ = e+ mu+
define l- = e- mu-
define vl = ve vm vt
define vl~ = ve~ vm~ vt~

Is the model 4FS or 5FS. 306000 in 5FS but maxjetflavor 4. 

The directory is hh -> w w w w, but i asumme this will be done in pythia8. There is no decay added.

AndrewLevin:
The director and process name have been updated and the default lines have been removed.

Normally a model is not specific to 4f or 5f, right? I have updated the 2016 run card to use a specific 4f PDF, and I have updated the 2017 run card to use $DEFAULT_PDF_SETS.

agrohsje:
@AndrewLevin It is in the sense that the default param_card is either mb=0 or not. Can you check that all is 4FS and confirm?

AndrewLevin:
the b mass is not 0 in the default param card, so I have updated the cards to use the 5 flavor scheme

AndrewLevin:
actually, now that I think about it, b mass not 0 should correspond to 4f scheme, right?

agrohsje:
0 b mass means 5th flavor (i.e. the b) is active. So if it is not 0, the b will not be included in the proton and it is 4FS.

AndrewLevin:
ok, it is correct in the last commit

agrohsje:
please add also : 
 '\$DEFAULT_PDF_MEMBERS = reweight_PDF'
to the run_card

AndrewLevin:
would this have any effect? it seems like it will be just removed here

https://github.com/cms-sw/genproductions/blob/1103275db61a494716315057f2415b2da63050ba/bin/MadGraph5_aMCatNLO/Utilities/gridpack_helpers.sh#L16

agrohsje:
agreed. maybe you can quickly check once you have the gridpack? 
will merge.

PR #2637, opened 2020-05-19
Uploading cards for X to AA to 4 Photons

SClarkPhysics:
Uploading cards for the process X->aa->4gamma. These are sample cards for a single mass point, X 500 GeV, a 5 GeV


agrohsje:
Why is ickkw 0 ? Why is the top width 10^5 ? What about the width of 90000054 and 54 ?

SClarkPhysics:
Fixed several issues. 
Changed ickkw to 1
computed widths for 54 and 900054
Added dm inputs

Top width is so large in order to make the gg -> X production pointlike.

Was able to successfully generate gridpack using  gridpack_generation.sh and using these cards

agrohsje:
You are about to push 5000 files. Please correct.

SClarkPhysics:
Apologies. Unnecessary files have been removed.

agrohsje:
+1 
@qliphy feel free to merge if ok with you. removed systematics tools myself.

PR #2640, opened 2020-05-27
Cards for SM V(qq) gamma 

AndreasAlbert:
These cards are for SM W+gamma and Z+gamma production with enforced decays of the W/Z to hadrons.

(My previous PR #2639 did not have the right card versions, so I closed and recreated after fixing it)


qliphy:
You may add pta150 in the card names?

AndreasAlbert:
Thanks for checking the cards! I updated the name as per your suggestion.

qliphy:
+1 @agrohsje

PR #2641, opened 2020-05-27
Cards for NLO Gamma + 1jet pt binned

AndreasAlbert:
I added the cards for the lowest photon pt bin 100-250. The higher bins can be generated using the `make_cards.py` script.


qliphy:
+1 @agrohsje

agrohsje:
Hi Andreas,
why do you specify:  
292201   = PDF_set_min      ! First of the error PDF sets 
 292302   = PDF_set_max      ! Last of the error PDF sets ?
Will be done via : https://github.com/cms-sw/genproductions/blob/master/MetaData/pdflist_5f_2017.dat#L35
We usually removed this PDF weights as we had them twice. Now I changed the logic that PDF's get only evaluated by the systematics module, but nothing removed anymore. 
Cheers, Alexander.

AndreasAlbert:
Thanks for checking! The min/max parameters are leftover from copy/pasting the 2016 cards. If I understand correctly, they are ignored in our workflow, right? I can still remove them for consistency, of course.

PR #2642, opened 2020-05-30
LQ cmutau run card changes

Physicsjong1:
Changes flavour scheme from 5Fs to 4FS


swertz:
Don't forget to also update the top quark mass to the CMS central value using a customize card...

swertz:
Looks goot to me!

Physicsjong1:
I set the lhaid to $DEFAULT_PDF_SETS and is the PDF automatically selected as PDF for 4FS one?
Without that, the value is all the same for the both.

Physicsjong1:
@agrohsje I missed to check the top width when changing the top mass.
Yes, I think it should be changed also. Could you give me the top width? I will commit on the customise card.

agrohsje:
set param_card DECAY   6 auto
should give you the right value as everything is anyway LO 2 - > 2

Physicsjong1:
> set param_card DECAY 6 auto
Does this work for both, the top mass and width?
Then, the mass MT in param card may be useless.

agrohsje:
Wait the line you have now sets the mass. The line that I suggested to add will trigger mg5 to calculate the width for that mass. The settings are independent.

PR #2643, opened 2020-06-01
Remove fakevirt for parallel stages 2 and 3

AndrewLevin:
Shouldn't fakevirt be used in parallel stage 1 but not the later parallel stages? That is my understand from reading the POWHEG manual for several processes. For example: [manual-BOX-VBF_Wp_Wp.pdf](https://github.com/cms-sw/genproductions/files/4709336/manual-BOX-VBF_Wp_Wp.pdf)

### DIFF
--- a/bin/Powheg/run_pwg_condor.py
+++ b/bin/Powheg/run_pwg_condor.py
@@ -321,10 +321,6 @@ def runEvents(parstage, folderName, EOSfolder, njobs, powInputName, jobtag, proc
         if not 'manyseeds' in open(inputName).read() :
             runCommand("echo \'manyseeds 1\' >> "+ inputName)
 
-        if not 'fakevirt' in open(inputName).read() :
-            if process != 'b_bbar_4l':
-                runCommand("echo \'fakevirt 1\' >> "+inputName)
-    
     runCommand('cp -p ' + folderName + '/powheg.input ' + folderName + '/powheg.input.' + parstage)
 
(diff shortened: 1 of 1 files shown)


agrohsje:
I agree. @covarell maybe you can double check?

kdlong:
Do we really ever want to force this to be included? In some cases it's beneficial, but I thought the authors mostly advised against it in processes I've looked at.

In general I would rather require sensible settings in the cards than silently overwrite them.

agrohsje:
Indeed! I was wondering the same. Should we really include it as default. Maybe @covarell remembers the motivation.

covarell:
I remember this was added by Eiko for ttH because it did not work otherwise. I think that the setting can be replaced with if (process == ttH) etc. for stage 1 only

covarell:
I found the precise comment on the Twiki:
"The authors of ttH gave the following suggestions regarding the generation of gridpacks and LHE files.
     "We recommend using the option "fakevirt" for the generation of the grids. 
      To this end, activate the fakevirt option, but set ncall2 to zero.The program
      will then generate grids quickly, but stop after that stage. After this step is completed, 
      de-active fakevirt (this is important to get true NLO predictions), set ncall2 to 
      the number of your choice, and re-run pwhg_main."

So Andrew's correction is definitely OK, and should be merged, then we should decide process by process where we should activate fakevirt, (in a different PR)

PR #2645, opened 2020-06-04
Update scalar LQ cards for high-coupling LQ -> btau analysis

IzaakWN:
This PR updates an existing set of cards for EXO-19-016. The cards have been presented and approved before in context of EXO-17-029, but have never actually been used for production. The main changes are

1. Renamed directories and cards to more descriptive `(Pair|Single|NonRes)ScalarLQToBTau`.
2. Fix scale for resonant pair and single production to LQ mass (as recommended by theorists, and done in the vector LQ samples).
3. Update the script to create mass-lambda points. (lambda = LQ-btau coupling strength)

Links:
* `LQ_UFO` Model:
    https://cms-project-generators.web.cern.ch/cms-project-generators/LQ_UFO.tar.gz
* Previous presentations for EXO-17-029:
    https://indico.cern.ch/event/675004/#2-leptoquark-to-tau-b-mc-reque
    https://indico.cern.ch/event/678334/#5-3rd-gen-single-lq-follow-up
* AN-19-019, EXO-19-016:
    http://cms.cern.ch/iCMS/user/noteinfo?cmsnoteid=CMS%20AN-2019/019
    http://cms.cern.ch/iCMS/analysisadmin/cadilines?line=EXO-19-016


qliphy:
+1 @agrohsje

IzaakWN:
Hi @agrohsje, the new commit removes any of the redundancies regarding the PDF and scales in `run_card` and `customize_card` and `ProduceOtherMassPoints.sh` now edits those files directly with `sed` when looping over mass points.

IzaakWN:
I just had a talk with a MC expert. Regarding `NNPDF23_lo_as_0130_qed` (247000), this was a misunderstanding on my part. All the existing 2016 LQ samples use `NNPDF30_lo_as_0130` (263000), so I will use this instead.

Regarding the systematic variations, I noticed that nanoAOD seems to only save 33 variations of `PDF4LHC15_nnlo_30_pdfas`. Do you know if we can change this somewhere along the production, or can we use `PDF4LHC15_nnlo_30_pdfas` variations on top of the nominal PDF (i.e. 263000 or 315200)? Otherwise, I can include the variations by running LHAPDF.

agrohsje:
You were only partially wrong in what you were saying. It is indeed the PDF used for the Monash tune. It is just that in 2016 we used NNPDF3.0 and not 2.3. That is exactly why I was asking this question above. 
It is not directly steerable in production. You might need to run NANO-AOD yourself. But that is a different story beyond GEN. We should just make sure that all relevant info/variation is in MINI-AOD.

IzaakWN:
Thanks. I could change `1` to `101` in line 38 and 40 in [`MetaData/pdflist_5f_2017.dat`](https://github.com/IzaakWN/genproductions/blob/83e32dba569d1b032d90d2f3b8bea0373fd2e362/MetaData/pdflist_5f_2017.dat#L38-L40). This seems to include the right systematic variations in the lhe file produced by `runcmsgrid.sh` in the gridpack if I use the current setup. If you want I can include it in this PR, otherwise I'll can use it locally to get those variations.

Besides that is there anything else that should be improved in this PR?

agrohsje:
Right, you can locally change line 40 to 101. You might also set line 35 to 1 to avoid unnecessary computations.

agrohsje:
Hi @IzaakWN. Please update.

IzaakWN:
I made those changes to `MetaData/pdflist_5f_2017.dat` locally, so I think there's nothing more to change to this PR, is there?

PR #2646, opened 2020-06-04
Intermediate H+ -> taunu: added H- in the proc card

slehti:
Added H- in the intermediate H+->taunu proc card


qliphy:
+1 @agrohsje

PR #2647, opened 2020-06-05
Template MG cards for XToYYprime, and script to generate actual cards

gvonsem:
This PR contains template MG cards for XToYYprime (resonance X decaying to two non-SM bosons Y and Y'), and a script to generate the actual cards. 

- The cards are based on a Z' -> ZH model (see [these cards](https://github.com/cms-sw/genproductions/tree/339ced1cc4be2a73b701f0b750155d70b38010ba/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/exo_diboson/Spin-1/Zprime_Zh_Zhadhbb)), changing the Z', Z, and H masses. The widths of the Y (= modified Z) and Y' (= modified H) bosons are set to narrow widths. The decays of the Y and Y' are to be done in Pythia.

- The script to generate the actual cards from the template cards is based on the script in the [BulkGraviton_hh_granular](https://github.com/cms-sw/genproductions/tree/22d4141/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/BulkGraviton_hh_granular) model.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/XToYYprime/XToYYprime_MX1000_MY30_MYprime60_narrow/XToYYprime_MX1000_MY30_MYprime60_narrow_customizecards.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/XToYYprime/XToYYprime_MX1000_MY30_MYprime60_narrow/XToYYprime_MX1000_MY30_MYprime60_narrow_customizecards.dat
@@ -1,0 +1,9 @@
+set param_card mass 9000001 1000
+set param_card mass 9000002 1000
+set param_card rhoinputs 2 1000
+set param_card DECAY 9000001 1.000000e-03
+set param_card DECAY 9000002 1.000000e-03
+set param_card mass 25 60
+set param_card mass 23 30
+set param_card DECAY 25 1.000000e-03
+set param_card DECAY 23 1.000000e-03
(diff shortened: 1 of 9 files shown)


qliphy:
+1 @agrohsje

gvonsem:
Hi Alexander, thanks for checking. Some of these questions apply also to the standard Z' samples, as I took cards from that model and did only minimal changes. 

- As far as I know, the maxjetflavor 4 or 5 should have [no effect in LO without matching](https://answers.launchpad.net/mg5amcnlo/+question/235253), as there are no quarks or jets in the ME to cut on (and the number of flavors to include in a matching scheme is not relevant if there is no matching). In any case, the question remains if the model is 5FS and if the proton should be defined including b quarks. I am not sure where to check that, do you have suggestions? 
_I had a look at the particle masses in the [UFO model](https://github.com/syuvivida/DibosonBSMSignal_13TeV/tree/master/Models/Vector_Triplet_free_Width_UFO) but only u,d,s have zero masses, and not even c (or b); so I assume just looking at the quark masses here is too naive?_

-  Yes, the q definition can be removed as it is not used. This is a remnant from the Z' cards where the boson are decayed to q afterwards, but we plan to do that in Pythia.

- On the "set" statements: they were set in this way in the Z' cards so I did not touch them. If you confirm all of these are defaults, they can indeed be removed for simplicity.

agrohsje:
Hi Gerrit,
c is massless : https://github.com/syuvivida/DibosonBSMSignal_13TeV/blob/master/Models/Vector_Triplet_free_Width_UFO/parameters.py#L215
while b is indeed not and the default parameter card is for 4FS. For maxjetflavor you are right. I comment on this to avoid mistakes in case the proc_card gets later extended. But there is nothing wrong in what you committed. 
Think about which scheme you want/need and adjust. 
Cheers, Alexander.

gvonsem:
Hi Alexander, thanks for checking. So for my understanding: am I correct that 

- in case I would want to use 5FS; I would need to set the b quark mass to 0 via the customizecards card ("`set param_card mass 5 0`"), and in addition, set maxjetflavor to 5 in the run card (no effect but to avoid confusion)

- in case I would want to use 4FS; I would need to redefine the proton without b quarks in the proc card ("`define p = g u c d s u~ c~ d~ s~`"). Isn't this change necessary to use the correct [4FS PDF sets](https://github.com/cms-sw/genproductions/blob/master/MetaData/pdflist_4f_2017.dat) in that case?

agrohsje:
Hi Gerrit. About the second bullet: yes and yes. About the first one. This is more challenging. set blabla in customize_card will only change the option during event generation. I think you need already a proper restriction card when doing import to make that working.I never tested this myself, but this is what I remember from talks. Maybe you can check.

gvonsem:
Thanks for the info; for the purpose of these samples 4FS is sufficient. I updated the proc card accordingly.

PR #2650, opened 2020-06-10
sherpa VBS ZA

menglu21:
alternative sample for VBS Zgamma


agrohsje:
Aren't the cuts very close to isolation level? Otherwise looks good to me. 
+1 @qliphy

qliphy:
I agree isolation cut should be loosen: 
  IsolationCut 22 0.4 1 1;  -> 
  IsolationCut 22 0.05 1 1;

menglu21:
Hi Alexander, Qiang,

from the MG side, the default one is:
0.4    =  R0gamma  ! Radius of isolation code
  1.0    =  xn  ! n parameter of eq.(3.4) in hep-ph/9801442
  1.0    =  epsgamma  ! epsilon_gamma parameter of eq.(3.4) in hep-ph/9801442
  True    =  isoEM  ! isolate photons from EM energy (photons and leptons)

and in sherpa:
IsolationCut 22 <dR> <exponent> <epsilon>

should we change this value if we want to do comparison between two generators?

menglu21:
sorry, i just realized the 0.4 is the cone radius.

agrohsje:
My idea was to do whatever tight cut at the analysis level, but do gen cuts a little looser. When comparing, can't you apply offline cuts to bring them on equal footing? I am afraid the sample might not be useful for everybody in the end.

qliphy:
@menglu21   In MG settings,  if you set  "0.0 = ptgmin ! Min photon transverse momentum", then the other Photon-isolation cuts will be ignored.   These isolation cuts usually apply to NLO but not LO.

For LO,  pta, dral  and draj affect the isolation.

https://github.com/cms-sw/genproductions/blob/c52ed91d3fe5a7262246c3661c0732d3284dd79e/bin/MadGraph5_aMCatNLO/cards/production/13TeV/LLAJJ_EWK_MLL50_Mjj120_5f_LO/LLAJJ_EWK_MLL50_Mjj120_5f_LO_run_card.dat

menglu21:
@agrohsje @qliphy 
thanks, updated

PR #2652, opened 2020-06-11
tWZ sample with ttZ overlap removal

nadjieh:
DR and DS schemes ...


qliphy:
+1 @agrohsje

agrohsje:
Hi @nadjieh . Looks all good. Two small remarks. You are changing mtop and yukawa. You might also change the top width. Won't have a big effect, but just for full consistency this might be nice. 
What about "gamma contributions" to l+/l- ? You want the Z  only?

nadjieh:
Hi @agrohsje 
Can you tell what would be the corresponding value for top width?
On ll, as far as I know the plugin can only handle on-shell Z. I will double check.

nadjieh:
Update on this:
- I confirm that the PLUGIN dose not support off-shell dilepton production
- For the top width, since it is set to auto, I expect with a different mass it is recalculated automatically (referring to param card). 

Anything I miss?

nadjieh:
The witdth is now set to auto.
Can it be merged?

PR #2654, opened 2020-06-16
Powheg bbH template and script to generate generator cards

christopher-zzz:
Script to generate POWHEG-BOX-V2 bbH input cards from a template. For now only contains higgs masses from 60 to 200 GeV as parameters for higher masses still need to be validated.
The template is a modified from the version found [here](https://twiki.cern.ch/twiki/bin/viewauth/CMS/H2TauBBHSimulation#Information_about_bbH_LHE_event). `hwidth` and `ncall1` are chosen to guarantee numerical stability.


agrohsje:
Hi Christopher. Thanks for the PR. Is there anything we should modify on the Powheg side? Are you available to work on this or should we find somebody?

christopher-zzz:
If I remember correctly the only issue was that we needed to modify some code of the Powheg bbH package before compiling it to use the dynamical scales proposed in YR4. If this really is the case it is a one line fix in [run_pwg.py](https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/run_pwg.py), I can try to investigate and fix this in the next few days. 
However we will still continue with a pLHE campaign for our bbH production as we invested already a lot of time in that direction.

christopher-zzz:
Hello again, I have now implemented and commited the patch that allows for the usage of the `runningscale` settings from YR4.

agrohsje:
Do you want to make the change a real patch instead of just copying the whole file. This allows to be more version-independent. Why not using the default PDF and use the PDF4LHC from the variation: 
https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/make_rwl.py#L169 ?

qliphy:
Sorry I merged too quickly.  Please follow suggestions from Alexander and make a new PR.

PR #2663, opened 2020-07-02
four leptons added

nadjieh:
Similar to 3l case with Madspin changed to four lepton decay


agrohsje:
Hi @nadjieh. Looks good to me. If you can you could clean up the fragments a little bit. had and j are not used, you could move the lepton definitions to madspin, remove the lines with # in madspin to make it easier.

nadjieh:
Hi @agrohsje 
had is used in the decay. Do you mean the definition should go to madspin for that as well?

Also, do I need to regenerate gridpacks or only update the git?

agrohsje:
No, no. Don't update the gridpack. Unless @qliphy spots a mistake that I didn't. 
My comments are just cosmetics to make it easier. 
had is currently not used in madspin.

nadjieh:
@agrohsje 
" had is currently not used in madspin."

Ah yes, this is remnant of 3l. Will do ...

nadjieh:
@agrohsje can you please check it now?

Thanks!

agrohsje:
Thanks! +1 @qliphy

PR #2664, opened 2020-07-03
Request EL7 workers for codegen step in CMS Connect

khurtado:
Since `UL2019` branch works with EL7, this CMS Connect script needs to be updated in order to request the proper OS in the codestep step. Thank to @agrohsje for noticing this.


qliphy:
thanks!

PR #2665, opened 2020-07-03
New POWHEG compilation tests (for condor) 

covarell:
Also includes fixes for several processes failing before (in gcc7).

### DIFF
--- a/bin/Powheg/Utilities/helpers.py
+++ b/bin/Powheg/Utilities/helpers.py
@@ -19,7 +19,11 @@ cd ..",
     "WWJ" : "patch -l -p0 -i ${patches_dir}/wwj-weights.patch\n \
 cp ${patches_dir}/rwl_write_weights2_extra.f POWHEG-BOX/$process/",
     "Zj" : "patch -l -p0 -i ${patches_dir}/pwhg_write_weights_nnlo.patch",
-    "Wj" : "patch -l -p0 -i ${patches_dir}/pwhg_write_weights_nnlo.patch", 
+    "Wj" : "patch -l -p0 -i ${patches_dir}/pwhg_write_weights_nnlo.patch",
+    "VBF_H_smeft" : "cd POWHEG-BOX/VBF_H_smeft\n \
+head -n 966 pwhg_analysis.f | tail -n 12 > pwhg_analysis_new.f\n \
+mv pwhg_analysis_new.f pwhg_analysis.f\n \
+cd ../..", 
     }.get(process,"")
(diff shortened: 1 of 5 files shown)


covarell:
@Hooooon12 @yuanchao when this will be merged, that is the starting point for testing processes in CMSSW11 and gcc8

bin/Powheg/Utilities/helpers.py
```diff
@@ -19,7 +19,11 @@ def runGetSource_patch_1(process) :
     "WWJ" : "patch -l -p0 -i ${patches_dir}/wwj-weights.patch\n \
 cp ${patches_dir}/rwl_write_weights2_extra.f POWHEG-BOX/$process/",
     "Zj" : "patch -l -p0 -i ${patches_dir}/pwhg_write_weights_nnlo.patch",
-    "Wj" : "patch -l -p0 -i ${patches_dir}/pwhg_write_weights_nnlo.patch", 
+    "Wj" : "patch -l -p0 -i ${patches_dir}/pwhg_write_weights_nnlo.patch",
+    "VBF_H_smeft" : "cd POWHEG-BOX/VBF_H_smeft\n \
+head -n 966 pwhg_analysis.f | tail -n 12 > pwhg_analysis_new.f\n \
```
agrohsje, line 24: @covarell 966 and -n12 looks code specific or is this general enough for future versions?

covarell: It is code specific. For VBF_H_smeft, pwhg_analysis.f has basic analysis functions inside, which are identically defined in the common analysis file of the main POWHEG dir. So I get an error about subroutines with the same name in the linking phase. I don't think these functions will be removed from the main analysis file, so it should be general enough.

yuanchao: I see that the ANALYSIS is un-commented in the first patch 480390e. Is this a must for gridpack production?

covarell: If you have a closer look, you see that "_ANALYSIS" is uncommented, while just "ANALYSIS" is still commented.
This is because a few process have a section of the MakeFile called "LHEF_ANALYSIS" that was changed into "LHEF_#ANALYSIS" and this was enough to break the Makefile parsing.

covarell: Sorry for the awful writing, I do not know how to write a verbatim underscore in this comment window

agrohsje:
+1 @qliphy

PR #2667, opened 2020-07-07
update the patches for MGv273

shu-xiao:
It's a temporary working area with MG5v273.
If something is wrong, we can fix it.


qliphy:
please make a PR to mg27x branch instead of master.

PR #2670, opened 2020-07-08
Upload h to a1 a2 cards

hftsoi:
Hello, we are uploading card creation files and templates for the asymmetric decays h(125)->h1h2, with both cascade (h2->h1h1) and non-cascade cases, in the Benchmark Scenario 1 of the TRSM model (arXiv:1908.08554). Any comment would be much appreciated. Thanks.


qliphy:
+1

agrohsje:
Hi @hftsoi , is the b massive or massless in the model? Asking because of maxjetflavor and the jet pT and drjj cuts which could be different depending on the scheme.

hftsoi:
Hi @agrohsje , the b is massive in the model.

PR #2672, opened 2020-07-10
bbh cards & patch

christopher-zzz:
Updated how a patch is applied.


roger-wolf:
Also the work we are concerned with is mostly of relevance for BSM Higgs models with enhanced coupling to b-quarks.

agrohsje:
thanks so much! looks good to me! 
@roger-wolf that was indeed my question. will somebody else complain about the pdf choice? i mainly like to avoid that the setup produces a sample that is too narrow for others to use.

christopher-zzz:
We use the pdf sets recommended in the Yellow Report 4 and here [1]. So the pdf choice will not limit the use of the sample further than what Roger mentioned above. 
[1] https://twiki.cern.ch/twiki/bin/view/LHCPhysics/LHCHXSWGBBH#Recommendations_for_the_acceptan

PR #2679, opened 2020-07-20
Add cards of aMC@NLO AZh to ll tau tau

wywdiablo:


agrohsje:
5 = asrwgtflavor     ! highest quark flavor for a_s reweight
is different from your jet defintion which is 4fs.

what happens to the width when you are changing the mass? do you really want to keep it the same?

wywdiablo:
@agrohsje , Dear Alexander, thanks for the comments. I think the "asrwgtflavor" is kind of independent of favour scheme. It is a choice of analysts. As shown here (Thanks for Qiang digging the information):
https://answers.launchpad.net/mg5amcnlo/+question/269595

And we would like to align this setting with the same analysis but in bb final state so that future possible combination could be also easier:
https://github.com/cms-sw/genproductions/blob/e7ca10e7986783543e9990eab177d196315214b4/bin/MadGraph5_aMCatNLO/cards/production/13TeV/AToZhToLLBB_01j_4f/AToZhToLLBB_01j_4f_run_card.dat#L71


As for your second question, I think the model already did the job. The width would be calculated when setting the mass. If you open the model file and the "decays.py", then you can see.

Let me know if you have more questions/comments.

agrohsje:
As Rikkert explained, this will take out a qcd branching and replace the alphas at central scale with that at the jet one. In your case, the situation is a little different from what is discussed. In the MG5 thread, there are b's involved which are not part of the merging. But for you, this is not the case and you just define j without b , so I would assume 4 is the more natural choice.
About the consistency with the bb final state, I don't follow the argument. Anyway, feel free to leave as it is. 

The width will not be automatically adjusted, it can be adjusted. That's why I was asking. If you want to use this functionality,
you should add 
set param_card decay pdgid auto

wywdiablo:
Dear Alexander, thanks for your comments. I modify the asrwgtflavor as you suggest. And also thank you for pointing out the decay width is not automatically adjusted. We would like to keep the default value, which is 1 GeV I believe. We used 2HDMC to calculate the width and also the AUTO option to calculate the width, it is within a few GeV fulfilling the so-called narrow-width-approximation. The reason to keep it the same as default is that, we can later provide a exclusion plot of parameters as function of the decay width in model interpretation.

wywdiablo:
@agrohsje

bin/MadGraph5_aMCatNLO/cards/production/13TeV/AToZhToLLTT_01j_4f/AToZhToLLTT_01j_4f_run_card.dat
```diff
@@ -0,0 +1,270 @@
...
+# Automatic ptj and mjj cuts if xqcut > 0
+# (turn off for VBF and single top processes)
+#**********************************************************
+   T  = auto_ptj_mjj  ! Automatic setting of ptj and mjj
+#**********************************************************
+#                                                                    
+#**********************************
+# BW cutoff (M+/-bwcutoff*Gamma)
+#**********************************
+  15000  = bwcutoff      ! (M+/-bwcutoff*Gamma)
```
sihyunjeon, line 87: @qliphy @agrohsje 
Hi Qiang and Alexander,
Is it okay to use BW=15000? I thought this had to be < 20~30.
Perhaps there's something different for loop induced processes?

I was taking a look at these cards for [1].

[1] https://hypernews.cern.ch/HyperNews/CMS/get/generators/4776.html

sihyunjeon: Maybe it's okay cuz "p p > h3" h3 is defined onshell already?

qliphy: Right,  h3 is onshell,  and the decay was done through madspin.

PR #2680, opened 2020-07-20
ggWW->semilep(lvqq) cards

soarnsoar:
This is gluon,gluon -> WW -> lnuqq process using loop-induced mode.

There's no addtional emission in ME level. 

cut file is applied to avoid divergence where pt(WW) ~ 0.

Actually, upto 1jet emission was studied before but divergence problem is not fixed yet.

So, I want to go for 0j gridpack first.


qliphy:
Why bwcutoff is set to so large a value as 15000?

soarnsoar:
Dear Qiang,

Thanks for checking the detail! It's my mistake. Sorry.


I've fixed it to 15 now and

the xsec in log files are not changed before and after the fixing 
(1.248 +- 0.004944 pb)

qliphy:
+1 @agrohsje

PR #2690, opened 2020-07-27
Adding in cards for single top tW sample

leggat:
Adding madgraph cards for tW for CP5 tune samples


qliphy:
You run_card needs to be updated, for example,   store_rwgt_info is missing and  "$DEFAULT_PDF_MEMBERS  = reweight_PDF" should be there

https://twiki.cern.ch/twiki/bin/view/CMS/Moriond18MC


Please have a check with an example file here:
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/tzq_ll_4f_ckm_NLO_UL/tzq_ll_4f_ckm_NLO_UL_run_card.dat

leggat:
I have updated the run_card. Are there any other issues?

PR #2695, opened 2020-08-12
Add VHH cards summer2020

capalmer85:


qliphy:
There is no need to upload all cards corresponding to different {CV, C2V, C3}.  You can just upload one set of example cards.

capalmer85:
I'm coming back to this now.  Is there any other feedback?  Where should we put the list of coupling working points we'd like?

qliphy:
You can write a script to copy from your template cards and switch to other working points, e.g.
https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_bbbb
and
https://github.com/cms-sw/genproductions/pull/2670/files#diff-0946e09ccf894b70277ee0a16222cce1

agrohsje:
1.) Please remove these default lines: 
set group_subprocesses Auto
set complex_mass_scheme False.
2.) Please import the model, set new definitions and then add the process line. You now import the model twice.
3.) If the model per default is 4FS no need to define the p without the b. Or is the model 5 FS? 
4.) auto_ptj_mjj equal true will have no effect here. 
5.) Note that 100k events during pilot-run will require some time to finish. Why not just having 5k? The precision of the grid is not driven by that number.

PR #2696, opened 2020-08-18
mv dyfxfx to one directory and create cards for wjets_fxfx

menglu21:


agrohsje:
Hi Meng. Which MG5 version did you use for cuts.f? Older versions are different hence my question. 
Are run_cards always the same, modulo the cutting window? It is hard to check by eye.

menglu21:
> Hi Meng. Which MG5 version did you use for cuts.f? Older versions are different hence my question.
> Are run_cards always the same, modulo the cutting window? It is hard to check by eye.

Hi @agrohsje, to make these cuts.f consistent, i copied the fresh MG5_aMC_v2_7_3/Template/NLO/SubProcesses/cuts.f to each diectory and modified the corresponding lines for cuts (comment on the first line in cuts.f), and the run_cards are the same between the inclusive and binned cards.

menglu21:
Hi @agrohsje, @GurpreetSinghChahal , could you please have a further check on these cards.

bin/GenValidation/updated_validation_cards/dy_fxfx_mll50/dy012jfxfx_inc/dyellell012j_5f_NLO_FXFX_run_card.dat
```diff
@@ -0,0 +1,175 @@
...
+#***********************************************************************
+  tag_1     = run_tag ! name of the run 
+#***********************************************************************
+# Number of LHE events (and their normalization) and the required      *
+# (relative) accuracy on the Xsec.                                     *
+# These values are ignored for fixed order runs                        *
+#***********************************************************************
+ 1500 = nevents ! Number of unweighted events requested 
+ 0.001 = req_acc ! Required accuracy (-1=auto determined from nevents)
+ 20 = nevt_job! Max number of events per job in event generation.
```
agrohsje, line 36: Hi @menglu21 Why do you plan to use this option? I would set it to -1, no? But I am really asking. Don't know how this behaves in production.

bin/GenValidation/updated_validation_cards/dy_fxfx_mll50/dy012jfxfx_ptbinned/pt0_50/DYJetsToLL_012j_Zpt-0To50_5f_NLO_FXFX_run_card.dat
```diff
@@ -0,0 +1,175 @@
...
+#***********************************************************************
+  tag_1     = run_tag ! name of the run 
+#***********************************************************************
+# Number of LHE events (and their normalization) and the required      *
+# (relative) accuracy on the Xsec.                                     *
+# These values are ignored for fixed order runs                        *
+#***********************************************************************
+ 1500 = nevents ! Number of unweighted events requested 
+ 0.001 = req_acc ! Required accuracy (-1=auto determined from nevents)
+ 20 = nevt_job! Max number of events per job in event generation.
```
agrohsje, line 36: As above.

bin/GenValidation/updated_validation_cards/dy_fxfx_mll50/dy012jfxfx_ptbinned/pt100_250/DYJetsToLL_012j_Zpt-100To250_5f_NLO_FXFX_run_card.dat
```diff
@@ -0,0 +1,175 @@
...
+#***********************************************************************
+  tag_1     = run_tag ! name of the run 
+#***********************************************************************
+# Number of LHE events (and their normalization) and the required      *
+# (relative) accuracy on the Xsec.                                     *
+# These values are ignored for fixed order runs                        *
+#***********************************************************************
+ 1500 = nevents ! Number of unweighted events requested 
+ 0.001 = req_acc ! Required accuracy (-1=auto determined from nevents)
+ 20 = nevt_job! Max number of events per job in event generation.
```
agrohsje, line 36: As above.

bin/GenValidation/updated_validation_cards/dy_fxfx_mll50/dy012jfxfx_ptbinned/pt250_400/DYJetsToLL_012j_Zpt-250To400_5f_NLO_FXFX_run_card.dat
```diff
@@ -0,0 +1,175 @@
...
+#***********************************************************************
+  tag_1     = run_tag ! name of the run 
+#***********************************************************************
+# Number of LHE events (and their normalization) and the required      *
+# (relative) accuracy on the Xsec.                                     *
+# These values are ignored for fixed order runs                        *
+#***********************************************************************
+ 1500 = nevents ! Number of unweighted events requested 
+ 0.001 = req_acc ! Required accuracy (-1=auto determined from nevents)
+ 20 = nevt_job! Max number of events per job in event generation.
```
agrohsje, line 36: As above.

agrohsje:
Hi @menglu21 . I only have a question on the splitting. I started to list them all, but as it is always the same, I stopped at some point. I assume in case you want to modify, it is easy and doesn't help to see the full list. 
One more point: should we add madspin cards for tau decays. @NLO the spinup flag is not correctly filled and it seems P8 cannot correctly handle tau decays in multi-parton events. Should be minor but I realized this in a discussion with Josh.

menglu21:
> Hi @menglu21 . I only have a question on the splitting. I started to list them all, but as it is always the same, I stopped at some point. I assume in case you want to modify, it is easy and doesn't help to see the full list.
> One more point: should we add madspin cards for tau decays. @nlo the spinup flag is not correctly filled and it seems P8 cannot correctly handle tau decays in multi-parton events. Should be minor but I realized this in a discussion with Josh.

Hi @agrohsje i modified the option to no_splitting. As for the tau decay, i would like to not include it in these cards, as we usually apply mass cut on the dilepton, the leptons from tau would be small and the effect could be small as you said. on the other hand, as we need to produce inclusive and exclusive gridpacks for validation, i am worried about the complication introduced by the tau decay as the inclusive multi-parton process is already complicated.

agrohsje:
Hi @menglu21. There will be no modification in terms of decays or so. The point is to run the madspin tool with sethelicity only.
This will "reprocess" the events and add the spinup flag. Per default at NLO it will be 9, after running the tool you will see +-1. 
You generated already 
p p > l+ nu ...
so there will be no additional decay involved.

agrohsje:
@menglu21 please see also this request:
https://github.com/cms-sw/genproductions/pull/2714/files

menglu21:
> Hi @menglu21. There will be no modification in terms of decays or so. The point is to run the madspin tool with sethelicity only.
> This will "reprocess" the events and add the spinup flag. Per default at NLO it will be 9, after running the tool you will see +-1.
> You generated already
> p p > l+ nu ...
> so there will be no additional decay involved.

Hi @agrohsje  do i understand correctly, for these process with tau in the LHE, we need to include the madspin cards to set the tau spin information correctly for pythia. Since i use 

'define ell+ = e+ mu+ ta+' 
'generate p p > ell+ vl'

so the LHE has tau in it. I need to add the madspin for all the cards, even for the DY process, correct?

BTW, i do not think the all the cards we already used for the current production considered this point, we need to consider this from now on?

agrohsje:
The problem appears in NLO when the spinup info is missing. P8 can reconstruct the tau spin to proceed the decay, but it seems to have problems in events with several jets. Josh made this comment in a discussion with @sihyunjeon 
There is no issue at LO. The info is available.

menglu21:
> @menglu21 please see also this request:
> https://github.com/cms-sw/genproductions/pull/2714/files

Hi @agrohsje i add dummy madspin cards for fxfx wjets processes. i checked the sync with PR 2714, the variables for phase space are the same. Some other settings are slightly different, which i think should introduce no difference. I edit the card from the default card in MG273, maybe PR 2714 could have some slight modification if you want everything is sync.

sihyunjeon:
@menglu21 @agrohsje As I commented to #2714 , $$ t t~ h is needed to ignore other top and higgs contributions for W+Jets. And this $$ t t~ h is in Meng's cards but not in Matteo's. So this is more complete one. So I think you can close the request #2714 .

Aside from the cards in this directory : GenValidation/updated_validation_cards/dy_fxfx/dy012jfxfx_inc/dyellell012j_5f_NLO_FXFX_proc_card.dat, could we do something for cards in cards/examples/ repo?

The settings below makes errors when one tries LHE file generation (while gridpack is well made) during systematic PDF runs using ./runcmsgrid.sh, and this could be misleading for novice genproductions users (when modifying some lines in runcmsgrid.sh it does make you LHE files but still I don't think it's wise to show an example which is not perfectly working).

https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/examples/wellnu012j_5f_NLO_FXFX/wellnu012j_5f_NLO_FXFX_run_card.dat#L58
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/examples/wellnu012j_5f_NLO_FXFX/wellnu012j_5f_NLO_FXFX_run_card.dat#L96-L97

And also I see several cards with BW cutoff set to 15000, which is not a nice value to use if I remember correctly (bw <= 20?).

https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/examples/wplustest_5f_LO/wplustest_5f_LO_run_card.dat#L87

bin/GenValidation/updated_validation_cards/wjets_fxfx/w012jfxfx_inc/wellnu012j_5f_NLO_FXFX_madspin_card.dat
```diff
@@ -0,0 +1,27 @@
...
+#*                                                          *
+#************************************************************
+#Some options (uncomment to apply)
+#
+#directory for gridpack mode
+set ms_dir ./madspingrid
+
+#initialization parameters
+ set Nevents_for_max_weigth 250 # number of events for the estimate of the max. weight
+ set max_weight_ps_point 400  # number of PS to estimate the maximum for each event
```
agrohsje, line 18: I think this is not needed for onlyhelicity mode. Also the decay line has no impact. @sihyunjeon confirmed this.

menglu21: updated, removed these lines

agrohsje:
Hi @sihyunjeon , indeed I would use the validation cards as example and only have one reference. In fact I was wondering: @menglu21 Should we transform all these 159 cards into templates for LO and NLO and add a script that generates cuts etc. automatically. We could then use one or two examples and put them to examples folder? What do you think. I can also merge the request. This is already a huge step forward. Don't know how much time we should spend. Just thought when trying to approve. It is a challenging tasks to check 159 files for consistency. :-)

menglu21:
> Hi @sihyunjeon , indeed I would use the validation cards as example and only have one reference. In fact I was wondering: @menglu21 Should we transform all these 159 cards into templates for LO and NLO and add a script that generates cuts etc. automatically. We could then use one or two examples and put them to examples folder? What do you think. I can also merge the request. This is already a huge step forward. Don't know how much time we should spend. Just thought when trying to approve. It is a challenging tasks to check 159 files for consistency. :-)

Hi @agrohsje, sorry for the late,  i remade all the binned cards from the inclusive ones, you can check the shell script in those directories including both inclusive and exclusive cards.

agrohsje:
Thanks Meng. This is great. @kdlong had suggestions for sm parameters.

kdlong:
Thanks @menglu21. Indeed, I'll put together a param_card shortly.

One comment I can't resist: git has all past history recorded, so having folders like "updated_X" doesn't make much sense to me. Do we need to keep old and outdating things around in the master repo?

PR #2697, opened 2020-08-18
Add HToZdZd JHUGen files for displaced dimuon search

aminnj:
This PR adds HToZdZd (H->ZdZd->2mu+2X) JHUGen files for a displaced dimuon search using scouting data.

This model was presented in an EXO MC&I meeting [here](https://indico.cern.ch/event/936289/). Gridpacks are taken from the latest HZZ ones here: /cvmfs/cms.cern.ch/phys_generator/gridpacks/2017/13TeV/powheg/V2/gg_H_quark-mass-effects_ZZ_NNPDF31_13TeV/gg_H_quark-mass-effects_NNPDF31_13TeV_M125/v4/gg_H_quark-mass-effects_NNPDF31_13TeV_M125.tgz . For submission, the JHUGen files were edited within the gridpacks to have specific mZ and ctau values, which are the files added in this PR.

@usarica


agrohsje:
Wouldn't it be better to add something like 
WriteFailedEvents=2 
where 2 is the number of failing events? I mean this should be something that is at the level of 0.01*nevts you produce per job.

usarica:
@agrohsje This wouldn't be needed. DecayMode1=-9 ensures there is one Zd-.2mu, and DecayMode2=9 is the inclusive Zd part.

agrohsje:
not sure i understand. h > zd zd will have the same problem has e.g. h > W W SM samples, no ?

usarica:
No it wouldn't because a negative decay mode specification (DecayMode1=-9) in JHUGen is an explicit final state specification (Zd->2mu in this case). This would be different from applying a filter, where in WH, H->WW with a OS dilepton filter, for example, there is dependence on the final decay mode of the associated W. There is no associated Z/W in this case. All events generated by this data card would be Zd->2mu+Zd->2X.

agrohsje:
The setting is to avoid the abort of P8 in case JHUGEN fails to decay the event. JHUGEN only performs a limited number of trials. If the event cannot be decayed, then P8 will expect more events and production will fail. Why can this be avoided if the decaymode is negativ?

usarica:
From https://spin.pha.jhu.edu/Manual.pdf:
"WriteFailedEvents: By default, events that fail the lepton filter are not written to the LHE file."

There is no lepton filter. One needs NLepMin/Max options for the filter to be activated.  The settings of these new files should be similar to those already in genproductions with no filter, e.g. https://github.com/cms-sw/genproductions/blob/master/bin/JHUGen/cards/decay/v7.3.5/2017/ZZ2l2nu_notaus.input.

Have you seen any issues in later versions of JHUGen where a decay failure happens?

agrohsje:
My concern is not related to the lepton filter [1]. I am talking about the general problem of decaying an event. In this case the setting is needed so that the lhe file looks complete for P8. No? 
[1] From the manual. 
WriteFailedEvents: Write events that fail in the LHE file, but with a weight of 0

usarica:
In principle, this option could be turned on, but JHUGen has a 5M trial limit for each event. I don't think it would practically make any difference.

Have you observed any cases where decay failed in unfiltered data cards?

usarica:
Is there any action needed on this PR, or does it just need time for review?

agrohsje:
Hi @usarica , we had several JIRA tickets with problems due to JHUGEN NOT decaying all events. I don't have a complete list of all these tickets so that I can go through them and see if they had a filter or not. I pinged gen-ops but have no feedback yet.  Unless you can say 100% that not having this option will not cause problems in the production system (Heshy pointed that out in the past without mentioning the filter) I would really prefer setting a safety margin. If 1 event in 10k will fail, the whole job will be lost.

usarica:
I see. From our earlier discussions, I only remember these cases for samples with filter, but sure, it doesn't hurt to add this filter.

@aminnj Do you think you can add 'WriteFailedEvents=2' to the data cards and update the PR?

agrohsje:
do you know how many events you get in 8 hours? 
i would add instead of 2 something like 0,98 * this number. the point is: it will not bias the physics, but we are 100% safe in production.

usarica:
The option only allows integers, so it wouldn't support floats.

usarica:
Nick would know more about the event rate, but if I recall correctly, it is faster than H->ZZ since Zd has much narrower width, so phase space is faster to converge.

agrohsje:
not suggesting float. suggesting to estimate 98% of events in 8 hours and put (int) result .

usarica:
I understand, but the only options are 0, 1 and 2, so the numerical value of this estimate would not enter.

agrohsje:
Sorry, right. I understood it as the number of events to accept to fail, but it is writing out with weight 0 or not writing out at all. 
How will P8 handle if we set to 2?

usarica:
If set to 1 or 2, the event weight is set to 0. Choice 2 also removes the particles to save disk space, so it writes an empty event.

agrohsje:
I know. My question was what happens in P8 when using 1 or 2. Will it run without problems? 
Just to say. What we see. We produce 1000 events, 1 cannot be decayed, P8 expects 1000 and exists because it can only process 999. 
The option will not bias the results, but we should make sure P8 runs properly.

usarica:
Well, as far as I know, this option has been used in the past successfully, in fact this is how JHUGen-based filter is applied, so I wouldn't expect problems to come up here.

usarica:
e.g. https://github.com/cms-sw/genproductions/blob/master/bin/JHUGen/cards/decay/v7.3.5/2017/WplusLNu_WminusAny_withtaus_2losfilter_reweightdecay_CPS.input

agrohsje:
so both 1 and 2 are working well with p8? ok, then i have no preference. just not 0.

usarica:
Ok. Let's hear from @aminnj to update the PR then.

agrohsje:
Thanks! So let's use 2.

aminnj:
I updated the PR, adding `WriteFailedEvents=2` to each file

agrohsje:
Thanks. Looks good to me.
+1 @qliphy

PR #2699, opened 2020-08-20
PR for Type-1 heavy neutrino analysis production cards

Hooooon12:
Hi all,

This is the PR of production card for type-1 heavy neutrino full Run-II analysis.
Please let me know if you have any questions or suggestions.


agrohsje:
1.) I would suggest to remove default lines, like these ones: 
set group_subprocesses Auto
set loop_color_flows False
set gauge unitary
set complex_mass_scheme False
set max_npoint_for_channel 0
import model sm
define l+ = e+ mu+
define l- = e- mu-
vl = ve vm vt
vl~ = ve~ vm~ vt~

I would import the model first, then add your particle definitions followed by then generate syntax. 
2.) I see duplicate lines, like reweight_PDF. Note that the first one will be overwritten.  
3.) Can you comment on your setup with spin correlation turned off and on? 
4.) Is the model 4FS or 5FS?

agrohsje:
Hi @Hooooon12, please comment on the status.

Hooooon12:
Hi @agrohsje, sorry for the late reply.

1. I removed the default lines, imported model and then defined the particles.
please take a look at the latest commit.
2. I removed the duplicated reweight_PDF line.
3. In high mass regions(N mass > 80GeV), we use full spin correlation by letting spinmode be default.
Just with low masses(N mass < 80GeV) in which N decays into 3 body, we set spinmode none.
4. It is 5FS. After importing the model, I see
p = g u c d s u~ c~ d~ s~ b b~ by displaying multiparticles.

Thanks for the comments!

Best,
Jihun

PR #2675, opened 2020-07-14
Add VH powheg cards for 2017

alpakpinar:

PR #2702, opened 2020-08-26
Set bornonly to be 0 for ggHZ samples

alpakpinar:
For the cards included with PR #2675 , change bornonly option in the ggHZ cards to be 0.


agrohsje:
Hi @alpakpinar . In case you want NLO you also need to remove L0events = 1.

agrohsje:
ping @alpakpinar

alpakpinar:
Hi @agrohsje , sorry for the late reply. Now I also removed "LOevents 1 " option from 2016 and 2017/2018 Powheg input cards. Thank you!

PR #2709, opened 2020-09-02
Ul2019

whahmad:
Dear Experts,
I have added cards for LFV in Z boson decay model. Please check it if there are comments please let me know.

Regards,
Wael Haj Ahmad


agrohsje:
What is the reason to add the fragments?  Especially as they have a DY gridpack. I would just submit the MG5 cards and remove the rest. Or add one example if that is really needed. But we prefer not having them here. We will check the fragment, when the request is done. Here it is about the cards in the gridpack. 
Rest looks ok to me.

whahmad:
Well, one example is OK. We need the fragment when we make a central MC request, it nedd to be checked by the MC responsible persons.

agrohsje:
GEN L2 will check in the end. So that is us again. :-) If you are uncertain you can ping us before. Not really needed to have it here. It was done in the past, but is a little obsolete now. But I am also fine with having one example. But please replace the gridpack with a dummy.

whahmad:
LFV_Z_hadronizer_RunIISummer19UL17.py , has been updated with a dummy gridpack.

agrohsje:
Thanks for the changes. Please remove the 2 obsolete hadronizers.
One more point. When doing import model lfv using this model:
 https://cms-project-generators.web.cern.ch/cms-project-generators/lfv.tar.gz
I get an error: missing couplings.py. Did you test producing gridpacks with these cards ?

whahmad:
the 2 obsolete harmonizers are deleted. Regarding the lfv model yes i have produced a gridpack using this model without any error. The gridpack availabel here "/afs/cern.ch/cms/generators/www/Zjets_ll_LFV_tarball.tar.xz". May be we have to add this line to the proc_card "import model_v4 $modelfile" instead of this line "import model lfv".

agrohsje:
Thanks. Indeed, you need to adjust the import command so that it is working. one more comment, per default the model is 4FS, so the b will be massive. I recommend to add a customize_card and set the b mass and yukawa to 0 .

whahmad:
Thanks, your recommendation have been taken into account. Please check it.

agrohsje:
Thanks for the adjustments. 
Looks good to me. 
@GurpreetSinghChahal  please check and merge.

whahmad:
Names of the files are changed to be the same like the folder's name.

whahmad:
Dear Experts,
Could you please check and merge? Because it is urgent. I would like to start a gridpack generation, and make an Official MC request.

Regards,
Wael Haj Ahmad

PR #2710, opened 2020-09-10
Allow reweighting in decays

### LINKED ISSUE #2708: Gridpacks, LO reweighting, and missing files
Hello @agrohsje @qliphy, all,

I am trying to generate gridpacks in mg245 which involve EFT reweighting at LO. While the creation of gridpacks runs smoothly, the event generation results in an error due to some fortran files removed during the cleaning process. Putting these fortran files back solves the problem. Naively, I would assume that there is something fishy going on during the gridpack generation. If you have any ideas, please let me know!

Gridpack: 
/eos/user/k/kskovpen/EFT/eft_test_slc7_amd64_gcc700_CMSSW_10_2_18_tarball.tar.xz
https://cernbox.cern.ch/index.php/s/fWoX5DRCkroR8Sk

Log file from gridpack generation: 
https://cernbox.cern.ch/index.php/s/Y7JvG8jH5vi5SkI

Error when generating events:
"set max_npoint_for_channel 0" with error:
OSError : [Errno 2] No such file or directory: '/afs/cern.ch/user/k/kskovpen/EFT/genproductions/bin/MadGraph5_aMCatNLO/temp/mgbasedir/vendor/DiscreteSampler/DiscreteSampler.f'

This happens at:
echo "0" |./bin/madevent --debug reweight GridRun_${rnum}
in runcmsgrid_LO.sh

Thanks a lot in advance,

Kirill

kskovpen:
This PR resolves #2708 by computing decay widths during the pilotrun with updating reweight cards to include the computed values.


agrohsje:
Thanks so much for taking care. I have 2 questions. Can you see how different the logic for nlo would be? What about having a common block for the replacement  in MadGraph5_aMCatNLO/Utilities/gridpack_helpers.sh  and then just call the function in gridpack_generation.sh in the LO and NLO block?

kskovpen:
Added width extraction for NLO reweighting and moved the relevant function to helpers module.

bin/MadGraph5_aMCatNLO/Utilities/gridpack_helpers.sh
```diff
@@ -128,3 +128,46 @@ prepare_reweight () {
...
+        
+    hasauto=$(grep "auto" ${cdir}/${name}_reweight_card.dat | egrep -v "#")
+    
+    if [[ ${hasauto} != "" ]]; then
+      echo "extract computed widths and rewrite reweight card"
+
+      if [ "$isnlo" -gt "0" ]; then
+        cp ${wd}/processtmp/Events/pilotrun/events.lhe.gz temp.lhe.gz
+        gzip -d temp.lhe.gz
+        ./MG5_aMC_v2_6_5/Template/LO/bin/internal/extract_banner-pl temp.lhe banner.txt
```
agrohsje, line 148: Hi @kskovpen . Thanks for updating. I totally missed the mail. I suggest to make this generic so that it also works in future versions.

kskovpen: Updated!

sihyunjeon:
@kskovpen @agrohsje Sorry for my ignorance on this.
So this is for a problem when you change the physics parameters, decay widths change but this info is not properly propagated to MadSpin. And this helps you to extract the widths that you get from pilot run and use this as one of the inputs for MadSpin?

agrohsje:
the idea is to adjust the width in case of reweighting. so this allows beforehand to calculate the proper info and populate the reweight card with the correct value. going to merge.

PR #2714, opened 2020-09-15
inclusive Wjets NLO FXFX cards updated for MG265

mpresill:
Dear experts,
following the debugging on the update of WJets NLO FXFX cards happened in the HN (https://hypernews.cern.ch/HyperNews/CMS/get/generators/4795.html), I kindly ask to accept my pull request for this new working version of the cards.
Thanks, 
Matteo

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/WJetsToLNu/wellnu012j_5f_NLO_FXFX/wellnu012j_5f_NLO_FXFX_madspin_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/WJetsToLNu/wellnu012j_5f_NLO_FXFX/wellnu012j_5f_NLO_FXFX_madspin_card.dat
@@ -1,0 +1,37 @@
+#************************************************************
+#*                        MadSpin                           *
+#*                                                          *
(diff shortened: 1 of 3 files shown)


agrohsje:
Thanks for the request. I think @sihyunjeon confirmed that we can remove the decay lines in madspin ? 
Can you check the run card and compare to Meng: 
https://github.com/cms-sw/genproductions/pull/2696
Please list differences. I would suggest we sync settings. 
I would remove in the madspin card lines that are not used, e.g.  the # lines.

agrohsje:
Shouldn't the same problem apply for DY?

mpresill:
Hi @agrohsje , thanks for the comments. I have seen that also in #2696 there was an agreement in adding a dummy Madspin, so I guess it is ok to keep it as it is.

> I would remove in the madspin card lines that are not used, e.g. the # lines.
I can do that.

sihyunjeon:
@agrohsje  The one Meng Lu made is in more perfection. I haven't spotted that "p p > lep nu j j" could be made from top process "p p > w > t b > lep nu j b (== lep nu j j)" and so Matteo @mpresill (I owe you an apology). Maybe we can just keep Meng Lu's cards for merging.

agrohsje:
Thanks. Let me close this PR as included here: https://github.com/cms-sw/genproductions/pull/2696 .

PR #2716, opened 2020-09-16
Add WJetsToQQ_HT*, ZJetsToQQ_HT* cards for legacy

DryRun:
Adding updated cards for W/Z(qq) HT-binned samples in (ultra) legacy. Changes:
- The old cards have been moved to `.../preUL`.
- `pdfwgt` has been set to `F`.
- I tried to make the cards more consistent, fixing things like the bwcutoff (previously, ZJets had 15 and WJets had 15000; now all set to 15), the specification of irrelevant cuts (set to -1 instead of large number), and removing the dash in `HT-800toInf`.


agrohsje:
Hi @DryRun . The request looks good to me. We currently try to sync SM processes, see this PR: https://github.com/cms-sw/genproductions/pull/2696 . Maybe we should also sync here. Do you mind if we wait a little with merging or even see if we can include your cards there as a template. Let me know what you think.

DryRun:
Hi @agrohsje - certainly, this PR is not urgent, and would probably benefit from syncing with the other processes. I'll keep an eye on the other discussion.

menglu21:
Hi @DryRun is there any VToQQ sample with HT<400? 
@agrohsje i can take these cards as reference cards in the validation repo, maybe cards for HT<400 is also needed, do it make sense to you?

DryRun:
Hi @menglu21 - I'm not aware of any HT<400 VToQQ samples, I suppose because this would be rather hard to trigger. We could certainly add such a sample, but do you have an application in mind?

menglu21:
> Hi @menglu21 - I'm not aware of any HT<400 VToQQ samples, I suppose because this would be rather hard to trigger. We could certainly add such a sample, but do you have an application in mind?

I may include a 200~400 region, it's not very important. Anyway, i'll take your cards as reference to create the officials.

menglu21:
Hi all, i make a new PR #2726 for the hadronized decay Vjets process. BTW, @DryRun why don't you choose to decay the V in the madspin

DryRun:
Hi @menglu21 - the short answer is that I'm not the original creator of these cards and just copied what we used previously. But does using Madspin for the V decays make a difference here?

agrohsje:
The modelling via the decay chain syntax is superior:
https://answers.launchpad.net/mg5amcnlo/+question/242455

menglu21:
> The modelling via the decay chain syntax is superior:
> https://answers.launchpad.net/mg5amcnlo/+question/242455

thanks @agrohsje, it's clear to me now

sihyunjeon:
@DryRun Maybe this is [something](https://github.com/cms-sw/genproductions/pull/2770) might interest you

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/WJets_LO_MLM/WJetsToQQ_HT400to600/preUL/WJetsToQQ_HT400to600_run_card.dat
```diff
@@ -0,0 +1,275 @@
...
+ 1        = scalefact        ! scale factor for event-by-event scales
+#*********************************************************************
+# Matching - Warning! ickkw > 1 is still beta
+#*********************************************************************
+ 1        = ickkw            ! 0 no matching, 1 MLM, 2 CKKW matching
+ 1        = highestmult      ! for ickkw=2, highest mult group
+ 1        = ktscheme         ! for ickkw=1, 1 Durham kT, 2 Pythia pTE
+ 1        = alpsfact         ! scale factor for QCD emission vx
+ F        = chcluster        ! cluster only according to channel diag
+ T        = pdfwgt           ! for ickkw=1, perform pdf reweighting
```
agrohsje, line 71: should be false.

agrohsje:
Let me re-activate this chat. As SUSY and HIG pushed to get samples asap, we should also move this request forward. @dryrun would you mind removing the files and just provide templates for the ht splitting and then just replace this. I think it is sufficient to just upload the template and the script. I see some mismatches between bwcutoffs and pdfwgt which are sometimes modified, sometimes not. What do you think. Similar to this: https://github.com/cms-sw/genproductions/tree/master/bin/GenValidation/validation_cards/dylljetsmlm

DryRun:
Hi @agrohsje - sounds reasonable to me, working on it now.

sihyunjeon:
@DryRun Hey David, Any news on this?

DryRun:
Hi all, I made the some changes: 

- proc_cards and run_cards for different HT bins are generated with a script;
- HT400to600 cards are included as reference;
- Old cards saved in `preUL` folder;
- Everything placed under `genproductions/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/*Jets_LO_MLM/*JetsToQQ_HT` where *=W or Z.

A look-over would be helpful. I did a diff between one old and new card (ZJetsToQQ_HT400to600), and spotted a few differences that I think have no effect (see attached .txt file); I'll do this for all cards now, and then launch the gridpack production.

[ZJetsToQQ_HT400to600_run_card.diff.txt](https://github.com/cms-sw/genproductions/files/6070887/ZJetsToQQ_HT400to600_run_card.diff.txt)

DryRun:
Complete diffs:
[WJetsToQQ_HT.diff.txt](https://github.com/cms-sw/genproductions/files/6071046/WJetsToQQ_HT.diff.txt)
[ZJetsToQQ_HT.diff.txt](https://github.com/cms-sw/genproductions/files/6071047/ZJetsToQQ_HT.diff.txt)

sihyunjeon:
@DryRun @agrohsje I think the cards are fine

PR #2723, opened 2020-09-25
Explicitly set EW parameters for single V validation cards

kdlong:
Explicitly set the EW parameters for V+jets, following the setup used for the W mass samples. The values are set to the best PDG values, using the input scheme with G_fermi and either mW,mZ for W production and mZ, sinthetaW for Z production. MadGraph uses a fixed width scheme, so we account for the fact that the PDG values are sqrt(s) dependent using https://arxiv.org/pdf/1606.02330.pdf

I checked that this is technically working for DY MLM, but can someone make additional checks to be sure that it is always read properly and set, and that the cross sections are changed with or without the customization?


kdlong:
Sorry I messed up the history, will make it again.

PR #2725, opened 2020-10-01
add_min_mjj_invariant_mass

Soumyatifr:
@agrohsje I missed the minimum invariant mass cut for the light quark jets in our signal sample can you please check once. Extremely sorry for this mistake and please merge if you think.


agrohsje:
Why did you close?

Soumyatifr:
Sorry @agrohsje I forget that our signal topology is like that we don't need to put any harder cut in mjj as already they are with high mjj. We just put this cut to produce background only. I should have remember this unfortunately i made this PR. So I deleted that.

agrohsje:
Thanks for clarifying.

PR #2726, opened 2020-10-03
add cards for hadronized decay Vjets

menglu21:
The customized card Kenneth added is not included in this PR since they are from leptonic decay study.


agrohsje:
Hi @menglu21 .Should we merge this or ignore?

menglu21:
maybe we can ignore it, we don't really do the validation on the hadronic decay V_jets.

PR #2728, opened 2020-10-05
Ul2019

whahmad:
Dear Experts, 
Could you please merge the pull request , there were some updates required by MC&I_Exo group. Please it is urgent!!!


agrohsje:
I saw you removed the example tarball. But can you include the proper link to your cards in the fragment. I don't know how exo is using your example fragment, but if it is the base of the request, inlcuding the proper link to the cards instead of DY would be useful. Please update.

whahmad:
the gridpack path have been updated.

agrohsje:
Thanks. Please also adjust the info here: 
#Link to datacards:
#https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/DYJets_HT_LO_MLM_pdfwgt_F/DYJets_HT_mll50/DYJets_HT-incl
This is often useful info.

sihyunjeon:
@agrohsje @whahmad  The python files (pythia fragments) should not have externalLHEProducer, no? And since these pythia fragments are already there in genproductions, I don't think this should be merged to master.

whahmad:
I do not see any update except the "DEFAULT_PDF_SETS = lhaid" to set to "325300     = lhaid" in the run card is this what meant?

whahmad:
@sihyunjeon So you mean to remove the externalLHEProducer block! right?

agrohsje:
Dear all, let me clarify. GEN prefers not storing fragments in genproductions. The info is easily available in MCCM. 
However, if Exo wants to have the fragment in genproduction, no problem from our side. So if the tarball should be there or not is fully up to EXO. 
My comment above was about this line: 
https://github.com/cms-sw/genproductions/blob/9a9c3c58add765ac9d4114405baa844f5968d921/python/ThirteenTeV/LFV_Z_hadronizer_RunIISummer19UL17.py#L12
not the gridpack itself.

sihyunjeon:
@agrohsje I can confirm EXO doesn't need this part. And as you said the pythia fragment which Wael wants to use is the one used for LO DY+01234j productions which is available info in MCM so I don't think hadronizer should be added to genproductions.

@whahmad Yes, [1] would be taken care by Kai, you should not add this part to hadronizer. But can you remove the entire LFV_Z_hadronizer_RunIISummer19UL17.py file? This doesn't need to be merged in genproductions for sample submission.

[1] https://github.com/cms-sw/genproductions/blob/9a9c3c58add765ac9d4114405baa844f5968d921/python/ThirteenTeV/LFV_Z_hadronizer_RunIISummer19UL17.py#L1-L9

whahmad:
It is deleted

sihyunjeon:
@whahmad Thanks! I think everything is okay now

PR #2729, opened 2020-10-05
add gg_H_2HDM generator cards

christopher-zzz:
Input files for gg_H_2HDM


agrohsje:
Hi Christopher. Do you think you could just add a template and a script that replaces the relevant parameters ?

christopher-zzz:
Done. I did not remove the original datacards but they can be discarded as the script and template are now in place.

agrohsje:
Thanks @christopher-zzz . Would be nice if you could remove them on the command line. Thanks.

christopher-zzz:
Ok done.

covarell:
They look fine to me on the configuration side.
Please have a look at the result for one mass point, because I remember that many years ago we had problems in this process: unphysical pT(h) [and pT(H)] distributions.

agrohsje:
Maybe you can add some plots to your talk  and update : 
https://indico.cern.ch/event/951406/contributions/4041941/attachments/2115245/3558975/ggHProductionPlanGENmeet05102020.pdf

christopher-zzz:
Hello again,  

I added backup slides to my presentation with plots showing agreements between the 2HDM and SM Powheg-Box packages for the SM Higgs mass. The second slide compares generator level 2HDM vs the fixed order NLO result. The observed behavior can be verified in the cited theory paper.

https://indico.cern.ch/event/951406/contributions/4041941/attachments/2115245/3558975/ggHProductionPlanGEN.pdf

christopher-zzz:
@agrohsje @covarell Is there anything else needed for the PR to be merged?

wywdiablo:
Hi, @agrohsje , I was wondering whether it is allowed to used PDF4LHC15 PDF for the whole Run 2 period as the analysts said. As far as I understand, NNPDF3.0 is recommended for pre2017 and NNPDF3.1 for post2017. Can GEN please clarify this?

agrohsje:
Hi @wywdiablo. Thanks for pinging. I raised the same point in another thread with Christopher et al. This was the answer : 
"We use the pdf sets recommended in the Yellow Report 4 and here [1]. So the pdf choice will not limit the use of the sample further than what Roger mentioned above.
[1] https://twiki.cern.ch/twiki/bin/view/LHCPhysics/LHCHXSWGBBH#Recommendations_for_the_acceptan"

I personally would do a sample using the latest NNPDF 3.1. version for ALL three years and combine it with a CPx tune. This allows a simple statistical treatment and considers our latest knowledge of PDFs. However, as this is not a CMS-wide sample, I have no objections to follow a different strategy. The pwg-rwl dat file should, if not modified by hand, allow a large variety of re-weightings including my personal choice.

wywdiablo:
> Hi @wywdiablo. Thanks for pinging. I raised the same point in another thread with Christopher et al. This was the answer :
> "We use the pdf sets recommended in the Yellow Report 4 and here [1]. So the pdf choice will not limit the use of the sample further than what Roger mentioned above.
> [1] https://twiki.cern.ch/twiki/bin/view/LHCPhysics/LHCHXSWGBBH#Recommendations_for_the_acceptan"
> 
> I personally would do a sample using the latest NNPDF 3.1. version for ALL three years and combine it with a CPx tune. This allows a simple statistical treatment and considers our latest knowledge of PDFs. However, as this is not a CMS-wide sample, I have no objections to follow a different strategy. The pwg-rwl dat file should, if not modified by hand, allow a large variety of re-weightings including my personal choice.

OK, I created the BBH samples (already made use of this PDF set) + Tune CUETP8M1 for 16 and Tune CP5 for 17/18 and it was submitted. But now I get to know that a non-standard PDF set is used. Since these tunes were tuned based on NNPDF3.0(3.1), I was wondering whether this will tune make sense at all in this case. How should an analyst quantify the bias caused by this?

agrohsje:
The CUET tune is not based on NNPDF 3.0. We used it in 2016 together with NNPDF 3.0 in the ME's. But the PDF in CUET is NNPDF 2.3, like the Monash tunes from Peter Skands are using. A consistent choice of ME and PS PDF was something we introduced in 2017.

PR #2732, opened 2020-10-07
WJets NLO inclusive for tickets presented on 7Oct2020

mpresill:
Dear @agrohsje, all,

here are the cards I have used (and we have discussed together about on previous threads) for the submission of SMP-2020Oct07-00001 and SMP-2020Oct07-00002 tickets for legacy WJets NLO inclusive samples for 2017 and 2018. I have commented here and opened a PR to refresh all this thread! 
Let me know if anything is needed from my side.

Thanks for checking,
Matteo


agrohsje:
Your cards are similar enough to the final version: 
https://github.com/cms-sw/genproductions/tree/master/bin/GenValidation/updated_validation_cards/wjets_fxfx/w012jfxfx_inc
Customize card is not there but that is ok to be in sync with general 2017/2018 production.

mpresill:
thanks @agrohsje . Do I need to ping the prep-op HN for the submission of the related tickets?

agrohsje:
Yes. It is gen approved but needs submission. Thanks!

PR #2735, opened 2020-10-08
Fix bug in request script for UL20, make work from cmsenv

kdlong:
Something got messed up in the indention of the request script, this shows up when running after wget.

Also add a line to the get cookie call to clean the environment, except for the necessary kerberos variable, so if you call the script from a cmsenv it doesn't silently fail.

### DIFF
--- a/bin/utils/request_fragment_check.py
+++ b/bin/utils/request_fragment_check.py
@@ -8,7 +8,7 @@ import fnmatch
 #import json
 from datetime import datetime
 ###########Needed to check for ultra-legacy sample consistency check############################################
-os.system('cern-get-sso-cookie -u https://cms-pdmv.cern.ch/mcm/ -o cookiefile.txt --krb --reprocess')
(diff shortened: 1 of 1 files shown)


efeyazgan:
Thanks a lot! 
It is strange that I didn't have a problem (because of indentation) in the local tests I did in lxplus. 

The second modification is very good but just to be 100% sure it would be good if @justinasr checks it (to make sure it is OK for McM).

kdlong:
I suspect it is some combination of spaces and tabs that make it work in some cases and not others. 

For the other change, it just tells the script to ignore all environment variables than that one. I think it should really be safe. Do you mind merging and reverting if need be? I'd love to validate the mW samples overnight.

agrohsje:
Let me merge. We can always revert when Justinas has a different proposal.

justinasr:
I have not tested, but this makes sense and should work. I think everything will be fine.

PR #2737, opened 2020-10-08
add datacards for ultra low-mass DY sample

jingyucms:


agrohsje:
Dear @jingyucms , we are currently harmonizing the settings for DY production. 
The idea is to have a single template and use a script to generate all other files: 
https://github.com/cms-sw/genproductions/pull/2736
This will allow to avoid using different settings which caused a lot of problems in the past. For which campaign you wanted these samples? 20UL? The I would argue that it is better to use this new setup. Please let me know.

jingyucms:
> Dear @jingyucms , we are currently harmonizing the settings for DY production.
> The idea is to have a single template and use a script to generate all other files:
> #2736
> This will allow to avoid using different settings which caused a lot of problems in the past. For which campaign you wanted these samples? 20UL? The I would argue that it is better to use this new setup. Please let me know.

Does 20UL = UL for all 2016-2018 samples? If so, yes. Shall I do it using #2736?

agrohsje:
20UL means 20UL16, 20UL17, 20UL18. So one gridpack for all years. For which group are you doing this? Let me discuss on Monday a little more in the GEN coordination chat. I would propose to do central SMP samples.

jingyucms:
Then, yes, this is for 20UL all years. I am doing this for EXO (background sample for boosted low mass di-tau resonance search).

jingyucms:
@agrohsje Have you got a chance to talk to the GEN group? Please let me know if you need me to do anything.

agrohsje:
Dear @jingyucms . We discussed in coordination and will work on the production. Do you have a specific deadline in mind?

jingyucms:
@agrohsje I don't have a specific deadline, just would like it to be done ASAP :-)  You said the GEN coordination will work on the production? Does this mean that I can tell EXO MC contact that you are working on it and they do not need to worry about it anymore?

agrohsje:
Right. In fact, people like @sihyunjeon are following the effort and participating.

jingyucms:
Thanks a lot!

sihyunjeon:
@SanghyunKo @menglu21 @agrohsje 
Can you comment here about the HT bins? This is a request from EXO and wonder if gridpacks should be made differently not with HT binned

@jingyucms There were some studies from GEN group that HT binned might not be a wise thing to do for the moment

SanghyunKo:
@sihyunjeon @jingyucms maybe you can refer to [my answer](https://github.com/cms-sw/genproductions/pull/2770#issuecomment-741588865) to a similar topic. In short, X-sec from stitched HT-binned samples does not match properly with an inclusive one. So we're trying to understand the problem as much as possible at the moment before submitting them.

agrohsje:
@sihyunjeon. what should we do with this pr? I see pdfwgt = T. You could use the validation cards and adjust mll. If you prefer you can also discard the customize card with the PDG values to be in sync with other samples produced by susy/higg. let me know.

sihyunjeon:
@agrohsje I emailed Jingyu few days ago but there is no reply yet. Ah yes thanks for catching pdfwgt. And this is HT binned which reported to be a bit buggy by Sanghyun so I guess this can be CLOSED anyways. I'll contact Jingyu and ask him to change the binning variable and also change pdfwgt.

agrohsje:
Ok. Will close for now.

PR #2738, opened 2020-10-12
WWjj_OS EWK_noTop cards added

mlizzo:
These are the VBS OS cards for the legacy analysis. I just made a small fix, since bwcutoff was previously set to 15000 instead of 15

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/VBS/WWjj_OS/noTop/WWjj_OS_noTop_ewk_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/VBS/WWjj_OS/noTop/WWjj_OS_noTop_ewk_proc_card.dat
@@ -1,0 +1,37 @@
+#************************************************************
+#*                     MadGraph5_aMC@NLO                    *
+#*                                                          *
+#*                *                       *                 *
+#*                  *        * *        *                   *
+#*                    * * * * 5 * * * *                     *
+#*                  *        * *        *                   *
+#*                *                       *                 *
+#*                                                          *
+#*                                                          *
+#*         VERSION 2.2.1                 2014-09-25         *
+#*                                                          *
+#*    The MadGraph5_aMC@NLO Development Team - Find us at   *
+#*    https://server06.fynu.ucl.ac.be/projects/madgraph     *
+#*                                                          *
+#************************************************************
+#*                                                          *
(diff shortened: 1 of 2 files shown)


agrohsje:
Please remove this as it will cause problems in production and is not needed since 2.6:
#
#**************************************
# Parameter of the systematics study
#  will be used by SysCalc (if installed)
#**************************************                                  
#
0.5 1 2 = sys_scalefact  # factorization/renormalization scale factor
0.5 1 2 = sys_alpsfact  # \alpha_s emission scale factors
30 50 = sys_matchscale # variation of merging scale
# PDF sets and number of members (0 or none for all members).
NNPDF30_lo_as_0130.LHgrid = sys_pdf # matching scales
# MSTW2008nlo68cl.LHgrid 1  = sys_pdf 


Please remove unnecessary lines, like :
#************************************************************
#*                     MadGraph5_aMC@NLO                    *
#*                                                          *
#*                *                       *                 *
#*                  *        * *        *                   *
#*                    * * * * 5 * * * *                     *
#*                  *        * *        *                   *
#*                *                       *                 *
#*                                                          *
#*                                                          *
#*         VERSION 2.2.1                 2014-09-25         *
#*                                                          *
#*    The MadGraph5_aMC@NLO Development Team - Find us at   *
#*    https://server06.fynu.ucl.ac.be/projects/madgraph     *
#*                                                          *
#************************************************************
#*                                                          *
#*               Command File for MadGraph5_aMC@NLO         *
#*                                                          *
#*     run as ./bin/mg5_aMC  filename                       *
#*                                                          *
#************************************************************

set group_subprocesses Auto
set complex_mass_scheme False

Why do you plan to use 263000 as PDF? Why not NNPDF 3.1. This is UL, no? 
You could use this syntax, if you want 325500. You might also use LO if you prefer, but I would recommend to update to NNPDF 3.1 : 
lhapdf = pdlabel ! PDF set
 $DEFAULT_PDF_SETS = lhaid
see here https://twiki.cern.ch/twiki/bin/view/CMS/QuickGuideMadGraph5aMCatNLO

mlizzo:
I copied the old cards that were previously used for the 2016 sample, which I'm currently employing for the legacy analysis. I thought it would be possible to rerun the sample with just the bwcutoff correction

mlizzo:
I updated the cards with your comments, please let me know whether I need to implement more adjustments

agrohsje:
Looks good to me. @GurpreetSinghChahal please check and merge.

PR #2740, opened 2020-10-13
Output powheg job output to log and to stdout

kdlong:
The fact that the POWHEG output itself is hidden in a log inside a directory is confusing, and it also makes it impossible to debug jobs that run on the cluster or in central production. Also writing this to stdout is much more convenient.

### DIFF
--- a/bin/Powheg/runcmsgrid_powheg.sh
+++ b/bin/Powheg/runcmsgrid_powheg.sh
@@ -152,7 +152,7 @@ grep -q "pdfreweight 1" powheg.input ; test $? -eq 0 || produceWeights="false"
 grep -q "first runx" powheg.input ; test $? -ne 0 || produceWeights="true"
 
 cat powheg.input
-../pwhg_main &> log_${process}_${seed}.txt; test $? -eq 0 || fail_exit "pwhg_main error: exit code not 0"
+../pwhg_main 2>&1 | tee log_${process}_${seed}.txt; test $? -eq 0 || fail_exit "pwhg_main error: exit code not 0"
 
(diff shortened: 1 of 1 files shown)


agrohsje:
Thanks @kdlong. I fully agree. @covarell should I merge? What about UL branch. What is the plan for UL2019 and master for Powheg? Should we fully sync them?

kdlong:
WE should really consider having git submodules for different pieces. I believe @AndreasAlbert even set this up to some extent a long time ago.

AndreasAlbert:
I set up a splitting between code and cards once. This having been ~2 years ago, I'd have to look back to make myself familiar with it again. I still believe this works and is useful, so I'm all for that. One might also consider further submodules for different generators, etc (if they are reasonably independent).

agrohsje:
This would indeed be really nice to resume. I am wondering how to do this best. E.g. one could decouple PDFs, sqrts, ... from the actual cards. There is a priori no difference between a 14 tev and a 13 tev sample. I think we should replace relevant steering parameters similar to what we have already for pdfs.

covarell:
Hi, for the logging I surely agree to merge. 

About making UL2019 in sync with master, this is OK for me, but would you consider it safe already ? The new code by Jihun has never been tested yet on a large scale. We could have MC contacts test it, in fact, when they prepare UL requests.

agrohsje:
Thanks Roberto. Let me merge. 
Let's do the transition when we move to the more flexible repo scheme. 
@AndreasAlbert maybe you could think about a useful split and propose in an upcoming gen meeting? We can also chat with a few people independent of a GEN meeting. But maybe a talk is good to structure the discussion?

PR #2752, opened 2020-11-09
add cards Scalar Heavy Higgs to ttbar

kdeleo:


agrohsje:
What's the difference to this: 
https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Htottbar_dilepton ?

kdeleo:
> What's the difference to this:
> https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Htottbar_dilepton ?

It's the same process, only the ttbar decay is inclusive

agrohsje:
What is the reason for an inclusive sample? In 2016 we produced about the same stat for ll and ljets to have sufficient number of events. This will be challenging if you need to consider the BR.
For which campaign will you request MC? Note that we are currently preparing one for Ul. Why not coordinating with us (DESY,Rocherster people)?. 
Although assuming you will produce inclusive samples, there is no reason to add another path with partially duplicate info instead of extending the existing one. The more we spread info across genproductions the more settings will diverge.

ph2632:
Dear Ksenia and Alexander;
Let me add a bit more content here from B2G-RES group and MC-contact side.
The signal is supposed to be used in the tt-reso searches.
We have 3 searches in 0l, 1l, and 2l which will have the same signal interpretations and aim in common publication. 
This means we need tt-->inclusively. (The BR differences do not lead to the vast difference in statistics, at least all other interpretations are inclusive like this).
I know Ksenia is working on signal studies to motivate the number of events, masses, and widths needed.
She can also confirm that cards are the same as in ancestor production used in HIG-17-027, but couldn't find the git repository for them - perhaps you can point us.
The campaign will be mainly UL. While 3-6 points will be good to get promptly in preUL for early studies.
Thanks

agrohsje:
These are the previous cards: 
https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Htottbar_dilepton
You can rename the folder to heavyhiggstottbar and then have common cards as much as possible but add proper scripts to get yours. run_card should be the same for all if i am not mistaken.

kdeleo:
I renamed the directory Htottbar_dilepton as heavyhiggstottbar and I added my cards there. 
There are some differences in the run_card and in the gridpack_generation script, but if they are not significant, I will remove them to reduce redundant info

PR #2753, opened 2020-11-11
Add warning about instability of gridpack_genaration.sh in CERN EOS

juska:
See the warning message in the beginning of the script

### DIFF
--- a/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
+++ b/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
@@ -19,6 +19,24 @@
 #If QUEUE_SELECTION is omitted, then run on local machine only (using multiple cores)    #
 ##########################################################################################
 
+############################################################################
+# Warn about observed instability when running this script from a CERN EOS #
+# based path. By juska@cern.ch 11 Nov 2020                                 #
+############################################################################
+
+echo ""
+echo "Warning for CERN EOS and cernbox users!"
+echo ""
+echo "Running this script in a CERN EOS based directory has been observed"
+echo "to introduce instability which makes the majority of the processes to"
+echo "fail with varying obscure error messages pointing to missing directories etc."
+echo "The reason for this is not yet understood, but it looks like a race condition problem."
+echo "If you experience problems with EOS, please try in an AFS-based directory instead."
+echo "http://juska.web.cern.ch/juska/zprime/error_gallery.txt"
+echo "https://hypernews.cern.ch/HyperNews/CMS/get/generators/4761.html"
+echo ""
+sleep 5
+
 # Create tarball with very aggressive xz settings.
 # (trade memory and cpu usage for compression ratio)
 make_tarball () {


sihyunjeon:
Hi,
Could this not just be a problem that happens because some of the (or possibly all?) condor working nodes doesn't have eos area mounted?
I ran gridpack generation inside eos area locally (./gridpack_generation.sh PROCESSNAME /PATH/TO/CARD/) and this seems to work just fine.

juska:
I did all my test runs locally too, never played with condor.

Please take a look here if the error messages could tell you something: 
http://juska.web.cern.ch/juska/zprime/error_gallery.txt

sihyunjeon:
Could you also provide the card you used as well?
The cards I am testing doesn't fail after 30min run.

juska:
You can find example of the kind of cards we use here:
/afs/cern.ch/user/j/juska/public/zprime_cards

For me the failures occur mostly after roughly 50-120 minutes of running.

sihyunjeon:
@juska I tried 3 times with different cards (one of them with longer run time than the one you provided) which did not have problems at all. I was even able to create LHE files out of the gridpacks The only modification I made was removing the line in [1] (to avoid the escape due to slc version mismatch in lxplus) but I don't think this makes any difference. I put my results in /afs/cern.ch/work/s/shjeon/public/gridpack_in_EOS/.

Can someone else also check?
Anyhow, IF EOS IS INDEED UNSTABLE, I think it's still better to simply prevent people from using it rather than giving a warning message. @agrohsje What do you think?

[1] https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/gridpack_generation.sh#L622

kdlong:
My opinion/experience is that eos is absolutely unstable. It's definitely not impossible for it to work, but the mount quite often has read/write hangups. It might work, but it might not, and if it fails, the problem is very likely eos. I think the warning is a reasonable way to go.

kdlong:
Though I think it's a little bit of an overstatement to say that it's not understood. This is exactly why the afs --> eos plans were halted. It's just not reliable enough.

juska:
I have contacted also CERN service desk and described the problem and 
told my suspicions about race condition problems, but they said that 
they don't know that EOS would break any file operation rules. My 
empirical tests 10/10 jobs crashing on EOS but 9/10 succeeding in AFS is 
a fairly strong piece of evidence of the instability of EOS.
I wrote that the problem is not understood, because I have not been able 
to find any single reason for the errors, as the crashes happen with 
various different ways.

sihyunjeon:
> My opinion/experience is that eos is absolutely unstable. It's definitely not impossible for it to work, but the mount quite often has read/write hangups. It might work, but it might not, and if it fails, the problem is very likely eos. I think the warning is a reasonable way to go.

I thought echoing the message everytime and putting it to sleep for 5sec one runs the script is inefficient. Can we instead write the warning message inside the script like this? https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/gridpack_generation.sh#L1-L20

juska:
I do not trust that too many users bother to look inside the script and 
read the headers. If the normal running time of the script is tens of 
minutes, I don't think wasting five seconds for this is too big of a 
deal. And the user that bothers reading the script header can always 
comment out the sleep command. I'm not sure if a user would have enough 
time to react to the message if there's no pause. I have no strong 
opinion though.

agrohsje:
I fear so too. I don't think all MC contacts bother with reading the header, although they should.  They often don't even pay attention to warnings/failures from cout.

efeyazgan:
We can add a warning or error to the request checking script if you don’t have a better solution. Gridpacks are already checked for several things by the script anyway.

sihyunjeon:
@agrohsje Then do we want to allow people to use EOS from time to time, and in case if the gridpack generation fails, switch to a different place like AFS? I thought prventing EOS would be better due to its unstability, since it would be difficult to trace every generation logs especially when one makes multiple gridpacks at once for some signal samples.

I frankly don't agree that "warning messages should pop out from the generation script since users don't carefully read the headers" is an ideal solution for such cases since it's still users' responsibility to read the headers. But I guess it's only me who disagrees to this, so I think it's okay.

efeyazgan:
@sihyunjeon  no, if we make request checking script give an "error", then it stops the request to be validated. That's the point of the script. It is run automatically in mcm for every request before the validation step. So, it is not users responsibility. Of course, it is better to solve the issue at its root but until then we can do that.

kdlong:
There's nothing wrong with generating a gridpack on eos if it succeeds, so adding a check on an existing gridpack in the request checking script makes sense.

I really think the originally proposal is most reasonable: just warn the user that there is a chance of non-reproducible errors due to the file system. That way they won't go chasing an error that doesn't have a clear solution.

sihyunjeon:
@efeyazgan I was sticking to the gridpack generation only when I said "users (=MC contact's) responsibility" (1) to read the headers and (2) to check whether the generation ended up in failure or not (steps before injecting to MCM). So I thought echoing the error messages at the beginning of gridpack generation was rather redundant (for me I thought header would be enough since, again, it's their responsibility). Surely adding something to the request check script that can check whether the gridpack is corrupted or not due to EOS environment during MCM injection process would be a nice thing to do. I have no objection.

kdlong:
Oops I realized I missed a crucial word in my comment. I meant that the check in the request checking script _doesn't_ make sense.

agrohsje:
dear all. let me close the pr and add a comment on our twiki. i think the comment is also limited to /eos/home-*/ . in the scripts it will be confusing.

PR #2757, opened 2020-11-13
Change ggWW MCFM+JHUGen PDF for UL gridpacks

lviliani:
While redoing the MCFM+JHUGen gg->WW gridpacks for UL production, we also change the PDF set to NNPDF31_nnlo_as_0118_mc_hessian_pdfas.

Except for the PDF set, the cards are identical to the ones in the previous gridpacks.


agrohsje:
Will merge. Do you think you can do comparisons to the old setup? I would be curious to see the impact from nnlo as well as change in alphas. Would be an interesting talk for GEN. Let me know @lviliani.

PR #2744, opened 2020-10-19
Run/process card for 2HDM HPlus + h production

DaveBrun94:
These are the run and process cards for a 2HDM charged Higgs production in the association with a light CP even Higgs decaying to a lepton + four b-jet final state. It will be produced for UL.

It will be produced in the a grid of charged Higgs mass and light Higgs mass. My theorist colleagues provided SLHA files for each point in the grid, which are used as parameter cards. Should also these files be commited?

PR #2760, opened 2020-11-16
Pythia8 Hadronizer for 2HDM HPlus + h production decaying to lepton and 4b

DaveBrun94:
This is the the pythia8 hadronizer CMS python config for a 2HDM charged Higgs production in the association with a light CP even Higgs decaying to a lepton + four b-jet final state, see https://github.com/cms-sw/genproductions/pull/2744 for the madgraph cards.

I would appreciate if you could check if everything is proper configured.


agrohsje:
Hi @DaveBrun94. The question can only be answered in the context of the gridpack. L2 Gen signs off fragments before submission. What is the reason to store this here? We did in the past, but in principle it is hard to connect the info to the actual job, so it is a little useless, while all info can be consistently retrieved from mccm. let me know, alexander.

DaveBrun94:
Hey @agrohsje, thanks for your fast reply. My MC contacts and I were discussing the proper configuration, especially about the 'processParameters' block. They suggested to get in contact with you again because you also checked my madgraph cards. If you think everything can be done within mccm and its useless to put here, you can also close this PR.

agrohsje:
in your cards j doesn't include the b, so that part looks good. the only strange thing is that xqcut=qcut=20. how did you estimate this? qcut should be > xqcut.

DaveBrun94:
You are right, thank you. I updated the qcut

PR #2765, opened 2020-11-23
mg27x branch Madgraph5_aMCatNLO 0032*.patch fix

choij1589:
There was a mix of the patches between the master branch and the mg27x branch result in the failure of the patch. Fix for the mg27x branch.


agrohsje:
why did you close?

sihyunjeon:
@agrohsje  I found that mg27x patches weren't made properly and therefore not applied while toying with the mg27x branch. Jin said he made some mistakes while making the pull request few months ago. This one also had some typos so he is making a pull request again. I also asked him to double check whether the patch is applied or not so it should be good to go in for merging right away when the pull request is made.

PR #2766, opened 2020-11-23
fix mg27x 0032*.patch

choij1589:
There was a mix of 0032 patch between master branch and mg27x branch. Fix mg27x 0032*.patch.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/patches/0032-fix_madspin_seed_delivery.patch
+++ b/bin/MadGraph5_aMCatNLO/patches/0032-fix_madspin_seed_delivery.patch
@@ -24,18 +24,17 @@ index dd2be33..6796e51 100755
 +
              if cumul:
                  break
-         
+
 diff --git a/madgraph/interface/common_run_interface.py b/madgraph/interface/common_run_interface.py
-index 4265aea..da84476 100755
+index 49ae661..8efde64 100755
 --- a/madgraph/interface/common_run_interface.py
 +++ b/madgraph/interface/common_run_interface.py
-@@ -3678,8 +3678,9 @@ class CommonRunCmd(HelpToCmd, CheckValidForCmd, cmd.Cmd):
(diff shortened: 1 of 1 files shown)


bin/MadGraph5_aMCatNLO/patches/0032-fix_madspin_seed_delivery.patch
```diff
@@ -24,18 +24,17 @@ index dd2be33..6796e51 100755
...
+index 49ae661..8efde64 100755
 --- a/madgraph/interface/common_run_interface.py
 +++ b/madgraph/interface/common_run_interface.py
-@@ -3678,8 +3678,9 @@ class CommonRunCmd(HelpToCmd, CheckValidForCmd, cmd.Cmd):
-         for key, value in self.options.items():
+@@ -3697,7 +3697,8 @@ class CommonRunCmd(HelpToCmd, CheckValidForCmd, cmd.Cmd):
              if isinstance(value, str):
-                 madspin_cmd.mg5cmd.exec_cmd( 'set %s %s' %(key,value), errorhandling=False, printcmd=False, precmd=False, postcmd=True)
-+
+                 madspin_cmd.mg5cmd.exec_cmd( 'set %s %s --no_save' %(key,value), errorhandling=False, printcmd=False, precmd=False, postcmd=True)
```
agrohsje, line 34: spaces are irrelevant here, right? asking because you removed the for loop, so in principle everything should go to the left a bit. but i assume patching works still well. please confirm.

choij1589: I believe it's a difference between mg265 and mg273. The only line I changed is L38 as same as https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/patches/0032-fix_madspin_seed_delivery.patch

PR #2769, opened 2020-11-25
to check the ttz cards by Gen conveners

mmnajafabadi:
Dear Generator conveners,  
We would like to ask you please check the ttZ with Z > nu nu, for production. The gridpack is ready. 
Thanks a lot.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/ttznunu_5f_NLO/ttznunu_5f_NLO_run_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/ttznunu_5f_NLO/ttznunu_5f_NLO_run_card.dat
@@ -121,8 +121,8 @@ average = event_norm ! Normalize events to sum or average to the X sect.
 # (more specific gen cuts can be specified in SubProcesses/cuts.f)     *
 #***********************************************************************
    10.0  = ptl     ! Min lepton transverse momentum
-   3.0  = etal    ! Max lepton abs(pseudo-rap) (a value .lt.0 means no cut)
-   0.3  = drll    ! Min distance between opposite sign lepton pairs
+   5.0  = etal    ! Max lepton abs(pseudo-rap) (a value .lt.0 means no cut)
+   0.1  = drll    ! Min distance between opposite sign lepton pairs
    0  = drll_sf ! Min distance between opp. sign same-flavor lepton pairs
    0  = mll     ! Min inv. mass of all opposite sign lepton pairs
    0  = mll_sf  ! Min inv. mass of all opp. sign same-flavor lepton pairs
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/ttznunu_5f_NLO/ttznunu_5f_NLO_run_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/ttznunu_5f_NLO/ttznunu_5f_NLO_run_card.dat
@@ -25,9 +25,9 @@
 # (relative) accuracy on the Xsec.                                     *
 # These values are ignored for fixed order runs                        *
 #***********************************************************************
-  21000000 = nevents ! Number of unweighted events requested 
+  1500 = nevents ! Number of unweighted events requested 
  0.001 = req_acc ! Required accuracy (-1=auto determined from nevents)
-    1000000 = nevt_job! Max number of events per job in event generation. 
+    20 = nevt_job! Max number of events per job in event generation. 
                  !  (-1= no split).
 average = event_norm ! Normalize events to sum or average to the X sect.
 #***********************************************************************
@@ -55,7 +55,7 @@ average = event_norm ! Normalize events to sum or average to the X sect.
 # PDF choice: this automatically fixes also alpha_s(MZ) and its evol.  *
 #***********************************************************************
  lhapdf    = pdlabel   ! PDF set                                     
- 292200    = lhaid     ! if pdlabel=lhapdf, this is the lhapdf number
+ $DEFAULT_PDF_SETS    = lhaid     ! if pdlabel=lhapdf, this is the lhapdf number
 #***********************************************************************
 # Include the NLO Monte Carlo subtr. terms for the following parton    *
 # shower (HERWIG6 | HERWIGPP | PYTHIA6Q | PYTHIA6PT | PYTHIA8)         *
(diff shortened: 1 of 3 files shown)


bin/MadGraph5_aMCatNLO/cards/production/13TeV/ttznunu_5f_NLO/ttznunu_5f_NLO_run_card.dat
```diff
@@ -0,0 +1,146 @@
...
+#***********************************************************************
+# Tag name for the run (one word)                                      *
+#***********************************************************************
+  tag_1     = run_tag ! name of the run 
+#***********************************************************************
+# Number of events (and their normalization) and the required          *
+# (relative) accuracy on the Xsec.                                     *
+# These values are ignored for fixed order runs                        *
+#***********************************************************************
+  21000000 = nevents ! Number of unweighted events requested
```
sihyunjeon, line 28: @agrohsje This would pilot-run 2100K events right?

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ttznunu_5f_NLO/ttznunu_5f_NLO_run_card.dat
```diff
@@ -0,0 +1,146 @@
...
+#***********************************************************************
+    1   = lpp1    ! beam 1 type (0 = no PDF)
+    1   = lpp2    ! beam 2 type (0 = no PDF)
+ 6500   = ebeam1  ! beam 1 energy in GeV
+ 6500   = ebeam2  ! beam 2 energy in GeV
+#***********************************************************************
+# PDF choice: this automatically fixes also alpha_s(MZ) and its evol.  *
+#***********************************************************************
+ lhapdf    = pdlabel   ! PDF set                                     
+ 292200    = lhaid     ! if pdlabel=lhapdf, this is the lhapdf number
```
sihyunjeon, line 58: Is there are a reason to specify this 292200?
If not please change this to

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ttznunu_5f_NLO/ttznunu_5f_NLO_run_card.dat
```diff
@@ -0,0 +1,146 @@
...
+# Reweight flags to get scale dependence and PDF uncertainty           *
+# For scale dependence: factor rw_scale_up/down around central scale   *
+# For PDF uncertainty: use LHAPDF with supported set                   *
+#***********************************************************************
+ .true.   = reweight_scale   ! reweight to get scale dependence
+  0.5     = rw_Rscale_down   ! lower bound for ren scale variations
+  2.0     = rw_Rscale_up     ! upper bound for ren scale variations
+  0.5     = rw_Fscale_down   ! lower bound for fact scale variations
+  2.0     = rw_Fscale_up     ! upper bound for fact scale variations
+ .true.  = reweight_PDF     ! reweight to get PDF uncertainty
```
sihyunjeon, line 95: $DEFAULT_PDF_MEMBERS = reweight_PDF

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ttznunu_5f_NLO/ttznunu_5f_NLO_run_card.dat
agrohsje: :-D 
I was wondering how this gridpack was ever finished.

sihyunjeon:
@agrohsje I wonder if the tt+X sample cards could be also synchronised just like the DY and W samples. If you look at the two links below, one ttW is done with FXFX matching 1 jet, but ttZ with 0 jets. gridpack generation time won't be too much different right?

https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TTWJets/TTWJetsToQQ_5f_NLO_FXFX/TTWJetsToQQ_5f_NLO_FXFX_proc_card.dat

https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TTZJets/TTZJetsToQQ_5f_NLO/TTZJetsToQQ_5f_NLO_proc_card.dat

agrohsje:
Hi @sihyunjeon. I was wondering the same. I think this is generally a good idea. The question is how far we can push this. But yes, I agree with the proposal!

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ttznunu_5f_NLO/ttznunu_5f_NLO_run_card.dat
mmnajafabadi: Hello, the gridpack has been generated with few hundreds of events, I changed this number of events to point out here that we need around 20 M events. I did not know at which step I should request the number of events. I would be grateful if you let me know.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ttznunu_5f_NLO/ttznunu_5f_NLO_run_card.dat
mmnajafabadi: There is no specific reason for this PDFset, I got the run_card from an old production and did not change it. Now I set it to default as you recomment. thanks.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ttznunu_5f_NLO/ttznunu_5f_NLO_run_card.dat
mmnajafabadi: Yes, sure, I will change to default according to your recommendation.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ttznunu_5f_NLO/ttznunu_5f_NLO_run_card.dat
sihyunjeon: 2000-3000 should be okay (I think making this too small had some possibilities that it might be problematic, but not yet deeply investigated though, if I remember correctly)

The number of events you request for the samples would be something that you should send to PAG MC contacts.

sihyunjeon:
> Hi @sihyunjeon. I was wondering the same. I think this is generally a good idea. The question is how far we can push this. But yes, I agree with the proposal!

Hmmm yeah that's something we should discuss...

For now I think it would be nice to synchronise ttW, ttZ, ttH, ttG if possible.
And for like dibosons, WW, WZ, ZZ should be synchronised.
But I don't think it's necessary to synchronise ttW and WZ samples for example.

For the samples that belong in the same category (diboson, triboson, tt+X, ...) should be in sync : number of additional partons, lepton cuts, pdfs, cptunes, etc, this is what I have in mind now.

Maybe you can raise this in slack or bring menglu and sanghyun here

agrohsje:
Dear @mmnajafabadi. I cleaned up the cards and lowered cuts. Hope that is ok.
@sihyunjeon: Let me know if you are ok with merging?

sihyunjeon:
@agrohsje It is fine with me

mmnajafabadi:
Dear all, Many thanks for reviewing the cards. The modifications are fine to me.

PR #2770, opened 2020-11-26
adding DYjets and Wjets background samples VHbb

panwarlsweet:
- Adding DYJets, DYBJets, WJets, WBJets samples

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/DYBJets_HT-incl_Zpt-100to200/DYBJets_HT-incl_Zpt-100to200_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/DYBJets_HT-incl_Zpt-100to200/DYBJets_HT-incl_Zpt-100to200_proc_card.dat
@@ -1,0 +1,21 @@
+set group_subprocesses Auto
+set ignore_six_quark_processes False
+set loop_color_flows False
+set gauge unitary
+set complex_mass_scheme False
+set max_npoint_for_channel 0
+import model sm-ckm_no_b_mass
+define l+ = e+ mu+
+define l- = e- mu-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+define p = p b b~
+define j = p
+define bquark = b b~
+define ell+ = e+ mu+ ta+
+define ell- = e- mu- ta-
+generate p p > ell+ ell- bquark / h @0
+add process p p > ell+ ell- bquark j / h @1
+add process p p > ell+ ell- bquark j j / h @2
+add process p p > ell+ ell- bquark j j j / h @3
+output DYBJetsToLL_M-50_Zpt-100to200_5f_LO_MLM -nojpeg
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/DYBJets_HT-incl_Zpt-100to200/DYBJets_HT-incl_Zpt-100to200_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/DYBJets_HT-incl_Zpt-100to200/DYBJets_HT-incl_Zpt-100to200_proc_card.dat
@@ -1,16 +1,4 @@
-set group_subprocesses Auto
-set ignore_six_quark_processes False
-set loop_color_flows False
-set gauge unitary
-set complex_mass_scheme False
-set max_npoint_for_channel 0
 import model sm-ckm_no_b_mass
-define l+ = e+ mu+
-define l- = e- mu-
-define vl = ve vm vt
-define vl~ = ve~ vm~ vt~
-define p = p b b~
-define j = p
 define bquark = b b~
 define ell+ = e+ mu+ ta+
 define ell- = e- mu- ta-
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/DYBJets_HT-incl_Zpt-100to200/DYBJets_HT-incl_Zpt-100to200_run_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/DYBJets_HT-incl_Zpt-100to200/DYBJets_HT-incl_Zpt-100to200_run_card.dat
@@ -1,0 +1,271 @@
+#*********************************************************************
+#                       MadGraph5_aMC@NLO                            *
+#                                                                    *
+#                     run_card.dat MadEvent                          *
+#                                                                    *
+#  This file is used to set the parameters of the run.               *
+#                                                                    *
+#  Some notation/conventions:                                        *
+#                                                                    *
+#   Lines starting with a '# ' are info or comments                  *
+#                                                                    *
+#   mind the format:   value    = variable     ! comment             *
+#*********************************************************************
+#
+#*******************
+# Running parameters
+#*******************
+#
+#*********************************************************************
+# Tag name for the run (one word)                                    *
+#*********************************************************************
+  tag_1     = run_tag ! name of the run
+#*********************************************************************
+# Run to generate the grid pack                                      *
+#*********************************************************************
+  .false.     = gridpack  !True = setting up the grid pack
+#*********************************************************************
+# Number of events and rnd seed                                      *
+# Warning: Do not generate more than 1M events in a single run       *
+# If you want to run Pythia, avoid more than 50k events in a run.    *
+#*********************************************************************
+  250 = nevents ! Number of unweighted events requested
+      0       = iseed   ! rnd seed (0=assigned automatically=default))
+#*********************************************************************
+# Collider type and energy                                           *
+# lpp: 0=No PDF, 1=proton, -1=antiproton, 2=photon from proton,      *
+#                                         3=photon from electron     *
+#*********************************************************************
+        1     = lpp1    ! beam 1 type
+        1     = lpp2    ! beam 2 type
+     6500     = ebeam1  ! beam 1 total energy in GeV
+     6500     = ebeam2  ! beam 2 total energy in GeV
+#*********************************************************************
+# Beam polarization from -100 (left-handed) to 100 (right-handed)    *
+#*********************************************************************
+        0     = polbeam1 ! beam polarization for beam 1
+        0     = polbeam2 ! beam polarization for beam 2
+#*********************************************************************
+# PDF CHOICE: this automatically fixes also alpha_s and its evol.    *
+#*********************************************************************
+ lhapdf    = pdlabel     ! PDF set
+ $DEFAULT_PDF_SETS = lhaid
+ $DEFAULT_PDF_MEMBERS  = reweight_PDF
+#*********************************************************************
+# Renormalization and factorization scales                           *
+#*********************************************************************
+ F        = fixed_ren_scale  ! if .true. use fixed ren scale
+ F        = fixed_fac_scale  ! if .true. use fixed fac scale
+ 91.1880  = scale            ! fixed ren scale
+ 91.1880  = dsqrt_q2fact1    ! fixed fact scale for pdf1
+ 91.1880  = dsqrt_q2fact2    ! fixed fact scale for pdf2
+ 1        = scalefact        ! scale factor for event-by-event scales
+#*********************************************************************
+# Matching - Warning! ickkw > 1 is still beta
+#*********************************************************************
+ 1        = ickkw            ! 0 no matching, 1 MLM, 2 CKKW matching
+ 1        = highestmult      ! for ickkw=2, highest mult group
+ 1        = ktscheme         ! for ickkw=1, 1 Durham kT, 2 Pythia pTE
+ 1        = alpsfact         ! scale factor for QCD emission vx
+ F        = chcluster        ! cluster only according to channel diag
+ F        = pdfwgt           ! for ickkw=1, perform pdf reweighting
+ 5        = asrwgtflavor     ! highest quark flavor for a_s reweight
+ T        = clusinfo         ! include clustering tag in output
+ 3.0      = lhe_version       ! Change the way clustering information pass to shower.
+#*********************************************************************
+#**********************************************************
+#
+#**********************************************************
+# Automatic ptj and mjj cuts if xqcut > 0
+# (turn off for VBF and single top processes)
+#**********************************************************
+   T  = auto_ptj_mjj  ! Automatic setting of ptj and mjj
+#**********************************************************
+#
+#**********************************
+# BW cutoff (M+/-bwcutoff*Gamma)
+#**********************************
+  15  = bwcutoff      ! (M+/-bwcutoff*Gamma)
+#**********************************************************
+# Apply pt/E/eta/dr/mij cuts on decay products or not
+# (note that etmiss/ptll/ptheavy/ht/sorted cuts always apply)
+#**********************************************************
+   T  = cut_decays    ! Cut decay products
+#*************************************************************
+# Number of helicities to sum per event (0 = all helicities)
+# 0 gives more stable result, but longer run time (needed for
+# long decay chains e.g.).
+# Use >=2 if most helicities contribute, e.g. pure QCD.
+#*************************************************************
+   0  = nhel          ! Number of helicities used per event
+#*******************
+# Standard Cuts
+#*******************
+#
+#*********************************************************************
+# Minimum and maximum pt's (for max, -1 means no cut)                *
+#*********************************************************************
+ 0.01  = ptj       ! minimum pt for the jets
+  0  = ptb       ! minimum pt for the b
+ 10  = pta       ! minimum pt for the photons
+  0  = ptl       ! minimum pt for the charged leptons
+  0  = misset    ! minimum missing Et (sum of neutrino's momenta)
+  0  = ptheavy   ! minimum pt for one heavy final state
+  0  = ptonium   ! minimum pt for the quarkonium states
+ -1  = ptjmax    ! maximum pt for the jets
+ -1  = ptbmax    ! maximum pt for the b
+ -1  = ptamax    ! maximum pt for the photons
+ -1  = ptlmax    ! maximum pt for the charged leptons
+ -1  = missetmax ! maximum missing Et (sum of neutrino's momenta)
+#*********************************************************************
+# Minimum and maximum E's (in the center of mass frame)              *
+#*********************************************************************
+  0  = ej     ! minimum E for the jets
+  0  = eb     ! minimum E for the b
+  0  = ea     ! minimum E for the photons
+  0  = el     ! minimum E for the charged leptons
+ -1   = ejmax ! maximum E for the jets
+ -1   = ebmax ! maximum E for the b
+ -1   = eamax ! maximum E for the photons
+ -1   = elmax ! maximum E for the charged leptons
+#*********************************************************************
+# Maximum and minimum absolute rapidity (for max, -1 means no cut)   *
+#*********************************************************************
+   5  = etaj    ! max rap for the jets
+  -1  = etab    ! max rap for the b
+  -1  = etaa    ! max rap for the photons
+  -1  = etal    ! max rap for the charged leptons
+  -1  = etaonium ! max rap for the quarkonium states
+   0  = etajmin ! min rap for the jets
+   0  = etabmin ! min rap for the b
+   0  = etaamin ! min rap for the photons
+   0  = etalmin ! main rap for the charged leptons
+#*********************************************************************
+# Minimum and maximum DeltaR distance                                *
+#*********************************************************************
+ 0.0 = drjj    ! min distance between jets
+ 0   = drbb    ! min distance between b's
+ 0   = drll    ! min distance between leptons
+ 0.0   = draa    ! min distance between gammas
+ 0   = drbj    ! min distance between b and jet
+ 0.1 = draj    ! min distance between gamma and jet
+ 0   = drjl    ! min distance between jet and lepton
+ 0   = drab    ! min distance between gamma and b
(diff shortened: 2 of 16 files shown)


agrohsje:
Can you comment on the purpose of the request. Is this for UL? For DYJets and Wjets, we will generate gridpacks based on the validation cards. Due to the problems we had in the past we prefer harmonizing. 

About Wb and Dyb. Was this tested? Does it work with the merging? j also has b's. for p8 you will specify the number of jets to merge. won't the b cause confusion?

panwarlsweet:
Yes, this is for UL. Sorry but, I don't have that much knowledge about these samples as we have been asked to get these samples submitted ASAP. So I found 2017 cards and I have copied the process cards from here 
https://github.com/cms-sw/genproductions/tree/mg242legacy/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/DYBJets_HT-incl_Zpt-100to200

All the cards I have taken from 2017 ones which were already merged with genproduction mg242legacy branch. So I assumed everything has been defined properly.

Could you please explain a bit what the is issue with these WJets and DY samples?

agrohsje:
Hi @panwarlsweet . You mean the issue with the Wb,DYb samples?  Who requested this? Can you provide some background info?  

For the others you don't have to worry, GEN will take care.

panwarlsweet:
The VHbb analysts are using these b-enriched samples from non-legacy production. And these are required from UL campaign.

Thanks for making clear the case with DYJets and WJets. I assume ZJets is also in the same queue. Right? Also could you please let us know by when it will be available on eos so that we could make the mcm requests?

panwarlsweet:
Hello @agrohsje !

Could you please comment on b-enriched cards? Should we follow up with analysts for this?

agrohsje:
Yes. Please. Would be curious to understand more.

capalmer85:
@agrohsje Yes, we have a detailed stitching procedure that makes use of all of the events in both sets of samples (HT-binned and b-enriched).  Our AN (http://cms.cern.ch/iCMS/jsp/openfile.jsp?tp=draft&files=AN2019_229_v6.pdf) has a full description in section 4.4.  We can also point you to presentations if you like, but the short answer is:  yes, we need these and are able to use them effectively.

panwarlsweet:
@agrohsje could you please let us know by when the WJet, ZJets and DYJets grid-packs will be available? Also, the AN shared by Chris for b-enriched samples, does this look good to go ahead with the PR? We have been asked to put all these samples in the production before the holidays. So if you could provide feedback that would be helpful for us. Thanks :)

agrohsje:
yes. that looks reasonable. let me check with the team, if we also want to produce the b-enriched samples. we have the same plan: getting everything into the production system before the break. it is tough and a lot of work.

SanghyunKo:
Hi @panwarlsweet, I quickly looked through the AN and have a comment and a question from GEN point of view.

For the HT-binned samples, I'm afraid we may not be able to produce them in a short time scale, since historically we've received countless reports for the inconsistency between inclusive & HT-binned samples during the legacy production, and unfortunately the issue still seems to persist. So I wonder it is ok for you that sticking with inclusive samples at the moment and coming back to HT-binned later.

Another question related to Zpt binned sample: we used to have `DY M-50 Zpt-150toInf MLM` sample centrally, and we would like to keep Zpt binning of flavour enriched samples consistent with (flavour-)inclusive sample as much as possible to prevent diverging between binned and inclusive ones. Would you mind for us to sync the Zpt binning?

capalmer85:
Hi @SanghyunKo.  Thanks for following up.  Both of these changes are actually quite problematic at the analysis level.  Let me address them in turn.

1) using only inclusive samples.  The VHbb relies on tagging a very boosted Z or W.  The most sensitive region is where P_T,V is 250-400 GeV.  The inclusive sample does not cover this at all and even the b-enriched samples tend to start to run out of event in the higher P_T part of this regime. The HT-binned samples very necessary for the phase space of the analysis.  Can you give some details about the issues with these samples?

2) adjusting P_T,V binning.  We could adjust the binning, but changing to one sample with only >150 GeV would also be very sub-optimal for the VHbb analysis.  Besides trying to line up with other samples, is there another goal that you'd like to achieve in changing the binning?  Can we achieve harmonizing the run cards with some flags for start/stop P_T,V perhaps?

SanghyunKo:
@capalmer85 I understand your point. For the HT-binned sample issue, maybe you can refer to [this presentation](https://indico.cern.ch/event/781231/contributions/3263952/subcontributions/274982/attachments/1795600/2926947/190213_mc_discrep_ppd.pdf). It historically persisted quite a while, and my naive guess is that the reweighting procedure from LO to NLO describe in AN might have been shadowed(or fixed?) the issue. It's X-sec issue and since you've reweighted HT-binned sample to another sample, I would assume the issue did not cause any problem on your analysis.

Of course, this does not mean that we're not going to produce HT-binned sample: I understand it is one of the most useful and popular samples in CMS, just want to make sure in this round that we don't put it in the production queue which we already know that's problematic (as long as the time permits). In the worst case, we may produce it as it already is, and provide some reweighting recipe to sync with inclusive ones.

For the Pt binning, I wonder if it is the case that you need high ZpT events and eventually reweights LO events to NLO, why don't you try to use `DYJetsToLL_Pt-*To*_TuneCP5_13TeV-amcatnloFXFX-pythia8`? They're NLO samples with ZpT binning [0, 50, 100, 250, 400, 650, Inf].

capalmer85:
Sorry for the long wait.  I didn't realize the loop wasn't closed.  We would like to merge this right away because we still want the b-enriched samples in this campaign.  

As for your suggestion about the NLO samples, we are aware of them, but the large fraction of negative weights and relatively small fraction of events with b-jets makes them difficult to use as our nominal samples.  We do of course need to reweight LO samples to match NLO samples in important analysis variables (e.g. particle-level vector boson PT).  

The HT-binned UL samples should not have the issues that exist in the current samples because there has been a big bug-fix in madgraph and the affected setting (pdfweight) is now set to "false", which should have been done before.

capalmer85:
@SanghyunKo can we merge this?

sihyunjeon:
@capalmer85 You mean with MGv26x HT binned has no such issues? If I remember correctly, I think @SanghyunKo still observed such behaviour with recent MG releases.

And besides pdfwgt=T/F was somewhat irrelevant to this HT binned specific issue. It was also shown in inclusive samples

capalmer85:
This PR isn't about the HT-binned samples.  I don't think there is a suggestion for any change of code that is relevant for these b-enriched samples.  Maybe I missed it?

sihyunjeon:
Ah apologies, I just naivley assumed this was only about HT binned samples after reading Sanghyun's latest comment.

The cards seem fine @agrohsje What do you think?

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/DYBJets_HT-incl_Zpt-100to200/DYBJets_HT-incl_Zpt-100to200_run_card.dat
```diff
@@ -0,0 +1,271 @@
...
+#*******************
+#
+#*********************************************************************
+# Tag name for the run (one word)                                    *
+#*********************************************************************
+  tag_1     = run_tag ! name of the run
+#*********************************************************************
+# Run to generate the grid pack                                      *
+#*********************************************************************
+  .false.     = gridpack  !True = setting up the grid pack
```
SanghyunKo, line 26: Isn't `gridpack` supposed to be `true`?

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/DYBJets_HT-incl_Zpt-100to200/DYBJets_HT-incl_Zpt-100to200_run_card.dat
```diff
@@ -0,0 +1,271 @@
...
+#
+#**********************************
+# BW cutoff (M+/-bwcutoff*Gamma)
+#**********************************
+  15  = bwcutoff      ! (M+/-bwcutoff*Gamma)
+#**********************************************************
+# Apply pt/E/eta/dr/mij cuts on decay products or not
+# (note that etmiss/ptll/ptheavy/ht/sorted cuts always apply)
+#**********************************************************
+   T  = cut_decays    ! Cut decay products
```
SanghyunKo, line 93: supposed to be `False`, I guess? (to be synchronized with [inclusive sample](https://github.com/cms-sw/genproductions/blob/master/bin/GenValidation/updated_validation_cards/dy_mlm_mll50/dy01234mlm_inc/dyellell01234j_5f_LO_MLM_run_card.dat#L98))

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/DYBJets_HT-incl_Zpt-100to200/DYBJets_HT-incl_Zpt-100to200_run_card.dat
panwarlsweet: Hello @SanghyunKo! As I remember it is set to true automatically while using gridpack production script.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/DYBJets_HT-incl_Zpt-100to200/DYBJets_HT-incl_Zpt-100to200_run_card.dat
sihyunjeon: This won't have any effect to current setup where no mediate resonance exists.

1. If it's T=cut_decays with ptl=20
p p > z, z > e+ e- ==> electrons have pt > 20

2. If it's F=cut_decays with ptl=20
p p > z, z > e+ e- ==> electrons have no ptcut

3. If it's T/F=cut_decays with ptl=20
p p > e+ e- ==> electrons have pt > 20

panwarlsweet: It was set to true by following the non-legacy cards
https://github.com/cms-sw/genproductions/blob/mg242legacy/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/DYBJets_HT-incl_Zpt-100to200/DYBJets_HT-incl_Zpt-100to200_run_card.dat#L93

I am not that sure about it. Should I modify this?

capalmer85: Probably not great to cut on the lepton PTs.  However, it shouldn't matter too much because the Zpt is at least 100 GeV.  I don't think I've ever seen any issue here in practice.  I'll check now though.

sihyunjeon: @capalmer85 I didn't mean to say you should give cuts to lepton pts. ptl is just an **example**. What I meant to say is that "cut_decays" decide whether decay particles from mediate resonances are applied with cuts or not. And for your case, cut_decays = T won't really do anything since you haven't explicitly write the mediate resonance.

capalmer85: I actually see plenty of events where the lepton pt is less than 20 GeV and the parent is the Z boson in the DYB 100-200 GeV sample.  So lepton cuts weren't there before and if the configuration hasn't changed, I guess I don't expect them there now.

SanghyunKo: @panwarlsweet @capalmer85 @sihyunjeon Thanks for the swift responses. I was checking the run cards by 1-1 comparison with inclusive ones apart from processes so @sihyunjeon's comment should be right.

capalmer85: hmmm... does look like there is some effect afterall.

https://capalmer.web.cern.ch/capalmer/DYB100to200_leptonpt.png

sihyunjeon: @capalmer85 I think my comments have confused you. ptl is just an example. Not somthing that is asked to give to your gridpacks. I just wanted to explain to Sanghyun what cut_decays=T or F can or cannot do when (for instance) ptl=20 is given for "p p > z, z > e+ e-" or "p p > e+ e-" processes irrelevant to your cards.

sihyunjeon: @capalmer85 Your cards are fine. Either you give cut_decays = T or cut_decays = F, what you get should stay the same.

sihyunjeon: Resolved through Skype as I caused some confusions with the ptl comments. Resolving this.

sihyunjeon:
@agrohsje Can you take a look?

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/DYBJets_HT-incl_Zpt-100to200/DYBJets_HT-incl_Zpt-100to200_run_card.dat
capalmer85: Thanks!  All good for me.

capalmer85:
@agrohsje and @sihyunjeon still not merged? :(

capalmer85:
@agrohsje and @sihyunjeon

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/DYBJets_HT-incl_Zpt-100to200/DYBJets_HT-incl_Zpt-100to200_proc_card.dat
```diff
@@ -0,0 +1,21 @@
...
+import model sm-ckm_no_b_mass
+define l+ = e+ mu+
+define l- = e- mu-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+define p = p b b~
+define j = p
+define bquark = b b~
+define ell+ = e+ mu+ ta+
+define ell- = e- mu- ta-
```
agrohsje, line 16: please remove everything that is not relevant for you. defintions/models you don't use, default mg5 settings ...

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/DYBJets_HT-incl_Zpt-200/DYBJets_HT-incl_Zpt-200_proc_card.dat
```diff
@@ -0,0 +1,21 @@
...
+import model sm-ckm_no_b_mass
+define l+ = e+ mu+
+define l- = e- mu-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+define p = p b b~
+define j = p
+define bquark = b b~
+define ell+ = e+ mu+ ta+
+define ell- = e- mu- ta-
```
agrohsje, line 16: as above

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/DYJets_HT-incl_Zpt-100to200/DYJets_HT-incl_Zpt-100to200_proc_card.dat
```diff
@@ -0,0 +1,21 @@
...
+set max_npoint_for_channel 0
+import model sm-ckm_no_b_mass
+define l+ = e+ mu+
+define l- = e- mu-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+define p u c s d b u~ c~ s~ d~ b~ g
+define l+ = e+ mu+ ta+
+define l- = e- mu- ta-
+define j  = p
```
agrohsje, line 15: same here.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/DYJets_HT-incl_Zpt-200/DYJets_HT-incl_Zpt-200_proc_card.dat
```diff
@@ -0,0 +1,21 @@
...
+set max_npoint_for_channel 0
+import model sm-ckm_no_b_mass
+define l+ = e+ mu+
+define l- = e- mu-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+define p u c s d b u~ c~ s~ d~ b~ g
+define l+ = e+ mu+ ta+
+define l- = e- mu- ta-
+define j  = p
```
agrohsje, line 15: and here

agrohsje:
dear lata, chris. the pr is not easy to review. we always recommend to clean up the process card and only keep what you are really doing. also we encourage people to remove default settings. there might nothing be wrong with the lines, but it is really wasting our time to review and make sure things are consistent. same for the run_card.dat. why not using a script and committting what is needed?  if all run_cards are the same, why having them multiple times?  and if one parameter or 2 are changing, why not replacing this in a script? 
we know that prs have been made in the past like this one and it is not your fault. but committing several files, mistakes sneek in and we have a hard time to review.  
morevoer. did you make sure you run card is and parameters are in sync with this dataset in UL: 
https://cms-pdmv.cern.ch/mcm/requests?prepid=TAU-RunIISummer20UL16wmLHEGEN-00001&page=0&shown=524319
people might combine files to  increase stat with that sample. for the time being I just did the check and it is in sync. But it would certainly help if you could do first and let us know you did. too often people start from random gridpacks and settings go out of sync.

agrohsje:
@panwarlsweet . If you tell me that the run cards are identical modulo the corresponding V pT cut, then I can also merge for the time being, once you cleaned up the process cards but would strongly encourage to use scripts in the future.

panwarlsweet:
@agrohsje, thanks for reviewing the cards. I agree with you that having a script is more convenient but I made this PR last year (as I used to do usually). I will follow your suggestion in future.

Regarding the run cards cross-check part, I found there is some difference in the run card settings, I am adding a screenshot of that (red is my card and the green one is from the link you shared). Although I don't think these parameters will impact much. However, it is better to get feedback if these should be exactly the same.

image


Apart from this, I also found some additional lines in our cards

======================================
   0 = ptgmin ! Min photon transverse momentum
 0.4 = R0gamma ! Radius of isolation code
 1.0 = xn ! n parameter of eq.(3.4) in hep-ph/9801442
 1.0 = epsgamma ! epsilon_gamma parameter of eq.(3.4) in hep-ph/9801442
 .true. = isoEM ! isolate photons from EM energy (photons and leptons)
===========================
  -1    =  ktdurham
 0.4  =  parameter
 ===========================
 
 But as far as I understand this should also be fine.
 
 Lata

agrohsje:
Thanks @panwarlsweet . Indeed I saw this, however, either the particle doesn't appear (like the photon) or the cuts have the same effect, like  max 100 or -1 .  But thanks for being careful. We really need to pay more attention as often also relevant settings are different.

agrohsje:
Hi @sihyunjeon . I also checked that the Wjets run cards are in sync with what we have in UL:
https://cms-pdmv.cern.ch/mcm/requests?prepid=BTV-RunIISummer20UL17wmLHEGEN-00001&page=0&shown=524319.
I removed the mll cut for Wb and cleaned up the process cards. Do you mind quickly checking a last time and merging this PR? Thanks so much!

agrohsje:
As Sihyun took a day off, @panwarlsweet can you cross check my fixes and ping me to merge. Thanks.

panwarlsweet:
Hello @agrohsje!

Sorry for the delay. Run cards changes are fine. But I have a question for the proc_cards of the B-enriched sample.

I find this line "define p = p b b~" for b-enriched one which you have removed while the default definition of p in MG
"Defined multiparticle p = g u c d s u~ c~ d~ s~
Defined multiparticle j = g u c d s u~ c~ d~ s~"

why did you prefer to go with the default one? The same line is there In old samples (non-legacy) as well.

Apart from this, I am fine with other changes.

agrohsje:
Hi @panwarlsweet, 
if you do: 
import model sm-ckm_no_b_mass 
the b is massless and the b is added to p as well as j.

agrohsje:
If you load that model and do 
p p > t t~ j 
you will get e.g. 
INFO: Process b g > t t~ b added to mirror process g b > t t~ b 
Let me know if you are ok with merging.

panwarlsweet:
Okay, thanks for clarifying this. 
Yes, I am fine with all the changes you have made. 
You could merge it now.

PR #2775, opened 2020-11-29
MonoHiggs 2HDMa for gamma gamma

DebabrataBhowmik:
Adding parameter scan points and gridpack producer in monoHiggs-2HDMa for gamma gamma channel


sihyunjeon:
@DebabrataBhowmik @agrohsje Though this is merged already, I have one question to ask.
Is there a reason not to use the $DEFAULT_PDF_SETS for lhaid and instead use 305800 in the run cards?
The basic recommendation for ME PDF + PS Tune should be NNPDF31NNLO + CP5

DebabrataBhowmik:
@sihyunjeon @agrohsje As you can see these cards do not refer to MonoH to gamma gamma only rather the whole monoH. 
The gridpacks were not also generated separately for gamma gamma, it's been used for MonoH to bb also which is already in production if I am not mistaken and also to other channels most probably 
So, for gamma gamma channel we wanted to use the same gridpacks.

agrohsje:
Hi @DebabrataBhowmik . Are you saying we have in UL production samples with PDF 305800 ? I don't remember approving...

DebabrataBhowmik:
@agrohsje @sihyunjeon No I am not saying ultra legacy, I am saying mono higgs to bb production for 2HDMa model 
The same gridpacks are being used for gamma gamma also just with few different parameter values

agrohsje:
Ok.

PR #2776, opened 2020-11-29
tt+ll lowmass cards

cericeci:
Fixed the cuts file for low mass TT+LL production to run on master branch. Gridpack can now be produced without fortran compilation problems

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TTZJets/TTZJetsToLL_M-1to10_5f_NLO/TTZJetsToLL_M-1to10_5f_NLO_cuts.f
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TTZJets/TTZJetsToLL_M-1to10_5f_NLO/TTZJetsToLL_M-1to10_5f_NLO_cuts.f
@@ -1,4 +1,3 @@
-c *This file contains maximum cuts(M<10) on the dilepton mass: line 136-142*
 c
 c This file contains the default cuts (as defined in the run_card.dat)
 c and can easily be extended by the user to include other.  This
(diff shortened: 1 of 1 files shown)


agrohsje:
Is this based on 2.6.5 cuts.f ? 
What are the production plans for this sample?

cericeci:
I adapted the old 2.4.2 file to run on 2.6.5 rather than starting from scratch with the 2.6.5 template.  In retrospective it might not be the best idea in case the process card is modified, so I can refactor it to the later for consistency if needed.

The plan was to use this one for UL production.

agrohsje:
Please see my proposal here: 
https://www.desy.de/~agrohsje/hidden/TTZJetsToLL_M-1to10_5f_NLO_cuts.f
It starts from 2.6.5 and should include your cuts. Please double-check and update the PR if ok with you.

cericeci:
Many thanks for the help @agrohsje. I merged your proposal, reproduced the gridpack and checked that it matches what we need.

sihyunjeon:
@agrohsje @cericeci 
I have one naive question about the naming of the gridpacks. It's named TTZJets while it doesn't have multiparton process added (surely you'd get additional partons from QCD NLO calculations but I am talking about the ME level added jets) while TTWJets has additional jets as in [1]. Wouldn't it be better to add one jet and do FxFx matching for this as well?

[1] https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TTWJets/TTWJetsToLNu_5f_NLO_FXFX/TTWJetsToLNu_5f_NLO_FXFX_proc_card.dat

agrohsje:
Hi @cericeci . I brought the same up in MCM. Let us know if the proposal above of adding one jet is what you like or not. If you don't want to redo production, we could still sync the settings and have one additional jet and ickkw=3 here.

sihyunjeon:
@agrohsje should we not close this?

agrohsje:
Will merge now.

PR #2777, opened 2020-11-30
VBF_H_WW PDF updated for UL gridpack

mlizzo:
PDF set updated to 325300 in order to use the card for the production of VBF_H_WW gridpack for the UL campaign. Other than that, no changes have been implemented

### DIFF
--- a/bin/Powheg/production/2017/13TeV/Higgs/VBF_H_WW_NNPDF31_13TeV/VBF_H_WW_NNPDF31_13TeV_M125.input
+++ b/bin/Powheg/production/2017/13TeV/Higgs/VBF_H_WW_NNPDF31_13TeV/VBF_H_WW_NNPDF31_13TeV_M125.input
@@ -1,0 +1,49 @@
+numevts NEVENTS    ! number of events to be generated
+ih1   1           ! hadron 1 (1 for protons, -1 for antiprotons)
+ih2   1           ! hadron 2 (1 for protons, -1 for antiprotons)
+ebeam1 6500d0     ! energy of beam 1
+ebeam2 6500d0     ! energy of beam 2
+
+lhans1  325300         ! pdf set for hadron 1 (LHA numbering)
+lhans2  325300         ! pdf set for hadron 2 (LHA numbering)	
+
+! Parameters to allow or not the use of stored data
+use-old-grid    1 ! if 1 use old grid if file pwggrids.dat is present (<> 1 regenerate)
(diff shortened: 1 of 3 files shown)


gourangakole:
Dear all;
I think I made a similar request earlier: why not modifying the template file instead of adding extra file:

https://github.com/cms-sw/genproductions/tree/master/bin/Powheg/production/2017/13TeV/Higgs/VBF_H_WW_NNPDF31_13TeV

agrohsje:
Hi @mlizzo. What about the comment from Gouranga and updating the template. @gourangakole: Why don't you guys clean up a bit and just stick to templates and scripts. I think having the explicit cards causes confusion. Let me know if I should merge.

mlizzo:
Hi @agrohsje, I used the makecards.py script to generate the card from the template file, I thought it was there exactly for this purpose (as I said in the comment, I just changed the PDF of the output file). I can update the PDF set in the template file, if that's what you mean, maybe it's useful since new gridpacks must be in compliance with UL requirements. Please let me know whether you agree.

agrohsje:
Right. I had in mind to update the template as well. Ideally one would just work on the template, run scripts locally but don't commit the output. The reason of the template is to have everything in one place. If we now commit all the time the cards, then this is not true anymore and people will start to change the files but not the template, exactly as you did. Anyway. Adjusting the template is good and @gourangakole and Alexandros might clean up a bit in a separate PR.

mlizzo:
Ok thanks for the clarification, maybe I misunderstood how the script was intended to be used

agrohsje:
Ok. Please update the template and I merge.

gourangakole:
> Right. I had in mind to update the template as well. Ideally one would just work on the template, run scripts locally but don't commit the output. The reason of the template is to have everything in one place. If we now commit all the time the cards, then this is not true anymore and people will start to change the files but not the template, exactly as you did. Anyway. Adjusting the template is good and @gourangakole and Alexandros might clean up a bit in a separate PR.

Yes, we shall do that. but for the time being as you suggested @mlizzo can update the template card with PDF.

mlizzo:
Done! Please let me know whether it's fine like this

agrohsje:
Happy with the PR. @gourangakole let me know and I merge.

gourangakole:
> Happy with the PR. @gourangakole let me know and I merge.

I meant to say not immediate 😄 ..may in a week we shall need to talk to other contacts. so for the time being if you are happy with the content of this PR. Then please merge. we shall proceed with UL grid pack production.

PR #2778, opened 2020-12-02
adjusted mergepl to distinguish between sum and average norm in mg5

agrohsje:


agrohsje:
Hi @qliphy and @efeyazgan . Does one of  you know why the version of merge.pl in cvmfs was used ? I now moved to a local one and adjusted the logic to accommodate average and sum norm: 
if (index(lc($event_norm), "average") != -1) {
    $uwgt = sprintf('%0.5E', $totxsec );
} elsif (index(lc($event_norm), "sum") != -1) {
    $uwgt = sprintf('%0.5E', $totxsec / $totevents);   
} else {
    die ("Unknown normalization mode \n");
}
The official version has  this line:
$uwgt = sprintf('%0.5E', $totxsec / $totevents);   
which gives wrong results when average event_norm (default in MG5) is used. It shouldn't matter as long as there are always the same number of events produced, but can introduce a bias and is confusing.
Let me know if you are ok with this PR.

kdlong:
@agrohsje: I modified this from the one in MadGraph in 2017. When you say a local one, do you mean a version distributed with MadGraph? Has MadGraph changed it at all since 2017 (I believe it was mostly deprecated at the time)? It remains the only Perl I've ever written, so their absolutely could be bugs. The plan at the time was that this would be a stopgap of a few months... There is some python code that could be adapted to accomplish this that the authors suggested at the time.

The reason it needed to be modified is because in some cases MadGraph would not generate events from ever subprocess, and those would not be represented in the file, and the merging had no mechanism to initialize them. Did you diff the new and current versions? Can you see any differences that are likely related?

kdlong:
Here is the diff I have from the time I edited it. There is also an email thread with Olivier that I could forward: https://github.com/kdlong/genproductions/compare/93dce985e2215cf3dd036f025cb326faf1c6b086...kdlong:d9081e4f639a6a779709036a69334b05fc5448a4

agrohsje:
Hi @kdlong. With local I mean included in genproductions. I was a bit confused that we are using a file from cvmfs. The merge.pl is rather short, so size of the gridpack is not an issue, but I see some risk in not being properly updated like it happened here, when event_norm sum/average where introduced in MG5. 
When I diff the version I commit with the one in cvmfs I see your changes above and my one to adjust the norm procedure. So I think this is ok.
To cross check I produced 350 events, once split in 4 jobs a 3x100+1x50 and once from a single run. A diff between the outputs for both (event_norm=sum or average) didn't show any difference in numbers (modulo statistics).

sihyunjeon:
@agrohsje I double checked the results giving same numbers.
@kdlong @qliphy @efeyazgan A question a little bit aside from this out of curiosity, does anyone know why for LO we split the event generations, but not for NLO?

colizz:
Hello, I have a quick question on this merge.pl script.

In MG LO gridpack produced by V2.6.5 I found that the \<MGGenerationInfo\> block does not seem to give the correct number of events.
E.g. I am producing 10 or 30 events from `/cvmfs/cms.cern.ch/phys_generator/gridpacks/2017/13TeV/madgraph/V5_2.6.5/VLQ/Tprime/TprimeToTZ/tZ_b_LH_M600_slc6_amd64_gcc630_CMSSW_9_3_16_tarball.tar.xz` (prepid: B2G-RunIISummer19UL16wmLHEGEN-00001), the \<MGGenerationInfo\> block is always

```
<MGGenerationInfo>
#  Number of Events        :       1000
#  Integrated weight (pb)   : 15.8224032545
</MGGenerationInfo>
```
The number of events is not 10 or 30.

Do we have any discussion/solution on this? It may influence LHE merging because this merge.pl script uses the above event number to calculate the new nEvent & Xsec which is actually not correct. (btw, it seems this issue is fixed in the latest MG version but I cannot trace back to when it is fixed)

agrohsje:
Hi @colizz. Should we update the script merge.pl in genproduction to the python version to simplify maintenance?

colizz:
> Hi @colizz. Should we update the script merge.pl in genproduction to the python version to simplify maintenance?

Thanks @agrohsje I think this is a good idea. Would you suggest to refer to the script from `cmssw` by URL or put a copy here?

agrohsje:
Maybe it is easier to decouple. The problem is that the lhe part can run in a different cmssw version. So it might be more stable and easier. What do you think?

colizz:
Yes I agree, thanks!

PR #2784, opened 2020-12-08
Minor fixes for running gridpack generation on CMS Connect 

SanghyunKo:
Now running gridpack generation on CMS Connect is almost compatible for both `slc6` (by the help of command `cmssw-slc6-condor`) and `slc7` (native), this PR attempts to fix two minor issues:

1. Fix hard-coded OS version in the `condor` submit card, detect either OS or user-specified `scram_arch` automatically
2. Command `git` does not exist occasionally when using singularity container inside `condor` sandbox environment. Try to check whether `git` exists first, and add a backup solution for locating directories when `git` does not exist.

Tested it by generating several simple gridpacks (e.g. `dy01jMLM`) on CMS Connect


agrohsje:
Thanks so much @SanghyunKo.

PR #2787, opened 2020-12-11
samples for topWidth - ST

michael-pitt:
Samples to be used for TopWidth measurement 
Validation of the cards is at [14.12.20@TMG](https://indico.cern.ch/event/985233/) meeting
The samples are single top production in t-channel at LO using 4FS include different top mass and width values
**The samples work with MG version >= 2.7** due to width variation in the reweight_card
Default width = 10.0GeV to allow reweighting to work


sihyunjeon:
image

I ran your cards with some suspicion that it might include several top-unrelated diagrams as well. And actually it did include them. Is this intended? Of course it's not always correct to get rid of all the interference effects but I am curious whether these processes inclusion is something you really wanted.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ST_4f_w_lo/ST_4f_w_lo_run_card.dat
```diff
@@ -0,0 +1,180 @@
...
+# WARNING: Do not use for interference type of computation           *
+#*********************************************************************
+  True	= use_syst ! Enable systematics studies
+#
+  systematics	= systematics_program ! none, systematics [python], SysCalc [depreceted, C++]
+['--mur=0.5,1,2', '--muf=0.5,1,2', '--pdf=errorset'] = systematics_arguments ! see: https://cp3.irmp.ucl.ac.be/projects/madgraph/wiki/Systematics#Systematicspythonmodule
+# Syscalc is deprecated but to see the associate options type'update syscalc'#********************************************************************* 
+#  Additional hidden parameters
+#*********************************************************************
+  ['--mur=0.5,1,2', '--muf=0.5,1,2', '--pdf=errorset']	= systematics_arguments # Choose the argment to pass to the systematics command. like --mur=0.25,1,4. Look at the help of the systematics function for more details.
```
agrohsje, line 180: Please remove this from the card.

michael-pitt: @sihyunjeon: these are W+jets EW, contributions, we had impression that it is not safe to remove them (otherwise we are using `/` syntax which in some cases not safe). In overall they don't affect the kinematics, from tests we did with and w/o these processes the overall result was <0.3%.
@agrohsje I fixed the `run_card.dat`

sihyunjeon:
@michael-pitt I see your point that it would might be not safe to drop out EW Wjets. But looking at one of the samples from RunII Legacy in the past made with POWHEG that seems to be targeting the same process [1] I am not sure what the correct way is to make the samples. The POWHEG-BOX used for [1], "ST_tch" is citing [2] as a reference and based on my quick glimpse, "ST_tch" seems to be removing other processes unrelated to single top. So I won't say what is wrong or correct frankly I don't know, especially given that I have zero experience on POWHEG. But maybe you can give it a thought about this.

@agrohsje Can you comment about the POWHEG as well? Maybe this makes us fall back into the whole "SYNCHRONISATION" issue again (so let's forget about that). But at least I would like to understand if these differences can be allowed (same ST tchannel samples but different processes between POWHEG and MadGraph).

[1] https://cms-pdmv.cern.ch/mcm/requests?prepid=TOP-RunIIFall17wmLHEGS-00026&page=0&shown=127
[2] https://arxiv.org/pdf/0907.4076.pdf

sihyunjeon:
I just realised this line
>> The samples work with MG version >= 2.7 due to width variation in the reweight_card

@agrohsje Should this go into the master branch or mg27x branch?

michael-pitt:
Hello, we observe a large disagreement when comparing to standard Powheg sample in jet multiplicity. We've ran a RIVET routine with standard pythia fragment and looked at the nJet distributions. 
I also tried to use MLM matching as in [W+3jets fragment](https://cms-pdmv.cern.ch/mcm/public/restapi/requests/get_fragment/BTV-RunIIFall17wmLHEGS-00002) with nJet=2, and set `1 = ickkw` in runcard, but the disagreement gets worse.
We will bring it today (14/12, 16:00) to the TopGroupModeling meeting for feedback (https://indico.cern.ch/event/985233/contributions/4149088/attachments/2162254/3648562/topWidth_samples.pdf)
Should we check better the Generator settings, or is it all about Parton shower matching?
compare_njets_norm

sihyunjeon:
This is interesting.
Though I don't think this is a matter of PS matching. Your samples made with MadGraph should not be doing any jet matchings here since there are no additional partons... So perhaps that's why when you did MLM, the differences got worse.
Can you share the Pythia fragment you used for MadGraph and also Powheg (did you also make the Powheg samples) when you made plot?

agrohsje:
Hi @michael-pitt. Can you comment on the fix for 2.7, i.e. this point:  
 The samples work with MG version >= 2.7 due to width variation in the reweight_card
I discussed with Lukas and Kirill and wondering if this is the same. 

Powheg is using NWA for the top,

sihyunjeon:
@agrohsje Do we have any idea on that njet difference though?

michael-pitt:
Here is the pythia code used in processing the Madgraph files here:
https://github.com/WbWbX/Rivet/blob/main/TOP/test/runRivetAnalyzer_wbwbx_cfg.py#L45

for the standard top, I run [RIVET routine](https://github.com/WbWbX/Rivet/blob/main/TOP/src/CMS_WbWb_internal.cc) on mini-aods

Regarding 2.7: running the older version of MG is not working with the reweight card we used.

agrohsje:
Hi @michael-pitt. You run LO production without additional jets. That means there is no matching and no merging involved and you can just add the parton shower. For Powheg you would need to run some matching procedure, a la main31. Now I assume the Powheg sample is a centrally produced sample, which should be done correctly? Or did you produce the Powheg yourself?

michael-pitt:
Hi @agrohsje, the single top is from a centrally produced sample. I added a plot of pT spectra of the leading light/b jet in the event. While the Powheg sample has more physical pt spectra (diverge at low pT) in madgraph, this singularity is missing. Maybe some additional generator cuts present in MG that are missing in Powheg? 
compare_jets_norm

michael-pitt:
Dear @agrohsje  and @sihyunjeon, we have solved the discrepancy between Powheg and MG RIVET plots. In the Rivet cfg the `MergedGenParticleProducer` was missing when Powheg mini-aods were accessed. The agreement is much better now. Now we have discussed a possible implementation of custom scale choice following a pheno paper ([1207.5391](https://arxiv.org/pdf/1207.5391v2.pdf)). The new scale is 4 times the transverse mass of the spectator b-quark. We consulted with the NLO example from [st_tch_4f_ckm_NLO](https://github.com/cms-sw/genproductions/tree/c41ab29f3d86c9e53df8b0d76c12cd519adbf013/bin/MadGraph5_aMCatNLO/cards/production/13TeV/st_tch_4f_ckm_NLO) and implemented the costume scale in the LO card. Unfortunately, we see a rather large effect on the basic kinematics. We computed the transverse mass of `nexternal-1` particle in the scale card, where the generated process is `p p > w+ b b~ j + CC`. btw: we can always stick to the default scale implementation since the MG LO with a default scale gives very similar results to the Powheg NLO sample. 
compare_lj1pt_norm
compare_b1pt_norm

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ST_4f_w_lo/ST_4f_w_lo_setscales.f
```diff
@@ -0,0 +1,204 @@
...
+cc      USER-DEFINED SCALE: ENTER YOUR CODE HERE                                 cc
+cc      to use this code you must set                                            cc
+cc                 dynamical_scale_choice = 0                                    cc
+cc      in the run_card (run_card.dat)                                           cc
+ccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc
+c        transverse mass of spectator b quark (keep it last in the proc card)
+c        m^2+pt^2=p(0)^2-p(3)^2=(p(0)+p(3))*(p(0)-p(3))
+         i=nexternal-1
+         rscale=(P(0,i)+P(3,i))*(P(0,i)-P(3,i))
+         rscale=4d0*dsqrt(max(0d0,rscale))
```
agrohsje, line 92: Not sure I understand the implementation. Is that supposed to be eq 3.1 of the paper?

michael-pitt: Yes, it is supposed to be eq3.1, I followed other examples from the LO card where `m^2+pt^2=p(0)^2-p(3)^2=(p(0)+p(3))*(p(0)-p(3))`, where the mass and the pT are of the  spectator b-quark

agrohsje:
@sihyunjeon ready from my side to be merged.

sihyunjeon:
merged now, this looks okay

michael-pitt:
Hello.  A follow-up on this ticket.
I checked the issue with the custom scale choice. Apparently, if generating both on-shell and off-shell single top production in 4FS we cannot set the scale to be _q=4*sqrt(pt^2+m^2)_ where _pT_ and _m_ are pf the spectator b-quark in the `setscales.f` since in our process it is hard to pick up the spectator b-quark correctly. There is a solution of how to do it: https://answers.launchpad.net/mg5amcnlo/+question/695706, but I think we prefer to keep the default scale choice rather than mess with the fortran code.

PR #2789, opened 2020-12-13
add cards for ZPrimeToZRToZWW process

xdlyu:
We want to produce an additional signal by replacing wkk with zkk. The cards are similar with the previous ones: https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/pre2017/13TeV/Triboson%20search


sihyunjeon:
Hello,
I have one generic comment on the process (just asking for my lesson actually).
So here only decay to W or Z bosons are handled with MadGraph, and probably the rest is handled in Pythia right? Is there a reason not to fully decay bosons to final state objects? As far as I know decaying with Pythia would lose spin correlations (let me refer to MG lauchpad [1]).
Perhaps it's because using Pythia is much faster and spin correlations don't make a big difference to your samples? But still, making full decay chain gridpacks don't seem to be that much time-consuming so I wonder what the reason is.

[1] https://answers.launchpad.net/mg5amcnlo/+question/662752

sihyunjeon:
@agrohsje to me it's good to merge

xdlyu:
> Hello,
> I have one generic comment on the process (just asking for my lesson actually).
> So here only decay to W or Z bosons are handled with MadGraph, and probably the rest is handled in Pythia right? Is there a reason not to fully decay bosons to final state objects? As far as I know decaying with Pythia would lose spin correlations (let me refer to MG lauchpad [1]).
> Perhaps it's because using Pythia is much faster and spin correlations don't make a big difference to your samples? But still, making full decay chain gridpacks don't seem to be that much time-consuming so I wonder what the reason is.
> 
> [1] https://answers.launchpad.net/mg5amcnlo/+question/662752

Hi, we want to have inclusive decay Z for our signal.

sihyunjeon:
@xdlyu that can also be done in MadGraph by writing the decay channels, no?

xdlyu:
> @xdlyu that can also be done in MadGraph by writing the decay channels, no?

Do you mean to use madspin? Actually in our analysis, we do not use variables related with spin correlations. So it is ok to be handled in Pythia.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Triboson_ZPrimeToZRToZWW/Triboson_M_R/Triboson_M_R_customizecards.dat
```diff
@@ -0,0 +1,6 @@
+set param_card kkinputs 5 <MASS>
+set param_card kkinputs 2 6.708000e+00
+set param_card mass 9000025 <MASS_R>
+set param_card mass 9000023 <MASS>
+set param_card DECAY 9000023 1.195600e+01
+set param_card DECAY 9000025 0.31035
```
agrohsje, line 6: Do you consider the width stable across the mass range you are probing? The values seems very specific.

xdlyu: We use different width for different signal point. I upload file "width.txt" which contains widths of zkk and r for different signal points and they are calculated in the way as the author of the model recommended.

agrohsje:
Thanks so much! Can you extend the script so that it correctly fills the values?

xdlyu:
> Thanks so much! Can you extend the script so that it correctly fills the values?

Hi, you can find the file called "gen_card.py".

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Triboson_ZPrimeToZRToZWW/Triboson_M_R/Triboson_M_R_customizecards.dat
```diff
@@ -0,0 +1,6 @@
+set param_card kkinputs 5 <MASS>
+set param_card kkinputs 2 6.708000e+00
```
agrohsje, line 2: Which param is this? kkinputs 2 ? Maybe I misunderstood.

xdlyu: It is one parameter in the coupling.

xdlyu:
@agrohsje Hi, are there still some issues remaining?

PR #2791, opened 2020-12-14
WW card - PDF updated for UL gridpack

mlizzo:
PDF set updated to 325300 in order to use the card for the production of WW gridpack for the UL campaign. Other than that, no changes have been implemented


agrohsje:
Make sure that the 325301,... variations are in pwg-rwl.dat when producing the gridpack. There is currently a bug in 4FS. I think 5FS works but please check when preparing!

mlizzo:
Hi @agrohsje, indeed you're right, among "PDF_variation1/2" there are no 32530x PDFs. How should I do to fix this?

agrohsje:
This one:
https://raw.githubusercontent.com/cms-sw/genproductions/2e887687c55b4a5c18f60957153dd29d6ca96ca7/bin/Powheg/make_rwl.py
should give you all needed PDF's. Do you mind running this and update the pwg-rwl.dat in the gridpack. There is no need to re-run. You can just untar, replace the pwg-rwl.dat and tar again. 
Please let me know if something is wrong or didn't work. I made a PR to fix the issue.

mlizzo:
Done! Thanks for your help, I think it worked properly and the gridpack should now be fixed

Issue #2783, opened 2020-12-04
PDF variation list in Powheg


gvonsem:
In the reweighting script for Powheg gridpacks there seems to be no line for the default recommended PDF set in UL, NNPDF31_nnlo_as_0118_mc_hessian_pdfas (325300): 
https://github.com/cms-sw/genproductions/blob/f46d2a98123e38c4ae88c640bd2752f13be61825/bin/Powheg/make_rwl.py#L102-L132

Also, for the current first line in this list, 'NNPDF31_nnlo_hessian_pdfas' is associated to a wrong LHAPDF ID. Instead of 320900 (NNPDF31_nnlo_as_0118_nf_4) it should be 306000, and should have 103 variations instead of 101. 

Can you confirm these should be modified?

agrohsje:
hi @gvonsem . indeed this is wrong in master. is this affecting current production? or is this taken from the ul branch. ul2019 looks good to me.

agrohsje:
hi @covarell. are you available to double-check and mak a pr to update? otherwise I can also do it. please let me know.

PR #2792, opened 2020-12-16
fixed pdf list for powheg

agrohsje:


agrohsje:
Hi @gvonsem . This should fix the problem you pointed out (https://github.com/cms-sw/genproductions/issues/2783). Note that I generally re-ordered and assigned 4fs and 5fs properly as well as hessian vs replicas. Do you mind checking? 
@covarell can you check my PR? If ok, I will merge.

covarell:
Hello, thanks for taking care of this! Somehow for a long time this list was incorrect, because Jihun's merges always put back the wrong version... 

The pdf names and codes look OK to me now, I wonder if the weight numbers could be reordered a little (e.g. in 5f the main pdf has the code 3000, with 2000 coming later, in 4f this is different. But it is not so important.

agrohsje:
Hi @covarell . Thanks for checking. I didn't want to change the numbers, maybe somebody is using the info, don't know. Do you suggest to re-order the numbers or change?

covarell:
Can we use the same numbering as MadGraph? That would be ideal. Though I don't think people are really using these weight numbers for analysis, since they changed so many times during Run2.

agrohsje:
In MG5 the number just goes through: 
<weight id="1353" MUR="1.0" MUF="1.0" PDF="23796" > PDF=23790 MemberID=6 </weight>

<weight id="1354" MUR="1.0" MUF="1.0" PDF="25410" > PDF=25410 MemberID=0 </weight>

covarell:
OK, mmm, we cannot do that. Then maybe it is not worth changing the order.

PR #2793, opened 2020-12-16
fix of pdfs for ul2019

agrohsje:


agrohsje:
Same here for UL. Will merge if @covarell is ok with https://github.com/cms-sw/genproductions/pull/2792 .

PR #2796, opened 2020-12-21
Adding LH + SM interaction and mixed coupling cards for Wprime to tb 

adeiorio:


bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LH_run_card.dat
```diff
@@ -0,0 +1,269 @@
...
+     1        = lpp1    ! beam 1 type 
+     1        = lpp2    ! beam 2 type
+     6500.0     = ebeam1  ! beam 1 total energy in GeV
+     6500.0     = ebeam2  ! beam 2 total energy in GeV
+# To see polarised beam options: type "update beam_pol"
+#*********************************************************************
+# PDF CHOICE: this automatically fixes also alpha_s and its evol.    *
+#*********************************************************************
+     nn23lo1    = pdlabel     ! PDF set                                     
+     230000    = lhaid     ! if pdlabel=lhapdf, this is the lhapdf number
```
sihyunjeon, line 46: Is there a reason to use NNPDF23 instead of NNPDF31?

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LH_run_card.dat
```diff
@@ -0,0 +1,269 @@
...
+#
+# For display option for energy cut in the partonic center of mass frame type 'update ecut'
+#
+#*********************************************************************
+# Maximum and minimum absolute rapidity (for max, -1 means no cut)   *
+#*********************************************************************
+  5.0 = etaj    ! max rap for the jets 
+  3.0  = etab    ! max rap for the b
+ 2.5  = etaa    ! max rap for the photons 
+ 2.5  = etal    ! max rap for the charged leptons
```
sihyunjeon, line 132: Can we loosen this cut a little bit?

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LH_run_card.dat
```diff
@@ -0,0 +1,269 @@
...
+ -1.0  = mmnlmax ! max invariant mass for all letpons (l+- and vl) 
+#*********************************************************************
+# Minimum and maximum pt for 4-momenta sum of leptons                *
+#*********************************************************************
+ 0.0   = ptllmin  ! Minimum pt for 4-momenta sum of leptons(l and vl)
+ -1.0  = ptllmax  ! Maximum pt for 4-momenta sum of leptons(l and vl)
+#*********************************************************************
+# Inclusive cuts                                                     *
+#*********************************************************************
+ 250.0  = ptheavy   ! minimum pt for at least one heavy final state
```
sihyunjeon, line 191: What is the purpose of this cut? Are you trying to look for boosted tops?

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LR_proc_card.dat
```diff
@@ -0,0 +1,16 @@
+import model sm
+define p = g u c d s u~ c~ d~ s~
+define j = g u c d s u~ c~ d~ s~
+define l+ = e+ mu+
+define l- = e- mu-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
```
sihyunjeon, line 7: Please clean up the unneeded lines

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LH_proc_card.dat
```diff
@@ -0,0 +1,19 @@
+import model sm
+define p = g u c d s u~ c~ d~ s~
+define j = g u c d s u~ c~ d~ s~
+define l+ = e+ mu+
+define l- = e- mu-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+import model WEff_UFO
```
sihyunjeon, line 8: Please clean up the lines that are not needed

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LH_customizecards.dat
```diff
@@ -0,0 +1,5 @@
+set MADSPIN=ON
```
sihyunjeon, line 1: Doesn't have to turn on madspin yourself

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LH_run_card.dat
adeiorio: We want to suppress the SM production in favour of  the BSM one

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LH_customizecards.dat
sihyunjeon: But I just realised this, where is your madspin card?

adeiorio: It is the same already present in the folder. It was approved in our previous commit for the RH production.

adeiorio: By the way you can find it [here](https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_madspin_card.dat)

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LH_run_card.dat
adeiorio: We already use this set for the RH production, already approved. We need an LO set and we already checked that this not cause any serious issue.
[Here](https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_run_card.dat) the link to the run card for the RH production.

sihyunjeon: lhapdf    = pdlabel     ! PDF set                                     
     315200    = lhaid     ! if pdlabel=lhapdf, this is the lhapdf number
     $DEFAULT_PDF_SETS = lhaid
     $DEFAULT_PDF_MEMBERS  = reweight_PDF     ! reweight to get PDF uncertainty

Actually this card is not quite right. Guess 315200 = lhaid would be overwritten by the line below,  $DEFAULT_PDF_SETS = lhaid. But anyways unlike what you said you did not use the same set for RH production based on the cards you linked. lhaid 3XXXXX is NNPDF31, not NNPDF23

sihyunjeon: What do you mean by "this set"? For the samples you already produced in the past used different PDF set and you ONLY want to change the PDF to LO? 
And you'd be using CP1-5 tunes for pythia showerings right? CP1-5 tunes is made to complement MC-data disagreements when using NNPDF31. So I think NNPDF23 will not be a reasonable choice.

sihyunjeon: So you had RH production in the past, and now you want LH production, is this correct?
But would it not be better to synchronize the PDF settings with RH then? The PDF set used for [1] is clearly different, it's not NNPDF23

[1] https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_run_card.dat#L47-L50

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LH_customizecards.dat
sihyunjeon: @adeiorio Can you add full set of your cards that will go into gridpacks as an input?

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LH_run_card.dat
adeiorio: You are right indeed. The PDF used in the past was the 315200. In the previous cards the line 49 has been commented out when producing the GPs, but it was never updated here. I will update this card

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LH_customizecards.dat
adeiorio: I have now pushed the madspin and extramodels cards, so that everything we use is here.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LH_run_card.dat
```diff
@@ -0,0 +1,269 @@
...
+  average =  event_norm       ! average/sum. Normalization of the weight in the LHEF
+#*********************************************************************
+# https://twiki.cern.ch/twiki/bin/view/CMS/Moriond18MC#Checks_to_the_datacards_BEFORE_T 
+#*********************************************************************
+ 1        = ktscheme         ! for ickkw=1, 1 Durham kT, 2 Pythia pTE
+ F        = pdfwgt           ! for ickkw=1, perform pdf reweighting
+#*********************************************************************
+# Matching parameter (MLM only)
+#*********************************************************************
+ 1 = ickkw            ! 0 no matching, 1 MLM
```
sihyunjeon, line 77: You are not doing any jet matching, but why is ickkw set to 1? there's no additional parton for your process

generate p p > wp+ > t b~
add process p p > wp- > t~ b
add process p p > t b
add process p p > t~ b~
add process p p > t b~ 
add process p p > t~ b

sihyunjeon:
I have now a different question. Your old cards [1] are doing MLM jet matching with multipartons (you see the additional js in your process). Isn't it also better to synchronize additional jets for your new samples with W interference as well? Anyways if you don't want multiparton samples, ickkw should be 0 and the jet matching related parameters should be turned off.

[1] https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_proc_card.dat#L11-L16

adeiorio:
Here we don't consider additional jets because these open the t-channel production that we don't want in the production

> I have now a different question. Your old cards [1] are doing MLM jet matching with multipartons (you see the additional js in your process). Isn't it also better to synchronize additional jets for your new samples with W interference as well? Anyways if you don't want multiparton samples, ickkw should be 0 and the jet matching related parameters should be turned off.
> 
> [1] https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_proc_card.dat#L11-L16

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LH_run_card.dat
sihyunjeon: 1 = ickkw            ! 0 no matching, 1 MLM
This should be 0

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LH_run_card.dat
```diff
@@ -0,0 +1,269 @@
...
+ 1.0 = alpsfact         ! scale factor for QCD emission vx
+ False = chcluster        ! cluster only according to channel diag
+ 5 = asrwgtflavor     ! highest quark flavor for a_s reweight
+ True  = auto_ptj_mjj  ! Automatic setting of ptj and mjj if xqcut >0
```
sihyunjeon, line 81: This should be False

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LH_run_card.dat
```diff
@@ -0,0 +1,269 @@
...
+                                   ! (turn off for VBF and single top processes) 
+ 30.0   = xqcut   ! minimum kt jet measure between partons
```
sihyunjeon, line 83: This should be 0

sihyunjeon:
I commented on MLM matching parameters that should be turned off if you don't additional partons.

sihyunjeon:
Thanks for the quick fix. I don't see any obvious issues now.
@agrohsje might comment more on the PDF sets or some other things.

ph2632:
Hello, is there an obstacle for not merging these cards?
It would be good to know as we wanted to make the requests asap. Thanks

sihyunjeon:
@ph2632 We need Alexander to sign off after looking at the cards but he might be away from work for a while due to the break.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LR_proc_card.dat
agrohsje, line 7: Please remove.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LR_proc_card.dat
```diff
@@ -0,0 +1,16 @@
...
+import model WEff_UFO
+define p = p b b~
+define j = j b b~
+generate p p > wp+ > t b~
+add process p p > wp- > t~ b
+add process p p > t b
+add process p p > t~ b~
+add process p p > t b~ 
+add process p p > t~ b
```
agrohsje, line 16: What about the output line?

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LR_run_card.dat
```diff
@@ -0,0 +1,269 @@
...
+#*********************************************************************
+# Tag name for the run (one word)                                    *
+#*********************************************************************
+  tag_1     = run_tag ! name of the run 
+#*********************************************************************
+# Number of events and rnd seed                                      *
+# Warning: Do not generate more than 1M events in a single run       *
+# If you want to run Pythia, avoid more than 50k events in a run.    *
+#*********************************************************************
+  20000 = nevents ! Number of unweighted events requested
```
agrohsje, line 28: 5 k are enough for the pilot

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LR_run_card.dat
```diff
@@ -0,0 +1,269 @@
...
+#
+#*********************************************************************
+# Store info for systematics studies                                 *
+# WARNING: Do not use for interference type of computation           *
+#*********************************************************************
+   T  = use_syst      ! Enable systematics studies
+#
+systematics = systematics_program ! none, systematics [python], SysCalc [depreceted, C++]
+['--mur=0.5,1,2', '--muf=0.5,1,2', '--pdf=errorset'] = systematics_arguments ! see: https://cp3.irmp.ucl.ac.be/projects/madgraph/wiki/Systematics#Systematicspythonmodule
+# Syscalc is deprecated but to see the associate options type'update syscalc'
```
agrohsje, line 269: Please remove. Is sometimes causing problems.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LH_proc_card.dat
```diff
@@ -0,0 +1,15 @@
+import model sm
+define p = g u c d s u~ c~ d~ s~
+define j = g u c d s u~ c~ d~ s~
```
agrohsje, line 3: Please remove.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LH_run_card.dat
```diff
@@ -0,0 +1,269 @@
...
+#*********************************************************************
+# Tag name for the run (one word)                                    *
+#*********************************************************************
+  tag_1     = run_tag ! name of the run 
+#*********************************************************************
+# Number of events and rnd seed                                      *
+# Warning: Do not generate more than 1M events in a single run       *
+# If you want to run Pythia, avoid more than 50k events in a run.    *
+#*********************************************************************
+  20000 = nevents ! Number of unweighted events requested
```
agrohsje, line 28: 5 k is sufficient for pilot.

agrohsje:
Sorry. Taking a few days off. Added some comments. Please fix.

ph2632:
Thank you for the input @agrohsje and for helping us in the middle of the break.
@adeiorio has made the GPs and me the request with this setup.
If I'm not mistaken, there are some redundant lines in the cards but the physics is fine, correct? So we can stick to the GPs already produced, correct?

agrohsje:
I assume the output line is there, otherwise it will not work? So physics wise this looks ok to me.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LR_run_card.dat
adeiorio: When producing the RH samples we don't have these lines and they were produced without the systematic variations. Was the problem caused by the absence of these lines?

adeiorio:
> I assume the output line is there, otherwise it will not work? So physics wise this looks ok to me.

Yes the output line was there in the GP production, but maybe I missed to copy it when uploading the cards.

ph2632:
Here is how a proc cards looks untarring a GP:

import model sm
import model WEff_UFO
define p = p b b~
define j = j b b~
generate p p > wp+ > t b~
add process p p > wp- > t~ b
add process p p > t b
add process p p > t~ b~
add process p p > t b~
add process p p > t~ b
output Wprimetotb_M4000W40_LH -nojpeg
output Wprimetotb_M2000W200_LH -nojpeg

sihyunjeon:
> Here is how a proc cards looks untarring a GP:
> 
> import model sm
> define p = g u c d s u~ c~ d~ s~
> define j = g u c d s u~ c~ d~ s~
> import model WEff_UFO
> define p = p b b~
> define j = j b b~
> generate p p > wp+ > t b~
> add process p p > wp- > t~ b
> add process p p > t b
> add process p p > t~ b~
> add process p p > t b~
> add process p p > t~ b
> output Wprimetotb_M4000W40_LH -nojpeg
> output Wprimetotb_M2000W200_LH -nojpeg

Does one gridpack contain two output lines?

adeiorio:
> > Here is how a proc cards looks untarring a GP:
> > import model sm
> > define p = g u c d s u~ c~ d~ s~
> > define j = g u c d s u~ c~ d~ s~
> > import model WEff_UFO
> > define p = p b b~
> > define j = j b b~
> > generate p p > wp+ > t b~
> > add process p p > wp- > t~ b
> > add process p p > t b
> > add process p p > t~ b~
> > add process p p > t b~
> > add process p p > t~ b
> > output Wprimetotb_M4000W40_LH -nojpeg
> > output Wprimetotb_M2000W200_LH -nojpeg
> 
> Does one gridpack contain two output lines?

This happen only in the LH but the last line is always the correct one. Does this cause any harm?

sihyunjeon:
I guess it should be just something redundant, but no harm

ph2632:
Happy new year all!
Is this thread closed=Merged?
All relevant requests are successfully validated and already on the gdoc.
About half of the cards do have a double output like above, but always the correct one as the latter.

ph2632:
Let me follow up here to let analyzers interfere.
GEN (@agrohsje) commented on gdoc
"Just to be sure this is supposed to be pure W’ as well as W/W’ interference? The syntax gives you interferences because NP vertex has weight 1? "
Your notation "W/W’" might signifies a signal twice accounted, also not sure where you refer for "vertex weight 1".
To my understanding is: 
W'-->tb & 
(SM) W->tb & 
(SM)scattering bt, b~t~
Perhaps analyzers (@adeiorio) can clarify?

adeiorio:
> Let me follow up here to let analyzers interfere.
> GEN (@agrohsje) commented on gdoc
> "Just to be sure this is supposed to be pure W’ as well as W/W’ interference? The syntax gives you interferences because NP vertex has weight 1? "
> Your notation "W/W’" might signifies a signal twice accounted, also not sure where you refer for "vertex weight 1".
> To my understanding is:
> W'-->tb &
> (SM) W->tb &
> (SM)scattering bt, b~t~
> Perhaps analyzers (@adeiorio) can clarify?

Hi all. If I understood correctly your question is about a possible double presence of the BSM process in the generation. To this question I can reply that without explicitly including the line W'-->tb the process is not taken into account.

agrohsje:
Let's merge the threads and indeed continue here as suggested by @ph2632 .
In the MC request, I was reading something about interference. Can you clarify what you have in mind to generate with these lines: 
generate p p > wp+ > t b~
add process p p > wp- > t~ b
add process p p > t b
add process p p > t~ b~
add process p p > t b~ 
add process p p > t~ b

adeiorio:
> Let me follow up here to let analyzers interfere.
> GEN (@agrohsje) commented on gdoc
> "Just to be sure this is supposed to be pure W’ as well as W/W’ interference? The syntax gives you interferences because NP vertex has weight 1? "
> Your notation "W/W’" might signifies a signal twice accounted, also not sure where you refer for "vertex weight 1".
> To my understanding is:
> W'-->tb &
> (SM) W->tb &
> (SM)scattering bt, b~t~
> Perhaps analyzers (@adeiorio) can clarify?

Hi all. If I understood correctly your question is about a possible double presence of the BSM process in the generation. To this question I can reply that without explicitly including the line W'-->tb the process is not taken into account. 

> Let's merge the threads and indeed continue here as suggested by @ph2632 .
> In the MC request, I was reading something about interference. Can you clarify what you have in mind to generate with these lines:
> generate p p > wp+ > t b~
> add process p p > wp- > t~ b
> add process p p > t b
> add process p p > t~ b~
> add process p p > t b~
> add process p p > t~ b

Here we are generating the s-channel BSM production, s-channel SM production and their interference.  Lines add process p p > t~ b~ and add process p p > t b should be unnecessary.

agrohsje:
Wait. This is the BSM resonance part: 
generate p p > wp+ > t b~
add process p p > wp- > t~ b
This is t-channel:
add process p p > t b
add process p p > t~ b~
and this s-channel:
add process p p > t b~
add process p p > t~ b

Are you sure interference is included in the last 4 processes? I think this will only give you the SM W, but not the output of the W' amplitude \times  W amplitude. When doing NP^2==2 (which should give me 2 NP vertices at the ME level) I get a super small cross section, while pure generate p p > t b~ give some 5 pb cross section, which looks like the SM value, no?

adeiorio:
Hi @agrohsje I checked a bit by playing with the NP parameter. I tried:
generate process p p > t b~ NP^2<=2 
add process p p > t~ b NP^2<=2 
and these lines produce 2 Feynman diagrams for each one of the 8 subprocesses. Maybe this was the easiest way to produce the samples. Anyway this is identical to what you get by  
generate p p > wp+ > t b~
add process p p > wp- > t~ b
add process p p > t b~
add process p p > t~ b
The t-channel 
add process p p > t b
add process p p > t~ b~
should not contribute at this level since no extra jet is present at ME level. Am I wrong?
I also produce a plot with the tb mass system by vetoing the t and b coming from a W' (at LHE level) and it seems that the interference is correctly taken into account.
image_2020_12_16T08_07_35_254Z

adeiorio:
Hi @agrohsje, after the study we did, I updated the cards. Now They should be ok and the effects of the interference are now taken into account.

adeiorio:
Hi @agrohsje, interacting with the MadGraph support  (thread [[here][https://answers.launchpad.net/mg5amcnlo/+question/695301]) we updated the cards. Now the process is correctly decayed with MadSpin and the interference is correctly taken into account.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LR_customizecards.dat
```diff
@@ -0,0 +1,5 @@
+set MADSPIN=ON
```
agrohsje, line 1: That should be correctly handled by our scripts. Did you had problems when not having this line?

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LR_run_card.dat
```diff
@@ -0,0 +1,269 @@
...
+# (note that etmiss/ptll/ptheavy/ht/sorted cuts always apply)
+#*********************************************************************
+   False  = cut_decays    ! Cut decay products 
+#*********************************************************************
+# Standard Cuts                                                      *
+#*********************************************************************
+# Minimum and maximum pt's (for max, -1 means no cut)                *
+#*********************************************************************
+ 20.0  = ptj       ! minimum pt for the jets 
+ 100.0  = ptb       ! minimum pt for the b
```
agrohsje, line 112: what is your jet pT cut in the end ?

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LR_run_card.dat
```diff
@@ -0,0 +1,269 @@
...
+  0.0  = el     ! minimum E for the charged leptons 
+  -1.0   = ejmax ! maximum E for the jets
+ -1.0   = ebmax ! maximum E for the b
+ -1.0   = eamax ! maximum E for the photons
+ -1.0   = elmax ! maximum E for the charged leptons
+#*********************************************************************
+# Maximum and minimum absolute rapidity (for max, -1 means no cut)   *
+#*********************************************************************
+  5.0 = etaj    ! max rap for the jets 
+  3.0  = etab    ! max rap for the b
```
agrohsje, line 136: Same here. Taking cone radius etc into account: Is this loose enough?

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LR_run_card.dat
agrohsje: No problem when not present.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LR_run_card.dat
adeiorio: these cuts are to reduce the number of SM events

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LR_run_card.dat
adeiorio: Yes because we select jets in |eta| < 2.4

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LR_run_card.dat
adeiorio: these cuts are to reduce the number of SM events as also the ptheavy. At the end we select only jets with pt > 100, leadind jet pt >350 and subleading jet pt > 150

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LR_customizecards.dat
adeiorio: I don't think so, but we already produced GP with this line in

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LR_run_card.dat
agrohsje: wait. in this case 100 gev is too tight. i mean you can easily move a 90 gev parton to a 100 gev jet.

adeiorio: For signal we have only two jets... the lowest one has pt >150 so this should be good

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LR_customizecards.dat
agrohsje: do you mind removing from the pr. just not to cause confusion. olivier added this because he doesn't know about our scripts.

agrohsje:
So these gridpacks are now to get SM + interference + BSM. All together?

adeiorio:
> So these gridpacks are now to get SM + interference + BSM. All together?

Yes they are all together

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Wprimetotb/Wprimetotb_LR_customizecards.dat
adeiorio: Done

agrohsje:
Hi @sihyunjeon . Do you mind a final check from your side and merge if ok?

sihyunjeon:
@adeiorio
Let me see if I get this correctly.

So you are making |W+W'| interference effect existing "p p > t b'" sample right, and you mostly want to get the contributions from W' thus applying cuts on both ptj and ptb, or just ptb?
1. Maybe you already know this, ptj won't do anything here since cut_decays is set to False.
2. According to your comment, it seems that the second (b)jet that comes from decay also require high pt cuts (>150) in the RECO level. Then wouldn't it be better to turn on the cut_decays and thus apply the bjet ptcut?
3. What is the final state object you are considering in the analysis? bjet+(top tagged AK8jet) or 2bjets+(other top decays)?

adeiorio:
> @adeiorio
> Let me see if I get this correctly.
> 
> So you are making |W+W'| interference effect existing "p p > t b'" sample right, and you mostly want to get the contributions from W' thus applying cuts on both ptj and ptb, or just ptb?
> 
>     1. Maybe you already know this, ptj won't do anything here since cut_decays is set to False.
> 
>     2. According to your comment, it seems that the second (b)jet that comes from decay also require high pt cuts (>150) in the RECO level. Then wouldn't it be better to turn on the cut_decays and thus apply the bjet ptcut?
> 
>     3. What is the final state object you are considering in the analysis? bjet+(top tagged AK8jet) or 2bjets+(other top decays)?

Hi @sihyunjeon, yes you are right. The ptb cut is not applied but the ptheavy is. This is enough to significantly reduce the contribution from SM. For the selection of the analysis we require one lepton (pt>50-55), ptmiss (pt>120), leading jet pt >350 ande subleading jet pt >150. The top is reconstructed with AK4 jets. We use categories with both b-tagged and non-b-tagged jets, as signal and control regions. We do not use AK8 top tagging but we require that the AK4 uses for the top reconstruction is reconstructed inside an AK8 jet as well the one used to reconstruct the W'

sihyunjeon:
@adeiorio
Hello, The ptb cut WOULD BE APPLIED to the bjet that comes out from W'. (p p > t **b**). But the ptb cut WOULD NOT BE APPLIED to the bjet that comes out from top. (t > **b** w, w > all all) since cut_decays is set to false. Are you aware of this? Based on your comment _The ptb cut is not applied but the ptheavy is_ I am not sure if you are. Can you make sure of this?

And if you have one lepton requirement, why don't you decay the W to leptons in MadSpin instead of making it inclusive? Do you plan to apply filters during sample production? Though changing this would still be much more efficient given that the W > j j BR is higher.

So you use for top reconstruction "lepton + MET + 1 bjet that comes from top", right? You don't consider the hadronic top decay final state ("2 jets + 1 bjet). Again, I am not sure of the reason you are making W > j j decays.

adeiorio:
> @adeiorio
> Hello, The ptb cut WOULD BE APPLIED to the bjet that comes out from W'. (p p > t **b**). But the ptb cut WOULD NOT BE APPLIED to the bjet that comes out from top. (t > **b** w, w > all all) since cut_decays is set to false. Are you aware of this? Based on your comment _The ptb cut is not applied but the ptheavy is_ I am not sure if you are. Can you make sure of this?
> 
> And if you have one lepton requirement, why don't you decay the W to leptons in MadSpin instead of making it inclusive? Do you plan to apply filters during sample production? Though changing this would still be much more efficient given that the W > j j BR is higher.
> 
> So you use for top reconstruction "lepton + MET + 1 bjet that comes from top", right? You don't consider the hadronic top decay final state ("2 jets + 1 bjet). Again, I am not sure of the reason you are making W > j j decays.

Yes ptheavy cut is applied even if cut_decays is set to False. 
We decays the W also in quarks since we should also perform the combination between leptonic and hadronic channels.

sihyunjeon:
Hey sorry, I wasn't pointing out ptheavy, I mean **ptb cut** is going to be applied to the first bjet (from W') but not to the second bjet (from top). My question was whether this asymmetric cut was intended or not.

adeiorio:
Yes exactly. We expect the b-jet coming from the W' to be the leading between the two b-jets in the event

sihyunjeon:
Thanks for the answers, and sorry for not following up earlier. merged

PR #2797, opened 2020-12-24
Adding Cards for pair Produced RPV Stops decaying to jj

gabhijith:


sihyunjeon:
Hello,
Can you make a pull request with template cards, card generating script, and set of cards for one grid point? Like as an example in [1].
[1] https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/WRtoNLtoLLJJ

PR #2808, opened 2021-01-16
patch to run h7 for nlo fxfx with mg 265

agrohsje:
only for 265, not needed in 27x


sihyunjeon:
Should this be merged after the test?
I was going to try out MG265+H7, I can test it if you want

agrohsje:
One should test it with FxFx. If you are available, that would be nice!

sihyunjeon:
Okay just started making the gridpacks with patches.
Tried generating MG265 with HERWIGPP for sanity check and indeed it fails.

Will let you know once I have results.

sihyunjeon:
Okay this is good to go in.

Checked the gridpacks are produced & LHE files being made with it

PR #2812, opened 2021-01-20
Madgraph cards for XToJPsiG_JPsiToMuMu

RoccoA97:
Cards for the processes
- H/Z -> JPsi Gamma, JPsi -> Mu Mu
- H/Z -> PsiPrime Gamma, PsiPrime -> Mu Mu

For the H decay case, production modes considered:
- ggH, VBF, WH, ZH, ttH, bbH

For the Z decay case, only one production mode (from qq).

The cards for a single process are in a dedicated folder (total: 14)


sihyunjeon:
Hi there are so many cards to go through...
Can you first start with one of the H production modes and pull request the others one by one?
There are several trivial mistakes already I see and I think it will be better for us to go through one by one, instead of picking out repeated mistakes for every H production modes.

agrohsje:
I agree with @sihyunjeon . Can you base this on templates and replace relevant settings, so that we only have to review the minimum and are sure everything is consistent and ok.

RoccoA97:
Hi! I have remove some cases. I have left only the basic ones, namely the ggH mode for the Higgs and the qqZ for the Z. I hope it is fine for now and that I have not done mistakes...
Thank you!

bin/MadGraph5_aMCatNLO/cards/production/13TeV/XToJPsiG_JPsiToMuMu/GluGluToH_HToJPsiG_JPsiToMuMu/GluGluToH_HToJPsiG_JPsiToMuMu_proc_card.dat
```diff
@@ -0,0 +1,40 @@
...
+set gauge unitary
+set complex_mass_scheme False
+set max_npoint_for_channel 0
+import model sm
+define p = g u c d s u~ c~ d~ s~
+define j = g u c d s u~ c~ d~ s~
+define l+ = e+ mu+
+define l- = e- mu-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
```
sihyunjeon, line 37: Please clean up the unused lines

import model sm
define l+ = e+ mu+
define l- = e- mu-
define vl = ve vm vt
define vl~ = ve~ vm~ vt~

bin/MadGraph5_aMCatNLO/cards/production/13TeV/XToJPsiG_JPsiToMuMu/GluGluToH_HToJPsiG_JPsiToMuMu/GluGluToH_HToJPsiG_JPsiToMuMu_run_card.dat
```diff
@@ -0,0 +1,281 @@
...
+     1        = lpp1    ! beam 1 type 
+     1        = lpp2    ! beam 2 type
+     6500.0     = ebeam1  ! beam 1 total energy in GeV
+     6500.0     = ebeam2  ! beam 2 total energy in GeV
+# To see polarised beam options: type "update beam_pol"
+#*********************************************************************
+# PDF CHOICE: this automatically fixes also alpha_s and its evol.    *
+#*********************************************************************
+     nn23lo1    = pdlabel     ! PDF set                                     
+     230000    = lhaid     ! if pdlabel=lhapdf, this is the lhapdf number
```
sihyunjeon, line 47: Are these PDF settings what you want to use for your signal sample generation?

bin/MadGraph5_aMCatNLO/cards/production/13TeV/XToJPsiG_JPsiToMuMu/ZToJPsiG_JPsiToMuMu/ZToJPsiG_JPsiToMuMu_proc_card.dat
```diff
@@ -0,0 +1,40 @@
...
+set gauge unitary
+set complex_mass_scheme False
+set max_npoint_for_channel 0
+import model sm
+define p = g u c d s u~ c~ d~ s~
+define j = g u c d s u~ c~ d~ s~
+define l+ = e+ mu+
+define l- = e- mu-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
```
sihyunjeon, line 37: Here as well,
please clean up the unused lines

bin/MadGraph5_aMCatNLO/cards/production/13TeV/XToJPsiG_JPsiToMuMu/ZToJPsiG_JPsiToMuMu/ZToJPsiG_JPsiToMuMu_run_card.dat
```diff
@@ -0,0 +1,281 @@
...
+     1        = lpp1    ! beam 1 type 
+     1        = lpp2    ! beam 2 type
+     6500.0     = ebeam1  ! beam 1 total energy in GeV
+     6500.0     = ebeam2  ! beam 2 total energy in GeV
+# To see polarised beam options: type "update beam_pol"
+#*********************************************************************
+# PDF CHOICE: this automatically fixes also alpha_s and its evol.    *
+#*********************************************************************
+     nn23lo1    = pdlabel     ! PDF set                                     
+     230000    = lhaid     ! if pdlabel=lhapdf, this is the lhapdf number
```
sihyunjeon, line 47: Same as above,
Is this the PDFset you wish to use for signal sample?

sihyunjeon:
Overall,
Are your kinematic cuts (drll for instance) set with reasons?
Your current kinematic cuts will probably remove all the boosted jpsi events because of drll.

RoccoA97:
Hi, thank you again for the various points. We have:
* removed from proc_cards the unused lines
* updated the run_cards with the correct pdf sets (it should be the default NNPDF30_nlo_nf_5_pdfas)
* removed (now they are commented) almost all the kinematics cuts from the run_cards, since the signal efficiency at generator level is high, so we though they were useless for our study

sihyunjeon:
@RoccoA97 Hi, you should not "comment out" the cuts with #. It will apply the default cuts. Can you remove the cuts by writing the values (drll = 0.4 -> drll =0.0).

sihyunjeon:
Just to make sure with the PDF set, Is this not going to be used for Ultra Legacy analysis?

RoccoA97:
Sorry, my fault. I have explicitely disabled the kinematics cuts and we have also written in more clearly the run cards. We are going to use them for Ultra Legacy analysis, we though that the pdfs were the same for the Legacy version.
Now we have added some shortcuts to have the PDF sets automatically added to the run_card at run time to avoid specifying them directly.

sihyunjeon:
Thanks for fixing things.
              lhapdf = pdlabel      ! PDF set
   $DEFAULT_PDF_SETS = lhaid
Should give you the correct PDFsets for UL.

One more thing. Is your process "p p > X, (X > jpsi a)" safe with pta = 0 cut?
I am just worried this pta = 0 will cause a divergence from lowpt gamma radiations and lead to generation failure.

RoccoA97:
Thank you for the tips!
Concerning the cut on pta, we have tried to generate a test dataset of about ~100k events directly from the gridpack with these cards. From the `.lhe` file results, the pt and angular distributions for the final products (namely, the two muons and the photon) do not exhibit strange behaviours.
Moreover, in our case we would avoid a cut on `pta` since it would change the efficiency.
However, with this test it seems that these cuts are not dangerous only at Madgraph level.

sihyunjeon:
@agrohsje For me the cards are fine

RoccoA97:
Hi @sihyunjeon! If it is ok also for @agrohsje, I would commit the other production modes for the H and then the version of the processes with the Psi(2S) in place of the J/Psi. The run_cards would be the same of the ones already committed.
Moreover, do I need to commit the other sets of cards in this pull request or is it better to open another pull request?

RoccoA97:
Hi @sihyunjeon and @agrohsje! I have committed the other version with Psi(2S) (in place of the J/Psi) of the two already committed processes (gg>H and qq>Z productions). The run card is again the same as for the previous processes. The only variations are in the customizecards, where we change the mass and the width of the JPsi to the values for Psi(2S)

sihyunjeon:
@RoccoA97 It seems that I missed one question.
How is your proton defined in the process card?
Does it include b or not?

RoccoA97:
No, it doesn't include `b`. It includes `g u c d s u~ c~ d~ s~` when specified in the process card

agrohsje:
@sihyunjeon I just removed the systematics lines as discussed. are you ok with merging?

sihyunjeon:
Yes this looks fine

PR #2813, opened 2021-01-21
feat: add xs cal collect script

andrzejnovak:
Short script that collects the outputs of `./calculateXSectionAndFilterEfficiency.sh -f datasets.txt `  into one file.


agrohsje:
Thanks @andrzejnovak !

PR #2820, opened 2021-02-03
Cards for ptLL binned ultra-low mass (1 < mll < 10) DY sample

jingyucms:
Cards created for DY sample with 1 GeV < mll < 10 GeV. Both inclusive and pT ll binned cards are created. Please let me know if you have any comments @sihyunjeon @agrohsje

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/DYJetsToLL_M-1to10/DYJetsToLL_M-1to10_PtLL-1200toInf_13TeV-madgraphMLM-pythia8/DYJetsToLL_M-1to10_PtLL-1200toInf_13TeV-madgraphMLM-pythia8_run_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/DYJetsToLL_M-1to10/DYJetsToLL_M-1to10_PtLL-1200toInf_13TeV-madgraphMLM-pythia8/DYJetsToLL_M-1to10_PtLL-1200toInf_13TeV-madgraphMLM-pythia8_run_card.dat
@@ -1,0 +1,280 @@
+
+#*********************************************************************
+#                       MadGraph5_aMC@NLO                            *
+#                                                                    *
+#                     run_card.dat MadEvent                          *
+#                                                                    *
+#  This file is used to set the parameters of the run.               *
+#                                                                    *
+#  Some notation/conventions:                                        *
+#                                                                    *
+#   Lines starting with a '# ' are info or comments                  *
+#                                                                    *
+#   mind the format:   value    = variable     ! comment             *
+#*********************************************************************
+#
+#*******************                                                 
+# Running parameters
+#*******************                                                 
+#                                                                    
+#*********************************************************************
+# Tag name for the run (one word)                                    *
+#*********************************************************************
+  tag_1	= run_tag ! name of the run 
+#*********************************************************************
+# Run to generate the grid pack                                      *
+#*********************************************************************
+  True	= gridpack !True = setting up the grid pack
+#*********************************************************************
+# Number of events and rnd seed                                      *
+# Warning: Do not generate more than 1M events in a single run       *
+# If you want to run Pythia, avoid more than 50k events in a run.    *
+#*********************************************************************
+  1000	= nevents ! Number of unweighted events requested 
+  0	= iseed ! rnd seed (0=assigned automatically=default))
+#*********************************************************************
+# Collider type and energy                                           *
+# lpp: 0=No PDF, 1=proton, -1=antiproton, 2=photon from proton,      *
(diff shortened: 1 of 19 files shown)


agrohsje:
Why don't you commit a template for the run_card.dat where the pt bins get filled by a script and then add the proc_card. This makes 3 simple files that can be easily reviewed and compared. 
Where did you get your run_card from? Did you compare to existing UL requests for the 10-50 and 50 to infinity? Are your specifications with what is already in production in sync?

jingyucms:
I do use a python script to loop over pt_ll bins to produce these cards. Shall I upload the script instead?
The template of the run_card comes from https://github.com/cms-sw/genproductions/blob/master/bin/GenValidation/updated_validation_cards/dy_mlm_lowmass/inc/dyellell01234j_mll10to50_5f_LO_MLM_run_card.dat
so it should be comparable to UL requests for 10-50 and 50 to infinity.

agrohsje:
Yes. Please upload the script and the template. This is best to review. Thanks!

bin/MadGraph5_aMCatNLO/cards/DYJetsToLL_M-1to10/makeDatacards.py
```diff
@@ -0,0 +1,325 @@
...
+  True	= use_syst ! Enable systematics studies
+#
+#********************************************************************* 
+#  Additional parameter
+#*********************************************************************
+  0.0	= ptonium # hidden parameter
+  -1.0	= etaonium # hidden parameter
+    """)
+    
+
```
agrohsje, line 325: can you remove this. my point here is about parsing info. e.g. \alphas will cause problems. not sure about the 3 " and the ) .  same for the cards you generated from the template. if updated, this is good to go from my side. @sihyunjeon  let me know what you think.

sihyunjeon: This can be some common background sample for many other analysis so I think it should  be fixed if troublesome for NanoAODs

bin/MadGraph5_aMCatNLO/cards/DYJetsToLL_M-1to10/DYJetsToLL_M-1to10_PtLL-1200toInf_13TeV-madgraphMLM-pythia8/DYJetsToLL_M-1to10_PtLL-1200toInf_13TeV-madgraphMLM-pythia8_run_card.dat
```diff
@@ -0,0 +1,280 @@
...
+#*********************************************************************
+# Minimum and maximum invariant mass for all letpons                 *
+#*********************************************************************
+  0.0	= mmnl ! min invariant mass for all letpons (l+- and vl) 
+  -1.0	= mmnlmax ! max invariant mass for all letpons (l+- and vl) 
+#*********************************************************************
+# Minimum and maximum pt for 4-momenta sum of leptons                *
+#*********************************************************************
+  1200	= ptllmin ! Minimum pt for 4-momenta sum of leptons(l and vl)
+  Inf	= ptllmax ! Maximum pt for 4-momenta sum of leptons(l and vl)
```
sihyunjeon, line 192: Does "Inf" work in Madgraph? Instead of -1?

jingyucms: It doesn't. Thanks for the catch. Working on the fix...

bin/MadGraph5_aMCatNLO/cards/DYJetsToLL_M-1to10/makeDatacards.py
jingyucms: Will remove these lines.

sihyunjeon:
#********************************************************************* 
#  Additional parameter
#********************************************************************* 

@agrohsje Does these comment out blocks have to be removed as well? to avoid nanoaod parsing issues

agrohsje:
Hi @sihyunjeon that is not an issue. i was just uncertain about quotation marks or slashes. But there is also a check  in our scripts. just to avoid wasting time with gridpack production.

sihyunjeon:
ok i guess this is fine to go in then

PR #2763, opened 2020-11-17
ggH PDF updated for UL gridpack

mlizzo:
PDF set updated to 325300 in order to use the card for the production of ggH gridpack for the UL campaign. Other than that, no changes have been implemented

PR #2821, opened 2021-02-09
Signal samples for UL: Update PDFs

alpakpinar:
Hello, I was working on producing gridpacks for UL signal samples (VBF, ggH, WH, ZH, ggZH) and I used the Powheg input cards that are in genproductions/master, only with the PDFs updated. You can find the cards that I used in this PR. Can I ask if these cards look good and if any further update would be needed for UL (I ran with CMSSW_10_6_20 and all the gridpacks worked fine for me, but wanted to ask if I'm missing any details)? Thank you!

### DIFF
--- a/bin/Powheg/production/2017/13TeV/Higgs/HZJ_HanythingJ_NNPDF31_13TeV/HZJ_HanythingJ_NNPDF31_13TeV_M125_Vhadronic.input
+++ b/bin/Powheg/production/2017/13TeV/Higgs/HZJ_HanythingJ_NNPDF31_13TeV/HZJ_HanythingJ_NNPDF31_13TeV_M125_Vhadronic.input
@@ -21,8 +21,8 @@ ebeam2 6500d0     ! energy of beam 2
 bornktmin 0.26d0      ! (default 0d0) generation cut. Minimum kt in underlying Born
 bornsuppfact 0d0 ! (default 0d0)  mass param for Born suppression factor. If < 0 suppfact = 1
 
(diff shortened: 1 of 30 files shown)


agrohsje:
Dear @alpakpinar . Thanks for the commit. Note that top is still using 306000. Will HIG completely switch to 325300? Was the impact studied? I assume when probing high masses it makes a difference? Please let me know.

alpakpinar:
Dear @agrohsje , thanks a lot for looking into this! I'm not sure about the PDF choice but can double check with HIG MC contacts, I updated the PDF because of the error message given in [1]. Thanks!

[1]: https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/run_pwg_condor.py#L779

agrohsje:
Yes. Please check. In fact, in SM phase space we expect little change. Big impact is expected for phase space that is not well determined in the PDF's. Please check the HIG choice and I am happy to merge.

gourangakole:
Hello @agrohsje , yes from HIG side we expect almost no change between two PDF sets for SM samples.
From HWW side we already switched to 325300.
https://github.com/cms-sw/genproductions/pull/2771
https://github.com/cms-sw/genproductions/pull/2763

so we are happy to move with 325300.

alpakpinar:
Thank you @gourangakole, I also recently pushed 9e50a5c to update the PDF in ttH cards as well, if we're going to move forward with 325300.

agrohsje:
Thanks. Going to merge.

PR #2823, opened 2021-02-09
run2 cards for H ->ZA->llbb process with UFO model 2HDMtII_NLO

kjaffel:
These card are for 2HDM type II, H->ZA->ll bb process, for different tan beta values and for different mA and mH masses. The gg fusion and the b-associated production are treated separately due to the different order of computation :
The UFO model: https://feynrules.irmp.ucl.ac.be/attachment/wiki/2HDM/2HDMtII_NLO.tar.gz
The name of the card :
- HToZATo2L2B_MH_MA_tb_bbH4F_TuneCP5_13TeV-amcatnloFXFX-pythia8 [[for b associated production, 4F-scheme ]
- HToZATo2L2B_MH_MA_tb_ggH_TuneCP5_13TeV-madgraphMLM-pythia8     [for ggfusion ]

The pdf set used is NNPDF31 positive definite, following the recommendation on the usage of PDFs and CPX Tunes for UL campaigns:  https://cms-pdmv.gitbook.io/project/mccontact/info-for-mc-production-for-ultra-legacy-campaigns-2016-2017-2018 
- bb-associated production, 4FS : NLO computation
  LHA Name = NNPDF31_nnlo_as_0118_nf_4_mc_hessian
  LHA ID = 325500
- ggfusion with top/bottom loop induced : LO computation
  LHA Name = NNPDF31_nnlo_as_0118_mc_hessian_pdfas
  LHA ID = 325300


sihyunjeon:
Hi,

Can you upload the card generating script and just one set of cards so that we can review it easier?
There are too many cards we have to look at here.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/HToZATo2L2B_ggfusion_b-associatedproduction/HToZATo2L2B_1000p00_200p00_0p50_bbH4F_TuneCP5_13TeV-amcatnloFXFX-pythia8/HToZATo2L2B_1000p00_200p00_0p50_bbH4F_TuneCP5_13TeV-amcatnloFXFX-pythia8_madspin_card.dat
```diff
@@ -0,0 +1,26 @@
...
+#*    Manual:                                               *
+#*    cp3.irmp.ucl.ac.be/projects/madgraph/wiki/MadSpin     *
+#*                                                          *
+#************************************************************
+#Some options (uncomment to apply)
+#
+# set seed 1
+# set Nevents_for_max_weight 75 # number of events for the estimate of the max. weight
+# set BW_cut 15                 # cut on how far the particle can be off-shell
+# set spinmode onshell          # Use one of the madspin special mode
```
agrohsje, line 19: When just uploading an example, please clean these lines up. They just complicate the review.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/HToZATo2L2B_ggfusion_b-associatedproduction/HToZATo2L2B_1000p00_200p00_0p50_bbH4F_TuneCP5_13TeV-amcatnloFXFX-pythia8/HToZATo2L2B_1000p00_200p00_0p50_bbH4F_TuneCP5_13TeV-amcatnloFXFX-pythia8_proc_card.dat
```diff
@@ -0,0 +1,48 @@
...
+set group_subprocesses Auto
+set ignore_six_quark_processes False
+set gauge unitary
+set complex_mass_scheme False
+define l+ = e+ mu+
+define l- = e- mu-
+define p = g u c d s u~ c~ d~ s~
+define j = g u c d s u~ c~ d~ s~
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
```
agrohsje, line 46: Same here. Please remove everything that is already default and keep the relevant lines: 
import yourmodel
special particle definition
generate your process
output.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/HToZATo2L2B_ggfusion_b-associatedproduction/HToZATo2L2B_1000p00_200p00_0p50_bbH4F_TuneCP5_13TeV-amcatnloFXFX-pythia8/HToZATo2L2B_1000p00_200p00_0p50_bbH4F_TuneCP5_13TeV-amcatnloFXFX-pythia8_run_card.dat
```diff
@@ -0,0 +1,196 @@
...
+#***********************************************************************
+  1	= lpp1 ! beam 1 type (0 = no PDF)
+  1	= lpp2 ! beam 2 type (0 = no PDF)
+  6500.0	= ebeam1 ! beam 1 energy in GeV
+  6500.0	= ebeam2 ! beam 2 energy in GeV
+#***********************************************************************
+# PDF choice: this automatically fixes also alpha_s(MZ) and its evol.  *
+#***********************************************************************
+  lhapdf	= pdlabel ! PDF set
+325500 = lhaid ! if pdlabel=lhapdf, this is the lhapdf number
```
agrohsje, line 68: Please use default pdf syntax.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/HToZATo2L2B_ggfusion_b-associatedproduction/HToZATo2L2B_1000p00_200p00_0p50_bbH4F_TuneCP5_13TeV-amcatnloFXFX-pythia8/HToZATo2L2B_1000p00_200p00_0p50_bbH4F_TuneCP5_13TeV-amcatnloFXFX-pythia8_run_card.dat
```diff
@@ -0,0 +1,196 @@
...
+            ! specify for which choice to include scale dependence.
+  False	= reweight_pdf ! Reweight to get PDF uncertainty. Should be a
+            ! list booleans of equal length to LHAPDF ID to specify for
+            !  which PDF set to include the uncertainties.
+  -1.0	= rw_fscale_down  ! lower bound for fact scale variations
+  -1.0	= rw_rscale_down  ! upper bound for fact scale variations
+  -1.0	= rw_fscale_up   ! upper bound for fact scale variations
+  -1.0	= rw_rscale_up   ! upper bound for ren scale variations
+  325501   = PDF_set_min      ! First of the error PDF sets 
+  325600   = PDF_set_max      ! Last of the error PDF sets
```
agrohsje, line 112: We don't use this logic but the systematics module in rumcmsgrid_NLO.sh. Please remove.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/HToZATo2L2B_ggfusion_b-associatedproduction/HToZATo2L2B_1000p00_200p00_0p50_bbH4F_TuneCP5_13TeV-amcatnloFXFX-pythia8/HToZATo2L2B_1000p00_200p00_0p50_bbH4F_TuneCP5_13TeV-amcatnloFXFX-pythia8_run_card.dat
```diff
@@ -0,0 +1,196 @@
...
+# ickkw parameter:                                                     *
+#   0: No merging                                                      *
+#   3: FxFx Merging - WARNING! Applies merging only at the hard-event  *
+#      level. After showering an MLM-type merging should be applied as *
+#      well. See http://amcatnlo.cern.ch/FxFx_merging.htm for details. *
+#   4: UNLOPS merging (with pythia8 only). No interface from within    *
+#      MG5_aMC available, but available in Pythia8.                    *
+#  -1: NNLL+NLO jet-veto computation. See arxiv:1412.8408 [hep-ph].    *
+#***********************************************************************
+  3	= ickkw
```
agrohsje, line 135: Do you have additional jets at NLO? Don't see that in the proc card.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/HToZATo2L2B_ggfusion_b-associatedproduction/HToZATo2L2B_1000p00_200p00_0p50_bbH4F_TuneCP5_13TeV-amcatnloFXFX-pythia8/HToZATo2L2B_1000p00_200p00_0p50_bbH4F_TuneCP5_13TeV-amcatnloFXFX-pythia8_run_card.dat
```diff
@@ -0,0 +1,196 @@
...
+  15.0	= bwcutoff 
+#***********************************************************************
+# Cuts on the jets. Jet clustering is performed by FastJet.            *
+#  - When matching to a parton shower, these generation cuts should be *
+#    considerably softer than the analysis cuts.                       *
+#  - More specific cuts can be specified in SubProcesses/cuts.f        *
+#***********************************************************************
+  1.0	= jetalgo ! FastJet jet algorithm (1=kT, 0=C/A, -1=anti-kT)
+  1.0	= jetradius ! The radius parameter for the jet algorithm
+  20.	= ptj ! Min jet transverse momentum
```
agrohsje, line 151: Is this soft enough compared to cut at analysis level?

agrohsje:
Please consider my comments when uploading the example cards as suggested by Sihyun.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/HToZATo2L2B_ggfusion_b-associatedproduction/HToZATo2L2B_1000p00_200p00_0p50_bbH4F_TuneCP5_13TeV-amcatnloFXFX-pythia8/HToZATo2L2B_1000p00_200p00_0p50_bbH4F_TuneCP5_13TeV-amcatnloFXFX-pythia8_run_card.dat
kjaffel: maybe not sure ! i set ptj = 0.01 as suggested [here](https://twiki.cern.ch/twiki/bin/view/CMS/QuickGuideMadGraph5aMCatNLO#Troubleshooting_and_Suggestions) !
at the analysis level we cut on j.pt > 20.  for resolved jets and j.pt> 200. for the boosted one

bin/MadGraph5_aMCatNLO/cards/production/13TeV/HToZATo2L2B_ggfusion_b-associatedproduction/HToZATo2L2B_1000p00_200p00_0p50_bbH4F_TuneCP5_13TeV-amcatnloFXFX-pythia8/HToZATo2L2B_1000p00_200p00_0p50_bbH4F_TuneCP5_13TeV-amcatnloFXFX-pythia8_run_card.dat
kjaffel: yes h3 decay to b b~ i add it in madspin_card for both NLO and LO

agrohsje:
What is the difference between the 2 gg cards you committed, same for the 2 bb associated. Strictly speaking if you do bbX, then the 2 b's are not considered as part of the merging, i.e. there is no 0, 1 or 3 jets contribution. So indeed the parton shower would just add soft jets. Why do you call them additional partons and use merging algorithms?

sihyunjeon:
@kjaffel Hi, on the side note, did you have time to check out the proposed solution for you on hypernews? It will be a great help if you can test this out as we are assuming the /eos/home-/ areas are problematic while /eos/users/ aren't

https://hypernews.cern.ch/HyperNews/CMS/get/generators/4945/1/1/1.html

kjaffel:
> What is the difference between the 2 gg cards you committed, same for the 2 bb associated. Strictly speaking if you do bbX, then the 2 b's are not considered as part of the merging, i.e. there is no 0, 1 or 3 jets contribution. So indeed the parton shower would just add soft jets. Why do you call them additional partons and use merging algorithms?

The difference between cards is just that one is a template while the other is the final  .dat card, I pass the template to prepare_MG5_cards.py in order to get the final cards needed for a bunch of signal mass points(MH, MA )and for different tb values. 
I updated README.md  !

For the 2nd question : do you mean why I am using FXFX for NLO process  with matching and jet merging (multiple jet multiplicities, ickkw=3 in run card)? if so, the process is ` p p > h3 Z b b~  , h3> b b~ `  so, I do expect 4 bjets.
So for you these are not considered as additional jets and I should use instead NLO with matching but no jet merging (single jet multiplicity, ickkw=0 in run card) , is this what you mean ?

kjaffel:
> @kjaffel Hi, on the side note, did you have time to check out the proposed solution for you on hypernews? It will be a great help if you can test this out as we are assuming the /eos/home-/ areas are problematic while /eos/users/ aren't
> 
> https://hypernews.cern.ch/HyperNews/CMS/get/generators/4945/1/1/1.html
Hi @sihyunjeon I just replied sorry for the delay !

agrohsje:
I would suggest to use master in the scripts instead of UL2019. For MG5, master should have everything that you need for UL. Also PDF's should be the right ones. Master gives you 2.6.5 with all bug fixes we know of. 2.6.1 in ul2019 is less well maintained. 
I would suggest you just subit the templates and not the outputs of your scripts. One can always get the final cards from mcm. 
ickkw is not for jets from decay or "associated production". this is really pp > coreprocess ; add pp > coreprocess + 1 jet ; add pp > coreprocess +2 jets. 
Please fix.

kjaffel:
side note for : for gg fusion i am using `set spinmode none `  which suppose to handle the loop induced process and in this mode the BR is computed as the ratio of the width computed via the decay syntax (computed by MadEvent --standard LO phase-space integration without any cut--) and the total one which  i provide in the param_card.
From Madspin I get such warning: 
 `CRITICAL: Branching ratio larger than one for 36  `
I don't think that the total width I provide is very small(0.00323503787 )  to have BR > 1 
I thought width still can be handled up to 10-12 GeV  !!

agrohsje:
I am not sure I understand. What is just means is that the partial width is > total width. This statement can be true independent of the overall size of them.

sihyunjeon:
You do this in customizecard,
 set param_card width 36 0.00323503787

BR is calculated by the ratio of the widths (partial/total).
And what madgraph is giving you is that, as alexander commented, partial>total.
Which might indicate NWA is not really working for you, or perhaps the width you gave to 36 is somewhat wrong.

Can you try it by doing
 set param_card width 36 Auto
 and see what happens?

kjaffel:
> I am not sure I understand. What is just means is that the partial width is > total width. This statement can be true independent of the overall size of them.

This a decay table i get from Sushi: 
```
Decay table for A
Total width:   2.264e-02 GeV      BR
A  -> s  s     4.413e-06      1.949e-04
A  -> c  c     1.092e-04      4.822e-03
A  -> b  b     1.227e-02      5.419e-01
A  -> t  t     1.297e-03      5.731e-02
A  -> e  e     1.157e-10      5.110e-09
A  -> mu mu    4.946e-06      2.185e-04
A  -> ta ta    1.399e-03      6.178e-02
A  -> ga ga    1.953e-05      8.628e-04
A  -> Z  ga    4.386e-06      1.937e-04
A  -> g  g     7.241e-03      3.199e-01
A  -> Z  h     2.898e-04      1.280e-02
---------------------------------------
```
I was expecting BR 36 to be close to : 5.419e-01

agrohsje:
You can compare sushi to mg5 by doing 
 set param_card width 36 Auto
 and see the values mg5 provides. I assume that partial width sushi > total width mg5 and hence BR > 1 .

kjaffel:
you mean; partial (from mg5 ) > total (from Sushi)
I have set the width of 36 to Auto. 
No warning that the BR is larger than 1.
This's the decay table i get from madgraph
```
#      PDG        Width
DECAY  36   4.355538e+00
#  BR             NDA  ID1    ID2   ...
   4.964375e-01   2    37  -24 # 2.16225231817
   4.964375e-01   2   -37   24 # 2.16225231817
   6.737280e-03   2    5    -5   # 0.0293444770343
   3.211854e-04   2    15  -15 # 0.00139893505873
   6.650920e-05   2    25   23 # 0.000289683329877
   1.396889e-21   2    22   25 # 6.08420150293e-21
```
So in this case, should i set the width to "Auto" for all gg fusion cards ?
Thank a lots for all helpful feedback btw !

kjaffel:
Another question about the ggfusion cards: 
since we set ` ickkw = 0 ` what value I should use for `  10.0	= xqcut ! minimum kt jet measure between partons`
I keep getting from mg5 such warning  : `xqcut>0 but ickkw=0. Potentially not fully consistent setup. Be carefull`
The old card [here](https://github.com/cms-sw/genproductions/blob/60013422c59c7c56c39441896f296cb371094777/bin/MadGraph5_aMCatNLO/cards/production/13TeV/higgs/HToZATo2L2B/HToZATo2L2B_200_100/HToZATo2L2B_200_100_run_card.dat#L259) use xqcut =10 for LO cards with no matching/merging (single jet multiplicity, ickkw=0 in run card)

sihyunjeon:
1. If you want physically making sense BRs with MadGraph, you should use "Auto". If you don't care what numbers MadGraph gives and plan to just take the cross sections from literature, I don't think this won't create a problem with event distributions.
2. xqcut will not be doing anything as long as auto_ptj_mjj is turned off. But for your case to be 100% careful and precise, the correct setting will be xqcut=0,ickkw=0, and auto_ptj_mjj = False.

kjaffel:
Working on converting one of the gridpacks to NanoGen: 
I stumble through this error , I am wondering if I am still missing something in madspin_card.dat !

The log here : https://cp3.irmp.ucl.ac.be/~kjaffel/gridpack_to_nanogen.log
The fragment : https://cp3.irmp.ucl.ac.be/~kjaffel/Hadronizer_TuneCP5_13TeV_generic_LHE_pythia8_PSweights_cff.py
The script is avery simple `cmsDriver.py` command that I tried on old/approved cards and it worked just fine.
```bash
cmsDriver.py "${fragment}" --fileout "file:${output}" --mc --eventcontent NANOAODSIM --datatier NANOAOD --conditions auto:mc --step LHE,GEN,NANOGEN "${nEvents}"
```
Thanks in advance, if someone can have look. 

btw: I saw that someone had already opened a similar issue with similar error on [launchpad: #670471](https://answers.launchpad.net/mg5amcnlo/+question/670471), but I can't find it on the HN ! ( the problem solved after he contact the mc group ! )

sihyunjeon:
@kjaffel Were the gridpacks at least successfully made without error in the generation logs?

kjaffel:
> @kjaffel Were the gridpacks at least successfully made without error in the generation logs?

yes the gridpacks are fine .

sihyunjeon:
@kjaffel Can you upload your gridpack to lxplus public repo so that I can take a look and post this to GEN hypernews maybe?
There are some other experts on NanoGEN, and this thread might get extensively long.

kjaffel:
> @kjaffel Can you upload your gridpack to lxplus public repo so that I can take a look and post this to GEN hypernews maybe?
> There are some other experts on NanoGEN, and this thread might get extensively long.
Thanks, the gridpack here : `/afs/cern.ch/user/k/kjaffel/public/ZArun2_gridpacks`
I post my issue here as well  : https://hypernews.cern.ch/HyperNews/CMS/get/generators/4955.html

sihyunjeon:
> > @kjaffel Can you upload your gridpack to lxplus public repo so that I can take a look and post this to GEN hypernews maybe?
> > There are some other experts on NanoGEN, and this thread might get extensively long.
> > Thanks, the gridpack here : `/afs/cern.ch/user/k/kjaffel/public/ZArun2_gridpacks`
> > I post my issue here as well  : https://hypernews.cern.ch/HyperNews/CMS/get/generators/4955.html

I tested your gridpacks and the gridpack itself seems to be broken.

kjaffel:
> > > @kjaffel Can you upload your gridpack to lxplus public repo so that I can take a look and post this to GEN hypernews maybe?
> > > There are some other experts on NanoGEN, and this thread might get extensively long.
> > > Thanks, the gridpack here : `/afs/cern.ch/user/k/kjaffel/public/ZArun2_gridpacks`
> > > I post my issue here as well  : https://hypernews.cern.ch/HyperNews/CMS/get/generators/4955.html
> 
> I tested your gridpacks and the gridpack itself seems to be broken.

ooh no, did you see the log file attached in the same dir. I get no error  !
I did set `run_mode = 0 ` so I get such warning as I explained here in the [HN](https://hypernews.cern.ch/HyperNews/CMS/get/generators/4945/1/1/1.html)
``` 
Single-core mode not supported for loop-induced processes.
Beware that MG5aMC now changes your runtime options to a multi-core mode with only one active core.
```
and because run_mode= 0 is not supported for loop induced, I bypass the events generation but i still get the grid-pack which doesn't seem to me of a big issue. 
see Olivier comment about this in [launchpad]( https://answers.launchpad.net/mg5amcnlo/+question/695546) and if there's anything need to be fixed about loop-induced setting on condor cluster.

Can you check for me please, the other process , if the gridpack correct.
The path here :  `/afs/cern.ch/user/k/kjaffel/public/ZArun2_gridpacks/HToZATo2L2B_500p00_300p00_15p00_bbH4F_TuneCP5_13TeV-amcatnlo_pythia8_slc7_amd64_gcc700_CMSSW_10_6_19_tarball.tar.xz`

Thanks a lots .

agrohsje:
@kjaffel Please see my answer on HN.

sihyunjeon:
@kjaffel 
NLO : p p > h2 > h3 Z b b~, 4f scheme
LO : p p > h2 > h3 Z, 4f scheme

Is this what you want? Would it be okay for your analysis to have 2 signal samples with different ME level calculation orders?

sihyunjeon:
And one has mll=30 and the other has mll=0 cut. Is this for some reason?

kjaffel:
> Is this what you want? Would it be okay for your analysis to have 2 signal samples with different ME level calculation orders?
NLO : p p > h2 > h3 Z b b~, 4f scheme
LO : p p > h2 > h3 Z, loop induced via top/bottom loop ( I don't know if this case counted as 4FS  to be honest , my guess not ! ) 
Analysis wise for having to different signal with different ME level , won't be a big of problem because they will be treated separately anyway with 2 different signal modelling strategies.

kjaffel:
> And one has mll=30 and the other has mll=0 cut. Is this for some reason?
There's no reason to have different mll cut, and I don't think the cut matter, because at the analysis level ( I am asking for 2 leptons OSSF in range  70. < mll <120. ).
[
Screenshot from 2021-02-26 13-19-13
](url)
Thank you for bringing this up, because  I look to the distribution at the genlevel and isn't great for ggfusion (with  mll = 0 ): right plot.
The left plot for b-associated production , seems to me fine.
I am not sure a cut on the mll could do that.
The only thing I could think off is the special mode I am using `set spinmode none ` for gg fusion.

sihyunjeon:
Oh wait sorry
The mll cuts won't do anything because you are using madspin. The cuts in runcard aren't applied there

sihyunjeon:
If I remember correctly, "spinmode none" doesn't provide BW smearings so you would only get a "delta-function" for mZ. Can you check m(Z) instead of m(ll)?

kjaffel:
> If I remember correctly, "spinmode none" doesn't provide BW smearings so you would only get a "delta-function" for mZ. Can you check m(Z) instead of m(ll)?
z_fromGenPart_Mass_logy
- Does the same thing apply for h3 ?? I see the same thing when I plot the mass for the ggfusion process, see the plot below : 
h3_Mass_logy

sihyunjeon:
Yes this is totally reasonable in terms of MadGraph computing

sihyunjeon:
The cards itself doesn't have problems (at least for me now).
Should we merge this? @agrohsje

PR #2824, opened 2021-02-10
Add higgsino case, PSweights to Disappearing Tracks AMSB fragments

bpfrancis:
This adds a few things to the 13 TeV disappearing tracks AMSB fragments. The Higgsino-like LSP case is introduced (same as the private generation for EXO-19-010), PSweights is added, and the card creation script is slightly cleaned up.


agrohsje:
Thanks for the commit. Would it make sense to have the slha files as template and just replace relevant parameters? 
What is the idea behind setting isresonance to false for the chargino?  Not saying you should change, just like to better understand.

bpfrancis:
@agrohsje We took these SLHA files as Isajet produced them, and we want this repo to reflect what was historically done for EXO-19-010. Plus this way we can just rely on saying "mass spectra from Isajet" in publications. Perhaps for future searches we could do such a thing, but I would consider it a different sample even if very very similar.

The `isResonance = false` method is a quirk of Pythia8 that we need to avoid; the difference between M(chargino) and M(neutralino + pion) is smaller than the MSAFETY parameter that Pythia8's resonance decays require. The Pythia8 developers suggested this was the correct way to achieve our decay, by avoiding the ResonantDecay class altogether, and said in the future they'd fix things so an explicit `isResonance = False` isn't necessary.

agrohsje:
Thanks for all the info. Going to merge.

PR #2828, opened 2021-02-14
archive MG5 v2.7.3 and fix 0032-patch

sihyunjeon:
Instead of wget-ing MG releases from launchpad everytime, fetch it from cms-project-generators web.
And also fixed 0032-patch which was broken.


sihyunjeon:
patching file MadSpin/interface_madspin.py
Hunk #1 succeeded at 1282 with fuzz 1.
patching file madgraph/interface/common_run_interface.py
Hunk #1 succeeded at 3697 with fuzz 1.

@agrohsje Checked the patches applying correctly now (before, it was broken)

agrohsje:
Thanks so much. Let me merge.

PR #2831, opened 2021-02-16
typeI seesaw model heavy neutrino (N) from ttbar

sihyunjeon:
We have 4 different set of process final states.
+01j gridpacks with MLM jet matching.

**2L** : 2 lepton
**LepSMTop3L** : 3 lepton where 2 of the leptons are from N decaying top and the remaining 1 from SM top decay -> given as example
**HadSMTop3L** : 3 lepton where all 3 of the leptons are from N decaying top (the other top decays hadronically)
**4L** : 4 lepton where 3 of the leptons are from N decaying top and the remaining 1 from SM top


agrohsje:
Hi @sihyunjeon . Did you add the card to download the UFO?  I cannot spot it.

agrohsje:
Small comment: If you want you can addjust the top/Higgs width after changing the mass. Impact will of course be negligible.

sihyunjeon:
@agrohsje The model cards are uploaded there in feynrules wiki page and when madgraph doesn't find the model in models/ directory it looks up there and wgets it automatically. So it was intentional but sure I will add it, since archiving the model file also seems nice.
And thanks for the comment on the widths. We are also thinking of top mass uncertainty systematics so that should be a nice touch. I will update this in few hours.

sihyunjeon:
@agrohsje Took a while to make updates :(
Changed the top mass, yukawa, and widths to pdgid values instead.
Added extramodels card.

And forgot to mention this, p = 4f scheme

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TTbar01Jets_TypeIHeavyN/TTbar01Jets_TypeIHeavyN-El_LepSMTop_3L_LO_MN40_XQCUT20/TTbar01Jets_TypeIHeavyN-El_LepSMTop_3L_LO_MN40_XQCUT20_customizecards.dat
```diff
@@ -0,0 +1,18 @@
...
+set param_card ven2 0.0
+set param_card ven3 0.0
+set param_card vmun2 0.0
+set param_card vmun3 0.0
+set param_card vtan1 0.0
+set param_card vtan2 0.0
+set param_card vtan3 0.0
+set param_card wn2 100
+set param_card wn3 100
+set param_card mass 6 172.9
```
agrohsje, line 16: sorry. small remark. the background will have 172.5. wouldn't it be better to make a consistent choice of mtop. of course there will be no impact on the result. more a general remark also for the write-up and description of what was done.

sihyunjeon: I just checked one other TTbar samples, and that didn't change the width for top. So I think giving mtop=yukawatop=172.5 without modifying the widths should be okay. Right?

agrohsje:
You could also just use auto to sync it with your mass hypothesis. I assume in the end you use Powheg ttbar for background, which cannot easily be synced with MG5. I checked for heavy higgs but this is almost impossible.

sihyunjeon:
@agrohsje mt=172.5 and wt=Auto set!

PR #2834, opened 2021-02-17
Update EWK Z and W run cards

alpakpinar:
Update on Madgraph run cards for EWK Z(vv) and Z(ll): Set xqcut = 0 and auto_ptj_mjj = F

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ewkwmlvjj_5f_LO/ewkwmlvjj_5f_LO_run_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ewkwmlvjj_5f_LO/ewkwmlvjj_5f_LO_run_card.dat
@@ -29,7 +29,7 @@
 # Warning: Do not generate more than 1M events in a single run       *
 # If you want to run Pythia, avoid more than 50k events in a run.    *
 #*********************************************************************
-  250 	      = nevents ! Number of unweighted events requested 
+  2500	     = nevents ! Number of unweighted events requested 
       0       = iseed   ! rnd seed (0=assigned automatically=default))
 #*********************************************************************
 # Collider type and energy                                           *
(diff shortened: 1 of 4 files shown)


sihyunjeon:
A comment and perhaps a question in the run card.

1. mc_grouped_subproc = False
Is this really needed? [link](https://answers.launchpad.net/mg5amcnlo/+question/271601) says _"2) One new way to handle multi-particle in the initial state which was suppose to be neutral for normal case and improve the speed in presence of CKM. This do not create any bias at all and this can be turned off by adding in the run_card the following lines: False = mc_grouped_subproc"_
p p > w j j (2 -> 3) doesn't look really like "multi-particle", it's pretty standard. No?

2. Why don't you consider applying wbf cuts instead of drjj = 0.4 which makes more sense to me for t-channel diagrams?

3. Consider increasing the number of events 250 to 2-3000 for pilot run to be safe.

gourangakole:
> 1\. mc_grouped_subproc = False
Is this really needed? link says "2) One new way to handle multi-particle in the initial state which was suppose to be neutral for normal case and improve the speed in presence of CKM. This do not create any bias at all and this can be turned off by adding in the run_card the following lines: False = mc_grouped_subproc"
p p > w j j (2 -> 3) doesn't look really like "multi-particle", it's pretty standard. No?

I agree from HIG UL backgrounds production perspective. @alpakpinar  please update

> 2\. Why don't you consider applying wbf cuts instead of drjj = 0.4 which makes more sense to me for t-channel diagrams?

@alpakpinar can you clarify this point with your group (or probably who ever is going to use that?) so that we move forward. 

> 3\. Consider increasing the number of events 250 to 2-3000 for pilot run to be safe.

I agree on this point, @alpakpinar what you think?

alpakpinar:
Hello, sorry for my late reply. Thanks for these comments, for the first two items, I'll check with my group. I agree on the third point, I'll update the run cards with higher number of events for the pilot run in a shortly upcoming commit, I can update to something like 2.5K events if it works. Thanks!

alpakpinar:
Hello @sihyunjeon , @gourangakole , I updated the EWK cards now. Here is a list of items I changed since the last set of commits:

- I removed the mc_grouped_subproc = False option since it is not particularly needed here, as you suggested.
- I removed the minimum drjj requirement, I checked with my analysis team about this and we didn't introduce this cut on purpose, just copied it over. We were worried this cut might introduce problems in the cases where the jets we pick at reco do not correspond to these LHE jets. Since we don't have a big stat problem with these samples either, we are okay to move on without any min drjj requirement, if you think that is okay.
- I increased the number of events in the pilot run to 2.5K.

Thanks all for the suggestions If you think more changes will be needed before the merge, please let me know and I can work on those.

sihyunjeon:
@agrohsje I think the cards are fine

PR #2835, opened 2021-02-18
Cards for Drell-Yan (Z to EE) UL samples for HToEE analysis

Joe-W-Davies:
Adding cards for Drell-Yan (Z->ee) UL samples for the H->ee search with full Run 2 dataset. The cards are intended for producing two DY samples with selection on the dilepton invariant mass (105 < MLL < 160), one of which has enhanced VBF topology (VBFFilter). These samples will be requested for all three years. The sample names are listed below:

/DYJetsToEE M-105To160 TuneCP5 PSweights 13TeV-amcatnloFXFX-pythia8/*-v1/MINIAODSIM 
/DYJetsToEE M-105To160 VBFFilter TuneCP5 PSweights 13TeV-amcatnloFXFX-pythia8/*/MINIAODSIM


bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/dyee012j_5f_NLO_FXFX_M105to160/dyee012j_5f_NLO_FXFX_M105to160_proc_card.dat
```diff
@@ -0,0 +1,17 @@
...
+
+#switch to diagonal ckm matrix if needed for speed
+#import model loop_sm-no_b_mass
+
+define p = p b b~
+define j = j b b~
+
+define ell+ = e+ mu+ ta+
+define ell- = e- mu- ta-
+
```
agrohsje, line 11: Suggest to remove as pure ee.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/dyee012j_5f_NLO_FXFX_M105to160/dyee012j_5f_NLO_FXFX_M105to160_run_card.dat
```diff
@@ -0,0 +1,144 @@
...
+#***********************************************************************
+# Cuts on the charged leptons (e+, e-, mu+, mu-, tau+ and tau-) *
+# (more specific gen cuts can be specified in SubProcesses/cuts.f) *
+#***********************************************************************
+0 = ptl ! Min lepton transverse momentum
+-1 = etal ! Max lepton abs(pseudo-rap) (a value .lt.0 means no cut)
+0 = drll ! Min distance between opposite sign lepton pairs
+0 = drll_sf ! Min distance between opp. sign same-flavor lepton pairs
+0 = mll ! Min inv. mass of all opposite sign lepton pairs
+105 = mll_sf ! Min inv. mass of all opp. sign same-flavor lepton pairs
```
agrohsje, line 128: You use this as a lower cut and then run filers during production?

sihyunjeon: I'd rather instead recommend adding the cuts.f for maxmll, since some LHE events would be a waste with current setups.

sihyunjeon:
By "two samples" in your description, do you mean one with DY mll 105-160 and the other with DY mll 105-160 + VBF filters?

If this is the case, current setup won't give you VBF DY feynman diagrams. p p > e+ e- j j QCD=0 [QCD] might give you what you need. What you now have will be giving 2 gluon splitting to 2 pairs of qq~ and 2 of the 4 quarks making Z boson, not the VBF. So even if you attach VBF filter to that (with mjj or detajj cuts maybe?), there will be no VBF contribution. To make VBF you have to specify the number of QCD couplings by writing "QCD=0".

sihyunjeon:
And wouldn't it be worth adding ta+ ta- final states (then decaying to electrons)? Given that higgs BR to ta+ta- is much more bigger than the light leptons? Maybe you already did the cross section checks, but I am asking just in case.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/dyee012j_5f_NLO_FXFX_M105to160/dyee012j_5f_NLO_FXFX_M105to160_proc_card.dat
```diff
@@ -0,0 +1,17 @@
+import model loop_sm-ckm_no_b_mass
+
+#switch to diagonal ckm matrix if needed for speed
+#import model loop_sm-no_b_mass
```
sihyunjeon, line 4: Also remove the line that won't be used

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/dyee012j_5f_NLO_FXFX_M105to160/dyee012j_5f_NLO_FXFX_M105to160_proc_card.dat
Joe-W-Davies: sure, we will remove this

Joe-W-Davies:
@sihyunjeon,
the VBF-filtered sample we require is not from VBF Z production, rather just DY events with two jets from radiated gluons. This is a background in the VBF(H->ee) analysis category, where we require more events with this topology for MVA training.

For the Z->tau tau, there are samples already generated which we will use for this analysis. We dont require additional MC stats here, since the contribution from these events is small.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/dyee012j_5f_NLO_FXFX_M105to160/dyee012j_5f_NLO_FXFX_M105to160_proc_card.dat
Joe-W-Davies: sure, this will be removed

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/dyee012j_5f_NLO_FXFX_M105to160/dyee012j_5f_NLO_FXFX_M105to160_run_card.dat
Joe-W-Davies: for the max mLL cut, is it possible to just add this in the run card rather than creating a dedicated cut card?

sihyunjeon:
Okay that does makes sense but maybe when you request for official samples, the namings should be made with care.
@agrohsje I think the cards are okay

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/dyee012j_5f_NLO_FXFX_M105to160/dyee012j_5f_NLO_FXFX_M105to160_run_card.dat
sihyunjeon: No it won't work you need to add the cuts.f
And just out of curiosity, what would be your analysis level mll selections for the signals?

Joe-W-Davies: OK sure. Could you point us to an example for how to do this please?
The analysis level selection is currently 110 < mll < 150 GeV, however this is still subject to change.

sihyunjeon: https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/DYJetsToLL_jetbinned_vptbinned_NLO/dyellell1j_ZpT-120to280_5f_NLO_FXFX/dyellell1j_ZpT-120to280_5f_NLO_FXFX_cuts.f#L382-L401

This might be helpful? Then I would suggest you to loosen the cut a little bit more 100-160

Joe-W-Davies:
As you suggest, I have added a cut card to implement the Mll < 160 GeV selection. This was borrowed from a similar H->mumu card here: 

Please let me know if the cut card looks sensible.

I have also updated the proc and run cards with your other comments, including setting the lower Mll cut to 100 GeV.

agrohsje:
Was this using 2.6.5 ? Best is to start from a setcuts.f card that is in MG5.2.6.5. They are not always compatible with previous versions.

Joe-W-Davies:
I'm not sure sorry - is there an example with MG5.2.6.5 that I can compare to, to check?

agrohsje:
I copied the default from 2.6.5 here: https://www.desy.de/~agrohsje/hidden/cuts.f
You should add your cuts there.

Joe-W-Davies:
ok thanks, I've tried to modify this default card to add the Mll selection. Let me know how it looks.

sihyunjeon:
I think they look fine
@agrohsje

agrohsje:
Hi @Joe-W-Davies . Looks good to me. When you produce the gridpack you might double-check that mll is between 100 and 160 with a few events from the gridpack.

Joe-W-Davies:
Hi @agrohsje, just a follow up to this thread.

With the new cut card, I can't run the gridpack generation successfully - is it possible I am missing something to do with the cut card in my command? Currently I'm doing the following: 

./gridpack_generation.sh dyee012j_5f_NLO_FXFX_M100to160/ cards/production/2017/13TeV/dyee012j_5f_NLO_FXFX_M100to160/

Thanks for your help

agrohsje:
That should be copied by the script: 
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/gridpack_generation.sh#L425
Do you observe problems with that?

agrohsje:
can you send the full log ?

sihyunjeon:
check the proc_card please
the output is set to "dyee012j_5f_NLO_FXFX_M105to160" not "dyee012j_5f_NLO_FXFX_M**100**to160"

Joe-W-Davies:
ok the original issue is now solved, however the gridpack is still not getting produced.

Its not obvious to me what fails this time, so I have attached the log file for the attempted production. Interestingly, the execution does not stop - instead it just remains idle, after a warning about a sub-process termination is printed (L 2323)

[dyee012j_5f_NLO_FXFX_M100to160.log](https://github.com/cms-sw/genproductions/files/6142350/dyee012j_5f_NLO_FXFX_M100to160.log)

Joe-W-Davies:
Hi,
Just to say, this error still persists and causes the gridpack generation to fail. 

It's still not obvious to me what the issue is, so any help/pointers would be greatly appreciated.

sihyunjeon:
How/where did you run this?
In lxplus or cmsconnect? Using condor or interactively?

Joe-W-Davies:
this was when running interactively in a screen session on lxplus

sihyunjeon:
@Joe-W-Davies Hi, Sorry for late replies (a bit busy right now).
Can you post your errors to generator tools hypernews and put your inputcards in lxplus public repo?
It's a bit difficult for me to keep up with the debugging issues here.

PR #2836, opened 2021-02-19
XToJPsiG_JPsiToMuMu:: VBF H production mode cards for JPsi and PsiPrime

RoccoA97:
Cards for other production modes for the Higgs in the context of the analysis H->JPsiGamma->MuMuGamma. I have added the VBF case. Then, if it is ok, I will commit the remaining ones (WH, ZH, ttH and bbH).
The pull request for main production modes has already been merged recently. The run cards for the VBF are the same for the GluGluH case.


sihyunjeon:
Just a minor suggestion.
If these are for VBF H, wouldn't it be better to apply small ptj cut or wbf eta cut? Maybe the impact would be small but I am just curious

RoccoA97:
We have tried some test generations without cuts, but it seems working without anomalies. There is just a small fraction of events that we would lose by putting, for instance, a cut on the ptj of about 20.
Anyway, in our case, we would prefer to not have cuts and keep these settings since we are more interested in calculation of the efficiency of the signal rather than the calculation of the acceptance...

sihyunjeon:
The cards look fine to me @agrohsje

PR #2841, opened 2021-02-24
XToJPsiG_JPsiToMuMu:: WH and ZH production modes cards

RoccoA97:
Cards for other minor production modes for the Higgs in the context of the analysis H->JPsiGamma->MuMuGamma. I have added the WH and ZH production for both the cases of decay in JPsi and PsiPrime.
Again, as in the pull request #2836, the run cards are identical to the ones for the other production modes of the H.
If the cards for WH and ZH production modes are ok, then I will commit the last ones for bbH and ttH.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/XToJPsiG_JPsiToMuMu/bbH_HToJPsiG_JPsiToMuMu/bbH_HToJPsiG_JPsiToMuMu_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/XToJPsiG_JPsiToMuMu/bbH_HToJPsiG_JPsiToMuMu/bbH_HToJPsiG_JPsiToMuMu_proc_card.dat
@@ -8,6 +8,5 @@ set complex_mass_scheme False
 set max_npoint_for_channel 0
 import model jpsi_sm
 define p = g u c d s u~ c~ d~ s~
-generate p p > b b~ h $$ u c d s b t u~ c~ d~ s~ b~ t~ g / u c d s t u\
-~ c~ d~ s~ t~ g, (h > jpsi a, jpsi > mu+ mu-)
+generate p p > b b~ h HIG<=0 QED<=1, (h > jpsi a, jpsi > mu+ mu-)
 output bbH_HToJPsiG_JPsiToMuMu -nojpeg
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/XToJPsiG_JPsiToMuMu/bbH_HToJPsiG_JPsiToMuMu/bbH_HToJPsiG_JPsiToMuMu_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/XToJPsiG_JPsiToMuMu/bbH_HToJPsiG_JPsiToMuMu/bbH_HToJPsiG_JPsiToMuMu_proc_card.dat
@@ -1,0 +1,13 @@
+set default_unset_couplings 99
+set group_subprocesses Auto
+set ignore_six_quark_processes False
+set loop_optimized_output True
+set loop_color_flows False
+set gauge unitary
+set complex_mass_scheme False
+set max_npoint_for_channel 0
+import model jpsi_sm
+define p = g u c d s u~ c~ d~ s~
+generate p p > b b~ h $$ u c d s b t u~ c~ d~ s~ b~ t~ g / u c d s t u\
+~ c~ d~ s~ t~ g, (h > jpsi a, jpsi > mu+ mu-)
+output bbH_HToJPsiG_JPsiToMuMu -nojpeg
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/XToJPsiG_JPsiToMuMu/bbH_HToPsiPrimeG_PsiPrimeToMuMu/bbH_HToPsiPrimeG_PsiPrimeToMuMu_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/XToJPsiG_JPsiToMuMu/bbH_HToPsiPrimeG_PsiPrimeToMuMu/bbH_HToPsiPrimeG_PsiPrimeToMuMu_proc_card.dat
@@ -1,0 +1,13 @@
+set default_unset_couplings 99
+set group_subprocesses Auto
+set ignore_six_quark_processes False
+set loop_optimized_output True
+set loop_color_flows False
+set gauge unitary
+set complex_mass_scheme False
+set max_npoint_for_channel 0
+import model jpsi_sm
+define p = g u c d s u~ c~ d~ s~
+generate p p > b b~ h $$ u c d s b t u~ c~ d~ s~ b~ t~ g / u c d s t u\
+~ c~ d~ s~ t~ g, (h > jpsi a, jpsi > mu+ mu-)
+output bbH_HToPsiPrimeG_PsiPrimeToMuMu -nojpeg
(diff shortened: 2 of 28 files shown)


sihyunjeon:
Hello,

I might have asked this already (apologies if I did).
Are you not considering decaying W/Z through MadGraph? If you do it with Pythia you will lose spin correlations (please correct me if I am wrong @agrohsje)

RoccoA97:
This is a good point. In principle, this should not affect significantly our results, since we are focusing on other angular distributions, namely the ones for the muons produced from the decay of the JPsi (with respect to the flight direction of the JPsi) and for the JPsi (with respect to the flight direction of the H).
However, it is better that we fix this by decaying the W/Z inside Madgraph in order to be safe. Maybe, in the meanwhile, I commit the cards for the bbH and ttH so that you check if these ones are correct (the run cards are always the same, the differences are just in the proc cards)

bin/MadGraph5_aMCatNLO/cards/production/13TeV/XToJPsiG_JPsiToMuMu/ttH_HToPsiPrimeG_PsiPrimeToMuMu/ttH_HToPsiPrimeG_PsiPrimeToMuMu_proc_card.dat
```diff
@@ -0,0 +1,13 @@
...
+set ignore_six_quark_processes False
+set loop_optimized_output True
+set loop_color_flows False
+set gauge unitary
+set complex_mass_scheme False
+set max_npoint_for_channel 0
+import model jpsi_sm
+define p = g u c d s u~ c~ d~ s~
+generate p p > t t~ h $$ u c d s b t u~ c~ d~ s~ b~ t~ g / u c d s b u\
+~ c~ d~ s~ b~ g, (h > jpsi a, jpsi > mu+ mu-)
```
sihyunjeon, line 12: Is there a reason to discard processes like g g > g > t t~ h?

bin/MadGraph5_aMCatNLO/cards/production/13TeV/XToJPsiG_JPsiToMuMu/ttH_HToJPsiG_JPsiToMuMu/ttH_HToJPsiG_JPsiToMuMu_proc_card.dat
```diff
@@ -0,0 +1,13 @@
...
+set ignore_six_quark_processes False
+set loop_optimized_output True
+set loop_color_flows False
+set gauge unitary
+set complex_mass_scheme False
+set max_npoint_for_channel 0
+import model jpsi_sm
+define p = g u c d s u~ c~ d~ s~
+generate p p > t t~ h $$ u c d s b t u~ c~ d~ s~ b~ t~ g / u c d s b u\
+~ c~ d~ s~ b~ g, (h > jpsi a, jpsi > mu+ mu-)
```
sihyunjeon, line 12: Is there a reason to discard processes like g g > g > t t~ h?

bin/MadGraph5_aMCatNLO/cards/production/13TeV/XToJPsiG_JPsiToMuMu/bbH_HToPsiPrimeG_PsiPrimeToMuMu/bbH_HToPsiPrimeG_PsiPrimeToMuMu_proc_card.dat
```diff
@@ -0,0 +1,13 @@
...
+set ignore_six_quark_processes False
+set loop_optimized_output True
+set loop_color_flows False
+set gauge unitary
+set complex_mass_scheme False
+set max_npoint_for_channel 0
+import model jpsi_sm
+define p = g u c d s u~ c~ d~ s~
+generate p p > b b~ h $$ u c d s b t u~ c~ d~ s~ b~ t~ g / u c d s t u\
+~ c~ d~ s~ t~ g, (h > jpsi a, jpsi > mu+ mu-)
```
sihyunjeon, line 12: Same comments here as well, for which purpose is $$ and / used for?

bin/MadGraph5_aMCatNLO/cards/production/13TeV/XToJPsiG_JPsiToMuMu/bbH_HToJPsiG_JPsiToMuMu/bbH_HToJPsiG_JPsiToMuMu_proc_card.dat
```diff
@@ -0,0 +1,13 @@
...
+set ignore_six_quark_processes False
+set loop_optimized_output True
+set loop_color_flows False
+set gauge unitary
+set complex_mass_scheme False
+set max_npoint_for_channel 0
+import model jpsi_sm
+define p = g u c d s u~ c~ d~ s~
+generate p p > b b~ h $$ u c d s b t u~ c~ d~ s~ b~ t~ g / u c d s t u\
+~ c~ d~ s~ t~ g, (h > jpsi a, jpsi > mu+ mu-)
```
sihyunjeon, line 12: Same comments here as well, for which purpose is $$ and / used for?

RoccoA97:
Hi! Actually, there were too much unnecessary $$ and /, as well as missing processes (like the g g > g > t t~ h). Thank you for the suggestions. We are fixing all these things, I will commit the corrected versions of the cards for WH, ZH, ttH and bbH as soon as possible

RoccoA97:
Hi! I have just committed the fixes in the process cards of the WH, ZH, ttH and bbH modes for both JPsi and PsiPrime:
* for WH and ZH, we have cleaned the process cards and added the decay of the W into q,q'/l,nu and of the Z into q,q~/l,l~, so that it is handled by MadGraph and not by Pythia (in this case, we would lose the spin correlations);
* for bbH and ttH, we have fixed the part with $$ and /, now there are all the main diagrams we needed and the `generate` line is more clean.

I hope it is fine and more clear now. Thank you again for all the tips!

sihyunjeon:
Hi,

Maybe one last comment. What are your targeting final states (for instance ZH)? Are you also considering both cases where Z decays hadronically and leptonically? BRs are pretty much different so you won't get equal stats.

Why don't you separate the gridpacks?

RoccoA97:
In our case this is not so relevant, since we are very efficient with the signal and also a small fraction of events where the Z decays leptonically are fine. Moreover, we do not have a sufficient statistics in the observed data in order to deal also with a dedicated search region for the case of ZH (for example), with the Z decaying through Z->ll...

sihyunjeon:
Okay I think the cards are fine @agrohsje

RoccoA97:
Hi @sihyunjeon, @agrohsje! Is there any news concerning the merge? If it is all ok, can we proceed with it?

agrohsje:
Dear @RoccoA97 . Sorry for the delay. I have two final questions before merging. 
What is the purpose of doing sometimes add sm and sometimes not ? 
About these lines: 
generate p p > b b~ h HIG<=0 QED<=1, (h > jpsi a, jpsi > mu+ mu-)
What is the labeling of the bbH vertex in the UFO? Or differently asked why do you put HIG=0 and QED=1 ?

RoccoA97:
1. the purpose of adding sm is due to the definition of the model `jpsi_sm`. It is a minimal version of sm with the inclusion of jpsi particle and of the decays H/Z->jpsi+photon and  jpsi->mu+mu-. Therefore, some basic sm interactions are not included and when we expect them to be in the diagrams, we just add the model `sm`.
For the other cards already merged, adding or not `sm` does not provide changes neither in the considered diagrams, nor in the couplings. Concerning WH/ZH production modes, we need to add `sm` in order to include all the diagrams given by the decay of W/Z. This is the only case where we need to add it. I hope this is clear.
2. we put `HIG<=0 QED<=1` in order to avoid the inclusion of diagrams in which there is an effective vertex ggH, which is defined in the model `jpsi_sm`. However, we have performed some tests with and without this constraint and we have seen that their cross section is irrelevant, so their presence is not a big deal

agrohsje:
Thanks for the info about 1. 
Small follow-up on 2: How is the ttH vertex defined? You get this from having QED=1?

agrohsje:
Let me merge. I assume you get results, so the answer to my question will not affect the merging. :-)

PR #2849, opened 2021-03-04
Adding template run card with 2016 pdf settings

jotadram6:


agrohsje:
Hi Jose,
wouldn't it make more sense to update : 
https://github.com/cms-sw/genproductions/tree/master/MetaData and include the systematic variations of the desired pdf as well  ? 
One question unrelated to your but triggered by your pr: wouldn't the analysis benefit from using MM ?

sihyunjeon:
@agrohsje I think we can close this request without merging.
This is the one I had to fix 263400@0 to 263400 in the runcmsgrid.sh.

agrohsje:
Going to close.

PR #2850, opened 2021-03-08
Added cards for production of X5/3 to top and charged Higgs

farrah-simpson:


sihyunjeon:
Hi please only make a pull request for 1 example mass point + card generating script.
It's difficult to review all the cards in here

PR #2855, opened 2021-03-15
MuMuJet_mll_0to60_LO for Higs Dalitz analysis

hrjheng:
Pull request for MuMuJet_mll_0to60_LO sample for the Higgs Dalitz analysis


bin/MadGraph5_aMCatNLO/cards/production/13TeV/MuMuJet_mll_0to60_LO/dyellell01234j_5f_LO_MLM_proc_card.dat
```diff
@@ -0,0 +1,20 @@
...
+define ell+ = mu+
+define ell- = mu- 
+
+#generate p p > ell+ ell- @0
+#add process p p > ell+ ell- j @1
+#add process p p > ell+ ell- j j @2
+#add process p p > ell+ ell- j j j @3
+#add process p p > ell+ ell- j j j j @4
+
+generate p p > ell+ ell- j @1
```
sihyunjeon, line 17: The process you want is DY mumu with one hard jet from matrix element?

sihyunjeon, line 17: Please remove unnecessary lines that are comment outed with #

bin/MadGraph5_aMCatNLO/cards/production/13TeV/MuMuJet_mll_0to60_LO/dyellell01234j_5f_LO_MLM_proc_card.dat
```diff
@@ -0,0 +1,20 @@
...
+
+output dyellell01234j_5f_LO_MLM
```
sihyunjeon, line 19: output directory name should be the same with card directory name

bin/MadGraph5_aMCatNLO/cards/production/13TeV/MuMuJet_mll_0to60_LO/dyellell01234j_5f_LO_MLM_run_card.dat
```diff
@@ -0,0 +1,270 @@
...
+#*********************************************************************
+# Beam polarization from -100 (left-handed) to 100 (right-handed)    *
+#*********************************************************************
+        0     = polbeam1 ! beam polarization for beam 1
+        0     = polbeam2 ! beam polarization for beam 2
+#*********************************************************************
+# PDF CHOICE: this automatically fixes also alpha_s and its evol.    *
+#*********************************************************************
+ 'lhapdf'    = pdlabel     ! PDF set                                  
+  263000    = lhaid     ! if pdlabel=lhapdf, this is the lhapdf number
```
sihyunjeon, line 52: Is this for RunII-prelegacy or Ultra legacy?

bin/MadGraph5_aMCatNLO/cards/production/13TeV/MuMuJet_mll_0to60_LO/dyellell01234j_5f_LO_MLM_run_card.dat
```diff
@@ -0,0 +1,270 @@
...
+#*************************************************************
+   0  = nhel          ! Number of helicities used per event
+#*******************                                                 
+# Standard Cuts
+#*******************                                                 
+#                                                                    
+#*********************************************************************
+# Minimum and maximum pt's (for max, -1 means no cut)                *
+#*********************************************************************
+ 20  = ptj       ! minimum pt for the jets
```
sihyunjeon, line 107: Your analysis level minimum jet pt cuts is bigger than 20 GeV?

bin/MadGraph5_aMCatNLO/cards/production/13TeV/MuMuJet_mll_0to60_LO/dyellell01234j_5f_LO_MLM_run_card.dat
```diff
@@ -0,0 +1,270 @@
...
+#*********************************************************************
+# Minimum and maximum DeltaR distance                                *
+#*********************************************************************
+ 0   = drjj    ! min distance between jets 
+ 0   = drbb    ! min distance between b's 
+ 0   = drll    ! min distance between leptons 
+ 0   = draa    ! min distance between gammas 
+ 0   = drbj    ! min distance between b and jet 
+ 0.1 = draj    ! min distance between gamma and jet 
+ 0.4 = drjl    ! min distance between jet and lepton
```
sihyunjeon, line 151: I would suggest removing this cut to be safe from smearing

sihyunjeon:
Sorry I did not follow up after your GEN meeting presentations.
Did you find your gridpacks stable although with low mll cuts?

bin/MadGraph5_aMCatNLO/cards/production/13TeV/MuMuJet_mll_0to60_LO/dyellell01234j_5f_LO_MLM_proc_card.dat
hrjheng: Yes, Drell-Yan process to a muon pair with a hard jet from ME.

hrjheng: Done in the new commit

bin/MadGraph5_aMCatNLO/cards/production/13TeV/MuMuJet_mll_0to60_LO/dyellell01234j_5f_LO_MLM_proc_card.dat
hrjheng: Done in the new commit

bin/MadGraph5_aMCatNLO/cards/production/13TeV/MuMuJet_mll_0to60_LO/dyellell01234j_5f_LO_MLM_run_card.dat
hrjheng: Yes, we will have pT cut > 20GeV

bin/MadGraph5_aMCatNLO/cards/production/13TeV/MuMuJet_mll_0to60_LO/dyellell01234j_5f_LO_MLM_run_card.dat
hrjheng: Could you point out which cut specifically?

hrjheng:
> Sorry I did not follow up after your GEN meeting presentations.
> Did you find your gridpacks stable although with low mll cuts?

Yes, the gridpack is stable with additional kinematic cuts.

sihyunjeon:
> > Sorry I did not follow up after your GEN meeting presentations.
> > Did you find your gridpacks stable although with low mll cuts?
> 
> Yes, the gridpack is stable with additional kinematic cuts.

Maybe this is a lot outdated [1]. But anyways in slide 8, with same mll=0.211317GeV cut, everything except +1jet gridpack was not produced. Is this now understood? Which kinematic cut helped you resolve this?

[1] https://indico.cern.ch/event/961675/contributions/4084800/attachments/2134551/3595633/HDalitz-bkg-GENPOG-20201102.pdf

bin/MadGraph5_aMCatNLO/cards/production/13TeV/MuMuJet_mll_0to60_LO/dyellell01234j_5f_LO_MLM_run_card.dat
sihyunjeon: Then I'd suggest you loosen the ptj cut to 10 or 15, 20 GeV ME level cut is too tight

bin/MadGraph5_aMCatNLO/cards/production/13TeV/MuMuJet_mll_0to60_LO/dyellell01234j_5f_LO_MLM_run_card.dat
sihyunjeon: I meant this drjl

hrjheng:
> > > Sorry I did not follow up after your GEN meeting presentations.
> > > Did you find your gridpacks stable although with low mll cuts?
> > 
> > 
> > Yes, the gridpack is stable with additional kinematic cuts.
> 
> Maybe this is a lot outdated [1]. But anyways in slide 8, with same mll=0.211317GeV cut, everything except +1jet gridpack was not produced. Is this now understood? Which kinematic cut helped you resolve this?
> 
> [1] https://indico.cern.ch/event/961675/contributions/4084800/attachments/2134551/3595633/HDalitz-bkg-GENPOG-20201102.pdf

Yes, the problem of producing gridpacks with errors was solved by using a non-zero pT cut on leptons. This solution was suggested by Armando.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/MuMuJet_mll_0to60_LO/dyellell01234j_5f_LO_MLM_run_card.dat
hrjheng: Could you explain why 20GeV is too tight?

hrjheng: Could you explain why a 20GeV cut is too tight? Thanks

bin/MadGraph5_aMCatNLO/cards/production/13TeV/MuMuJet_mll_0to60_LO/dyellell01234j_5f_LO_MLM_run_card.dat
hrjheng: We put this cut on purpose. Our analysis selection is actually not DY+jet, but it is DY+gamma, where gamma comes from jet. So we want to have a sample with a jet, which would fake as a photon. Further, we want this jet to separate from the muons in angle. We checked that if this drjl cut was not imposed, a significant amount of events would have overlapped hard leptons and jet, which we do not want to have.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/MuMuJet_mll_0to60_LO/dyellell01234j_5f_LO_MLM_run_card.dat
hrjheng: We would like to have Ultra-legacy samples

hrjheng:
Hi Sihyun,
Do you have other comments and suggestions that we should address in order to proceed the PR? 
Thanks!

sihyunjeon:
Why does ptl cut solve issues?
ptl=3 => If it wasn't, I think this is some sort of adhoc fix.
With higher ptl, you might just be practically removing low mll contributions near the low threshold 0.2 GeV, which is not stable and getting away with problems by restricting phase spaces with that cut.

What is your analysis reco level ptcut on muons?

hrjheng:
> Why does ptl cut solve issues?
ptl=3 => If it wasn't, I think this is some sort of adhoc fix.
With higher ptl, you might just be practically removing low mll contributions near the low threshold 0.2 GeV, which is not stable and getting away with problems by restricting phase spaces with that cut.

Adding a non-zero ptl cut was the suggestion from Armando. He commented that a non-zero ptl cut could resolve the 
problem in the "check_invariant" function. I also checked the error message from the failed gridpack production and found that errors did occur from the "check_invariant" function with a zero ptl cut. Once I added the pTl cut, I no longer saw errors. Therefore, I think it's not a adhoc solution.

---
> What is your analysis reco level ptcut on muons?

The minimum pT cut on the reco level muons is 4GeV. This can also justify the use of 3GeV on ptl cut in the gridpack production.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/MuMuJet_mll_0to60_LO/dyellell01234j_5f_LO_MLM_run_card.dat
sihyunjeon: That makes sense

bin/MadGraph5_aMCatNLO/cards/production/13TeV/MuMuJet_mll_0to60_LO/dyellell01234j_5f_LO_MLM_run_card.dat
sihyunjeon: There could be some smearing effects.
ME level ptj = 19 -> RECO level ptj = 21
Things like this could happen especially for the jets where the momentum resolution is not so great compared to the leptons.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/MuMuJet_mll_0to60_LO/dyellell01234j_5f_LO_MLM_run_card.dat
sihyunjeon: Then is there any reason not to use NNPDF31?

hrjheng: No, there is no particular reason. I used the default settings in the card and did not make any change to it. Do you suggest to use NNPDF31? 
(Technically, do I simply put 'nnpdf31' = pdlabel for this option?)

bin/MadGraph5_aMCatNLO/cards/production/13TeV/MuMuJet_mll_0to60_LO/dyellell01234j_5f_LO_MLM_run_card.dat
hrjheng: Thanks. Then I think I should have make it clearer in my response earlier. Sorry for the confusion.
What I meant a pT cut > 20GeV was, we would apply a cut like 30GeV above, and this cut value is larger than 20GeV. 
Since we would like to have a higher cut than what we put in the card (or at ME level), I think it's safe to use 20GeV.

hrjheng:
Hi Sihyun,
I have replied to your questions earlier. Do you have further comments?
Apart from the gridpack, I also have a fragment file that I use to generate GEN events with the EM-enriched filter. Do I also need to commit that fragment file somewhere in this repository?

Thanks

bin/MadGraph5_aMCatNLO/cards/production/13TeV/MuMuJet_mll_0to60_LO/dyellell01234j_5f_LO_MLM_run_card.dat
agrohsje: Please use the default pdf setting as specified on the MG5 subpage of the GEN TWiki.Thanks!

agrohsje:
Hi @hrjheng , please follow-up on the above comment. Thanks!

hrjheng:
Hi @agrohsje and @sihyunjeon,
I believe the twiki page you mentioned is this one, am I right?: [https://twiki.cern.ch/twiki/bin/view/CMS/QuickGuideMadGraph5aMCatNLO#Specific_information_for_2017_pr](https://twiki.cern.ch/twiki/bin/view/CMS/QuickGuideMadGraph5aMCatNLO#Specific_information_for_2017_pr)
There I can see information regarding the 2016 and 2017 productions (but different for these two cases). Is there any special treatment for 2018 production?

sihyunjeon:
> Hi @agrohsje and @sihyunjeon,
> I believe the twiki page you mentioned is this one, am I right?: https://twiki.cern.ch/twiki/bin/view/CMS/QuickGuideMadGraph5aMCatNLO#Specific_information_for_2017_pr
> There I can see information regarding the 2016 and 2017 productions (but different for these two cases). Is there any special treatment for 2018 production?

For Ultra-legacy, the settings would likely be following the 2017 case for all 3 years (2016 is when TuneCUETP is used; for UL it's TuneCP so you don't need to worry about this at all).

sihyunjeon:
My one last concern/question would be the cut drjl = 0.4.
But as long as you feel safe with this, I don't have any other issues

hrjheng:
If I did not put drjl = 0.4 cut in my card, I saw a majority of events in which the leading gen jet would just be the leading hard muon. This is not what we want in our analysis. That's why I put this cut. 

I also have a customized fragment file that I used privately to generate GEN events with the EM-enriched filter. Do I also need to commit this fragment file somewhere in the repository?

sihyunjeon:
> If I did not put drjl = 0.4 cut in my card, I saw a majority of events in which the leading gen jet would just be the leading hard muon. This is not what we want in our analysis. That's why I put this cut.
> 
> I also have a customized fragment file that I used privately to generate GEN events with the EM-enriched filter. Do I also need to commit this fragment file somewhere in the repository?

I think you misunderstood the definition of "genjet" here. "genjet" you are referring to in CMSSWs is the "clustering of energy using jet algorithms", not "generator level quark/gluon QCD jet". The drjl cut you applied is "deltaR cut between lepton and QCD jet in matrix element level". So the genjet distribution doesn't have anything to do with drjl cut here.

hrjheng:
Thanks for the clarification. I think having a "deltaR cut between lepton and QCD jet in matrix element level" is still ok for our analysis as we need a hard jet, which would fake as a photon, being back-to-back to the di-muon system.

sihyunjeon:
https://github.com/cms-sw/genproductions/tree/master/genfragments/ThirteenTeV/Hadronizer

You can add your EM filter there if it's not already in the repository

genfragments/ThirteenTeV/Hadronizer/Hadronizer_TuneCUETP8M1_13TeV_EMEnrichedFilter_LHE_pythia8_cff.py
```diff
@@ -0,0 +1,70 @@
...
+        tkIsoMax=cms.double(5.0),
+        caloIsoMax=cms.double(10.0),
+        requireTrackMatch=cms.bool(False),
+        genParSource=cms.InputTag("genParticlesForFilter"),
+    ),
+)
+
+ProductionFilterSequence = cms.Sequence(
+    generator * (genParticlesForFilter + emenrichingfilter)
+)
```
hrjheng, line 70: Hi @sihyunjeon, I am not sure if this is the correct file for the fragment to be put in this directory.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/MuMuJet_mll_0to60_LO/MuMuJet_mll_0to60_LO_run_card.dat
```diff
@@ -0,0 +1,271 @@
...
+ 0.4  =  dparameter 
+#*********************************************************************
+# maximal pdg code for quark to be considered as a light jet         *
+# (otherwise b cuts are applied)                                     *
+#*********************************************************************
+ 5 = maxjetflavor    ! Maximum jet pdg code
+#*********************************************************************
+# Jet measure cuts                                                   *
+#*********************************************************************
+ 10   = xqcut   ! minimum kt jet measure between partons
```
agrohsje, line 260: can you set to 0 to be in sync with ickkw=0

genfragments/ThirteenTeV/Hadronizer/Hadronizer_TuneCUETP8M1_13TeV_EMEnrichedFilter_LHE_pythia8_cff.py
```diff
@@ -0,0 +1,70 @@
...
+                                 'JetMatching:scheme = 1',
+                                 'JetMatching:merge = on',
+                                 'JetMatching:jetAlgorithm = 2',
+                                 'JetMatching:etaJetMax = 5.',
+                                 'JetMatching:coneRadius = 1.',
+                                 'JetMatching:slowJetPower = 1',
+                                 'JetMatching:qCut = 30.',
+                                 'JetMatching:nQmatch = 5',
+                                 'JetMatching:nJetMax = 1',
+                                 'JetMatching:doShowerKt = off',
```
agrohsje, line 37: Why do you use matching? Matching assumes that you get +1 jet contributions from 0 jets @ ME + PS. However, this is not produced.

hrjheng:
Hi @agrohsje and @sihyunjeon, do you have further concerns and comments that need clarifications?

sihyunjeon:
For Ultralegacy, you should use CP5 for showering tunes but what you have is CUETP. Please change this

hrjheng:
The file for hadronizer (Hadronizer_TuneCP5_13TeV_EMEnrichedFilter_LHE_pythia8_cff.py) is updated to use CP5 for showering tunes.

sihyunjeon:
This is fine for me @agrohsje

agrohsje:
I see that Sihyun asked above it but let me double-check. drjl = 0.2 would not work for you? 0.4 seems tight to me. What is your jet algorithm? Similar eta parton 3.0 could be loosened to say 3.5 to be sure.

hrjheng:
I still do not see any reason why the cut value of drjl would be a serious problem here. The cut values are chosen according to the analysis selections. First, we want the photon (in the DY+jet sample, the jet fakes as a photon) to be angularly separated from the muons. Second, we only select photon within |eta| < 2.5, and therefore a cut at 3.0 was set in producing the events.

The point is not to make a perfect sample at this stage. What's important is we need a background sample **as a start** to compare the data. In previous iterations of analysis, we do not have such a sample. Once we have a sample, we then compare it to the data and see if it's reasonable or not.

sihyunjeon:
You mean "analysis selections" as in RECO level right? One possible concern is that there could be some smearing effects (like, your muon pT @ GEN level = 14 GeV -> RECO level = 16 GeV; but if you give GEN ptl cut to 15, you won't get such events). And since you probably will be using AK4 jets with RelIso04, separated with dR=0.4, this value is just at the boundary.
So we would prefer to take a safer path instead of risking it with tight values. IF indeed you find it's not reasonable, you should then resubmit your samples for your analysis which would be waste of computing, time, efforts from the MC contacts

hrjheng:
the drjl cut is modified to be 0.2.
the |eta| cut on the parton is not changed because we want the cross-section as small as possible

sihyunjeon:
Makes sense to me now @agrohsje

bin/MadGraph5_aMCatNLO/cards/production/13TeV/MuMuJet_mll_0to60_LO/MuMuJet_mll_0to60_LO_run_card.dat
```diff
@@ -0,0 +1,271 @@
...
+#*************************************************************
+   0  = nhel          ! Number of helicities used per event
+#*******************                                                 
+# Standard Cuts
+#*******************                                                 
+#                                                                    
+#*********************************************************************
+# Minimum and maximum pt's (for max, -1 means no cut)                *
+#*********************************************************************
+ 20  = ptj       ! minimum pt for the jets
```
agrohsje, line 108: Why do you want to apply such a hard cut?

bin/MadGraph5_aMCatNLO/cards/production/13TeV/MuMuJet_mll_0to60_LO/MuMuJet_mll_0to60_LO_run_card.dat
```diff
@@ -0,0 +1,271 @@
...
+# Inclusive cuts                                                     *
+#*********************************************************************
+ 0  = xptj ! minimum pt for at least one jet  
+ 0  = xptb ! minimum pt for at least one b 
+ 0  = xpta ! minimum pt for at least one photon 
+ 0  = xptl ! minimum pt for at least one charged lepton 
+#*********************************************************************
+# Control the pt's of the jets sorted by pt                          *
+#*********************************************************************
+20   = ptj1min ! minimum pt for the leading jet in pt
```
agrohsje, line 199: Same here. Btw. why needed. Why not turning off here and just use above setting.

hrjheng:
I've already answered the same question from Sihyun here https://github.com/cms-sw/genproductions/pull/2855#pullrequestreview-611780516. We will apply a cut at 30GeV or higher in our analysis, so low pT jets will not be selected in our selection requirements and hence producing them is just a waste of resources and time.
Since we would like to have a higher cut than what we put in the card (or at ME level), it's safe to use a 20GeV cut for partons in the sample.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/MuMuJet_mll_0to60_LO/MuMuJet_mll_0to60_LO_run_card.dat
hrjheng: will this cause any problem?

agrohsje: no. just warning as it means you have to change twice. let me merge.

PR #2857, opened 2021-03-15
Fix customizecards compute widths

sihyunjeon:


sihyunjeon:
@agrohsje Sorry these lines should never have been there. With this line it "returns 1" and exits make_gridpack () function and doesn't run properly.

PR #2858, opened 2021-03-18
Adding back some 2016 cards for 3 types of diboson resonances.

sregnard:
Adding back some cards to produce gridpacks for additional 2016 samples (RunIISummer16MiniAODv3) for 3 types of high-mass diboson resonances (GBulk->WlepWhad, W'->WlepZhad, and W'->Wlephbb) used in the B2G-19-002 analysis. 
(Only the mass points from 0.6 TeV to 4.5 TeV had been produced in the past, we will now produce the 5 TeV to 8 TeV points.)

The cards were apparently not present anymore in the mg242legacy branch, not even in the pre2017/ folder, but I found them at the pre2017 tag in the 13TeV/exo_diboson/ folder. I added the extramodels.dat files, and for each of the 3 models I only kept one set of example cards together with a script that changes the resonance mass. I didn't make any other change. Can you please confirm that these pre2017 cards have the proper setup for the production of 2016 samples?


sihyunjeon:
MGv242 is seriously outdated. Is there any reason to use this?

sregnard:
I am just trying to stay close to the configuration that was used for the already existing 2016 samples (other mass points), which I think were generated in v2.2.2. Should I do things differently?

agrohsje:
Are you saying you are working on a 2016 search and need more signal samples for this? This will not go to UL? Are you sure you still get that approved?

sregnard:
This a full Run2 search (B2G-19-002) that just got approved and doesn't use UL. Yes, the conveners asked us to request these missing high-mass samples so that a future combination analysis can reach at least 5 TeV.

sregnard:
Hi all, may I ask again if these cards are fine, and if you need any other information from me?

PR #2859, opened 2021-03-19
runcards for Hgg interference

echapon:
Run cards for studying the Higgs boson width through interference between signal and the QCD continuum (see also [this presentation](https://indico.cern.ch/event/973661/#3-a-first-look-at-ggh-interfer) at the GEN meeting in November -- run cards have since been modified following [this thread](https://hypernews.cern.ch/HyperNews/CMS/get/generators/4960.html) ). One request has already been [validated](https://cms-pdmv.cern.ch/mcm/requests?prepid=HIG-RunIISummer20UL16GEN-00018&page=0&shown=127) by Chu Wang in McM. Corresponding Sherpacks are also available at `/afs/cern.ch/work/e/echapon/public/HGG_interference/`.

Please let me know if the run cards would better be named differently or put in another place, or if you have any other comment.


sihyunjeon:
If this is for Sherpack,
Why do we need to have Rivet routine, and PrintGraphs functionality included in the cards?

sihyunjeon:
And can you removed the unused/unnecessary lines in cards?
HIM and HIO descriptions are fine for future references but lines like gamma pt selectors doesn't seem to be used at all

sihyunjeon:
Is this sample going to be used for real analysis, and if it does, do you not need :
1) muF and muR variations
2) PDF error set variations
3) underlying event effects
4) hadronizations

Or would this only be used to look at GEN level distributions?

echapon:
I think you are right. I removed the unncessary lines and PrintGraphs commands.

echapon:
Yes this is for real analysis. Would you have a pointer to how to include all these relevant variations into the run cards?

sihyunjeon:
Sherpa2.2 should be tuned with NNPDF30 (or possibly lower), definitely not NNPDF31+NNLO.
1) It would be better to use matching NNPDF versions
2) Using NNLO PDF for LO or NLO hard process has never been validated (for Pythia, we have CPtunes paper). So I don't think this could be justified.

sihyunjeon:
EVENT_GENERATION_MODE W;

This will give you weighted events which can sometimes cause ill behaviours. If there was no good reason for this, suggest to change it to unweighted option for equal weights or partially unweighted option instead.

sihyunjeon:
VARIATIONS 4.0,4.0 1.0,1.0 0.25,0.25 4.0,1.0, 1.0,4.0 0.25,1.0 1.0,0.25;
PDF_SET NNPDF31_nnlo_hessian_pdfas[all];

These two are needed

echapon:
About the PDF: sure, sorry for my ignorance. I'll use whatever PDF is recommended / CMS standard for our case. I can update to NNPDF30 NLO (or LO for qg interference?).

About the weights: I think we do need this. This is because the interference has this odd shape (see e.g. https://indico.cern.ch/event/973661/contributions/4108292/attachments/2148321/3621624/chapon_GEN_Hgg_20201123.pdf#page=9).

Thank you for the variations. I will add them once the PDF to be used is confirmed, I will discuss with Hgg MC contacts.

sihyunjeon:
Sorry from my side, I actually don't see at first sight what the issue is in the slides... Can you tell me what is odd?

echapon:
What I meant is "odd" as opposed to "even", with respect to MH. If you look at the bottom left plot (mass distribution for weighted events), it goes positive and then... negative. This is because this sample is interference-only, naturally. We cannot get such a locally negative cross section without weights.

sihyunjeon:
@echapon Let me ask one more question on this weighted/unweigthed events, just want to make sure since things are not understandable for me here. It says on the y-axis "Entries" for the upper unweighted plots. Did you also take into account the negative weights when you plotted the ones above?
What makes me confusing is, either you do unweighted or weighted for event generations, the kinematic distributions should be the same ==> physics should be the same. From what I see, your eta(photon) for instance, is peaked around the barrel when event is unweighted but flat with weighted events? Which is a clear indication that physics is treated differently.

echapon:
In the plots on a given slide (e.g. slide 9), the top and bottom row correspond to the exact same events, except that weights were applied only when filling the bottom histograms, while top plots really just correspond to raw unweighted events (not checking whether events have a positive or negative weight). This is why distributions are so different. Unweighted histograms don't really have a physical meaning, they are there only for illustration.

sihyunjeon:
@echapon 
From now what I understand, is that you did not draw histograms with "unweighted events", what you actually did is "Unweighted entries with weighted events". "Unweighted entries with weighted events" is not same with "unweighted events". I think there's some confusion with "unweighted" and "weighted" events. Unweighted events only means that each event will receive equal weights while weighted events will receive diversely different weights. With the unweighted events, especially when you use MC@NLO method for NLO matchings, you would still have -1 event weights and you should necessarily take this into account when you draw histograms.
In other words, you should be able to get what you want if you draw your histograms with "unweighted events with weights".

echapon:
OK I'm giving it a try in "partially unweighted" mode.

@GurpreetSinghChahal since you know about these requests, any opinion about what we have been discussing above? (e.g. about the PDF order to use, also about the weighting mode)

echapon:
@sihyunjeon I have tested the "partially unweighted" mode, it seems like it is giving the same results as the "weighted" mode so we can use it if this is preferred. Also, I have added the variations. I am only not sure which order to use for PDFs. Perhaps we can add NNPDF3.0 LO and NLO as PDF variations?

sihyunjeon:
Thanks for the check, the partial setting is okay, would be a bit more efficient in the sense of computations.
For LO hard process, LO PDF is correct and for NLO->NLO in physics terms.

You can store the PDF variations if you are not too sure.
But at least I'd suggest you to match the hard process and PDF set orders for the nominal productions

bin/Sherpa/cards/production/13TeV/higgs/Run.dat_13TeV_gamgam_forint_interference
```diff
@@ -0,0 +1,46 @@
+ (run){
+ EVENTS 100000;
+ EVENT_GENERATION_MODE P;
+ ME_SIGNAL_GENERATOR Amegic Higgs;
+ RANDOM_SEED XX; XX:=0; WH:=0.00407;
+ ANALYSIS Rivet; ANALYSIS_OUTPUT results/Analysis_XX_WH_QSF; ANALYSIS_WRITEOUT_INTERVAL 0.1;
```
sihyunjeon, line 6: Please remove this line,
For the sherpacks I don't think Rivet analysis output is necessary

echapon:
The remaining line about Rivet in the cards is now removed. Also, I switched to NNPDF3.0 NLO everywhere, also for qg interference since, even if it is labelled "LO", it is meant to be analysed together with the other processes so I believe it would be better to use the same PDF everywhere, so that we have a meaningful signal+interference set of events in the end with all 3 sub-processes together.

GurpreetSinghChahal:
Thanks @sihyunjeon. Sherpa's default pdf is recommended for all processes: NNPDF30_nnlo_as_0118 Even if the process is only LO, the rapidity information will be better. @echapon Please update NNPDF30_nnlo_as_0118 for all 

Secondly, YUKAWA[6] 164 was used in the original cards given by authors and you're using YUKAWA[6] 172
Perhaps you're right but can you please remind me the reason?

echapon:
Thank you @GurpreetSinghChahal . About the YUKAWA parameter for the top quark, I thought you were the one forwarding the run cards to Rajdeep. So maybe you know the motivation for it, or perhaps who to ask?

Once this is settled I will update all cards back to NNPDF3.0 NNLO.

GurpreetSinghChahal:
Thanks! I have sent him [these runcards](https://www.dropbox.com/s/8fzy19ibp77n7fs/Re__Higgs_interference_MC%40NLO_cards_in_2.2.x.zip?dl=0) that are using YUKAWA[6] 164 . Please confirm with him.

echapon:
ok yes indeed, but was there a reason why you set YUKAWA[6] 164 there?

echapon:
@GurpreetSinghChahal I am confused, Rajdeep does not know why this parameter was set, since you were the one to send the runcards.

rchatter:
> Thanks! I have sent him [these runcards](https://www.dropbox.com/s/8fzy19ibp77n7fs/Re__Higgs_interference_MC%40NLO_cards_in_2.2.x.zip?dl=0) that are using YUKAWA[6] 164 . Please confirm with him.

Dear Gurpreet,
Just sent you mail in the thread where we had originally discussed and exchanged these Run cards.
Thanks,
Rajdeep

GurpreetSinghChahal:
Hi Rajdeep, thanks! This value was suggested in sherpa author's run card that I have sent you. But you decided to use different value, i.e. YUKAWA[6] 172 along with RUN_MASS_BELOW_POLE. I expect  RUN_MASS_BELOW_POLE should require YUKAWA[6] 164 instead of YUKAWA[6] 172. However, I am clarifying the affect of these conditions on your process. Sorry, things were slow during the break. I'll come back to you soon with the updates.

echapon:
hi Gurpreet, Rajdeep, thank you for the clarification. I don't mind using YUKAWA[6] 164 of course as far as I understand why this is needed. Now I see that this is needed because of RUN_MASS_BELOW_POLE (which I guess is found to be needed for modelling the interference), then I'll switch back to YUKAWA[6] 164.

echapon:
hi @rchatter @GurpreetSinghChahal @sihyunjeon I updated the run cards back to YUKAWA[6] 164 and NNPDF3.0 NNLO. I believe with this we should be all set. Let me know if anything else needs to be fixed.

rchatter:
> hi @rchatter @GurpreetSinghChahal @sihyunjeon I updated the run cards back to YUKAWA[6] 164 and NNPDF3.0 NNLO. I believe with this we should be all set. Let me know if anything else needs to be fixed.

Thanks @echapon . This sounds good to me as these were the settings in the original cards from @GurpreetSinghChahal .

bin/Sherpa/cards/production/13TeV/higgs/Run.dat_13TeV_gamgam_forint_interference
```diff
@@ -0,0 +1,45 @@
+ (run){
+ EVENTS 100000;
+ EVENT_GENERATION_MODE P;
+ ME_SIGNAL_GENERATOR Amegic Higgs;
+ RANDOM_SEED XX; XX:=0; WH:=0.00407;
+ QSF:=1.0; SCALES METS{MU_F2}{MU_R2}{QSF*MU_Q2}; CORE_SCALE VAR{Abs2(p[2]+p[3])};
+ HEPMC_USE_NAMED_WEIGHTS=1;
+ SCALE_VARIATIONS 4.0,4.0 1.0,1.0 0.25,0.25 4.0,1.0 1.0,4.0 0.25,1.0 1.0,0.25;
```
sihyunjeon, line 8: One last comment, is SCALE_VARIATIONS 1,1 needed? This would be same with your nominal one.

echapon: I am not sure, I left it there because I found it in all other Sherpa run cards in this repository (e.g. [DY](https://github.com/cms-sw/genproductions/blob/master/bin/Sherpa/cards/production/13TeV/DY_ew/Run.dat_DYee#L8), [exo GJets](https://github.com/cms-sw/genproductions/blob/master/bin/Sherpa/cards/production/13TeV/exo/GJets_Pt-100To200_13TeV-sherpa_cff.py#L30), [VBS](https://github.com/cms-sw/genproductions/blob/master/bin/Sherpa/cards/production/2017/13TeV/VBS/Run.dat_ZA#L8)...).

sihyunjeon: I believe this would be redundant but still fine to be in there.
I think the card is fine @agrohsje @GurpreetSinghChahal

PR #2861, opened 2021-03-22
Update EWK W proc cards to account for non-resonant l+jets final states

alpakpinar:
Hello, for the EWK W cards merged in #2834, we received some comments saying that we're not taking non-resonant lepton+jets final states into account, one such contributing diagram is shown in [1]. Looking at XS before and after, this seems to be a small impact, around 1% or so, but we were advised to update the proc cards as in this PR. We wanted to make this PR and ask for your opinion, if you think these look good, is it possible to merge these cards? Thanks a lot!  

[1]: http://cms-results.web.cern.ch/cms-results/public-results/publications/SMP-17-011/CMS-SMP-17-011_Figure_001-c.png


sihyunjeon:
The cards look fine to me @agrohsje

PR #2867, opened 2021-03-30
XS calculation script for girdpack->NanoGEN

menglu21:
This script is only for calculating XS for gridpack->NanoGEN, based on calculateXSectionAndFilterEfficiency.sh


agrohsje:
Thanks for the commit!

PR #2870, opened 2021-03-31
tW 5F NLO DR inclusive MadGraph5 sample cards for gridpack generation

vrbouza:
In this PR I add cards for generate tW 5F NLO inclusive sample gridpack with the diagram removal (DR) scheme for MadGraph5, following @agrohsje suggestion on Mattermost.

The cards are based on the analogous for the tWZ already present in the same folder. The only modifications are to the process themselves, the name in the process card, and the madspin decay information. They have been tested locally and a gridpack has been successfully produced.

We have, nevertheless, a question on whether the way we generalised the cards for all decay channels is the best one (changing in the madspin file the decays to "all" in all cases): is it really? Should we forget about using madspin?

@epalencia

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/ST_tW_5f_NLO_inclusive_DR/ST_tW_5f_NLO_inclusive_DR_run_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/ST_tW_5f_NLO_inclusive_DR/ST_tW_5f_NLO_inclusive_DR_run_card.dat
@@ -1,0 +1,188 @@
+#***********************************************************************
+#                        MadGraph5_aMC@NLO                             *
+#                                                                      *
+#                      run_card.dat aMC@NLO                            *
+#                                                                      *
+#  This file is used to set the parameters of the run.                 *
+#                                                                      *
+#  Some notation/conventions:                                          *
+#                                                                      *
+#   Lines starting with a hash (#) are info or comments                *
+#                                                                      *
+#   mind the format:   value    = variable     ! comment               *
+#                                                                      *
+#   Some of the values of variables can be list. These can either be   *
+#   comma or space separated.                                          *
+#                                                                      *
+#   To display additional parameter, you can use the command:          *
+#      update to_full                                                  *
+#***********************************************************************
+#
+#*******************                                                 
+# Running parameters
+#*******************                                                 
+#
+#***********************************************************************
+# Tag name for the run (one word)                                      *
+#***********************************************************************
+    tag_1   = run_tag ! name of the run
+#***********************************************************************
+# Number of LHE events (and their normalization) and the required      *
+# (relative) accuracy on the Xsec.                                     *
+# These values are ignored for fixed order runs                        *
+#***********************************************************************
+    10000   = nevents   ! Number of unweighted events requested
+    -1.0    = req_acc   ! Required accuracy (-1=auto determined from nevents)
+    -1      = nevt_job  ! Max number of events per job in event generation.
+                        !  (-1= no split).
+#***********************************************************************
+# Normalize the weights of LHE events such that they sum or average to *
+# the total cross section                                              *
+#***********************************************************************
+    average = event_norm ! valid settings: average, sum, bias
+#***********************************************************************
+# Number of points per itegration channel (ignored for aMC@NLO runs)   *
+#***********************************************************************
+    0.01    = req_acc_fo    ! Required accuracy (-1=ignored, and use the
+                            ! number of points and iter. below)
+# These numbers are ignored except if req_acc_FO is equal to -1
+    5000    = npoints_fo_grid   ! number of points to setup grids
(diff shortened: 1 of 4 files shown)


bin/MadGraph5_aMCatNLO/cards/production/13TeV/ST_tW_5f_NLO_inclusive_DR/ST_tW_5f_NLO_inclusive_DR_run_card.dat
```diff
@@ -0,0 +1,188 @@
...
+#***********************************************************************
+# Tag name for the run (one word)                                      *
+#***********************************************************************
+    tag_1   = run_tag ! name of the run
+#***********************************************************************
+# Number of LHE events (and their normalization) and the required      *
+# (relative) accuracy on the Xsec.                                     *
+# These values are ignored for fixed order runs                        *
+#***********************************************************************
+    10000   = nevents   ! Number of unweighted events requested
```
sihyunjeon, line 34: reduce this to 2-3000, this would be enough for pilot run

vrbouza: Done

sihyunjeon:
Sorry for my ignorance on this madstr. What's the motivation for removing several diagrams?

agrohsje:
Hi @sihyunjeon . At NLO there are contribution from bWt, which overlap with the resonant bw production from ttbar.This overlap can be removed with this new plugin. I introduced it in the context of the t-channel DM models where the same problem of overlapping resonant and non-resonant production appears. 
Let  me know. I am ok with merging the PR.

sihyunjeon:
image

@agrohsje Thanks for the explanations. One more question (just for my lesson, I am fine with the cards). So I did "p p >  t w [QCD]" and this will remove the diagrams like above if I plugin MadSTR=1 right? So with this option does it search for onshell resonances formed by "t+j" or "w+j" where j is the jet from NLO calculations? And thus removes "w+j(=b) = t" onshell diagrams?

sihyunjeon:
merging this in advance

vrbouza:
Hi @agrohsje and @sihyunjeon , thanks for merging the PR.

However, I would like to re-ask again whether the decays specifications of these cards are correct, or not: should we be using no madspin card? Or should we use what is in the cards (a madspin file, but decaying to all possible final states)? Is there a difference (in this case)?

Regarding the problem of overlapping, @sihyunjeon , you can find more information on the details for this case (and similars) e.g. in https://arxiv.org/abs/1607.05862 (although it is dedicated for tWH/ttH, it explains the issue quite well).

sihyunjeon:
Hi, without madspin, your t/W decays are going to be handled with Pythia. Now for your case, since you inject the MadSpin card, your t/W will be decayed using MadSpin. And from what I see in your MadSpin card, I don't see anything wrong.

But one minor concern is the BR of Ws. Since W->qq consumes up most of the BRs, your gridpack/samples will at the end consist of a lot of fully hadronic process events "p p > t W > b j j j j".  Is this still fine for you? Otherwise I will try splitting the gridpacks by modifying the MadSpin card leptonic/semileptonic/hadronic.

And thanks for the nice reference :)

vrbouza:
Hi @sihyunjeon and apologies for answering so late. Thanks for the explanation (although I still don't know what would be the difference between using MadSpin and purely MadGraph, using the syntaxis for the decays directly in the proc_card file).

In any case, and following your suggestion, I am going to open a new PR with the cards for tW NLO 5F with the DR, DR2 (DR+interf.), DS, and the other variations of DS (reshuffling with a running Breit-Wigner, reshuffling only in initial state particles for the artificial term, and in the IS particles and with a running BW) for both the inclusive and dileptonic case (we know for sure at least one current developing analysis that would be suitable for it: actually, my analysis).

Cheers!

sihyunjeon:
> Hi @sihyunjeon and apologies for answering so late. Thanks for the explanation (although I still don't know what would be the difference between using MadSpin and purely MadGraph, using the syntaxis for the decays directly in the proc_card file).
> 
> In any case, and following your suggestion, I am going to open a new PR with the cards for tW NLO 5F with the DR, DR2 (DR+interf.), DS, and the other variations of DS (reshuffling with a running Breit-Wigner, reshuffling only in initial state particles for the artificial term, and in the IS particles and with a running BW) for both the inclusive and dileptonic case (we know for sure at least one current developing analysis that would be suitable for it: actually, my analysis).
> 
> Cheers!

MadSpin will work only when narrow width approximation can be safely applied. If you add additional decay chains in the proc_card, it would be safer in that sense. But for your process, since mass gaps between top and W or b, and between W and it's decay particles are pretty big enough, either you use MadSpin to decay things or use pure MadGraph, physics won't change.

vrbouza:
I see, I understand now, thanks for the explanation! I already created the PR that I mentioned before, it's #2892 .

PR #2872, opened 2021-04-02
Adding ZZTo4l0To3jets

welmeten:


bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZZTo4l0To3jets/ZZNJets_Zll_VBF_LO_proc_card.dat
```diff
@@ -0,0 +1,14 @@
...
+define l- = e- mu- ta-
+define j = p
+# Specify process(es) to run
+generate p p > z z @1
+add process p p > z z j @2
+add process p p > z z j j @3
+add process p p > z z j j j @4
+#add process p p > z z j j j j @5
+# Output processes to MadEvent directory
+output ZZNJets_Zll_VBF_LO  -nojpeg
```
sihyunjeon, line 14: Please remove "VBF" in the output name, this could be misleading

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZZTo4l0To3jets/ZZNJets_Zll_VBF_LO_run_card.dat
```diff
@@ -0,0 +1,269 @@
...
+#*********************************************************************
+# Beam polarization from -100 (left-handed) to 100 (right-handed)    *
+#*********************************************************************
+        0     = polbeam1 ! beam polarization for beam 1
+        0     = polbeam2 ! beam polarization for beam 2
+#*********************************************************************
+# PDF CHOICE: this automatically fixes also alpha_s and its evol.    *
+#*********************************************************************
+ lhapdf = pdlabel ! PDF set
+ 316200 = lhaid
```
sihyunjeon, line 52: Is there a reason to use this pdf instead of the standard/recommended pdf?

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZZTo4l0To3jets/ZZNJets_Zll_VBF_LO_run_card.dat
```diff
@@ -0,0 +1,269 @@
...
+ 1        = scalefact        ! scale factor for event-by-event scales
+#*********************************************************************
+# Matching - Warning! ickkw > 1 is still beta
+#*********************************************************************
+ 1        = ickkw            ! 0 no matching, 1 MLM, 2 CKKW matching
+ 1        = highestmult      ! for ickkw=2, highest mult group
+ 1        = ktscheme         ! for ickkw=1, 1 Durham kT, 2 Pythia pTE
+ 1        = alpsfact         ! scale factor for QCD emission vx
+ F        = chcluster        ! cluster only according to channel diag
+ T        = pdfwgt           ! for ickkw=1, perform pdf reweighting
```
sihyunjeon, line 71: set this to F, setting this to T sometimes give not understandable distributions

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZZTo4l0To3jets/ZZNJets_Zll_VBF_LO_run_card.dat
```diff
@@ -0,0 +1,269 @@
...
+# Automatic ptj and mjj cuts if xqcut > 0
+# (turn off for VBF and single top processes)
+#**********************************************************
+   T  = auto_ptj_mjj  ! Automatic setting of ptj and mjj
+#**********************************************************
+#                                                                    
+#**********************************
+# BW cutoff (M+/-bwcutoff*Gamma)
+#**********************************
+  10  = bwcutoff      ! (M+/-bwcutoff*Gamma)
```
sihyunjeon, line 86: default value is 15 I suppose, no? If yes, is there a reason to tighten this?

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZZTo4l0To3jets/ZZNJets_Zll_VBF_LO_run_card.dat
```diff
@@ -0,0 +1,269 @@
...
+ -1  = dralmax ! maxdistance between gamma and lepton
+#*********************************************************************
+# Minimum and maximum invariant mass for pairs                       *
+# WARNING: for four lepton final state mmll cut require to have      *
+#          different lepton masses for each flavor!                  *           
+#*********************************************************************
+ 0   = mmjj    ! min invariant mass of a jet pair 
+ 0   = mmbb    ! min invariant mass of a b pair 
+ 0   = mmaa    ! min invariant mass of gamma gamma pair
+ 50   = mmll    ! min invariant mass of l+l- (same flavour) lepton pair
```
sihyunjeon, line 172: Are you aware that this won't be implemented? MadSpin (where Z > l l decay is handled) doesn't read the implemented cuts so this won't do anything

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZZTo4l0To3jets/ZZNJets_Zll_VBF_LO_proc_card.dat
```diff
@@ -0,0 +1,14 @@
...
+
+define p = u c s d b u~ c~ s~ d~ b~ g
+define l+ = e+ mu+ ta+
+define l- = e- mu- ta-
+define j = p
+# Specify process(es) to run
+generate p p > z z @1
+add process p p > z z j @2
+add process p p > z z j j @3
+add process p p > z z j j j @4
```
sihyunjeon, line 11: If you produce Z in ME level, it would give you to onshell Zs that would have very limited phase space. For your analysis, in RECO level, do you cut on mll just to be around the Z peak for both Zs? What you would have with this card will be two Zs that looks like slide12 top right in [1].

[1] https://indico.cern.ch/event/234296/contributions/1547084/attachments/391432/544354/13_02_16_tutomg.pdf

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZZTo4l0To3jets/ZZNJets_Zll_VBF_LO_run_card.dat
prebello: @sihyunjeon  is pdfwgt disable now? in the beginning of 2020 I was not aware of it.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZZTo4l0To3jets/ZZNJets_Zll_VBF_LO_run_card.dat
prebello: @sihyunjeon  it was a suggestion from PdmV group in the beginning of 2020. This was the most recent PDF that time. Remembering these cards were used already in gridpack.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZZTo4l0To3jets/ZZNJets_Zll_VBF_LO_proc_card.dat
prebello: hi @sihyunjeon, these cards are used in gridpacks for the Higgs to ZZ VBF differential XS analysis. Indeed, the VBF topology assumes ZZ+2jets only. Is this the only reason to remove the VBF from the name? (in this case gridpacks have to be reproduced only because of the name)

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZZTo4l0To3jets/ZZNJets_Zll_VBF_LO_run_card.dat
prebello: @sihyunjeon  I used 10   = xqcut = bwcutoff . Is it a problem? I didn't know there is a default value for it. I tried to maintain the cuts as looser as possible.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZZTo4l0To3jets/ZZNJets_Zll_VBF_LO_run_card.dat
prebello: @sihyunjeon  the same cards were used for the "decay chain" process. Therefore the cuts were maintained.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZZTo4l0To3jets/ZZNJets_Zll_VBF_LO_proc_card.dat
prebello: @sihyunjeon  exactly. Nevertheless, the process with one of Z offshell (see bellow, to cover broader phase space) crashed due to computing issues while trying to decay the Z ( It is documented elsewhere in many HN msgs changed that time. These gridpacks were tough to be produced in slc6)

generate p p > z l+ l- @1
add process p p > z l+ l- j @2
add process p p > z l+ l- j j @3
add process p p > z l+ l- j j j @4

prebello:
@sihyunjeon the available gridpacks are stored at /store/group/phys_generator/cvmfs/gridpacks/slc6_amd64_gcc630/13TeV/madgraph/V5_2.6.5/ZZleptonicNJetsVBF_LO

ZZ0to3JetsDecayChain
ZZ0to3JetsMadSpin
ZZ0to4JetsNoDecay
Zll0to3jets_nodecay

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZZTo4l0To3jets/ZZNJets_Zll_VBF_LO_proc_card.dat
sihyunjeon: The gridpacks already have some settings that doesn't really seem OK for me (pdfwgt, lhaid, ...). So I would still ask you to change the name. The gridpack itself does not include VBF physics process, whether it is used for VBF H to ZZ analysis or not. I believe gridpack/sample info should be clear to all CMS people.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZZTo4l0To3jets/ZZNJets_Zll_VBF_LO_run_card.dat
sihyunjeon: I don't really recall the time when 316200 was the recommended PDF set among other NNPDF31 releases... 306000 was for a few years when NNPDF31 first came out and then after some issues on negative weights were found 325300 became the nominal one.

If you do "$DEFAULT_PDF_SETS=lhaid" this will automatically fetch the recommended PDFs from [1].

[1] https://github.com/cms-sw/genproductions/tree/master/MetaData

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZZTo4l0To3jets/ZZNJets_Zll_VBF_LO_run_card.dat
sihyunjeon: There was some issues since Summer/Fall in 2019 (and I still don't think we have understood this yet, @agrohsje can correct me if I am wrong) shown in [1], slide from Andreas Albert and others. The general guideline would be to givet it F.

[1] https://indico.cern.ch/event/855110/contributions/3613939/attachments/1930820/3197907/2019-09-27_vht_265.pdf

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZZTo4l0To3jets/ZZNJets_Zll_VBF_LO_run_card.dat
sihyunjeon: xqcut actually is not related to this cut at all.
This would decide how wide width your onshell Z bosons should have, and if you set this to smaller value, you'll only end up with narrower peaks

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZZTo4l0To3jets/ZZNJets_Zll_VBF_LO_proc_card.dat
sihyunjeon: I think it's because 2->6 decay is too heavy in your case. Anyhow with this syntax, you'll only get the phase space where mlls are very restricted "mll : 80-100". If your analysis is fine with this, it won't be a problem (== if your analysis look for two onshell Z bosons only)

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZZTo4l0To3jets/ZZNJets_Zll_VBF_LO_run_card.dat
sihyunjeon: Ah with MadSpin, this value might not matter at all.

@agrohsje If both Z bosons are produced onshell in MadGraph and then sent to MadSpin, the bwcutoff values in MadGraph will be ignored right?

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZZTo4l0To3jets/ZZNJets_Zll_VBF_LO_run_card.dat
prebello: @sihyunjeon as informed by PdmV, 316200 was used in MC central production for the 2020 campaigns. 
The MetaData only shows 2016 and 2017 cases. Is it not updated to newer versions?

sihyunjeon: Those 2016 and 2017 are just some flags for RunIISummer20 campaign's 2016 and 2017 data taking scenarios. And this is because (if I am correct again) CPX tunes were developed and used since 2017. Basically we have same scenarios in terms of hard process + showering/hadronization for 2017 and 2018 so that's why we don't have 2018 there.

For the RunIISummer20 campaign (==ultra legacy), it is recommended to follow the 2017 scenario (i.e. lhapdf=325300). And 316200 might be used for some cases which I am not aware of, but as far as I am aware, most of the samples follow 325300 recommendation.

For example, you can look at [1], EGM-RunIISummer20UL16wmLHEGEN-00003 at the top which was produced like 1-2 months ago, and as you can see it's **RunIISummer20** campaign for the DY sample. If you look at the "Fragment", you can see that it uses the gridpack in [2] which also uses 325300.

[1] https://cms-pdmv.cern.ch/mcm/requests?prepid=*RunIISummer20*wmLHEGEN*&dataset_name=DY*&page=0&shown=524415
[2] /cvmfs/cms.cern.ch/phys_generator/gridpacks/slc7_amd64_gcc700/13TeV/madgraph/V5_2.6.5/dy_fxfx_incl/dyellell012j_5f_NLO_FXFX_slc7_amd64_gcc700_CMSSW_10_6_19_tarball.tar.xz

prebello: no problem. 325300 is Much better.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZZTo4l0To3jets/ZZNJets_Zll_VBF_LO_proc_card.dat
sznajder: @sihyunjeon and @agrohsje , is it possible to break the generation in two gridpacks , leaving the heavy subprocess with 3 jets multiplicity to be generated separately ?

sihyunjeon: @sznajder It is possible,
So you mean dividing it to two gridpacks?
1) p p > z l+ l- & p p > z l+ l- j  & p p > z l+ l- j j
2) p p > z l+ l- j j j

The merging would be handled purely in Pythia as long as the xqcut and mlm settings are correctly turned on so you don't need to change anything here.

But to synchronise with other physics process's gridpack settings, I think dividing it in to 4 gridpacks (0j, 1j, 2j, 3j) just like other jet binned processes would be better. What do you think?

sznajder: Yes, that's what I meant.  Then I think it's worth a try to generate the gridpacks separately and check if Madgraph can handle p p > z l+ l- j j j and Madspin with Z>l+l- separetely.
By the way , in this case one must have different PYTHIA matching setting for 'JetMatching:nJetMin' and 'JetMatching:nJetMax', , both set to the given jet multiplicity , correct ?

sihyunjeon: @sznajder I totally agree that it's worth a try :)
No the Pythia fragments should also be kept the same. If you attempt to merge 0j, 1j, 2j, 3j samples, all these 4 Pythia fragments should have nJetMax = 3.

sznajder: Yes , that's correct about nJetMax, but what about nJet ?

PYTHIA  manual says:
JetMatching:nJet   (default = -1; minimum = -1)
When JetMatching:exclusive = 2, nJet indicates the minimum number of additional light jets in the incoming process. This value may be set automatically.

Shouldn't it be set to the given gridpack multiplicity, which is the minimum number of jets in each gridpack ? Or Pythia gets this automatically from the LHE information ?

sihyunjeon: As far as I am aware, yes they don't need to be specified (and other jet binned samples don't do it as well).

@agrohsje Please correct me if I am wrong

agrohsje: Right. You just need to specify the maximum number of additional partons in the P8 fragment. Independent of technically splitting gridpacks or not. This njetmax and inclusive/exclusive is e.g. needed for Alpgen production.

welmeten: @sihyunjeon @agrohsje Now I managed to do the process "p p > z l+ l- j j j" decaying with madspin which is the heaviest subprocess of our main process(*). Now I just need to understand how should we merge these different processes which with different XS to obtain our main process? Is the merge should be handled in Pythia, and how? maybe you can give an example config file?

(*)generate p p > z l+ l- @1
add process p p > z l+ l- j @2
add process p p > z l+ l- j j @3
add process p p > z l+ l- j j j @4

sihyunjeon: https://github.com/cms-sw/genproductions/blob/master/genfragments/ThirteenTeV/Hadronizer/Hadronizer_TuneCP5_13TeV_MLM_5f_max1j_LHE_pythia8_cff.py

You should change the nJetMax to 3. But you should do some careful studies on xqcut/qcut, checking the DJR plots.

welmeten: @sihyunjeon , sorry I don't ask now about the qcut, let me rewrite the question,
I mean now I generated the processes (*) as we agreed before, each one in a separate gridpack.
what I need to understand now,  at which step should we merge these processes in one process?  Is this should be handled in Pythia and how?
(*)
generate p p > z l+ l- @1
generate p p > z l+ l- j@1
generate p p > z l+ l- jj@1
generate p p > z l+ l- jjj@1

sihyunjeon: You mean you want to merge below?
X > Y
X > Y j
X > Y j j
X > Y j j j

You have to have separate datasets with same Pythia fragments.
ZZto4Lep_0jet
ZZto4Lep_1jet
ZZto4Lep_2jet
ZZto4Lep_3jet

You cannot merge separately generated gridpacks and combine them into one dataset (actually there is, but it is highly depreciated).

sihyunjeon:
Hi
https://answers.launchpad.net/mg5amcnlo/+question/696360
I have overlooked this issue that when doing "p p > z l+ l-", it could be problematic.

In RECO level (==your analysis), if you give strict cuts on mll to catch both of the Z bosons (mll ~ mZ for both pairs), "p p > z z (+@ j)" would be okay. If not, (== if you give strict Z mass cut only on one of the pairs and, say, give mll : 4 ~ 120 GeV to the other) you should use "p p > l+ l- l+ l- (+@ j)".

sznajder:
> Hi
> https://answers.launchpad.net/mg5amcnlo/+question/696360
> I have overlooked this issue that when doing "p p > z l+ l-", it could be problematic.
> 
> In RECO level (==your analysis), if you give strict cuts on mll to catch both of the Z bosons (mll ~ mZ for both pairs), "p p > z z (+@ j)" would be okay. If not, (== if you give strict Z mass cut only on one of the pairs and, say, give mll : 4 ~ 120 GeV to the other) you should use "p p > l+ l- l+ l- (+@ j)".

Hi @sihyunjeon,
There's a new post concerning this issue:
 https://answers.launchpad.net/mg5amcnlo/+question/696694
As in the analysis we impose a the 118 < m4l < 130GeV cut to define our signal region, it excludes the double resonant Z background contribution. I think we should run a test to compare p p > l+ l- l+ l-  and p p > z l+ l- , z > l+ l- cross sections and distributions in the signal region...

agrohsje:
May I ask: What is the plan. Different gridpacks for different jet partons? There is no problem with this approach. Will someone update the PR if this is the case.

sznajder:
Hi Alexander,
We have been discussing the problem with Olivier and although in the past he suggested himself to use p p > Z l+ l- , Z > l+l- syntax , now he is claiming this might lead to nonsense due to an ambiguity with the symmetry factor:

https://answers.launchpad.net/mg5amcnlo/+question/696694 <https://answers.launchpad.net/mg5amcnlo/+question/696694>

https://answers.launchpad.net/mg5amcnlo/+question/696742 <https://answers.launchpad.net/mg5amcnlo/+question/696742>

We are still studying the problem. I’m generating the samples: p p > l+l-l+l- and p p > Z l+l- , Z>l+l- to compare the distributions and Xsections after the analysis level cut 118<m4l<130GeV. My guess is that this will kill the double resonating (Z Z onshell) contribution due to the lack of phase space, but we need to check if this enough to make the two above processes equivalent.  
Definitely we need to create a separate gridpack for each jet multiplicity due to the number of final state particles.
Cheers,
Andre

menglu21:
ping @sznajder @welmeten

bbilin:

PR #2892, opened 2021-04-27
Addition of inclusive & dileptonic cards for the tW 5F NLO process with different interference removal schemes

vrbouza:
In this PR I add, following the already approved PR #2870 new cards that mimic those but 
1) also for the dileptonic case and
2) with all possible interference subtraction schemes in the MadStr plugin.

All the cards are almost identical, except for the madspin file (the decay), the proc_card (because of the name) and the integer that sets the interference removal scheme (in the run_card).

@epalencia


sihyunjeon:
Hi, can you split the pull requests containing <10 dat files in one pull request?
44 files in one pull request is a bit difficult to review efficiently.

vrbouza:
Hi @sihyunjeon , I reduced the changes to keep only the inclusive cards. I will add later the dileptonic ones in a different PR. As I mentioned before, all of them (as with the dileptonic cards) are copy one of each other, changing only the name in the proc_card, the index for the madstr plugin in the run_card, and in the case of the dileptonics, the decay particles in the madspin card. All of them are based on the already merged #2870 .

vrbouza:
Good morning @agrohsje @sihyunjeon .

I will do as you say, although I'm not sure to what kind of script are you refering... A script that automatically generates the DR2 and DS folders with their cards from the original DR/DS ones?

agrohsje:
Right. You have 20 files where the majority should be almost 100% identical except 1 parameter. In this case it is easier and less error prone if you just submit a template and a script that copies the template, modifies 1 or 2 parameters and produces the final cards to make the gridpack.  As this is NLO cards are somewhat easier to review compared to LO.

vrbouza:
Ok, I see. I will add this with a commit later today.

vrbouza:
Hi again @agrohsje and @sihyunjeon , you can see now my last commit: I keep the dileptonic "basic" DR & DS, and in both of them, as well as in the inclusive DR & DS, I included a small python script that dynamically generates the other variations (DR2, and the ones from DS). Executing it directly from the folder produces it.

agrohsje:
Thanks @vrbouza . What is the difference for DR vs DS in terms of cards, except the istr. I would not change 1 but make it a parameter 
SUBISTR and then replace this with 1, 5 or whatever you need. Even dilepton and inclusive can be a string 
SUBTOPDECAY and then you use lep nu or all all or whatever you want.

vrbouza:
Hi again @agrohsje . I kept DS & DR because they are the "standard" ones. If you want, I can do as you say (although I don't understand, a parameter to be replaced in the python script I wrote? In other place?), however, I would like to ask you to be precise, because we need urgently this samples for an analysis to be approved in the next month/next two and I was expecting to submit the ticket ASAP.

vrbouza:
To clarify: with "precise" I meant to say that please tell me how you would like exactly the PR to be.

agrohsje:
As all cards are identical and the only difference is the W decay and istr, I would just commit, one template for proc card, one for run card, one for madspin and customize and in each have e.g. . 
  SUBISTR = istr 
  and then replace according to the setup SUBISTR with 1 or 2 or whatever, adjust the name of the output accordingly. 
  Similar for the decay. 
  W > SUBDECAY 
  and then SUBDECAY can be all all or lep nu or whatever. 
  Why don't you produce gridpacks from the existing cards. I mean there is nothing wrong with it. And then you fix this PR with a single script and 4 templates. 
  @sihyunjeon can confirm that he is ok with the current cards and you can go ahead with gridpack production.

sihyunjeon:
The cards are fine.

@vrbouza If you are not 100% sure what Alexander means please skype me sihyun_jeon.
One question (just for my lesson), is there a reason to ignore istr=5?

@agrohsje The cards are fine.

agrohsje:
Thanks @sihyunjeon . @vrbouza are you going to use gridpacks from the current version and then we leave the PR open to further simplify. Let me know. Otherwise I can also just merge.

vrbouza:
Hi @sihyunjeon @agrohsje .

Thanks, I already created the gridpacks and uploaded them to CVMFS (though I noticed I will need to recreate them in SL6 for our campaign of interest... I didn't noticed before). You can leave the PR open, and I will commit whenever I have the time the changes you required.

Regarding your quiestion, @sihyunjeon , istr=5 is not ignored in the current code of the PR, I maintain as "standard" the DR (istr = 1) and DS (istr = 5) methods. The other ones are produced from those cards.

sihyunjeon:
@vrbouza Sorry for getting back to this so late. Your samples are for RunII-prelegacy, the old one, NOT ultra-legacy?

vrbouza:
> @vrbouza Sorry for getting back to this so late. Your samples are for RunII-prelegacy, the old one, NOT ultra-legacy?

Hi @sihyunjeon : for both. We have an analysis in review with the current rereco for which these samples might be interesting for comparison, and we are preparing for the UL (will prepare, actually, as UL is yet missing things) another analysis for which these samples will be also useful.

agrohsje:
Leave this PR open and wait for further simplifications from @vrbouza .

vrbouza:
Hi again @sihyunjeon and @agrohsje . I just simplified the code as you requested (apologies for taking so much time hehe, I forgot this PR).

sihyunjeon:
@agrohsje looks nice and neat

agrohsje:
Thanks for the PR. Will merge.

PR #2893, opened 2021-04-27
Polarized ZZ

susmitamondal29:


bin/MadGraph5_aMCatNLO/cards/examples/Z0Z0_LO/Z0Z0_LO_run_card.dat
```diff
@@ -0,0 +1,283 @@
...
+
+  .false.     = gridpack  !True = setting up the grid pack
+ -1.0	      = time_of_flight ! threshold (in mm) below which the invariant livetime is not written (-1 means not written)
+  average     = event_norm ! average/sum. Normalization of the weight in the LHEF
+#*********************************************************************
+# Number of events and rnd seed                                      *
+# Warning: Do not generate more than 1M events in a single run       *
+# If you want to run Pythia, avoid more than 50k events in a run.    *
+#*********************************************************************
+  10000 = nevents ! Number of unweighted events requested
```
sihyunjeon, line 35: reduce this to 2000. pilot runs don't have to be big that much

susmitamondal29: We kept it to 10000 events because we wanted to test the Gridpacks being generated from these cards.

sihyunjeon: Not sure what you mean... this is just pilot run. Once you get the gridpacks you can get as many events as you want with by using runcmsgrid.sh in the gridpack.

susmitamondal29: Okay, I am making the changes on the run cards now to have 2000 events.

sihyunjeon:
I think other than that cards are fine mostly.
Is model "sm" 5f or 4f in default?

agrohsje:
Is 4FS. This needs to be fixed.

susmitamondal29:
> Is 4FS. This needs to be fixed.

We agree that the sm model specified in the proc card is 4f and we should change it to specify 5f. I have checked cards for unpolarized ZZ production at NLO and model used there is "loop_sm-no-b-mass". We are trying to generate polarized ZZ production at LO. Do you have any suggestions as to which model in MadGraph might be best for this purpose? Is sm model in 5f the best for our purpose?

agrohsje:
You can check the param_card between loop_sm-no-b-mass and the sm-no-b-mass. If they are "in sync", I would use this.

susmitamondal29:
> You can check the param_card between loop_sm-no-b-mass and the sm-no-b-mass. If they are "in sync", I would use this.

Hello, sincerely sorry about the delayed reply. I have checked the parameter cards for the two models loop_sm-no_b_mass and sm-no_b_mass. There are a  few additional parameters in the loop_sm-no-b-mass model and they are as follows:
###################################
## INFORMATION FOR LOOP
###################################
Block loop
    1 9.118800e+01 # MU_R
#===========================================================
# QUANTUM NUMBERS OF NEW STATE(S) (NON SM PDG CODE)
#===========================================================

Block QNUMBERS 82  # gh
        1 0  # 3 times electric charge
        2 1  # number of spin states (2S+1)
        3 8  # colour rep (1: singlet, 3: triplet, 8: octet)
        4 1  # Particle/Antiparticle distinction (0=own anti)
I am having trouble finding the definitions of these parameters. Every other parameter other than the ones mentioned are defined similarly in both models. Could you guide me on the parameter definitions and in light of this information, which model would work best for generating polarized ZZ production at LO?

agrohsje:
Hi @susmitamondal29 . When you run this at LO you basically just have the quark line that radiates the z bosons. 4FS or 5FS isn't much difference in the end, it is mainly a question of cross section. So you can use sm-no_b_mass if you want 5 FS.

sihyunjeon:
any follow ups on this?

susmitamondal29:
I have made the suggested changes.

agrohsje:
I have one follow-up on the cards. You have some cuts. I assume there are all loose enough compared to analysis level cuts (eta_l < 2.6) ? If confirmed: @sihyunjeon  do you mind checking and merging?

susmitamondal29:
In the run cards, eta_l = 2.6,  pt_l = 4. Does that pose a problem?

agrohsje:
You should make sure that any cut in your card is loose enough so that there is no conflict with analysis level cuts. E.g. a parton eta of < 2.6 is not a good choice as the jet eta could be 2.4 in this case, so you remove something that you don't remove at analysis level. for a lepton 2.6 might be ok, but there is also no harm in going to 3.0 if you are in doubt about detector level cuts.

susmitamondal29:
We have checked with a previous ZZ analysis and we are okay with keeping the eta_l at 2.6 in the run card

agrohsje:
Ok. If cuts are not too tight, @sihyunjeon please check and merge. Thanks!

sihyunjeon:
This won't work for MGv265 because of the polarization assigning syntaxes, no? If so, I think this should go in to mg27x branch rather than master

sihyunjeon:
> This won't work for MGv265 because of the polarization assigning syntaxes, no? If so, I think this should go in to mg27x branch rather than master

Maybe @agrohsje @menglu21 might have better answers for this

menglu21:
> > This won't work for MGv265 because of the polarization assigning syntaxes, no? If so, I think this should go in to mg27x branch rather than master
> 
> Maybe @agrohsje @menglu21 might have better answers for this

Hi, 
yes, the polarization syntax is supported from 270. btw, why these cards are put in the example directory, shouldn't it be in the production directory

agrohsje:
indeed. maybe best to push the cards to the other polarization cards in 27x. we will not propagate the info from there, but we keep track of the review.

susmitamondal29:
Could you kindly let me if I need to do anything on my end?

menglu21:
> Could you kindly let me if I need to do anything on my end?

Hi, 
the polarization syntax is only support from MG27, the master branch is MG265, so you may push your cards to branch mg27x, and put them under the production directory rather than the example directory.

menglu21:
could you please close this PR.

susmitamondal29:
Closed PR due to request.

PR #2905, opened 2021-05-06
change dark higgs decay width from auto to a fixed value

y19y19:
This value was set to auto, but I realized that it might cause some unstable in calculation, so I set it to a fixed value. It happens when I ran the gridpack, the decay width of dark higgs to Z Z might be some tiny negative value, and madgraph quits. Since our signal process has nothing to do with dark higgs,  I set it to be a fixed value and it is solved.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/darkphotonDimuon/darkphotonDimuon_DPMass5e-1MeV_DF1e-2/darkphotonDimuon_DPMass5e-1MeV_DF1e-2_customizecards.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/darkphotonDimuon/darkphotonDimuon_DPMass5e-1MeV_DF1e-2/darkphotonDimuon_DPMass5e-1MeV_DF1e-2_customizecards.dat
@@ -2,5 +2,5 @@ set param_card HIDDEN 1 5e-4
 set param_card HIDDEN 2 1e4
 set param_card HIDDEN 3 1e-1
 set param_card HIDDEN 4 1e-10
-set param_card DECAY 35 auto
+set param_card DECAY 35 5.237950
 set param_card DECAY 1023 0


agrohsje:
What is the mass of 35? Is the mass always kept the same. Reason why I am asking the current value looks like a specific choice corresponding to a certain mass. If you change the mass you might pick something that looks like a fixed number and is close to the true so it has no impact.

y19y19:
Yes the mass of the 35 is always the same. It is set to a large value, 10000 GeV. We made such settings because we are not using 35 dark Higgs in our signal. Our study is a pure dark photon study. We only use particle 1023 and the coupling corresponding to it. -------- Original message --------From: agrohsje ***@***.***>Date: Thu, May 6, 2021, 12:22 AMTo: cms-sw/genproductions ***@***.***>Cc: Yao Yao ***@***.***>, Author ***@***.***>Subject: Re: [cms-sw/genproductions] change dark higgs decay width from auto to a fixed value (#2905)
What is the mass of 35? Is the mass always kept the same. Reason why I am asking the current value looks like a specific choice corresponding to a certain mass. If you change the mass you might pick something that looks like a fixed number and is close to the true so it has no impact.

agrohsje:
So this is the mass parameter : set param_card HIDDEN 2 1e4 ? Shouldn't we then just set the width to say 10, so that it is clear that this is a  technical choice rather than corresponding to the exact physics case.

y19y19:
Hi,

I just did a local gridpack generation again with the 10 GeV decay width, and cross section reported from Madgraph changed very tiny, and is within the uncertain range that it provided. So I believe it is fine.

Just to clarify, that the value ~5.23, is the decay width that I got from the madgraph calculation of those stable ones with “auto” decay width setting. For those stable ones, the decay width of 35 are all the same for whatever dark factor that I choose. So I believe 5.23 is a good value to be applied here. If there is any specific reason for using 10 GeV, please let me know.

Thank you.

agrohsje:
Agreed. @sihyunjeon should we merge?

sihyunjeon:
@agrohsje @y19y19 merged

PR #2906, opened 2021-05-06
Adding cards for SMEFT analysis in VH production 

Sumantifr:
We are uploading MadGraph5_aMCatNLO  cards for EFT analysis in WH and ZH production. The UFO model used is SMEFTsim_topU3l_MwScheme, which is uploaded in the central database [1]. 

Once the cards are uploaded, we would be requesting signal samples using the UltraLegacy MC campaign. The gridpacks for WH production are ready and checked. The gridpack production for ZH process is going on at the moment. 

[1] https://cms-project-generators.web.cern.ch/cms-project-generators/SMEFTsim_topU3l_MwScheme_UFO.tar.gz


bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/WH_WToLNu_1j_SMEFTsim_topU3l/WH_WToLNu_1j_SMEFTsim_topU3l_proc_card.dat
```diff
@@ -0,0 +1,24 @@
+set loop_color_flows False
+set gauge unitary
+set max_npoint_for_channel 0
+set group_subprocesses Auto
+set ignore_six_quark_processes False
+set loop_optimized_output True
+set complex_mass_scheme False
```
agrohsje, line 7: All MG5 default, please remove.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/WH_WToLNu_1j_SMEFTsim_topU3l/WH_WToLNu_1j_SMEFTsim_topU3l_proc_card.dat
```diff
@@ -0,0 +1,24 @@
...
+set max_npoint_for_channel 0
+set group_subprocesses Auto
+set ignore_six_quark_processes False
+set loop_optimized_output True
+set complex_mass_scheme False  
+import model SMEFTsim_topU3l_MwScheme_UFO-massless_VH
+define p = g u c d s u~ c~ d~ s~
+define j = g u c d s u~ c~ d~ s~
+define l+ = e+ mu+
+define l- = e- mu-
```
agrohsje, line 12: please remove

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/WH_WToLNu_1j_SMEFTsim_topU3l/WH_WToLNu_1j_SMEFTsim_topU3l_reweight_card.dat
```diff
@@ -0,0 +1,2761 @@
+# reweight card file created with https://github.com/schoef/gridpacks on 2021-04-18 22:34:04
+# Command line arguments: make_reweight_card.py --couplings 2 cHj1 1 cHj3 1 cHu 1 cHd 1 cHudRe 1 cHudIm 1 cHW 1 cHWtil 1 cHB 1 cHBtil 1 cHWB 1 cHWBtil 1 cHbox 1 cHDD 1 cbHRe 1 cbHIm 1 --filename SMEFTsim_VH_reweight_card.dat --overwrite
+# Polynomial order: 2
+change rwgt_dir rwgt
+
+launch --rwgt_name=dummy # Name of first argument seems to be rwgt_1. Add dummy to fix it.
+
```
agrohsje, line 7: Please remove. I fixed this bug already a while ago. No need to have a dummy.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/WH_WToLNu_1j_SMEFTsim_topU3l/WH_WToLNu_1j_SMEFTsim_topU3l_run_card.dat
```diff
@@ -0,0 +1,271 @@
...
+#*********************************************************************
+# Run to generate the grid pack                                      *
+#*********************************************************************
+  .true.     = gridpack  !True = setting up the grid pack
+#*********************************************************************
+# Number of events and rnd seed                                      *
+# Warning: Do not generate more than 1M events in a single run       *
+# If you want to run Pythia, avoid more than 50k events in a run.    *
+#*********************************************************************
+  5000000 = nevents ! Number of unweighted events requested
```
agrohsje, line 32: What is the reason for a large pilot run?

agrohsje:
The comments for WH apply to ZH. On top, why do you need the restrict card. Isn't it part on the ufo. Did you check that coupling 0.0 is not causing trouble? In you restriction card, are there couplings set to 0 that you turn on later? Can you comment more on the validation you did for reweighting?

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/WH_WToLNu_1j_SMEFTsim_topU3l/WH_WToLNu_1j_SMEFTsim_topU3l_proc_card.dat
Sumantifr: Done

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/WH_WToLNu_1j_SMEFTsim_topU3l/WH_WToLNu_1j_SMEFTsim_topU3l_proc_card.dat
Sumantifr: Done

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/WH_WToLNu_1j_SMEFTsim_topU3l/WH_WToLNu_1j_SMEFTsim_topU3l_reweight_card.dat
Sumantifr: Done

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/WH_WToLNu_1j_SMEFTsim_topU3l/WH_WToLNu_1j_SMEFTsim_topU3l_run_card.dat
Sumantifr: Changed to 1.e+5. 

IIRC, we need to produce 1% of total events in the pilot run, so this is the choice.

agrohsje: 1% of the total number of events you are going to produce? why? 
the pilot run will not have any affect on the pre-sampling or precision.

Sumantifr: Okay, then it should not matter. 

I thought this is the number of events used for phase-space sampling in gridpack. But that's not true, I'm fine to reduce it even further if you wish.

Sumantifr:
The restrict card present in the card directories actually doesn't have any role. The restrict card is always read from the UFO file. I've removed it now.

Sumantifr:
To validate reweighting, we produced NanoAODSIM files using FastSIM (Gridpack->AOD->MINAOD->NANOAOD), and checked the weights in LHEReweightingWeight branch. 
However, so far we have done this test by generating the gridpacks using UL2019 branch. Let me check the same with the gridpack generated using the master card. Will confirm you soon.

sihyunjeon:
SMEFT model should not be used with MLM jet matching if I remember correctly.
Is this validated/checked with authors of the model or somehow?

Sumantifr:
> SMEFT model should not be used with MLM jet matching if I remember correctly.
> Is this validated/checked with authors of the model or somehow?

In SMP-20-005 paper draft, I read that the authors have generated Wgamma events with up to 2 additional jets using SMEFTsim model. I can ask Andrew for the link to the cards.

sihyunjeon:
> > SMEFT model should not be used with MLM jet matching if I remember correctly.
> > Is this validated/checked with authors of the model or somehow?
> 
> In SMP-20-005 paper draft, I read that the authors have generated Wgamma events with up to 2 additional jets using SMEFTsim model. I can ask Andrew for the link to the cards.

I think I remember correctly. EFT with MGv27X gives this warning message below. Besides, during the LHC reinterpretation workshop I am pretty sure SMEFT tutorial session tutors mentioned this. But please ask the authors of SMP-20-005, they might have validated this.

CRITICAL: Model with non QCD emission of gluon (found 14 of those).
  This type of model is not fully supported within MG5aMC.
  Restriction on LO dynamical scale and MLM matching/merging can occur for some processes.
  Use such features with care.

Sumantifr:
> > > SMEFT model should not be used with MLM jet matching if I remember correctly.
> > > Is this validated/checked with authors of the model or somehow?
> > 
> > 
> > In SMP-20-005 paper draft, I read that the authors have generated Wgamma events with up to 2 additional jets using SMEFTsim model. I can ask Andrew for the link to the cards.
> 
> I think I remember correctly. EFT with MGv27X gives this warning message below. Besides, during the LHC reinterpretation workshop I am pretty sure SMEFT tutorial session tutors mentioned this. But please ask the authors of SMP-20-005, they might have validated this.
> 
> CRITICAL: Model with non QCD emission of gluon (found 14 of those).
> This type of model is not fully supported within MG5aMC.
> Restriction on LO dynamical scale and MLM matching/merging can occur for some processes.
> Use such features with care.

Thanks pointing to this reference. I have just written to Andrew, will follow up on this once I hear from him.

Sumantifr:
> Hi @agrohsje , I checked that weights are properly stored in the MiniAOD file.
> 
> Do you have any additional check for reweighting in mind?

I see an issue with reweighting at NanoAOD level. 

If I use the same cards in UL2019 and produce gridpacks, reweighting weights are stored properly. But with the gridpack produced with the master branch, I find that LHEReweightingWeight branch is empty. 

To understand it better I tried NanoGEN, so that nothing related to simulation and reconstruction affect. There also I find the same thing. Is there a known issue?

Sumantifr:
> > Hi @agrohsje , I checked that weights are properly stored in the MiniAOD file.
> > Do you have any additional check for reweighting in mind?
> 
> Hi @agrohsje,
> 
> I see an issue with reweighting at NanoAOD level.
> 
> If I use the same cards in UL2019 and produce gridpacks, reweighting weights are stored properly. But with the gridpack produced with the master branch, I find that LHEReweightingWeight branch is empty.
> 
> To understand it better I tried NanoGEN, so that nothing related to simulation and reconstruction affect. There also I find the same thing. Is there a known issue?

Following up on the absence of reweighting weights in NanoAOD, I could trace why these weights are not stored in the NanoAOD file when the gridpack is produced using the master branch, whereas those are stored when the gridpack is generated using the UL2019 branch. 

While comparing the LHE files produced from two gridpacks, I find that the 'weightgroup name' is different between two cases. 

For the CMSSW and MG5@NLO version of the master branch, the corresponding  'weightgroup name' is 'mg_reweighting' weight_name_strategy='includeIdInWeightName'. 

For the CMSSW and MG5@NLO version of the master branch, the corresponding  'weightgroup name' is only 'mg_reweighting'. 

The module used to read the weights while producing the NanoAOD file looks for the group name 'mg_reweighting' [1]. 

I have kept two LHE files in my public directory [2] for reference. 

So, this is not a problem with our cards. 

[1] https://github.com/cms-sw/cmssw/blob/master/PhysicsTools/NanoAOD/plugins/GenWeightsTableProducer.cc#L833

[2] LHE Files:
Produced with master branch:   /afs/cern.ch/user/c/chatterj/public/Higgs_EFT/LHE/Master_cmsgrid_final.lhe  (line 741)
Produced with UL2019 branch: /afs/cern.ch/user/c/chatterj/public/Higgs_EFT/LHE/UL2019_cmsgrid_final.lhe (line 731)

Sumantifr:
> > > > SMEFT model should not be used with MLM jet matching if I remember correctly.
> > > > Is this validated/checked with authors of the model or somehow?
> > > 
> > > 
> > > In SMP-20-005 paper draft, I read that the authors have generated Wgamma events with up to 2 additional jets using SMEFTsim model. I can ask Andrew for the link to the cards.
> > 
> > 
> > I think I remember correctly. EFT with MGv27X gives this warning message below. Besides, during the LHC reinterpretation workshop I am pretty sure SMEFT tutorial session tutors mentioned this. But please ask the authors of SMP-20-005, they might have validated this.
> > CRITICAL: Model with non QCD emission of gluon (found 14 of those).
> > This type of model is not fully supported within MG5aMC.
> > Restriction on LO dynamical scale and MLM matching/merging can occur for some processes.
> > Use such features with care.
> 
> Thanks pointing to this warning. I have just written to Andrew, will follow up on this once I hear from him.

Hi @sihyunjeon, 

Following up on the MLM matching. 

I got the cards from Andrew Gilbert (the analysis contact of SMP-20-005), who has produced Wgamma events with up to 2 jets and used MLM matching. These samples were produced privately up to parton-shower level. But, that should be still fine. 

For the final confirmation, do you want us to write to the SMEFTsim authors keeping you in cc? 

We can simply generate VH events without any additional jets, but there are some EFT-sensitive observables that are also affected by the additional jet radiation (as shown in SMP-20-005). So, the final sensitivity of the analysis can change if we include the additional jet or not. 

[1] https://gitlab.cern.ch/agilbert/EFT2Obs/-/blob/SMP-20-005/cards/WG-SMEFTsim-130-300/proc_card.dat
[2] https://gitlab.cern.ch/agilbert/EFT2Obs/-/blob/SMP-20-005/cards/WG-SMEFTsim-130-300/run_card.dat#L70

agrohsje:
Let me add a bit. I suggested for Top-19-001 to use MLM. The problem is that if an operator has QCD couplings but is not part of the merging routine:
 from https://feynrules.irmp.ucl.ac.be/wiki/dim6top:
 added a configuration option to normalize all chromo-dipole operator coefficients (ctG and FCNC cousins) with a factor of the strong coupling G. This is currently required for jet merging not to crash. Note however that the jet merging code was not written with non-SM interactions in mind. So results it produces with chromo-dipole operators should be validated. Note also that G runs and that its value at the scale MZ depends on the parton distribution function employed. See configuration.py for the hack, to be used with great care.
 So Olivier implemented a warning about this. An alternative opinion can be found in this paper: 
 https://arxiv.org/abs/1809.09744
 
 For Top-19-001 I was suggesting MLM as they split according to jets, so there is for sure a larger bias when not having jets from ME.

agrohsje:
For the CMSSW and MG5@NLO versions of the master branch, the corresponding 'weightgroup name' is only 'mg_reweighting'.
You mean for UL2019 ? You wrote twice master. 
We should check with @kdlong that both can be properly handled by nano-v9, currently in preparation.

Sumantifr:
> For the CMSSW and MG5@NLO versions of the master branch, the corresponding 'weightgroup name' is only 'mg_reweighting'.
> You mean for UL2019 ? You wrote twice master.
> We should check with @kdlong that both can be properly handled by nano-v9, currently in preparation.

You are correct that the  'weightgroup name' is only 'mg_reweighting' for the UL2019 branch. Sorry for typing it wrong, I have updated the comment.

Sumantifr:
> Let me add a bit. I suggested for Top-19-001 to use MLM. The problem is that if an operator has QCD couplings but is not part of the merging routine:
> from https://feynrules.irmp.ucl.ac.be/wiki/dim6top:
> added a configuration option to normalize all chromo-dipole operator coefficients (ctG and FCNC cousins) with a factor of the strong coupling G. This is currently required for jet merging not to crash. Note however that the jet merging code was not written with non-SM interactions in mind. So results it produces with chromo-dipole operators should be validated. Note also that G runs and that its value at the scale MZ depends on the parton distribution function employed. See configuration.py for the hack, to be used with great care.
> So Olivier implemented a warning about this. An alternative opinion can be found in this paper:
> https://arxiv.org/abs/1809.09744
> 
> For Top-19-001 I was suggesting MLM as they split according to jets, so there is for sure a larger bias when not having jets from ME.

This is very helpful to understand the issue! 

In our case, the operators we are interested in don't have QCD coupling [1], so I believe we will not be severely affected by the merging issue. 
At the analysis level, we use a veto on the number of additional jets (particularly in boosted phase space region). 
So, I think it's worth having one jet radiation from ME.

[1] https://indico.cern.ch/event/996940/contributions/4294588/attachments/2219222/3757710/Presentation_March31.pdf (slide 50)

sihyunjeon:
@agrohsje @Sumantifr 
I did some check several months ago :
SMEFT UFO with all BSM couplings turned **OFF** "p p > h" + "p p > h j"
SM UFO  "p p > h" + "p p > h j"
These two had severe differences (although I expected these two to be the same) and this difference was from "p p > h j". "p p > h" was rather comparable. I got answers from SMEFT authors that with SM process "p p > h j", pt(higgs) > 150 GeV would be screwed.

I know that above p p > h have QCD interactions, and what you are trying to do is irrelevant from QCD. But although it was turned OFF the expected SM results weren't reproduced with SMEFT UFO model.

So my concern is from this.
"p p > v h" + "p p > v h j" => SMEFT UFO BSM turned off
"p p > v h" + "p p > v h j" => SM UFO
Would these two be the same? And if not, is it still safe to add additional partons (j) for MLM matching although no QCD couplings play role?

sihyunjeon:
Of course,
if things are confirmed with the SMEFT authors, it won't be a problem at all... Let me know if my concerns are confusing you. Skype me (sihyun_jeon) and perhaps we can interactively talk there

Sumantifr:
Hi @sihyunjeon, 
I have the numbers you suggested using the restrict_SMlimit_massless.dat card in SMEFTsim where all Wilson coefficients (WC) are set to 0. The numbers I get are the following. 

                                     Z(->l+l-)h (with up to 1 additional jet):                             W-(->l-nu~)h (with up to 1 additional jet):
SM:                                0.08667 +- 0.0002127 pb                                                0.1966 +- 0.000581  pb
SMEFTsim:                    0.08468 +- 0.0002024 pb                                                0.1929 +- 0.0004909 pb
(all WCs set to 0)

So, the difference we get is ~2%. Do you think that this difference is large enough to worry about?

I have written the commands used in SMEFTsim in [1] for reference. 

[1]
Zh:
generate p p > z h SMHLOOP=0  NPprop=0 NP=0, z > l- l+ SMHLOOP=0  NPprop=0 NP=0
add process p p > z h j SMHLOOP=0  NPprop=0 NP=0, z > l- l+ SMHLOOP=0  NPprop=0 NP=0
W-h:
generate p p > w- h SMHLOOP=0  NPprop=0 NP=0, w- > l- vl~ SMHLOOP=0  NPprop=0 NP=0
add process p p > w- h j SMHLOOP=0  NPprop=0 NP=0, w- > l- vl~ SMHLOOP=0  NPprop=0 NP=0

sihyunjeon:
@Sumantifr Hello! Thanks for the swift checks. 2% is indeed a small difference. Can you also check below (just to make sure, while waiting for authors to reply)?
w/o any object cuts (using "set no_parton_cut") but only ptj cut (maybe 10 GeV, "set ptj 10")

SM model : generate p p > h j [QCD] # [QCD] to launch loop model
SMEFT model : generate p p > h j # not the SM limit cards, just the nominal one and turning off coupling by hand

Sumantifr:
Hi @sihyunjeon, 
I was not sure if you wanted me to write to the authors earlier. Anyways, as you have noticed, I wrote to Ilaria keeping you and @agrohsje in cc. 

Here are the numbers you asked for. 

SMEFTsim (with the option SMLOOP=1 NP=0 NPprop=0): 23.94 +- 0.04488 pb
SM (with [QCD]): 22.61 ± 0.058

Here the difference is ~5%, which is expected due to the fact that the modeling of the top quark loop fails at p > top quark mass, as Ilaria mentioned during the tutorial in the Re-interpretation workshop. This is also shown in Fig. 1 of the SMEFTsim manual (https://arxiv.org/pdf/2012.11343.pdf). 

If I use SMLOOP=0 in SMEFTsim, the result I get is two orders of magnitude less than the SM 1-loop result. 
SMEFTsim (with SMLOOP=0): 0.1836  +- 0.0003981 pb
This is sensible since here one does not take into account the top quark loop at all.

Sumantifr:
Hi @agrohsje @sihyunjeon , 

As Ilaria (the author of SMEFTsim) confirmed that our cards are fine, I hope that there is no showstopper anymore to merge this PR.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ZH_ZToLL_1j_SMEFTsim_topU3l/ZH_ZToLL_1j_SMEFTsim_topU3l_run_card.dat
```diff
@@ -0,0 +1,271 @@
...
+ 0   = ptj4min ! minimum pt for the fourth jet in pt
+ -1  = ptj1max ! maximum pt for the leading jet in pt 
+ -1  = ptj2max ! maximum pt for the second jet in pt
+ -1  = ptj3max ! maximum pt for the third jet in pt
+ -1  = ptj4max ! maximum pt for the fourth jet in pt
+ 0   = cutuse  ! reject event if fails any (0) / all (1) jet pt cuts
+#*********************************************************************
+# Control the pt's of leptons sorted by pt                           *
+#*********************************************************************
+ 10   = ptl1min ! minimum pt for the leading lepton in pt
```
sihyunjeon, line 212: This won't be applied since cut_decays is set to False. Are you aware of this?

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ZH_ZToLL_1j_SMEFTsim_topU3l/ZH_ZToLL_1j_SMEFTsim_topU3l_run_card.dat
```diff
@@ -0,0 +1,271 @@
...
+ -1  = dralmax ! maxdistance between gamma and lepton
+#*********************************************************************
+# Minimum and maximum invariant mass for pairs                       *
+# WARNING: for four lepton final state mmll cut require to have      *
+#          different lepton masses for each flavor!                  *           
+#*********************************************************************
+ 0   = mmjj    ! min invariant mass of a jet pair 
+ 0   = mmbb    ! min invariant mass of a b pair 
+ 0   = mmaa    ! min invariant mass of gamma gamma pair
+ 10   = mmll    ! min invariant mass of l+l- (same flavour) lepton pair
```
sihyunjeon, line 175: This as well, you produce onshell Z and then decay this to 2 leptons. So your signals will be already populated in the phase space m(ll)~m(Z) and there will be NO offshell contributions. Is this still fine for you? == Do you apply strong m(ll) cuts in your analysis RECO level to tag the Z bosons?

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ZH_ZToLL_1j_SMEFTsim_topU3l/ZH_ZToLL_1j_SMEFTsim_topU3l_run_card.dat
```diff
@@ -0,0 +1,271 @@
...
+#*********************************************************************
+# Minimum and maximum DeltaR distance                                *
+#*********************************************************************
+ 0.001 = drjj    ! min distance between jets 
+ 0   = drbb    ! min distance between b's 
+ 0   = drll    ! min distance between leptons 
+ 0   = draa    ! min distance between gammas 
+ 0   = drbj    ! min distance between b and jet 
+ 0   = draj    ! min distance between gamma and jet 
+ 0.4   = drjl    ! min distance between jet and lepton
```
sihyunjeon, line 153: This too, leptons are after decays so cut_decays=False will make this do nothing

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ZH_ZToLL_1j_SMEFTsim_topU3l/ZH_ZToLL_1j_SMEFTsim_topU3l_run_card.dat
```diff
@@ -0,0 +1,271 @@
...
+ -1   = ebmax ! maximum E for the b
+ -1   = eamax ! maximum E for the photons
+ -1   = elmax ! maximum E for the charged leptons
+#*********************************************************************
+# Maximum and minimum absolute rapidity (for max, -1 means no cut)   *
+#*********************************************************************
+   5  = etaj    ! max rap for the jets 
+  -1  = etab    ! max rap for the b
+  -1  = etaa    ! max rap for the photons 
+ 2.5  = etal    ! max rap for the charged leptons
```
sihyunjeon, line 138: This too

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ZH_ZToLL_1j_SMEFTsim_topU3l/ZH_ZToLL_1j_SMEFTsim_topU3l_run_card.dat
```diff
@@ -0,0 +1,271 @@
...
+#*********************************************************************
+# Run to generate the grid pack                                      *
+#*********************************************************************
+  .true.     = gridpack  !True = setting up the grid pack
+#*********************************************************************
+# Number of events and rnd seed                                      *
+# Warning: Do not generate more than 1M events in a single run       *
+# If you want to run Pythia, avoid more than 50k events in a run.    *
+#*********************************************************************
+  100000 = nevents ! Number of unweighted events requested
```
sihyunjeon, line 32: Please reduce this to 2-3000. For pilot run, you don't need to run this many events

sihyunjeon:
> Hi @agrohsje @sihyunjeon ,
> 
> As Ilaria (the author of SMEFTsim) has confirmed that our cards are fine, I hope that there is no showstopper anymore to merge this PR.

Hello, thanks for ccing me in the emails. Learned a lot! I just left some comments on computing side of the gridpack generations. I only left comments for ZH but are mostly applicable to WH case as well. Let me know what you think

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ZH_ZToLL_1j_SMEFTsim_topU3l/ZH_ZToLL_1j_SMEFTsim_topU3l_run_card.dat
Sumantifr: I was not actually aware of this. I can simply set cut_decays to True unless there is a known issue.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ZH_ZToLL_1j_SMEFTsim_topU3l/ZH_ZToLL_1j_SMEFTsim_topU3l_run_card.dat
Sumantifr: This should be fine, we apply a 60 GeV lower cut on the dilepton invariant mass at the analysis level.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ZH_ZToLL_1j_SMEFTsim_topU3l/ZH_ZToLL_1j_SMEFTsim_topU3l_run_card.dat
Sumantifr: Okay

Sumantifr:
Hi @sihyunjeon, 
I have reduced the number of events for the pilot run to 3000 and set the cut_decays to True. Hope this is fine.

sihyunjeon:
Hi,
If you apply dR(j,l) < 0.4, it means that it would reject all those events in GEN level, and this seems a bit harsh given that in RECO level there could be some smearing effects. So we usually recommend to loosen those cuts considering the RECO effects.
Same goes for etal = 2.5 -> leptons have coverage up to eta<2.5 in CMS, and there could also be leptons eta=2.51 in GEN level but eta 2.49 in RECO level.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ZH_ZToLL_1j_SMEFTsim_topU3l/ZH_ZToLL_1j_SMEFTsim_topU3l_run_card.dat
sihyunjeon: pointing this part once more,
Can you consider loosening these values? or not applying any cuts to this?

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ZH_ZToLL_1j_SMEFTsim_topU3l/ZH_ZToLL_1j_SMEFTsim_topU3l_run_card.dat
sihyunjeon: This too, I know there would be no drastic changes but it's always better to go safer

sihyunjeon:
Other than the too-tight cuts, for both WH and ZH, the cards look totally fine to me.
Sorry for multiple comments but this would likely be the last one

Sumantifr:
Hi @sihyunjeon , I have removed the cut on dR(l,j) and increased the maximum eta (etal) of leptons to 3.0. I don't think we need to change the cut on maximum eta of jets (etaj), since this is already quite relaxed (5). 

I think these are the cuts you referred to. Please let me know if I missed anything.

agrohsje:
@sihyunjeon I am happy with the request. Please merge.

sihyunjeon:
@Sumantifr @agrohsje All looks fine, merging
And thanks @Sumantifr

Sumantifr:
Many thanks to @sihyunjeon @agrohsje. It was a very good learning experience.

PR #2911, opened 2021-05-21
bug fix for cut card Mll selection and fix to proc card tag name

Joe-W-Davies:
Following email discussion with Alexander, I am making a PR with a bug fix for a proc card and cut card that were merged previously. The cards are for Drell-Yan (Z->ee) UL samples for the H->ee search with full Run 2 dataset. They were intended for producing two DY samples with selection on the dilepton invariant mass (105 < MLL < 160), one of which has enhanced VBF topology (VBFFilter). 

The original PR with the bugs was merged in this thread: https://github.com/cms-sw/genproductions/pull/2835

Please let me know if the new cards look sensible.

Thanks,
Joe


sihyunjeon:
What is the change?
If I understand correctly, there was some wrong implementations in the cuts.f file? mll < 100 GeV cut was applied when you had to do the opposite?

Joe-W-Davies:
Yes this is the change. The lower mll threshold is applied in the run card, while the upper threshold is now in the cut card. There was also a small update to the output name in the proc card to fix a typo (100 ->105)

sihyunjeon:
Looks fine to me @agrohsje

PR #2914, opened 2021-05-23
Adding Cards for EW production of Vector-Like Leptons in a benchmark 4321 model

kcormi:
Adding cards for gridpack production of Vector-like leptons (E - charged vector-like lepton, N - neutral vector-like lepton) in the "4321" model (see [1,2]). Only ElectroWeak production (No new physics coupling). For a brief overview of the analysis, see the recent presentation at the B2G VHF meeting[3].

Cards are separated between EE, EN and NN production, and specific final states including third generation leptons are specified so that madgraph performs the decay of the VLL. 

Cards are produced for mass points from 500-1050 GeV at 50 GeV intervals. These cards have been automatically produced with a script, so there should be no difference between any of the different mass points, except for the masses of the VLL. For each mass point the widths are computed automatically by madrgraph.

A couple of relevant points:

1) Here a default `param_card.dat` is also stored in the cards directory, which should then be imported during the gridpack generation (as specified by the `<basename>_customizecards.dat` file. I think this is not the standard way of running things, rather typically it seems each parameter which is different from the model's default `param_card.dat` setting should be specified individually in the `customizecards` file. It seemed a bit cumbersome in this case, where we were provided the full `param_card.dat` for use from our theorist colleagues who developed the model. However, I do worry that this is sensitive to the directory from which `gridpack_generation.sh` is run (since we need to give the path of the default `param_card.dat`). Perhaps there is another way to accomplish this? Or perhaps we are required to write out the differences explicitly. Please let me know in either case, and I will update.

2) The Automatic computation of widths by madgraph in this case is somewhat problematic for the neutral VLL, N. It includes for example, both the decay `N -> t (b~ tau-)` and the decay `N -> W b (b~ tau-)`, *i.e.* cases where the top is undecayed and cases where the `t->Wb` is considered. It appears that both of these decays are considered separately. Since we specify the final state in the generation (we always consider undecayed top-quarks in the specification), the reported cross-sections are incorrect (because the BR for the N differs by a factor of ~2). For our studies so far we have been correcting this by hand, multiplying the reported cross-section by the appropriate factor. Our colleagues who developped the model agree that this should still be a good approximation. Nonetheless, we plan to update the cards by pre-computing the widths for each mass point and inserting the pre-computed values manually into the cards. This is a work in progress.

Apologies for these slightly unconventional features. Our hope is that it is still useful to review the cards, and perhaps identify any other issues with the cards (e.g. PDF sets etc...) and to get feedback on the above mentioned issues. 

  

NB: There is an issue with the Madgraph automatic calculation of the neutral
vector-like lepton (N) decay, this should be updated in the future

[1] https://arxiv.org/abs/1808.00942
[2] https://arxiv.org/abs/1708.08450
[3] https://indico.cern.ch/event/1038139/contributions/4364219/attachments/2245529/3808171/VLL_VHF_210517.pdf


sihyunjeon:
Hi,
It's difficult to scroll down 145 files in this commit.
Can you just make a pull request with the minimal number of representative mass points + card generating scripts?

kcormi:
Okay, I've removed almost all of the cards, leaving only one process at one mass point (so four cards). In their place I have put scripts for producing the cards, which can be done by
```
./make_cards.py <process> <mass>
```
where process can be any of 'EE' 'EN' or 'NN'. As mentioned, above in the analysis we will use all three process, at mass point from 500 GeV to 1050 GeV in 50 GeV steps. 

The scripts commited here are not the same scripts used to actually produce the gridpacks which we have already produced for all 36 settings. Those are somewhat more involved to also make it easy to submit jobs, gather cross-sections etc. So I made new simplified sets which should be easier to review, but I ran gridpacks for each of the processes at the 900 GeV mass point to ensure they give the same cross-section as in the already produced cards.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/VLferm_EWcouplings_4321/VLferm_EWcouplings_4321_m900GeV_EE_to_4b_customizecards.dat
```diff
@@ -0,0 +1,7 @@
+../../../../cards/production/13TeV/VLferm_EWcouplings_4321/param_card.dat
```
agrohsje, line 1: please don't include relative paths that might not work. the default settings of the model will anyway be loaded and the customize card is just to adjust the rest.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/VLferm_EWcouplings_4321/VLferm_EWcouplings_4321_m900GeV_EE_to_4b_customizecards.dat
```diff
@@ -0,0 +1,7 @@
...
+set param_card mln 900
+set param_card mle 900
+update dependent
```
agrohsje, line 4: what will be changed by this command?

bin/MadGraph5_aMCatNLO/cards/production/13TeV/VLferm_EWcouplings_4321/VLferm_EWcouplings_4321_m900GeV_EE_to_4b_proc_card.dat
```diff
@@ -0,0 +1,6 @@
+import model VLferm_UFO
```
agrohsje, line 1: Is the model 4 FS or 5FS?

bin/MadGraph5_aMCatNLO/cards/production/13TeV/VLferm_EWcouplings_4321/VLferm_EWcouplings_4321_m900GeV_EE_to_4b_run_card.dat
```diff
@@ -0,0 +1,174 @@
...
+#
+#*********************************************************************
+# Store info for systematics studies                                 *
+# WARNING: Do not use for interference type of computation           *
+#*********************************************************************
+   True  = use_syst      ! Enable systematics studies
+#
+systematics = systematics_program ! none, systematics [python], SysCalc [depreceted, C++]
+['--mur=0.5,1,2', '--muf=0.5,1,2', '--pdf=errorset'] = systematics_arguments ! see: https://cp3.irmp.ucl.ac.be/projects/madgraph/wiki/Systematics#Systematicspythonmodule
+# Syscalc is deprecated but to see the associate options type'update syscalc'
```
agrohsje, line 174: please remove to avoid xml issues.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/VLferm_EWcouplings_4321/default_run_card.dat
```diff
@@ -0,0 +1,174 @@
...
+#
+#*********************************************************************
+# Store info for systematics studies                                 *
+# WARNING: Do not use for interference type of computation           *
+#*********************************************************************
+   True  = use_syst      ! Enable systematics studies
+#
+systematics = systematics_program ! none, systematics [python], SysCalc [depreceted, C++]
+['--mur=0.5,1,2', '--muf=0.5,1,2', '--pdf=errorset'] = systematics_arguments ! see: https://cp3.irmp.ucl.ac.be/projects/madgraph/wiki/Systematics#Systematicspythonmodule
+# Syscalc is deprecated but to see the associate options type'update syscalc'
```
agrohsje, line 174: please remove

agrohsje:
Is the param_card different from the default of the model? Please remove.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/VLferm_EWcouplings_4321/VLferm_EWcouplings_4321_m900GeV_EE_to_4b_customizecards.dat
kcormi: Okay, I've removed it and written out all of the differences explicitly.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/VLferm_EWcouplings_4321/VLferm_EWcouplings_4321_m900GeV_EE_to_4b_customizecards.dat
kcormi: I thought that this was necessary for the correct width computations (I'm not sure why). I was mistake, I've removed this.

agrohsje: I never tried this command. I would use  
set param_card decay PDGID auto 
but if you are telling me that the width is correctly updated I have no strong preference.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/VLferm_EWcouplings_4321/VLferm_EWcouplings_4321_m900GeV_EE_to_4b_proc_card.dat
kcormi: The model is in 5FS. In this case we are only considering EW production of the VLLs  the initial state b's don't really have an impact, since it process through s-channel Z/gamma and W  (though still they are included in the proton definition and we use massless b). However, for other states within the model, due to the highly generation-dependent structure, the initial state b's would be a dominant contribution.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/VLferm_EWcouplings_4321/VLferm_EWcouplings_4321_m900GeV_EE_to_4b_run_card.dat
kcormi: removed

bin/MadGraph5_aMCatNLO/cards/production/13TeV/VLferm_EWcouplings_4321/default_run_card.dat
kcormi: removed

kcormi:
> Is the param_card different from the default of the model? Please remove.

Yes, the param_card is different from the default of the model, but it is one of the theorists benchmarks. As per your comment above, I have no moved all differences w.r.t. the default `param_card.dat` into the `customizecards.dat` explicitly, and have removed the `param_card.dat`.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/VLferm_EWcouplings_4321/VLferm_EWcouplings_4321_m900GeV_EE_to_4b_customizecards.dat
kcormi: The width is updated, although see the note number 2 above, there does seem to be some difficulty in in the computation of the width for the neutral VLL, which we currently correct by hand after-the-fact. But I think this is unrelated to the syntax used for this command. 

I'm happy to try your syntax if you think there is any reason for it to be preferred.

kcormi: For completeness, I also tried the other syntax you suggested, and it made no difference

bin/MadGraph5_aMCatNLO/cards/production/13TeV/VLferm_EWcouplings_4321/default_run_card.dat
```diff
@@ -0,0 +1,170 @@
...
+ # Apply pt/E/eta/dr/mij/kt_durham cuts on decay products or not
+ # (note that etmiss/ptll/ptheavy/ht/sorted cuts always apply)
+ #*********************************************************************
+   False  = cut_decays    ! Cut decay products 
+#*********************************************************************
+# Standard Cuts                                                      *
+#*********************************************************************
+# Minimum and maximum pt's (for max, -1 means no cut)                *
+#*********************************************************************
+ 20.0  = ptj       ! minimum pt for the jets
```
sihyunjeon, line 97: The cuts you implemented in the run card will be ignored since you turned off cut_decays. Are you aware of this?

kcormi: Yes, there was no intention to have any cuts included in the generation. Is it preferred that we then set the values (e.g. -1 for jet pt) to make that more clear?

sihyunjeon: No not at all, wanted to make sure. Thanks

agrohsje:
@sihyunjeon please merge if you are happy with the PR. looks good to me. thanks!

PR #2918, opened 2021-05-31
Adding ZBJetsToNuNu and ZJetsToNuNu samples

tiwariPC:


bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ZJetsToNuNu/ZJetsToNuNu_Zpt-100to200_proc_card.dat
```diff
@@ -0,0 +1,9 @@
+import model sm-ckm_no_b_mass
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+generate p p > ve ve~ / h @0
+add process p p > ve ve~ j / h @1
+add process p p > ve ve~ j j / h @2
+add process p p > ve ve~ j j j / h @3
+add process p p > ve ve~ j j j j / h @4
```
sihyunjeon, line 8: Is there a reason to only do "ve ve~" for ZJets while you do "vl vl~" for ZJets+b?

tiwariPC: Thanks, @sihyunjeon for noticing this, I have fixed this bug now

agrohsje:
cards look good to me. what is the reason to have /h in the process line?

gourangakole:
> cards look good to me. what is the reason to have /h in the process line?

Hi @agrohsje , I agree with your point. To my understanding, 
in SM Higgs has no direct coupling with the neutrinos, so it should not matter 
but better to remove. 

thank you for catching that.

agrohsje:
@tiwariPC : can we quickly fix and then move forward.
@sihyunjeon are you ok with merging?

sihyunjeon:
@agrohsje Yes.
The higgs seems to be a null command but let's just wait for @tiwariPC to comment. After this, should be merged.

tiwariPC:
Hi @agrohsje  and @sihyunjeon, 
I can remove that part, ```/ h @1``` this whole part shall be removed right?
My second question is that, I made the grid packs with these current cards, as it seems the \h command is null, so shall I remake them or they are okay?

sihyunjeon:
You can keep the @1 @2 ... just remove "/ h".
And I am fine with the current gridpacks though they have "/ h" since the physics is the same.
But @agrohsje will let you know what to do.

agrohsje:
@tiwariPC it is fine to have it in the gridpacks. i would just remove it here in the pr to keep it clean and meaningful.

tiwariPC:
Okay Thanks @agrohsje, I have removed them here

PR #2919, opened 2021-05-31
JHUGen AC cards updated for UL production

mlizzo:
JHUGen AC cards have been updated with UL PDF set and WriteFailedEvents option + there is a specific production/decay card for each AC sample


mlizzo:
Dear @agrohsje @GurpreetSinghChahal would you please have a look to these cards? Thank you

agrohsje:
dear @mlizzo . thanks for the pr. may i ask, what is the idea? will you generally move to nnlo pdf with jhugen? you always use 325300, all requests are 5 flavor scheme?

mlizzo:
dear @agrohsje , thanks for your reply. Yes indeed, the corresponding gridpacks are in 5f scheme and have the 325300 PDF set in order to be in compliance with the UL campaign requirements, as it has already done for ggH Powheg AC samples. Is this reasonable?

agrohsje:
The ones you are changing are pure LO processes. I have no strong opinion in this case. I am also fine with LO PDFs and the corresponding CP tune. There are so many different arguments for one choice or the complete opposite that I don't think a generic statement makes sense. If NNLO showed no problem in performance, NNLO is good, if you spotted problems, LO is a good choice.

mlizzo:
Ok I see, thanks for the feedback. In any case I think it might be slightly better having the same PDF set across all AC samples (just for consistency purposes), hence if you agree we could keep these settings.

agrohsje:
keep nnlo ?

mlizzo:
Yes I mean keeping nnlo PDFs, sorry for the ambiguity.

PR #2937, opened 2021-07-05
Input Cards for Diphoton Mass 5-60 GeV

atsatsos97:
For very low mass diphoton search studies, mass points are in increments of 5 GeV.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/ggh012j_5f_NLO_FXFX_60/ggh012j_5f_NLO_FXFX_60_customizecards.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/ggh012j_5f_NLO_FXFX_60/ggh012j_5f_NLO_FXFX_60_customizecards.dat
@@ -1,0 +1,1 @@
+set param_card mass 25 60.
--- a/bin/MadGraph5_aMCatNLO/cards/ggh012j_5f_NLO_FXFX_60/ggh012j_5f_NLO_FXFX_60_run_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/ggh012j_5f_NLO_FXFX_60/ggh012j_5f_NLO_FXFX_60_run_card.dat
@@ -1,0 +1,131 @@
+#*******************                                                 
+# Running parameters
+#*******************                                                 
+#
+#***********************************************************************
+# Tag name for the run (one word)                                      *
+#***********************************************************************
+  tag_1     = run_tag ! name of the run 
+#***********************************************************************
+# Number of events (and their normalization) and the required          *
+# (relative) accuracy on the Xsec.                                     *
+# These values are ignored for fixed order runs                        *
+#***********************************************************************
+  1500 = nevents ! Number of unweighted events requested 
+ 0.001 = req_acc ! Required accuracy (-1=auto determined from nevents)
+    20 = nevt_job! Max number of events per job in event generation. 
+                 !  (-1= no split).
+average = event_norm ! Normalize events to sum or average to the X sect.
+#***********************************************************************
+# Number of points per itegration channel (ignored for aMC@NLO runs)   *
+#***********************************************************************
+ 0.01   = req_acc_FO       ! Required accuracy (-1=ignored, and use the 
+ 	                   ! number of points and iter. below)
+# These numbers are ignored except if req_acc_FO is equal to -1
+ 5000   = npoints_FO_grid  ! number of points to setup grids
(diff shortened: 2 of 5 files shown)


sihyunjeon:
Hi

Please commit only one mass point as an example + a script that generates the rest of the mass points

sihyunjeon:
Your directory name should match the prefix of the cards. <dirname>/<dirname>_run_card.dat
e.g. DYToLL/DYToLL_runcard.dat

sihyunjeon:
Does the x0 decay into photons in Pythia?

atsatsos97:
Hi,

I have updated my request to include only the cards for mass point 60 GeV in their appropriately-named directory, and a shell script that can generate cards for any other mass point. As an example, should someone want 50 GeV cards, they can go into this directory and type "./masspoint_gen.sh 50", and it will work.

bin/MadGraph5_aMCatNLO/cards/ggh012j_5f_NLO_FXFX_60/ggh012j_5f_NLO_FXFX_60_customizecards.dat
```diff
@@ -0,0 +1 @@
+set param_card mass 25 60.
```
sihyunjeon, line 1: Do you not set the widths of X0?

bin/MadGraph5_aMCatNLO/cards/ggh012j_5f_NLO_FXFX_60/ggh012j_5f_NLO_FXFX_60_run_card.dat
```diff
@@ -0,0 +1,131 @@
...
+#***********************************************************************
+  tag_1     = run_tag ! name of the run 
+#***********************************************************************
+# Number of events (and their normalization) and the required          *
+# (relative) accuracy on the Xsec.                                     *
+# These values are ignored for fixed order runs                        *
+#***********************************************************************
+  1500 = nevents ! Number of unweighted events requested 
+ 0.001 = req_acc ! Required accuracy (-1=auto determined from nevents)
+    20 = nevt_job! Max number of events per job in event generation.
```
sihyunjeon, line 16: please keep this to default -1

agrohsje:
Hi @sihyunjeon , if your concerns above are addressed and lines fixed, I am fine with merging. Rest looks good to me.

atsatsos97:
Hi @sihyunjeon @agrohsje ,

To address the comments:
- Yes, x0 will decay into photons, and that can be done using Pythia to simulate this decay while making the samples.
- We do not restrict the x0 width. For our analysis, we are searching for very low mass diphoton resonances from any possible ALPs, so I believe width restriction isn't necessary. Are you referring to the widths of the decay fragments done in Pythia, or something else entirely?

I noticed that you do include in [cards/examples/ggh012j_5f_NLO_FXFX](https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/examples/ggh012j_5f_NLO_FXFX) the cards I am requesting, with nevents, req_acc, and nevt_job  in run_card.dat as they are written in mine. The only difference between them is the mass point, and that I am also committing a shell script to add more mass points.
I can change the variables to the default values, but I did want to point out that my cards are cards that have been used before.

sihyunjeon:
@atsatsos97 If you look at the commit history, the runcard that you point out was created several years ago (even before RunII even started). A lot of changes have been made since. The settings that I tell you to change won't change any physics but how the madgraph computes its work so better to change them

sihyunjeon:
What is the default width setting in the UFO?

atsatsos97:
Hi @sihyunjeon,

I have updated the run_card.dat file to have the default values. Thank you for helping me to understand why!
Does this change mean I would have to redo my produced gridpacks, or would it be okay to proceed with making samples from them?

sihyunjeon:
> Hi @sihyunjeon,
> 
> I have updated the run_card.dat file to have the default values. Thank you for helping me to understand why!
> Does this change mean I would have to redo my produced gridpacks, or would it be okay to proceed with making samples from them?

@agrohsje I usually ask Alexander to reply on such questions but because he is absent for few weeks, I'll just suggest you to redo the gridpacks to be safe.

atsatsos97:
Hi @sihyunjeon @agrohsje,

I redid the gridpacks with the default values just in case. I now have gridpacks from 5-60GeV in increments of 5 for both sets of values.

PR #2938, opened 2021-07-06
bbll_4FermionCI

sihyunjeon:
bbll 4 fermion CI

5f scheme
mass binned


bin/MadGraph5_aMCatNLO/cards/production/13TeV/bbll_4FermionCI/bbll_4FermionCI_M-200To500_Lambda-1TeV_constLL_XQCUT-30/bbll_4FermionCI_M-200To500_Lambda-1TeV_constLL_XQCUT-30_proc_card.dat
```diff
@@ -0,0 +1,11 @@
+import model bbll_UFO
+define p = u d c s b u~ d~ c~ s~ b~ g
+define j = p
+generate p p > mu+ mu- NP^2>0 @0
```
agrohsje, line 4: when you do np^2>0, will you get the first order that achieves this? so the pure interference term?

sihyunjeon: What you get here is g * |NP X SM| + |NP X NP|.
So SM interference + pure NP terms.
And the g is either +4pi (constructive) or -4pi (destructive), if I understand your question correctly

sihyunjeon:
Added the PDF that I sent you a while ago @agrohsje  (We can remove this if you don't want PDF files in genproductions after you take a look)
**_Copy pasting old skype message_**


Hey Alexander,
Hope you had a nice rest. Sending you this now since I think you'll be back by now. I have some things to show you regarding [1].
Your question was, if I am correct, when doing NP^2>0 -> MadGraph first searches for the lowest NP order which is |SM X NP| term and stops so |NP|^2 might NOT BE INCLUDED. So I checked p p > mu+ mu- [+ NP settings].

a. NP^2==0 -> Not really sure if this gives only SM term, just guessed, at least I know that this process only contains SM diagrams if I check with "display diagrams"
b. NP^2==2 -> |SM X NP|, explained by Olivier in [2] (Exercise IV: Interference, page 52)
c. NP^2<=2 -> |SM|^2 + |SM X NP|, also in [2]
d. NP^2>0 -> |SM X NP| + |NP|^2, also in [2]
e. no arguments -> I guessed this would be either only |SM|^2 (same as a.) or  |SM|^2 + |SM X NP| +|NP|^2. This is a bit tricky, if I look at the diagrams generated on MadGraph, I only see |SM|^2 ones (no NP diagrams, just like a.). But the cross section as you can see in the slides e. and a. cross sections do not agree. So I just checked with SM UFO model and the cross section from SM UFO was closer to e. 877 pb.

Leaving a. and e. aside, which are just some studies based on my guesses for now, b. to d. all varies differently as I change the coupling values. So I think NP does not have restrictions searching for number of vertices in the same way as "QED" syntax does (it should still have some restrictions). ==> ""NP^2>0 can include both |SM X NP| and |NP|^2.""

So if it's okay for you maybe it's good to merge [1]? And I'll probably need to check the difference between a. and e. with Olivier and also ask about "NP" not having search restrictions like "QED" does.

[1] https://github.com/cms-sw/genproductions/pull/2938
[2] https://indico.cern.ch/event/796134/contributions/3560246/attachments/1908914/3157345/19_09_19_tuto_Vietnam.pdf

sihyunjeon:
I did some follow up checks and the difference in a. and e. (which both should be pure SM DY), comes from the difference in dynamic scale settings choice. If I synchronize this, I get the same result.

PR #2951, opened 2021-07-13
total_eff inclusion

jordan-martins:
Hi,

On the evaluation of the events/lumi a priori the math should be:

events/lumi = 8*3600*total_eff/time_per_evt,
where total_eff = filter_eff*match_eff

Thanks, Jordan

[Notes]:
Case of the study was: https://cms-pdmv.cern.ch/mcm/requests?prepid=HIG-RunIISummer20UL18wmLHEGEN-01665&page=0&shown=70368746274943 

In production, this is getting 22 events/lumi and it hasn't been flagged in gen check script.

### DIFF
--- a/bin/utils/request_fragment_check.py
+++ b/bin/utils/request_fragment_check.py
@@ -330,6 +330,7 @@ for num in range(0,len(prepid)):
         mem = r['memory']
         filter_eff = r['generator_parameters'][-1]['filter_efficiency']
         match_eff = r['generator_parameters'][-1]['match_efficiency']
+        total_eff = filter_eff*match_eff
 	cross_section = r['generator_parameters'][-1]['cross_section']
 	ext = r['extension']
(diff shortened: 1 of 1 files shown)


efeyazgan:
Hi, 

For that request, when I run the checking script, I get:
time per event (sec/event) =101.813121498
Expected number of events = 282.871201434 (if I only use filter efficiency)
Expected number of events = 25.4584081291 (if I use filter x matching efficiency)

I think in the past, MC contacts were writing the total efficiency to filter efficiency. 

I will update the code now but this should be changed in a few other places, so I will close your PR and I'll make a new one to merge. 

Thanks.

jordan-martins:
Hi Efe,
sorry, not sure where more to add. I see that I missed the change in line 648, only. In any case, filter eff and match eff should be placed properly by MC contacts as a result of genxsecanlzr output from the local test.

Thanks, Jordan

PR #2952, opened 2021-07-13
fix efficiency calculation error and its comment

efeyazgan:


efeyazgan:
fyi @jordan-martins

jordan-martins:
Thanks a lot, @efeyazgan !

PR #2957, opened 2021-07-19
NMSSM cards for a resonant search for di-Higgs production in the gamma gamma tau tau final state.

MatthewDKnight:
These cards are for a resonant search for di-Higgs production in the HH->gamma gamma tau tau final state. The intention is to use the full Run 2 dataset so UL samples will be asked for all 3 years. There are two sets of cards: NMSSM_ggtt and NMSSM_ggZZ. Both are based upon cards that already exist in genproductions [1].  The cards in [1] are for the 4b final state. The decay in the proc_card is changed correspondingly. 

The NMSSM_ggtt cards are supposed to correspond to a set of decays where the SM Higgs boson (h01) decays to a pair of photons and the other (h02) decays to a pair of taus. 

The NMSSM_ggZZ cards are supposed to account for the possibility for a Higgs boson to decay to two Z bosons, where at least one of them decays to a pair of taus, thereby leading to the same final state (+ a few extra particles). I chose to specify the decay as "H-> tau+ tau- all all" so that all possible contributions to this process (not necessarily mediated by Z's) are accounted for. I am not certain this is correct.

[1] https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_bbbb


bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_ggZZ/Template/proc_card.dat
```diff
@@ -0,0 +1,22 @@
...
+define l- = e- mu-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+set group_subprocesses Auto
+set ignore_six_quark_processes False
+set loop_optimized_output True
+set gauge unitary
+set complex_mass_scheme False
+import model NMSSMHET_UFO --modelname
+generate g g > h03 , (h03 > h02 h01, h02 > tau+ tau- all all, h01 > a a)
```
sihyunjeon, line 21: "h02 > tau+ tau- all all" is so vague syntax
Can you let us know what are the final states you target for?

MatthewDKnight: The final state we are interested in tau+ tau- gamma gamma. The reasoning behind the ggZZ cards is that suppose you have a process where H->gg and H->ZZ, then if one of the Z bosons decay into tau+ tau-, then you have your tau+ tau- gamma gamma final state + some 2 extra particles from the other Z boson. 

At least this was my initial intention/idea. I thought that I ought not to restrict myself to the H->ZZ decay in particular. For example, you could also get H->WW-> tau+ nu tau- nu which would lead to the same final state. By using "h02 > tau+ tau- all all", I am looking for tau+ tau- gamma gamma + some 2 extra particles from anywhere.

sihyunjeon: But you get bunch of contributions from diagrams that you might not want and increase the computing time so much.
If "m(h02) > 2 X m(Z or W)" would you not instead try to specify the decay channels you want?
By doing something like (define sth = leptons and quarks)
h02 > z z, (z > tau+ tau-), (z > sth sth)
h02 > w+ w-, (w+ > tau+ ......

image

MatthewDKnight: I originally tried these different process lines in the sm model and saw only approx. 200 diagrams for "h > tau+ tau- all all" which is why I thought it might not be an issue. I've since tried it in the NMSSMHET_UFO and now I get closer to 10,000 so I agree that this is too vague. 

I am now unsure what process line to write. I was confused about the suggestion of "h02 > z z, (z > tau+ tau-), (z > sth sth)" since my understanding is that this process line would require the z bosons to be on-shell. I do not think this requirement is necessary. An alternative would be "h02 > z z > ta+ ta- sth sth" but then I have also been told that this sort of syntax can lead to gauge invariance issues. What would you suggest to do?

sihyunjeon: Since I guess you'll be triggering on taus which would require some pT cuts, I don't think requiring onshell Z will give you too much loss (pTtau >~ 30 GeV). But then it would be safer to keep the two sths to come from offshell Z/gamma.
What I would try is
"h02 > z sth sth, z > ta+ ta-" where "sth = lepton, neutrino, and jet"
"h02 > w+ w-, w+ > ta+ vt, w- > ta- vt~"

Does this make sense to you?
If your are concerned, you might want to do a quick check with the cross section differences between a. and b. and see how much offshell contributions will be.
a. h02 > e+ e- ta+ ta-
b. h02 > z z, z > e+ e-, z > ta+ ta-

MatthewDKnight: We want to derive samples at multiple values of the h02 (Y) mass (see generate_grid.py L74). Some of these values are below 2*m_Z, in which case it won't be possible for both of the Z bosons to be on-shell. I tried calculating the partial widths you suggested anyway. 

For h02 mass = 90 GeV, and h03 mass = 300 GeV I got:
a. 0.425 pm 0.006 GeV
b. 7.38e-15 pm 2.43e-17 GeV
So assuming I haven't made a mistake, this shows the off-shell contributions are neccessary?

sihyunjeon: Sorry for the late response.

I am not sure what you did. If you set mh02 = 90, then
"h02 > z z, z > e+ e-, z > ta+ ta-" does not make sense at all, did you do this? If so, I am actually a bit surprised to see this worked. Because as you said, b. is only possible for MG to calculate when 2*mZ < mh02.

I can't think of any smart tricks to do mh02 = 90 GeV production yet.
What would be sensible to do for the mass points above will be checking the offshell contributions by comparing :
a. generate h02 > z e+ e-, z > ta+ ta-
b. generate h02 > ta+ ta- e+ e-
When MadGraph asks you to edit run card, you might have to type in "set no_parton_cut" and remove all kinematic cuts, and then apply minimal mmll or ptl cuts (~5 GeV maybe) as the width calculations will be affected by the default kinematic cut settings in MadGraph.

If you are not sure on what to do, please send me an email shjeon@cernREMOVETHIS.ch

agrohsje:
Was there a follow-up on this?

MatthewDKnight:
Whilst working on Sihyun's suggestions, I realised that the ggZZ samples may not be as necessary as first thought since the h02 > ZZ decay should be suppressed in the NMSSM. I am still in discussions with people (some are on holiday which is why it is taking time) about whether we should drop the ggZZ samples or not. I thought it best to wait until we had made that decision before continuing the discussion about the cards.

sihyunjeon:
@MatthewDKnight Hello, Any updates?

MatthewDKnight:
Hi, sorry for taking a while to get back to this. We've decided to drop the ggZZ samples so we only want samples where the Higgs decays directly to two taus.

What's the best way to do this? Should I close this pull request and create a new one with only the direct H->tautau cards?

sihyunjeon:
So you don't need the contributions from "higgs > z z > ta+ ta- nu nu"?

MatthewDKnight:
Correct. We do not need these contributions. 

Since the h02>zz decay is suppressed in the NMSSM we now know that this contribution should be negligible and therefore not necessary.

sihyunjeon:
I don't think you need to redo the pull request, just change it to the syntax you want in proc card.
h > ta+ ta- maybe

MatthewDKnight:
There are two sets of cards (two commits) in this pull request. One is already for the h > ta+ ta- process. I could delete all of the files from the "Add NMSSM HH->ggZZ" commit?

sihyunjeon:
Ah yes go for it

MatthewDKnight:
I've reverted the HH->ggZZ commit.

Please let me know if you have any questions about the HH->ggtt cards. They are the same as the ggZZ cards except for the decay process.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_ggtt/Template/proc_card.dat
```diff
@@ -0,0 +1,22 @@
+set group_subprocesses Auto
+set ignore_six_quark_processes False
+set loop_optimized_output True
+set loop_color_flows False
+set gauge unitary
+set complex_mass_scheme False
+set max_npoint_for_channel 0
+import model sm
```
sihyunjeon, line 8: please remove the unnecessary lines from here (import model sm, and several duplicated lines)

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_ggtt/Template/run_card.dat
```diff
@@ -0,0 +1,269 @@
...
+#*********************************************************************
+# Tag name for the run (one word)                                    *
+#*********************************************************************
+  tag_1     = run_tag ! name of the run 
+#*********************************************************************
+# Number of events and rnd seed                                      *
+# Warning: Do not generate more than 1M events in a single run       *
+# If you want to run Pythia, avoid more than 50k events in a run.    *
+#*********************************************************************
+  10000 = nevents ! Number of unweighted events requested
```
sihyunjeon, line 28: 3K events for pilot runs should be fine

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_ggtt/Template/proc_card.dat
```diff
@@ -0,0 +1,22 @@
...
+define l+ = e+ mu+
+define l- = e- mu-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+set group_subprocesses Auto
+set ignore_six_quark_processes False
+set loop_optimized_output True
+set gauge unitary
+set complex_mass_scheme False
+import model NMSSMHET_UFO --modelname
```
sihyunjeon, line 20: Is your model by default 4f scheme or 5f scheme?

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_ggtt/Template/run_card.dat
```diff
@@ -0,0 +1,269 @@
...
+ 0.4 = drjj    ! min distance between jets 
+ 0.0   = drbb    ! min distance between b's 
+ 0.4 = drll    ! min distance between leptons 
+ 0.4 = draa    ! min distance between gammas 
+ 0.0   = drbj    ! min distance between b and jet 
+ 0.4 = draj    ! min distance between gamma and jet 
+ 0.4 = drjl    ! min distance between jet and lepton 
+ 0.0   = drab    ! min distance between gamma and b 
+ 0.0   = drbl    ! min distance between b and lepton 
+ 0.4 = dral    ! min distance between gamma and lepton
```
sihyunjeon, line 157: You are aware the cuts here won't be applied to your samples since cut_decays is set to False, right?

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_ggtt/Template/proc_card.dat
MatthewDKnight: Done

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_ggtt/Template/run_card.dat
MatthewDKnight: Done

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_ggtt/Template/proc_card.dat
MatthewDKnight: When I import the model, the definition of the proton is 'p = g u c d s u~ c~ d~ s~'.

That would mean 4f scheme, right?

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_ggtt/Template/run_card.dat
MatthewDKnight: I originally took these cards from a equivalent HH->4b analysis [1] and assumed details, like the cuts (or lack thereof), should the same for our analysis... However, now that you have brought it up, I am uncertain whether the best idea is to apply cuts on the decay products or not.

On the one hand, applying cuts to the decay products makes sense because there is no point generating events which your analysis selection will remove anyway. On the other hand, I will need to calculate the efficiency of the selection criteria so that I can put a limit on the inclusive cross section of the process I am searching for. And this second point makes me think that we must **not** apply cuts to the decay products. Could you please advise?

[1] https://github.com/cms-sw/genproductions/blob/bf737d68ffaa086248c74256060f4cdbaa86948d/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_bbbb/Template/run_card.dat

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_ggtt/Template/proc_card.dat
agrohsje: yes

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_ggtt/Template/run_card.dat
agrohsje: I would go for an inclusive sample. I don't think the cut will help a lot with event generation efficiency.

agrohsje:
Hi @sihyunjeon @MatthewDKnight, I updated the proc card. Rest looks good to me if we agree not to apply cuts. Fine with merging.

MatthewDKnight:
I am also happy.

sihyunjeon:
merged

PR #2966, opened 2021-07-22
add exception for PPD for UL consistency checks

efeyazgan:


andreypz:
We need exceptions for the `Summer20UL17pp5TeV` campaign names, not only for PPD.
Something like this:
```
if "NULL" in pi_prime and ("APV" in pi or "Summer20UL18" in pi or ("Summer20UL17" in pi and  Summer20UL17pp5TeV not in pi) ):
```

efeyazgan:
Hi @andreypz thanks. Now the consistency check is ignored for Summer20UL17pp5TeV (see https://github.com/cms-sw/genproductions/commit/8fbec735b4e4e945f6c52f5dab1663320f575746)

PR #2970, opened 2021-07-26
semi-leptonic decay of ZVG

menglu21:
add cards of ZVG, v=z w+ w-, v > q q

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ZVG_4f_NLO/ZVG_4f_NLO_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ZVG_4f_NLO/ZVG_4f_NLO_proc_card.dat
@@ -1,0 +1,6 @@
+import model loop_sm
+define lep+ = e+ mu+ ta+
+define lep- = e- mu- ta-
+define v = z w+ w-
+generate p p > v lep+ lep- a $$ h [QCD]
+output ZVG_4f_NLO -nojpeg
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ZVG_4f_NLO/ZVG_4f_NLO_run_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ZVG_4f_NLO/ZVG_4f_NLO_run_card.dat
@@ -1,0 +1,145 @@
+#***********************************************************************
+#                        MadGraph5_aMC@NLO                             *
+#                                                                      *
+#                      run_card.dat aMC@NLO                            *
+#                                                                      *
+#  This file is used to set the parameters of the run.                 *
+#                                                                      *
+#  Some notation/conventions:                                          *
+#                                                                      *
+#   Lines starting with a hash (#) are info or comments                *
+#                                                                      *
+#   mind the format:   value    = variable     ! comment               *
+#***********************************************************************
+#
+#*******************                                                 
+# Running parameters
+#*******************                                                 
+#
+#***********************************************************************
+# Tag name for the run (one word)                                      *
+#***********************************************************************
+  ZVG_4f_NLO = run_tag ! name of the run 
+#***********************************************************************
+# Number of events (and their normalization) and the required          *
+# (relative) accuracy on the Xsec.                                     *
+# These values are ignored for fixed order runs                        *
+#***********************************************************************
+  3000 = nevents ! Number of unweighted events requested 
+ 0.001 = req_acc ! Required accuracy (-1=auto determined from nevents)
+    -1 = nevt_job! Max number of events per job in event generation. 
+                 !  (-1= no split).
+average = event_norm ! Normalize events to sum or average to the X sect.
+#***********************************************************************
+# Number of points per itegration channel (ignored for aMC@NLO runs)   *
+#***********************************************************************
+ 0.01   = req_acc_FO       ! Required accuracy (-1=ignored, and use the 
+ 	                   ! number of points and iter. below)
+# These numbers are ignored except if req_acc_FO is equal to -1
+ 5000   = npoints_FO_grid  ! number of points to setup grids
+ 4      = niters_FO_grid   ! number of iter. to setup grids
+ 10000  = npoints_FO       ! number of points to compute Xsec
+ 6      = niters_FO        ! number of iter. to compute Xsec
+#***********************************************************************
+# Random number seed                                                   *
+#***********************************************************************
+     0    = iseed       ! rnd seed (0=assigned automatically=default))
+#***********************************************************************
+# Collider type and energy                                             *
+#***********************************************************************
+    1   = lpp1    ! beam 1 type (0 = no PDF)
+    1   = lpp2    ! beam 2 type (0 = no PDF)
+ 6500   = ebeam1  ! beam 1 energy in GeV
+ 6500   = ebeam2  ! beam 2 energy in GeV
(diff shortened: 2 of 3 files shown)


bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ZVG_4f_NLO/ZVG_4f_NLO_proc_card.dat
```diff
@@ -0,0 +1,6 @@
+import model loop_sm
+define lep+ = e+ mu+ ta+
+define lep- = e- mu- ta-
+define v = z w+ w-
+generate p p > v lep+ lep- a $$ h [QCD]
```
sihyunjeon, line 5: Hi Meng, does this not have same issues that Olivier explained in [1]?

[1] https://answers.launchpad.net/mg5amcnlo/+question/696694

menglu21: Hi Sihyun, the semi-leptonic channel will not be affected by the ambiguity, only the identical decay of two Z will be affected,  there is the answer from Olivier, https://answers.launchpad.net/mg5amcnlo/+question/696360

sihyunjeon: Hmmm it's now more puzzling.

The cross section that you do with both a. and b. would be
sigma(p p > z lep+ lep-) that you get from MadGraph TIMES BR(Z > q q OR lep+ lep-) in MadSpin

a. MG : p p > z lep+ lep- & MS : z > q q
b. MG : p p > z lep+ lep- & MS : z > lep+ lep-

Then you mean only for d., the cross sections will be somewhat buggy? And c. will be OK?
c. MG : p p > z e+ e- & MS : z > mu+ mu-
d. MG : p p > z e+ e- & MS : z > e+ e-

menglu21: Hi Sihyun, from Olivier's answer, the issue is from the ambuiguity of same lepton pairs, if muon is different from ele in the module, then the 'c' should be ok. But in the semileptonic case, as Olivier said, there should be no ambuiguity as there are quark and lepton pair.

sihyunjeon: I don't fully get this issue frankly.
a. p p > z e+ e-, z > mu+ mu-
b. p p > z e+ e-, z > e+ e-
The cross sections of a. and b. are exactly the same when I give mmll=10 GeV
Is this expected? Sorry for my ignorance :|

menglu21: i think it's expected. if only consider two onshell Z, the phase space of final state e+e-e+e- should be half of the e+e-mu+mu-. As in 4e case, "p p > z e+ e-, z > e+ e-" is buggy, in which two onshell Z is counted twice, which make it agree with the correct syntax "p p > z e+ e-, z > mu+ mu-". i made some runs to confirm it:

all with "mmll=10", and with "True = cut_decay" if decay chain
syn1: p p > e+ e- e+ e-  XS: 0.005757 pb
syn2: p p > e+ e- z, z > e+ e- XS: 0.009628 pb (buggy one)
syn3: p p > z z, z > e+ e- XS: 0.004405 pb
syn4: p p > e+ e- e+ e- $z  XS: 0.0001746 pb
--> syn1 ~ syn2+syn4-syn3

syn5: p p > e+ e- mu+ mu- XS: 0.01151 pb
syn6: p p > e+ e- z, z > mu+ mu- XS: 0.009643 pb
syn7: p p > mu+ mu- z $z, z > e+ e- XS: 0.0008272 pb
syn8: p p > e+ e- mu+ mu- $z XS: 0.0002583 pb
--> syn5 ~ syn6+syn7+syn8

sihyunjeon: Ahhhh

Ok so the problem comes from the fact that when you have two Z bosons that decays into ee and mumu, you can distinguish the two. But for ee and ee case, there's no way to distinguish it because they are identical

sihyunjeon: Just out of curiosity, is there a sample that only include the higgs contributions? Or is the cross section negligible?

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ZVG_4f_NLO/ZVG_4f_NLO_run_card.dat
```diff
@@ -0,0 +1,145 @@
...
+# After showering an MLM-type merging should be applied as well.       *
+# See http://amcatnlo.cern.ch/FxFx_merging.htm for more details.       *
+#***********************************************************************
+ 0        = ickkw            ! 0 no merging, 3 FxFx merging
+#***********************************************************************
+#
+#***********************************************************************
+# BW cutoff (M+/-bwcutoff*Gamma)                                       *
+#***********************************************************************
+ 30  = bwcutoff
```
sihyunjeon, line 107: Again just some naive question, is there a reason to enlarge the BW cutoff instead of using the default value 15?

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ZVG_4f_NLO/ZVG_4f_NLO_proc_card.dat
menglu21: i did a test on following two processes (l=e,mu):
1. p p > z l+ l- a $$h [QCD]
2. p p > z l+ l- a [QCD]
with default cut setting, the xs are the same 1.855e-02 +- 5.3e-05 pb

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ZVG_4f_NLO/ZVG_4f_NLO_run_card.dat
menglu21: the xs of the signal is small, so we'd like to have loose requirement on BW to have more events

sihyunjeon: https://answers.launchpad.net/mg5amcnlo/+question/696189

Can you check this thread and see if it doesn't have any same issues?

menglu21: i did a test with runcmsgrid.sh from the gridpack, output is as expected.

agrohsje:
Thanks for clarifying @menglu21. I am happy with all your feedback. Going to merge. @sihyunjeon please ping in case you like to further follow-up. We can always update if needed.

PR #2972, opened 2021-07-30
Adding UL era VBF to Spin 0/1/2 to HH Madgraph cards 

atishelmanch:
Adding MadGraph cards for VBFToXToHH gridpacks which follow the latest GEN recommendations for the Run 2 UL era [1], including the latest recommended LHA PDFs and MadGraph + CMSSW versions. 

In addition, these are updated from previous VBFToXToHH cards in order to force the process to begin with quarks to remove an unwanted gluon diagram. Some basic LHE checks have been performed and can be found here: [2]. 

Thank you,
Abe

cc: @acarvalh 

[1] https://cms-pdmv.gitbook.io/project/mccontact/info-for-mc-production-for-ultra-legacy-campaigns-2016-2017-2018
[2] https://twiki.cern.ch/twiki/pub/CMS/AbrahamTishelmanCharnyHomepage/29_July_2021_Res_VBF_HH_Validation.pdf

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/VBFToXToHH_UL/VBFToBulkGravitonToHH_M/VBFToBulkGravitonToHH_M_extramodels.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/VBFToXToHH_UL/VBFToBulkGravitonToHH_M/VBFToBulkGravitonToHH_M_extramodels.dat
@@ -1,0 +1,2 @@
+## Models for searches of diboson resonances
+dibosonResonanceModel.tar.gz
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/VBFToXToHH_UL/VBFToBulkGravitonToHH_M/VBFToBulkGravitonToHH_M_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/VBFToXToHH_UL/VBFToBulkGravitonToHH_M/VBFToBulkGravitonToHH_M_proc_card.dat
@@ -1,0 +1,12 @@
+
+set group_subprocesses Auto
+set ignore_six_quark_processes False
+set loop_optimized_output True
+set complex_mass_scheme False
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+import model RS_bulk_ktilda
+define q = u c d s b u~ c~ d~ s~ b~
+define p = u c d u~ c~ d~
+generate p p > y q q / g h , ( y > h h )
+output VBFToBulkGravitonToHH_M<MASS> -nojpeg
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/VBFToXToHH_UL/VBFToBulkGravitonToHH_M/VBFToBulkGravitonToHH_M_run_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/VBFToXToHH_UL/VBFToBulkGravitonToHH_M/VBFToBulkGravitonToHH_M_run_card.dat
@@ -1,0 +1,268 @@
+#*********************************************************************
+#                       MadGraph5_aMC@NLO                            *
+#                                                                    *
+#                     run_card.dat MadEvent                          *
+#                                                                    *
+#  This file is used to set the parameters of the run.               *
+#                                                                    *
+#  Some notation/conventions:                                        *
+#                                                                    *
+#   Lines starting with a '# ' are info or comments                  *
+#                                                                    *
+#   mind the format:   value    = variable     ! comment             *
+#*********************************************************************
+#
+#*******************
+# Running parameters
+#*******************
+#
+#*********************************************************************
+# Tag name for the run (one word)                                    *
+#*********************************************************************
+  tag_1     = run_tag ! name of the run
+#*********************************************************************
+# Run to generate the grid pack                                      *
+#*********************************************************************
+  False     = gridpack  !True = setting up the grid pack
+#*********************************************************************
+# Number of events and rnd seed                                      *
+# Warning: Do not generate more than 1M events in a single run       *
+# If you want to run Pythia, avoid more than 50k events in a run.    *
+#*********************************************************************
+  100000 = nevents ! Number of unweighted events requested
+  0   = iseed   ! rnd seed (0=assigned automatically=default))
+#*********************************************************************
+# Collider type and energy                                           *
+# lpp: 0=No PDF, 1=proton, -1=antiproton, 2=photon from proton,      *
+#                                         3=photon from electron     *
+#*********************************************************************
+     1        = lpp1    ! beam 1 type
+     1        = lpp2    ! beam 2 type
+     6500.0     = ebeam1  ! beam 1 total energy in GeV
+     6500.0     = ebeam2  ! beam 2 total energy in GeV
+#*********************************************************************
+# Beam polarization from -100 (left-handed) to 100 (right-handed)    *
+#*********************************************************************
+     0.0     = polbeam1 ! beam polarization for beam 1
+     0.0     = polbeam2 ! beam polarization for beam 2
+#*********************************************************************
+# PDF CHOICE: this automatically fixes also alpha_s and its evol.    *
+#*********************************************************************
+ lhapdf = pdlabel ! PDF set
+ $DEFAULT_PDF_SETS = lhaid
+ $DEFAULT_PDF_MEMBERS  = reweight_PDF
+#*********************************************************************
+# Renormalization and factorization scales                           *
+#*********************************************************************
+ False = fixed_ren_scale  ! if .true. use fixed ren scale
+ False        = fixed_fac_scale  ! if .true. use fixed fac scale
+ 91.188  = scale            ! fixed ren scale
+ 91.188  = dsqrt_q2fact1    ! fixed fact scale for pdf1
+ 91.188  = dsqrt_q2fact2    ! fixed fact scale for pdf2
+ 1.0  = scalefact        ! scale factor for event-by-event scales
+#*********************************************************************
+# Matching - Warning! ickkw > 1 is still beta
+#*********************************************************************
+ 0 = ickkw            ! 0 no matching, 1 MLM, 2 CKKW matching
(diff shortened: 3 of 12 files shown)


bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/VBFToXToHH_UL/VBFToRadionToHH_M/VBFToRadionToHH_M_proc_card.dat
```diff
@@ -0,0 +1,12 @@
+
+set group_subprocesses Auto
+set ignore_six_quark_processes False
+set loop_optimized_output True
+set complex_mass_scheme False
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+import model heft_radion
+define q = u c d s b u~ c~ d~ s~ b~
+define p = u c d u~ c~ d~
```
sihyunjeon, line 10: I am just taking a super quick look.
But is 3 flavor scheme for proton definition what you really want?

atishelmanch: Hello @sihyunjeon - Yes this is intentional. Our reasoning is that the proton pdf for quarks other than u/d is very negligible, to a level only precision analyses in H couplings need. Since we are producing X->HH, we don't think this level of precision is necessary, and we considerably decrease the number of diagrams. 

-Abe

atishelmanch: Hello @sihyunjeon - A friendly ping/reminder. What do you think about this?

Thank you,
Abe

sihyunjeon: The number of diagrams will only affect the first step of sample production (wmLHEGEN) but this won't really make much difference in your full production as the most time consuming part is the SIM step where you run GEANT4 ==> It won't really make that much difference in time consumption (might help very little).
So I don't really disagree to your choice, it is valid point, but still it won't help you that much.

@agrohsje might give you answers on what to do.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/VBFToXToHH_UL/VBFToBulkGravitonToHH_M/VBFToBulkGravitonToHH_M_proc_card.dat
```diff
@@ -0,0 +1,12 @@
...
+set group_subprocesses Auto
+set ignore_six_quark_processes False
+set loop_optimized_output True
+set complex_mass_scheme False
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+import model RS_bulk_ktilda
+define q = u c d s b u~ c~ d~ s~ b~
+define p = u c d u~ c~ d~
+generate p p > y q q / g h , ( y > h h )
```
sihyunjeon, line 11: We observed some unhealthy behaviors when you use gauge unsafe syntaxes such as "/ g h".
Is it not possible to force ygg and hgg couplings to 0 and make the contributions negligible?
Also do you only need model parameter settings that are default in your model file's parameter card (I see LRS, ktilda, clight, ctop, cg, sw2)?

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/VBFToXToHH_UL/VBFToBulkGravitonToHH_M/VBFToBulkGravitonToHH_M_run_card.dat
```diff
@@ -0,0 +1,268 @@
...
+#*********************************************************************
+# Run to generate the grid pack                                      *
+#*********************************************************************
+  False     = gridpack  !True = setting up the grid pack
+#*********************************************************************
+# Number of events and rnd seed                                      *
+# Warning: Do not generate more than 1M events in a single run       *
+# If you want to run Pythia, avoid more than 50k events in a run.    *
+#*********************************************************************
+  100000 = nevents ! Number of unweighted events requested
```
sihyunjeon, line 32: You don't need to make this many events for pilot run. 2-3K would be enough.
Please change other run cards as well

atishelmanch: updated

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/VBFToXToHH_UL/VBFToRadionToHH_M/VBFToRadionToHH_M_proc_card.dat
atishelmanch: updated to 5F proton definition

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/VBFToXToHH_UL/VBFToBulkGravitonToHH_M/VBFToBulkGravitonToHH_M_proc_card.dat
atishelmanch: Hello @sihyunjeon : 

> We observed some unhealthy behaviors when you use gauge unsafe syntaxes such as "/ g h".

What kinds of unhealthy behaviors have you observed? 

> Is it not possible to force ygg and hgg couplings to 0 and make the contributions negligible?

We can do that if you feel strongly about it, however we already performed test runs and validated the simulation with this definition of process/parameters: [1]. Do you think these checks are sufficient? Or rather, do you think the unhealthy behavior you have observed may still be present and not visible in these checks? 

-Abe 

[1] https://twiki.cern.ch/twiki/pub/CMS/AbrahamTishelmanCharnyHomepage/29_July_2021_Res_VBF_HH_Validation.pdf

atishelmanch: Hello @sihyunjeon , @agrohsje - I would like to kindly ask if you have had a chance to view my reply above.

-Abe

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/VBFToXToHH_UL/VBFToBulkGravitonToHH_M/VBFToBulkGravitonToHH_M_proc_card.dat
```diff
@@ -0,0 +1,12 @@
+
+set group_subprocesses Auto
+set ignore_six_quark_processes False
+set loop_optimized_output True
+set complex_mass_scheme False
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+import model RS_bulk_ktilda
+define q = u c d s b u~ c~ d~ s~ b~
+define p = u c d s b u~ c~ d~ s~ b~
```
agrohsje, line 10: all models are 4 fs. you cannot add the massive b to the proton. also maxjetflavor and asrwgtflavor are set to 4. 
if you don't want to use 5fs, please remove the b's.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/VBFToXToHH_UL/VBFToBulkGravitonToHH_M/VBFToBulkGravitonToHH_M_proc_card.dat
agrohsje: Can you comment more which aspects you like to remove with /g h and why? I am not sure how you conclude from your slides that removing the photon is better.

sihyunjeon: Hey sorry for not responding. The possible issue is that this syntax induces possibilities of violating gauge invariance so it is typically much more safer to just turn off the couplings and avoid using such syntaxes. I mean, your slides look nice and everything seems to make sense, but who knows whether there could be differences or not?

atishelmanch:
Dear @agrohsje and @sihyunjeon,

We have made the following updates: 

- Created a new model at /afs/cern.ch/cms/generators/www/dibosonResonanceModel_toVBF.tar.gz which introduces the multipliers:

```
100 1.000000e+00 # cgluon
101 1.000000e+00 # chiggs
```

By default they're set to 1.0, but are set to zero in our customizecards in order to avoid possible violation of gauge invariance. 

- Reverted to 4F scheme 

Do you think anything else needs to be updated before these can be accepted? 

Thank you,
Abe

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/VBFToXToHH_UL/VBFToBulkGravitonToHH_M/VBFToBulkGravitonToHH_M_extramodels.dat
```diff
@@ -1,2 +1,2 @@
 ## Models for searches of diboson resonances
-dibosonResonanceModel_toVBF.tar.gz
+dibosonResonanceModel_toVBF_onefolder.tar.gz
```
atishelmanch, line 2: Pointing to an updated model which contains a single folder for necessary directory structure when importing models in madgraph cards.

sihyunjeon:
Hello this looks fine so will merge.

Just to check quickly, maybe it would be worth checking whether "cgluon, chiggs = 0.0" cross sections agree with "/ g h" cross sections. Would you let us know?

PR #2975, opened 2021-08-11
Add Radion To HH to bbtautau with a VBF production mode to 13 TeV cards (intended for UL)

aloeliger:
Radion to HH(bbtautau) cards for the VBF production mode for the B2G HH(bbtautau) analysis. 
Cards have been tested for making gridpacks and are intended for UL production.
Included is the template, and the script for making cards for relevant mass points,

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson/Spin-0/VBFToRadionToHH_htatahbb/Template/VBFToRadionToHH_htatahbb_MMASS_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson/Spin-0/VBFToRadionToHH_htatahbb/Template/VBFToRadionToHH_htatahbb_MMASS_proc_card.dat
@@ -1,0 +1,11 @@
+set group_subprocesses Auto
+set ignore_six_quark_processes False
+set loop_optimized_output True
+set complex_mass_scheme False
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+import model heft_radion
+define q = u c d s b u~ c~ d~ s~ b~
+define p = u c d u~ c~ d~
+generate p p > h2 q q  / g h , ( h2 > H H, H > ta+ ta-, H > b b~)
+output VBFToRadionToHH_htatahbb_MMASS -nojpeg
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson/Spin-0/VBFToRadionToHH_htatahbb/Template/VBFToRadionToHH_htatahbb_MMASS_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson/Spin-0/VBFToRadionToHH_htatahbb/Template/VBFToRadionToHH_htatahbb_MMASS_proc_card.dat
@@ -6,6 +6,6 @@ define vl = ve vm vt
 define vl~ = ve~ vm~ vt~
 import model heft_radion
 define q = u c d s b u~ c~ d~ s~ b~
-define p = u c d u~ c~ d~
+define p = g u c d s u~ c~ d~ s~
 generate p p > h2 q q, ( h2 > H H, H > ta+ ta-, H > b b~)
 output VBFToRadionToHH_htatahbb_MMASS -nojpeg
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson/Spin-0/VBFToRadionToHH_htatahbb/Template/VBFToRadionToHH_htatahbb_MMASS_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson/Spin-0/VBFToRadionToHH_htatahbb/Template/VBFToRadionToHH_htatahbb_MMASS_proc_card.dat
@@ -7,5 +7,5 @@ define vl~ = ve~ vm~ vt~
 import model heft_radion
 define q = u c d s b u~ c~ d~ s~ b~
 define p = u c d u~ c~ d~
-generate p p > h2 q q  / g h , ( h2 > H H, H > ta+ ta-, H > b b~)
+generate p p > h2 q q, ( h2 > H H, H > ta+ ta-, H > b b~)
 output VBFToRadionToHH_htatahbb_MMASS -nojpeg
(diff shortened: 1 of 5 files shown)


bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson/Spin-0/VBFToRadionToHH_htatahbb/Template/VBFToRadionToHH_htatahbb_MMASS_proc_card.dat
```diff
@@ -0,0 +1,11 @@
+set group_subprocesses Auto
+set ignore_six_quark_processes False
+set loop_optimized_output True
+set complex_mass_scheme False
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+import model heft_radion
+define q = u c d s b u~ c~ d~ s~ b~
+define p = u c d u~ c~ d~
+generate p p > h2 q q  / g h , ( h2 > H H, H > ta+ ta-, H > b b~)
```
sihyunjeon, line 10: The process with "/ g h" is unsafe in terms of gauge invariance and might lead to unphysical results.

image
If this was used to only take VBF diagrams (bottom) you also need to realize that you still get non-VBF diagrams (above).
So I think it would be much more safe/correct thing to do if you give WBF or mmjj cuts instead of using "/ g h". And this is more efficient as in your analysis level you might require deta(jj) or m(jj)

bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson/Spin-0/VBFToRadionToHH_htatahbb/Template/VBFToRadionToHH_htatahbb_MMASS_proc_card.dat
```diff
@@ -0,0 +1,11 @@
+set group_subprocesses Auto
+set ignore_six_quark_processes False
+set loop_optimized_output True
+set complex_mass_scheme False
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+import model heft_radion
+define q = u c d s b u~ c~ d~ s~ b~
+define p = u c d u~ c~ d~
```
sihyunjeon, line 9: What is the reason for taking this 3 flavour scheme?

bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson/Spin-0/VBFToRadionToHH_htatahbb/Template/VBFToRadionToHH_htatahbb_MMASS_proc_card.dat
aloeliger: This was taken from an example gridpack that was intended as an example for the lower mass analysis, but they upon review they warned us that some diagrams were incorrect. The `/ g h` has been removed in the latest commits

atishelmanch:
Hello - I would just like to comment that we have an open PR for a common set of UL HH VBF gridpacks [1], where the strategy within the HH working group under HIG is to use these among all HH channels where each channel decays HH with pythia. I'm not sure about B2G or this analysis in particular, but if there is a resonant HH combination I imagine it would be ideal to use the same gridpacks for consistency. 

-Abe 

[1] https://github.com/cms-sw/genproductions/pull/2972

aloeliger:
@atishelmanch Except for the most recent commits, and the dedicated bbtautau decay channel, these should be direct copies of yours, provided to us by the LLR group working on the low mass version of this analysis in HIG. I thought it might be better to try and create dedicated cards so that we could get our high mass points produced, and because I don't know what the overall efficiency/statistics might be after gen level cuts for our decay channel.

aloeliger:
@sihyunjeon Just wanted to bring this back to your attention. I have made the changes requested. Are there other changes that should be made before gridpacks are made from these cards? Thank you.

agrohsje:
sorry. i need to check a bit more with the mixing of massless and massive b's and the right settings. i think there are some inconsistencies. i need to better understand from the code.

sihyunjeon:
Sorry the blame is on me (I've been busy with other issues - still ongoing) But I will get back to you this week at least... Sorry

sihyunjeon:
So if I understood correctly what Alexander was concerned about :
"p" and "q" definitions are different ("p" including "b" and "q" not including "b").
MG run card automatically launches "b" related settings (drbb, drbj, ptb, ...) as p does not include b.
But because your "q" include "b", the concern was on the different cuts applied to drbb, drjj for instance.
If you have both "q"s to be "b" then there is no cut applied but when they are not "b", drjj>0.4 cut is applied.
So there could be some weird distortions in the phase spaces which you didn't intend.

I asked the MadGraph author if there is a systematic/elegant way to handle this. BUT the answer was NO. If you differently define your "p" and "q", you should check the consistency since MadGraph cannot handle this itself (in other words, it's not a safe way to generate something).

aloeliger:
@sihyunjeon Then am I understanding correctly that the best way to get the dedicated VBF production counterpart to our main signal (located here: https://github.com/cms-sw/genproductions/blob/5a224fc42aca2a097afb6fe79cdc7d11ecee106b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/exo_diboson/Spin-0/Radion_hh_htatahbb_incl_narrow/Template/Radion_hh_htatahbb_incl_narrow_MMASS_proc_card.dat) would be to add the "b" into the "p" definition, or is that not an option?

agrohsje:
Hi @aloeliger. The model as you invoke it now has a massive b, so you cannot add it to the p. You can make a 5FS version of the model where the b is massless. In this case you can also drop the b in the p, use a 4FS PDF, but add the b to the jets.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/exo_diboson/Spin-0/VBFToRadionToHH_htatahbb/Template/VBFToRadionToHH_htatahbb_MMASS_run_card.dat
```diff
@@ -0,0 +1,268 @@
...
+#*********************************************************************
+# Run to generate the grid pack                                      *
+#*********************************************************************
+  False     = gridpack  !True = setting up the grid pack
+#*********************************************************************
+# Number of events and rnd seed                                      *
+# Warning: Do not generate more than 1M events in a single run       *
+# If you want to run Pythia, avoid more than 50k events in a run.    *
+#*********************************************************************
+  100000 = nevents ! Number of unweighted events requested
```
sihyunjeon, line 32: reduce this to 2-3K. don't need 100K for pilot run

sihyunjeon:
@aloeliger ping. Let us know what your choice will be.
Synchronizing p and j OR synchronizing j and b related cuts

sihyunjeon:
@aloeliger Any news?

aloeliger:
@sihyunjeon Apologies, it has been a while since I looked at this, I will take a look at it again soon.

acarvalh:
Hi. As this is not merged yet I sincerely recommend using the set of cards pointed out by @atishelmanch . The physics is the same, but besides that being already merged, also the same set of gridpacks can be used (that may be already ready, you two can take it offline), and using the same gridpacks we are absolutely sure to be following the same mass grid for future comparisons.
As a plus, there is the case of spin 0 and 2 there, so it is more complete physics coverage.

bbilin:

PR #2976, opened 2021-08-12
updating SM VBS hadronic cards

stalbrec:
Hi,

this PR updates the SM VBS hadronic cards as they'll be used for the Summer20 samples:

 - it removes the final states `ZBBZJJnoBjj`, `ZNuNuZBBjj`, `ZNuNuZJJnoBjj` as those were and will not be produced,
 - `$ t t~` was removed for osWW (in cards for QCD and EWK+QCD production) after discussion with theorists,
 - adding additional final state `ZNuNuZJJjj`.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/VBS/VVjj_hadronic/EWK/ZNuNuZJJjj_EWK_LO/ZNuNuZJJjj_EWK_LO_run_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/VBS/VVjj_hadronic/EWK/ZNuNuZJJjj_EWK_LO/ZNuNuZJJjj_EWK_LO_run_card.dat
@@ -1,0 +1,272 @@
+#*********************************************************************
+#                       MadGraph5_aMC@NLO                            *
+#                                                                    *
+#                     run_card.dat MadEvent                          *
+#                                                                    *
+#  This file is used to set the parameters of the run.               *
+#                                                                    *
+#  Some notation/conventions:                                        *
+#                                                                    *
+#   Lines starting with a '# ' are info or comments                  *
+#                                                                    *
+#   mind the format:   value    = variable     ! comment             *
+#*********************************************************************
+#
+#*******************                                                 
+# Running parameters
+#*******************                                                 
+#                                                                    
+#*********************************************************************
+# Tag name for the run (one word)                                    *
+#*********************************************************************
+  tag_1     = run_tag ! name of the run 
+#*********************************************************************
+# Run to generate the grid pack                                      *
+#*********************************************************************
+  False     = gridpack  !True = setting up the grid pack
+#*********************************************************************
+# Number of events and rnd seed                                      *
+# Warning: Do not generate more than 1M events in a single run       *
+# If you want to run Pythia, avoid more than 50k events in a run.    *
+#*********************************************************************
+  1000 = nevents ! Number of unweighted events requested 
+  0   = iseed   ! rnd seed (0=assigned automatically=default))
+#*********************************************************************
+# Collider type and energy                                           *
+# lpp: 0=No PDF, 1=proton, -1=antiproton, 2=photon from proton,      *
+#                                         3=photon from electron     *
+#*********************************************************************
+     1        = lpp1    ! beam 1 type 
+     1        = lpp2    ! beam 2 type
+     6500.0     = ebeam1  ! beam 1 total energy in GeV
+     6500.0     = ebeam2  ! beam 2 total energy in GeV
+#*********************************************************************
+# Beam polarization from -100 (left-handed) to 100 (right-handed)    *
+#*********************************************************************
+     0.0     = polbeam1 ! beam polarization for beam 1
+     0.0     = polbeam2 ! beam polarization for beam 2
+#*********************************************************************
+# PDF CHOICE: this automatically fixes also alpha_s and its evol.    *
+#*********************************************************************
+     'lhapdf'    = pdlabel     ! PDF set                                     
+$DEFAULT_PDF_SETS  = lhaid     ! if pdlabel=lhapdf, this is the lhapdf number
+$DEFAULT_PDF_MEMBERS = reweight_PDF    
+#*********************************************************************
+# Renormalization and factorization scales                           *
+#*********************************************************************
+ False = fixed_ren_scale  ! if .true. use fixed ren scale
+ False        = fixed_fac_scale  ! if .true. use fixed fac scale
+ 91.188  = scale            ! fixed ren scale
+ 91.188  = dsqrt_q2fact1    ! fixed fact scale for pdf1
+ 91.188  = dsqrt_q2fact2    ! fixed fact scale for pdf2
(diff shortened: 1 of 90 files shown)


sihyunjeon:
Hi,
There is no need to remove unused final states. those cards could be used for future reference at any time so I'd suggest to include them back in although it won't be produced by your analysis team or PAG/POGs

stalbrec:
Hi, 
alright, thanks for the fast feedback.
How should i undo the commit? Using a `git revert ...` or should i rebase?

agrohsje:
You should be able to rebase. Otherwise I can also close and you do a new PR. Let me know.

stalbrec:
Okay, i managed to rebase. 
Is there something else to do to the cards?

stalbrec:
Hi, i wanted to follow up again, if there is something that needed to be done before this will be merged.
Thanks!

sihyunjeon:
Hello stalbrec, Sorry for the late reply. Is there a reason to use MadSpin?
The physics will be the same if you do "generate p p > z z j j, z > nu nu~, z > b b~" in your proc_card. But as far as I know, MadSpin only runs on single core so...

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/VBS/VVjj_hadronic/EWK/ZNuNuZJJjj_EWK_LO/ZNuNuZJJjj_EWK_LO_run_card.dat
```diff
@@ -0,0 +1,272 @@
...
+ 0.4 = drjj    ! min distance between jets 
+ 0.4 = drbb    ! min distance between b's 
+ 0.4 = drll    ! min distance between leptons 
+ 0.4 = draa    ! min distance between gammas 
+ 0.4 = drbj    ! min distance between b and jet 
+ 0.4 = draj    ! min distance between gamma and jet 
+ 0.4 = drjl    ! min distance between jet and lepton 
+ 0.0 = drab    ! min distance between gamma and b 
+ 0.0 = drbl    ! min distance between b and lepton 
+ 0.4 = dral    ! min distance between gamma and lepton
```
agrohsje, line 158: despite vbf topologies you might lower dr cuts?

agrohsje:
Like the proposal of @sihyunjeon . You might consider lowering dR cuts as you have mjj to assure vbf topology. 
Otherwise looks good to me.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/VBS/VVjj_hadronic/EWK/ZNuNuZJJjj_EWK_LO/ZNuNuZJJjj_EWK_LO_run_card.dat
stalbrec: Are you especially talking about lowering them (meaning 0.4->0.0 ?) for the ZNuNuZJJ final state, or for all of them in the directory?

agrohsje:
As the cuts are not on purpose (some even totally irrelevant) I would just set all to 0.01 or some tiny value that has no impact.

agrohsje:
hi @stalbrec , i cleaned up the process cards. i assume the dr cuts are not really used here, but a cut on the invariant jet mass. what about setting all drcuts to 0.01? or will you impose dr jj jet cuts at analysis level which are larger than 0.4 ?

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/VBS/VVjj_hadronic/EWK/ZNuNuZJJjj_EWK_LO/ZNuNuZJJjj_EWK_LO_run_card.dat
amarini: @agrohsje , these cuts were already in the current production for e.g. the single leptons. Do you think this could have been a problem that we should fix?

stalbrec:
Hi @agrohsje , sorry for the delay. Thank you very much!
I will change the respective dr mins and update the cards to not use madspin now.
I will also remove the redundant lines in the `proc_card.dat`, that you cleaned from the files already touched in this PR, from the remaining VBS_hadronic cards, correct?

agrohsje:
@stalbrec to answer @amarini question: do you mind running with and without the cuts and see if there is any impact on the cross section? if not the cuts are inferior to mmjj and the old production is fully ok.

stalbrec:
Hi @agrohsje - We actually already discussed this briefly on Mattermost, and i planned to circulate the plots here as well, but forgot.
I created a gridpack and some (in the plots 50k) LHE events (for the process `p p > z z j j QED=4 QCD=0, (z > j j)`) comparing:
- the setup with the Summer19 cards (`_withMadspin`), 
- with moving the decay from madspin to the proc_card (`_noMadspin`), 
- and  without Madspin and without dRcuts (or rather with your suggestion of a dr_min=0.01) (`_noMadspin_nodR`).

There is no significant difference in the distributions i looked at. Here for example the invariant mass of the diboson system:
mvv

For the  between the decay products of the first Z boson in the event, there is no difference visible:

dr

However, I'm wondering if it is expected to have this  to have values below 0.4 even for the Summer19 setup where the cut is still `drjjmin = 0.4`?

stalbrec:
Okay, regarding my last question. I didn't know the line `False  = cut_decays  ! Cut decay products` existed in the run_card.
I guess this answers my question completely.

agrohsje:
Right. The interesting quantity is the dR of the VBS jets. The question is if mmjj  makes the dR anyway to be above 0.4.

stalbrec:
Yes, thank you - I understand this better now.
This is the  of the two quarks with `status==1` that are not originating from a Z-decay.

Screenshot 2021-09-22 at 11 54 23

So here one can see that the mjj cut does the job, as you said.
In the sample without  (`drjjmin=0.01`), i get (what seems to be) **one** event with

sihyunjeon:
Hello, just one small note here. MadSpin won't really change the physics here as I said above.
1. MadGraph : generate p p > z + MadSpin : z > e+ e-
2. MadGraph : generate p p > z, z > e+ e-
These two are exactly the same things.

sihyunjeon:
I think the settings are fine (plots look great) "without dR cuts and without MadSpin" so I guess you can update your pull request accordingly

stalbrec:
Hi, thanks. I removed the madspin cards, added the decays into the line(s) in the proc_card, cleaned the remaining proc_cards and changed  the cuts to drmin=0.01

stalbrec:
Oh i missed some of the cards' directories, so there are decays missing from the proc_card. I will change it now

stalbrec:
Okay I fixed it.

agrohsje:
the pr hits a bit the limit of what a human being can review. but i think it is correct. @sihyunjeon do you mind a final check from your side? thanks!

amarini:
any update?

PR #2978, opened 2021-08-24
Changing location of madspin_card.dat in Madgraph LO gridpack

stalbrec:
Hi,

this changes the location of the `madspin_card.dat` in the LO Madgraph gridpacks.
I changed the respective paths in `gridpack_generation.sh` as well as `runcmsgrid_LO.sh` from
`$WORKDIR/process/madspin_card.dat` to `$WORKDIR/process/madevent/Cards/madspin_card.dat`.

This helps avoid problems when any of the runs in `runcmsgrid_LO.sh` fail to generate any events.
This would lead to the removal of `madspin_card.dat` in the `process` dir and consequently the skipping of madspin.
The card is removed from the `process` dir in those cases, since this is where madgraphs `run.sh` is called and this script is removing any `*.dat` cards from the current directory if `./madevent/Events/GridRun_${seed}/unweighted_events.lhe.gz` does not exists (see [Template/LO/bin/internal/Gridpack/run.sh#L60](https://bazaar.launchpad.net/~madteam/mg5amcnlo/3.x/view/head:/Template/LO/bin/internal/Gridpack/run.sh#L60) ).

I'm not sure if this is the preferred way of 'fixing' this - so let me know if I should change anyhting. 
(I'm not sure if the madspin_card is expected in the `process` dir by other scripts.)

Cheers
Steffen

### DIFF
--- a/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
+++ b/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
@@ -583,7 +583,6 @@ make_gridpack () {
         $WORKDIR/$MGBASEDIRORIG/MadSpin/madspin madspinrun.dat 
         rm madspinrun.dat
         rm -rf tmp*
-        cp $CARDSDIR/${name}_madspin_card.dat $WORKDIR/process/madspin_card.dat
       fi
     
       echo "preparing final gridpack"
--- a/bin/MadGraph5_aMCatNLO/runcmsgrid_LO.sh
+++ b/bin/MadGraph5_aMCatNLO/runcmsgrid_LO.sh
@@ -143,7 +143,10 @@ fi
 gzip -d $LHEWORKDIR/process/events.lhe.gz
 
 domadspin=0
-if [ -f ./madspin_card.dat ] ;then
+#checking for a madspin_card in the InputCards directory,
+#rather than checking for it in the process directory, since run.sh will remove *.dat from that directory,
+#if it fails to generate any event in the first iteration 
+if [ -f $LHEWORKDIR/InputCards/*madspin_card.dat ] ;then
     domadspin=1
     # extract header as overwritten by madspin 
     if grep -R "<initrwgt>" events.lhe ; then
@@ -152,7 +155,7 @@ if [ -f ./madspin_card.dat ] ;then
     echo "import events.lhe" > madspinrun.dat
     rnum2=$(($rnum+1000000))
     echo `echo "set seed $rnum2"` >> madspinrun.dat
-    cat ./madspin_card.dat >> madspinrun.dat
+    cat $LHEWORKDIR/InputCards/*madspin_card.dat >> madspinrun.dat
     $LHEWORKDIR/mgbasedir/MadSpin/madspin madspinrun.dat 
     # add header back 
     gzip -d events_decayed.lhe.gz  


stalbrec:
Hi,
sorry, i've been too quick to judge by thinking this works based on the tests i ran. With the madspin_card.dat in the madevent directory, MadGraph will start MadSpin on its own, correct?
I could change it, so `runcmsgrid.sh` would look for a `madspin_card.dat` in die `$LHEWORKDIR/InputCards/*madspin_card.dat` - would this be okay?

agrohsje:
Hi @stalbrec , I am not sure I understand your second concern.
madspin_card.dat is only in madevent  if we copy it there, i.e. if it is needed. what is the problem you have in mind ?

stalbrec:
My second concern is just with the way i tried to "fix" my first concern, by moving the madspin_card.dat to the madevent directory (as implemented in the PR changes right now).
I think it would be better to not move it at all and just check in [runcmsgrid_LO.sh#L146](https://github.com/cms-sw/genproductions/blob/c0a7caff12f6d831b2d47c463928b98f5e3b0c92/bin/MadGraph5_aMCatNLO/runcmsgrid_LO.sh#L146) if there is a madspin_card.dat in `$LHEWORKDIR/InputCards/` rather than in the `process` dir (as it is done at the moment), where it could be removed by `run.sh` whenever madgraph fails to generate any events (i.e. my first concern).
Does that make sense?

agrohsje:
Your point is that as soon as one of the run.sh in this part
https://github.com/cms-sw/genproductions/blob/e35d22b671fbcecccdfe117105fbee9f4aebe8f8/bin/MadGraph5_aMCatNLO/runcmsgrid_LO.sh#L78
fails the card is removed and madspin doesn't run anymore?
But you like to run it as there are still valid events from the rest?

stalbrec:
Yes. I noticed this, when i produced a LHE file for testing and it printed only an error in the first attempt to produce events. 
Later on it just finished up and i ended up with a LHE file with Events containing undecayed Z-bosons.

agrohsje:
You didn't get an error from web blabla still active please delete first? Do you still have the log?

stalbrec:
I know which error you are referring to, but no, as far as i can remember (unfortunately i don't have the log anymore - I'm trying to reproduce one right now), that error did not came up in those instances.
Also whenever i get that error asking to delete the RunWeb i end up with a failed run and no lhe at all.

stalbrec:
Here you can find the log where the situation i described above occurred: `/afs/desy.de/user/a/albrechs/public/PR2978/runcmsgrid.log`
The workdir with the full (untarred) gridpack is here: `/nfs/dust/cms/user/albrechs/VBS/August21_SMEFT_models/ZJJZJJjj_EWK_LO`

agrohsje:
Hi @stalbrec , I am fine with your updated proposal. Do you want to change the PR? Maybe you can add a short comment in the code why this change is done?   @sihyunjeon  are you ok with the proposal?

stalbrec:
Hi @agrohsje , i changed it and added a comment.
Should the line in [`gridpack_generation.sh#L586`](https://github.com/cms-sw/genproductions/blob/a1c9dd43d7ec49a56dc6aef3fcc94ef62196cbac/bin/MadGraph5_aMCatNLO/gridpack_generation.sh#L586), where the `madspin_card.dat` is moved to the `process` directory, be removed?

agrohsje:
You still need that for the pilot run.

stalbrec:
Isn't that what [`L580-585`](https://github.com/cms-sw/genproductions/blob/a1c9dd43d7ec49a56dc6aef3fcc94ef62196cbac/bin/MadGraph5_aMCatNLO/gridpack_generation.sh#L580-L585) are doing?

agrohsje:
sorry. you are right.

agrohsje:
@sihyunjeon : can you double check and merge if ok ? thanks!

sihyunjeon:
Apologize for my laziness, merging :) Thanks

PR #2983, opened 2021-09-11
update with filter efficiency

menglu21:


menglu21:
Hi @sihyunjeon @agrohsje could you please have a look on these codes, they are for the GEN validation

agrohsje:
Thanks for the PR. @sihyunjeon feel free to merge.

bin/GenValidation/scripts/Plot_code/histo.py
```diff
@@ -0,0 +1,364 @@
...
+  if tmproot<0:
+    pz=-1.*B/(2*A)
+  else:
+    tmpsol1 = (-B + math.sqrt(tmproot))/(2.0*A)
+    tmpsol2 = (-B - math.sqrt(tmproot))/(2.0*A)
+    if abs(tmpsol2-pzl)<abs(tmpsol1-pzl):
+      pz=tmpsol2
+    else:
+      pz=tmpsol1
+  if abs(pz)>300:
```
sihyunjeon, line 56: Just out of curiosity, is this how W mass is reconstructed usually in our CMS analysis?

sihyunjeon:
Sorry for the late respond, merged :)

bin/GenValidation/scripts/Plot_code/histo.py
menglu21: comparing with the HWW analysis, the methods are the same when no real solution exist, i.e., using the real part of the complex solution. The difference is that in the case of B*B-4AC>0, HWW always uses the smaller solution. But here i use the smaller one only when solution of neutrino pz is quite large, otherwise i use the one closer to the lepton pz.

sihyunjeon: Ah ok thanks :)

PR #2986, opened 2021-09-13
ZprimeTo2ChiTo2L cards for Snowmass Analysis

hjkwon260:
Cards for signal sample production(Z’->charginos->ll+MET) for Snowmass analysis. Imports extra model named 'UMSSM', where the UFO was given from theory authors. Also provided parameter card for this benchmark point.
'UMSSM.tar’ is not present in central directory yet(will need to be uploaded), have tested gridpack production from local files.
FYI: @Soumyatifr @Saptaparna @ssekmen


bin/MadGraph5_aMCatNLO/cards/production/2017/14TeV/ZprimeTo2ChiTo2L/ZprimeTo2ChiTo2L_extramodels.dat
```diff
@@ -0,0 +1 @@
+UMSSM.tar
```
agrohsje, line 1: Please provide the corresponding UFO. It is not on our model page.

bin/MadGraph5_aMCatNLO/cards/production/2017/14TeV/ZprimeTo2ChiTo2L/ZprimeTo2ChiTo2L_proc_card.dat
```diff
@@ -0,0 +1,40 @@
...
+set loop_optimized_output True
+set loop_color_flows False
+set gauge unitary
+set complex_mass_scheme False
+set max_npoint_for_channel 0
+import model sm
+define p = g u c d s u~ c~ d~ s~
+define j = g u c d s u~ c~ d~ s~
+define l+ = e+ mu+
+define l- = e- mu-
```
agrohsje, line 34: please remove what is not already default. especially the version 2.5.5 etc.

bin/MadGraph5_aMCatNLO/cards/production/2017/14TeV/ZprimeTo2ChiTo2L/ZprimeTo2ChiTo2L_proc_card.dat
```diff
@@ -0,0 +1,40 @@
...
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+import model UMSSM -modelname
+generate p p > zp > c1 c1bar, (c1 > wm n1, (wm > l- vl~)), (c1bar > wm\
```
agrohsje, line 38: should the l include a tau ? 
is the model 4SF of 5FS?

bin/MadGraph5_aMCatNLO/cards/production/2017/14TeV/ZprimeTo2ChiTo2L/ZprimeTo2ChiTo2L_run_card.dat
```diff
@@ -0,0 +1,282 @@
...
+ -1.0   = ebmax ! maximum E for the b
+ -1.0   = eamax ! maximum E for the photons
+ -1.0   = elmax ! maximum E for the charged leptons
+#*********************************************************************
+# Maximum and minimum absolute rapidity (for max, -1 means no cut)   *
+#*********************************************************************
+  5.0 = etaj    ! max rap for the jets 
+  -1.0  = etab    ! max rap for the b
+ 2.5  = etaa    ! max rap for the photons 
+ 2.5  = etal    ! max rap for the charged leptons
```
agrohsje, line 141: what are your  analysis cuts ? keep in mind gen cuts shouldn't be too close.

bin/MadGraph5_aMCatNLO/cards/production/2017/14TeV/ZprimeTo2ChiTo2L/ZprimeTo2ChiTo2L_run_card.dat
```diff
@@ -0,0 +1,282 @@
...
+ 0.4 = drjj    ! min distance between jets 
+ 0.0   = drbb    ! min distance between b's 
+ 0.4 = drll    ! min distance between leptons 
+ 0.4 = draa    ! min distance between gammas 
+ 0.0   = drbj    ! min distance between b and jet 
+ 0.4 = draj    ! min distance between gamma and jet 
+ 0.4 = drjl    ! min distance between jet and lepton 
+ 0.0   = drab    ! min distance between gamma and b 
+ 0.0   = drbl    ! min distance between b and lepton 
+ 0.4 = dral    ! min distance between gamma and lepton
```
agrohsje, line 158: same here. lower to 0.01 where not intended.

bin/MadGraph5_aMCatNLO/cards/production/2017/14TeV/ZprimeTo2ChiTo2L/ZprimeTo2ChiTo2L_run_card.dat
```diff
@@ -0,0 +1,282 @@
...
+#  will be used by SysCalc (if installed)
+#**************************************                                  
+#
+0.5 1 2 = sys_scalefact  # factorization/renormalization scale factor
+None = sys_alpsfact  # \alpha_s emission scale factors
+auto = sys_matchscale # variation of merging scale
+# PDF sets and number of members (0 or none for all members).
+NNPDF23_lo_as_0130_qed = sys_pdf # list of pdf sets
+# MSTW2008nlo68cl.LHgrid 1  = sys_pdf
+#
```
agrohsje, line 282: important to remove this part.

bin/MadGraph5_aMCatNLO/cards/production/2017/14TeV/ZprimeTo2ChiTo2L/ZprimeTo2ChiTo2L_extramodels.dat
hjkwon260: could you provide me the path? any write permission needed?

sihyunjeon: Upload it in public lxplus repo, we can handle it

bin/MadGraph5_aMCatNLO/cards/production/2017/14TeV/ZprimeTo2ChiTo2L/ZprimeTo2ChiTo2L_proc_card.dat
sihyunjeon: The model is 5FS, I checked

bin/MadGraph5_aMCatNLO/cards/production/2017/14TeV/ZprimeTo2ChiTo2L/ZprimeTo2ChiTo2L_proc_card.dat
sihyunjeon: Remove the lines that aren't used as well : e.g. import model sm

bin/MadGraph5_aMCatNLO/cards/production/2017/14TeV/ZprimeTo2ChiTo2L/ZprimeTo2ChiTo2L_proc_card.dat
sihyunjeon: If you are doing analysis for onshell zp, it is better to split the ME calculations like "p p > zp, zp > c1 c1bar". It gives you some unhealthy behaviors [1]

[1] https://twiki.cern.ch/twiki/bin/viewauth/CMS/MGSyntaxPitfall#Avoid_using_s_in_series_without

bin/MadGraph5_aMCatNLO/cards/production/2017/14TeV/ZprimeTo2ChiTo2L/ZprimeTo2ChiTo2L_run_card.dat
```diff
@@ -0,0 +1,282 @@
...
+#*********************************************************************
+# Tag name for the run (one word)                                    *
+#*********************************************************************
+  tag_1     = run_tag ! name of the run 
+#*********************************************************************
+# Number of events and rnd seed                                      *
+# Warning: Do not generate more than 1M events in a single run       *
+# If you want to run Pythia, avoid more than 50k events in a run.    *
+#*********************************************************************
+  10000 = nevents ! Number of unweighted events requested
```
sihyunjeon, line 28: You don't need 10K events for pilot run (would only be time consuming) 3K would be fine

bin/MadGraph5_aMCatNLO/cards/production/2017/14TeV/ZprimeTo2ChiTo2L/ZprimeTo2ChiTo2L_run_card.dat
sihyunjeon: I guess these cuts won't have an effect to the gridpacks since cut_decays is set to False. Though it would be nice if @hjkwon260 can confirm if you intended to give the phase space cuts OR did not consider changing the values here as you set cut_decays to False

bin/MadGraph5_aMCatNLO/cards/production/2017/14TeV/ZprimeTo2ChiTo2L/ZprimeTo2ChiTo2L_run_card.dat
sihyunjeon: Yup. remove line 270-282

bin/MadGraph5_aMCatNLO/cards/production/2017/14TeV/ZprimeTo2ChiTo2L/ZprimeTo2ChiTo2L_proc_card.dat
hjkwon260: > should the l include a tau ?

we don't need tau. I can remove vt for sure.

hjkwon260: > If you are doing analysis for onshell zp, it is better to split the ME calculations like "p p > zp, zp > c1 c1bar". It gives you some unhealthy behaviors [1]
> 
> [1] https://twiki.cern.ch/twiki/bin/viewauth/CMS/MGSyntaxPitfall#Avoid_using_s_in_series_without

It's onshell decay. Ok, I'll modify the syntax, thanks for the twiki

bin/MadGraph5_aMCatNLO/cards/production/2017/14TeV/ZprimeTo2ChiTo2L/ZprimeTo2ChiTo2L_run_card.dat
hjkwon260: we use much tighter cuts in analysis. in addition to it, I let disable to use cuts, so simply didn't modify further from default MG cuts

bin/MadGraph5_aMCatNLO/cards/production/2017/14TeV/ZprimeTo2ChiTo2L/ZprimeTo2ChiTo2L_run_card.dat
hjkwon260: ok, thanks

bin/MadGraph5_aMCatNLO/cards/production/2017/14TeV/ZprimeTo2ChiTo2L/ZprimeTo2ChiTo2L_proc_card.dat
agrohsje: what about taus that decay to an e or a mu in the end ?

hjkwon260: considering final lepton pt cuts, BR, etc, there seems not much sensitivity gain from tau contribution. though let me talk with my analysis team

sihyunjeon:
How do you vary zprime and chi masses, or do you only aim to make single mass point? I see only one parameter file and wonder how you control them

hjkwon260:
We use only 1 mass point for now

agrohsje:
Thanks. Seems only the tau question is left.  @sihyunjeon anything else ?

hjkwon260:
https://github.com/cms-sw/genproductions/pull/2986/commits/a3985075883f15417827819a268e54c3ca74b89d
@agrohsje does default l definition contains taus? we are fine to add tau decays as well, it's not harmful anyway. Thanks!

agrohsje:
I just added the relevant lines. Please test and let me know.

hjkwon260:
I checked calculated cross section, which increased ~9/4 times as expected

agrohsje:
why did you change the order ? 
a certain model involves definitions. first you load the model, then this comes with definitions and then you change whatever you don't like.

hjkwon260:
I wasn't able to sucess gridpack without the change. Error massage:

> interrupted in sub-command:
"define l+ = e+ mu+ ta+" with error:
InvalidCmd : No particle e+ in model
command not executed: define l- = e- mu- ta-

probably because the model define lepton differently, i.e e1,e2,e3

sihyunjeon:
You should then change the command lines accordingly
The current setting is actually meaningless

agrohsje:
If the model has a different naming convention you need to adjust. That was exactly my comment above: first you load the model, then you change. I just used the regular convention to help. If the model is different, this will not work.

hjkwon260:
I see below lines are equivalent to the latest commit:
> import model UMSSM -modelname
define p = g u1 u2 d1 d2 u1bar u2bar d1bar d2bar
define l- = e1 e2 e3 
define l+ = e1bar e2bar e3bar 
define vl = nu1 nu2 nu3
define vl~ = nu1bar nu2bar nu3bar

It seems if define cmd comes before the model calling, multiparticles are defined from MG convention and internally converted with model's naming at the end. However if this is not a good practice, I'll commit the above lines which are explicitly defining with imported model convention. could you confirm please?

sihyunjeon:
If definition is written before the model loading, it could be overwritten. If in your UFO model, "l+ = e1 e2" is set as default, your "l+ = e+ mu+ ta+" would be overwritten by the model default settings. So it's better to write down anything after loading the model.
The one you suggested above looks right.

sihyunjeon:
@hjkwon260 ping

sihyunjeon:
@agrohsje The cards look fine to me

PR #2994, opened 2021-09-23
Adding genfragments for QstarToQW process, adjusting Qstar and W mass

wpmccormack:
This is the 1+2 prong di-jet topology for the CASE analysis.  Scan over Q* and W masses.


sihyunjeon:
Is this for UL request?

wpmccormack:
Hi Sihyun, yes, this is for an UL request for the CASE analysis

sihyunjeon:
But why did you not use CPX tunes? I even never heard of 4C tune actually :|
@agrohsje Is this an old tune?

wpmccormack:
Ah, these genfragments are just an adaptation of https://github.com/wpmccormack/genproductions/tree/add_QstarQW_Wmass/genfragments/ThirteenTeV/QstarToQW.  This is what was suggested to me to modify.  I'm not really and expert on tunes.  Maybe @thaarres can comment?

genfragments/ThirteenTeV/QstarToQW_Wmass/QstarToQW_M_2000_mW_170_TuneCUETP8M1_13TeV_pythia8_cfi.py
```diff
@@ -0,0 +1,35 @@
...
+			'4000001:onIfMatch = 2 24',
+			'4000002:m0 = 2000',
+			'4000002:onMode = off',
+			'4000002:onIfMatch = 1 24',
+                        '24:m0 = 170',
+
+		),
+		parameterSets = cms.vstring('pythia8CommonSettings',
+		                            'pythia8CUEP8M1Settings',
+		                            'processParameters')
```
agrohsje, line 31: you always need something like this to get the proper tune.

agrohsje: i thought you wanted cp2

agrohsje:
what do you have in mind with this pr? in principle it is sufficient to have this in mcm. you could also commit a template and a script here if you prefer. but note that l2 will review this when submitting the tickets. 
note there is a bug. 
		'Tune:pp 5',
			'Tune:ee 3',
are not in sync with cp2. please remove. this will be called by the lines i suggested to add.

wpmccormack:
Hi @agrohsje , this was totally my fault, I had a typo in my fragment creation sh file.  The new fragments should be corrected, and I deleted the ones for unneeded tunes as well.  What do you mean by "mcm"?  Sorry, I'm pretty new to CMS.  Thank you!

agrohsje:
MCM is the tool used to steer mc production in cms. Take e.g. this example: https://cms-pdmv.cern.ch/mcm/requests?prepid=B2G-Fall13-00001&page=0&shown=2621471 There is the field fragment, from this you can get the fragment that was used to produce the events. Now in case of a scan, that would always be the same fragment but with just modified mass settings. So in principle there is no need to push the fragment here. You could push one example and a script if you prefer, but I think committing all cards with just a change in mass is a bit too much.

genfragments/ThirteenTeV/QstarToQW_Wmass/QstarToQW_M_2000_mW_170_TuneCP2_13TeV_pythia8_cfi.py
```diff
@@ -0,0 +1,38 @@
+import FWCore.ParameterSet.Config as cms
+
+from Configuration.Generator.Pythia8CommonSettings_cfi import *
+from Configuration.Generator.MCTunes2017.PythiaCP2Settings_cfi import *
+from Configuration.Generator.PSweightsPythia.PythiaPSweightsSettings_cfi import *
+
+generator = cms.EDFilter("Pythia8GeneratorFilter",
+	comEnergy = cms.double(13000.0),
+	crossSection = cms.untracked.double(1),
```
sihyunjeon, line 9: You might want to remove this line,
 	crossSection = cms.untracked.double(1),
The value won't be true and this will give you an error when it's submitted to the central cores for production since it checks whether the one you wrote in the fragment and the result matches or not

wpmccormack:
Hi @agrohsje , I just wanted to check the status of this PR.  I would like to request some samples using these genfragments relatively soon

agrohsje:
note that in official production you might need to switch to concurrent. going to merge. physics settings look good.

PR #2995, opened 2021-09-23
Adding cards for W prime to B prime + Top production for CASE analysis

wpmccormack:
This is the 3+3 prong topology for the CASE analysis.  Scan of W' and B' masses

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/WpToBpTop/WpToBpT_conf/Zt_madspin_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/WpToBpTop/WpToBpT_conf/Zt_madspin_card.dat
@@ -1,0 +1,26 @@
+#************************************************************
+#*                        MadSpin                           *
+#*                                                          *
+#*    P. Artoisenet, R. Frederix, R. Rietkerk, O. Mattelaer * 
+#*                                                          *
+#*    Part of the MadGraph5_aMC@NLO Framework:              *
+#*    The MadGraph5_aMC@NLO Development Team - Find us at   *
+#*    https://server06.fynu.ucl.ac.be/projects/madgraph     *
+#*                                                          *
+#************************************************************
+
+set ms_dir ./madspingrid
+set max_running_process 1
+
+set Nevents_for_max_weigth 250 # number of events for the estimate of the max. weight
+#set BW_cut 15                # cut on how far the particle can be off-shell
+set max_weight_ps_point 400  # number of PS to estimate the maximum for each event
+
+define dprod = g u c d s b u~ c~ d~ s~ b~ a ve vm vt e- mu- ve~ vm~ vt~ e+ mu+ ta- ta+
+
+# specify the decay for the final state particles
+decay t > w+ b, w+ > dprod dprod
+decay t~ > w- b~, w- > dprod dprod
+decay z > dprod dprod
+# running the actual code
+launch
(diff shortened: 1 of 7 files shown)


agrohsje:
hi @wpmccormack please check my corrections. mainly cleaned up and a small fixes for 4fs. @sihyunjeon do you prefer moving madspin to the me command line? we could also fix the syntax so that we don't have s-channel specifications. no strong preference. what do you think?

wpmccormack:
Hi @agrohsje , this looks fine with me thank you!  I had just copied these cards from a pre-existing set, with some minor modification

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/WpToBpTop/WpToBpT_conf/Zt_proc_card.dat
```diff
@@ -0,0 +1,5 @@
+import model VLQVP_UFO
+generate p p > wp+ bp~ > z t b~
```
sihyunjeon, line 2: Does "wp+ decay to bp top" and "bp to z and b"?
The syntax is not so safe in some sense so it would be my naive suggestion to change like below :
p p > t bp~, bp~ > z b~
Of course this case you are assuming onshell bp production so it only works when you are doing some bump search for m(bp)

sihyunjeon: MassesWp = [2000, 3000, 5000]
MassesBp = [25,80,170,400]
MassesT = [25,80,170,400]

But looking at your mass settings m(Wp) >> m(Bp) so I think the syntax suggested would be OK

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/WpToBpTop/WpToBpT_conf/Zt_madspin_card.dat
```diff
@@ -0,0 +1,14 @@
...
+set Nevents_for_max_weigth 250 # number of events for the estimate of the max. weight
+#set BW_cut 15                # cut on how far the particle can be off-shell
+set max_weight_ps_point 400  # number of PS to estimate the maximum for each event
+
+define dec = u c d s b u~ c~ d~ s~ b~ ve vm vt e- mu- ta- ve~ vm~ vt~ e+ mu+ ta+
+
+# specify the decay for the final state particles
+decay t > w+ b, w+ > dec dec
+decay t~ > w- b~, w- > dec dec
+decay z > dec dec
```
sihyunjeon, line 12: Are you also considering fully hadronic channel for your analysis as well?

wpmccormack:
Hi @sihyunjeon , first, many apologies for such a delayed response on my end.  I just pushed changes to make all-hadronic final states.
I also tried your recommended syntax of p p > t bp~, bp~ > z b~, but this gave 0 cross section for all masses.  As an alternate, I tried e.g. p p > wp+ and putting forced decays for wp and bp in the madspin file.  That seems to have hung in the "decaying the events" step of generation.  I went through the lhe files for the original process syntax, and they seem to have the correct parent-child relations among the particles.

agrohsje:
Hi @sihyunjeon , please feel free to merge if ok with you. It seems everything was addressed.

wpmccormack:
Hi all, the Z decays should be consistent between the two PRs now, thanks!

sihyunjeon:
I am quite surprised to see that all masses are giving you 0 cross sections. I've realized that you have some m(bp) < 90 GeV as well so bp~ > z b~ might not work. But for higher masses it should because it is giving you the same diagrams but only forcing bp~ to be onshell produced from wp+. Otherwise it might be some other diagrams that you DON'T WANT as your signals contributing to the process and that results in non-zero cross section. If you are confident with your physics modellings I can merge it (anything can happen with BSM...), but this behavior is not something I can't understand

wpmccormack:
Hi @sihyunjeon , so specifically the syntax p p > t bp~, bp~ > z b~ gives 0 cross section.  When I tried e.g. p p > wp+, wp+ > t bp~ or p p > wp+, wp+ > t bp~, bp~ > z b~, that did not give 0 cross section, but there were errors in the decay (from madspin I think).  When I did p p > wp+ with the decays specified in the madspin file, the decay process seemed to be taking forever (I let it run for several hours without finishing).  The LHE files that I generate with the current syntax seems to have the correct relationships among parent and daughter particles.  We really only care about the pronginess of the final state di-jet system, so particular diagrams aren't extremely important

PR #2996, opened 2021-09-23
Adding Triboson cards for CASE analysis

wpmccormack:
This is the 4+2 prong dijet topology for the CASE analysis.  The process is W_kk -> W_r + W, where W_r -> WW.  Scan over W_kk and W_r masses


bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Triboson_WkkWr/Triboson_M3000_R0.5/Triboson_M3000_R0.5_run_card.dat
```diff
@@ -0,0 +1,270 @@
...
+#*********************************************************************
+# Beam polarization from -100 (left-handed) to 100 (right-handed)    *
+#*********************************************************************
+     0.0     = polbeam1 ! beam polarization for beam 1
+     0.0     = polbeam2 ! beam polarization for beam 2
+#*********************************************************************
+# PDF CHOICE: this automatically fixes also alpha_s and its evol.    *
+#*********************************************************************
+     lhapdf    = pdlabel   ! PDF set                                     
+     263400    = lhaid     ! if pdlabel=lhapdf, this is the lhapdf number#*********************************************************************
```
agrohsje, line 52: what is the reason to use this old pdf ? if you want to use nnpdf 3.0 lo, make sure the corresponding one appears in runcmsgrid.sh as systematic variation.

agrohsje:
Main comment is on pdf. 4fs is correct, gauge and lorentz checks ok.

agrohsje:
I see that in your other PR's you are using: $DEFAULT_PDF_SETS = lhaid which will give you NNPDF 3.1 NNLO. This is the current default. For LO ME's some people prefer a LO PDF but the topic is complicated and really analysis specific.

wpmccormack:
Hi @agrohsje , I updated the PDF choice and verified that the generated events look fine

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Triboson_WkkWr/Triboson_M3000_R0.5/Triboson_M3000_R0.5_run_card.dat
```diff
@@ -0,0 +1,270 @@
...
+#*******************                                                 
+#                                                                    
+#*********************************************************************
+# Tag name for the run (one word)                                    *
+#*********************************************************************
+  tag_1     = run_tag ! name of the run 
+#*********************************************************************
+# Run to generate the grid pack                                      *
+#*********************************************************************
+  #False     = gridpack  !True = setting up the grid pack
```
agrohsje, line 26: don't enable settings. there will always be a default. better to know. anyway, the setting will be centrally handled. i will change the line.

wpmccormack: Ah actually, the template file is this one: bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Triboson_WkkWr/Triboson_M_R/Triboson_M_R_run_card.dat

agrohsje:
Hi @wpmccormack . Please check the template run card which I cleaned up a bit. Please fix the example run card. It still has the buggy PDFs.  Please apply your and my changes. Thanks!

wpmccormack:
Hi @agrohsje , The new cuts look fine to me, and I updated the example run card

wpmccormack:
Hi @agrohsje , just a quick ping here similar to PR 2994

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Triboson_WkkWr/Triboson_M_R/Triboson_M_R_customizecards.dat
```diff
@@ -0,0 +1,6 @@
+set param_card kkinputs 5 <MASS>
+set param_card kkinputs 2 6.708000e+00
+set param_card mass 9000025 <MASS_R>
+set param_card mass 9000024 <MASS>
+set param_card DECAY 9000024 1.195600e+01
+set param_card DECAY 9000025 0.31035
```
agrohsje, line 6: Small comment on these settings? Where are these values coming from? As it seems not just to be a small fixed number, maybe the same logic can be applied to all mass points?

agrohsje:
Hi @sihyunjeon @wpmccormack, I just have one small follow-up. Otherwise I am fine with the PR to be merged. Please check @sihyunjeon. Thanks !

wpmccormack:
Hi @agrohsje , good point with the weird numbers.  I used https://github.com/cms-sw/genproductions/tree/dace43b7b79c6c7d01d80bb17b96db6d84326830/bin/MadGraph5_aMCatNLO/cards/production/pre2017/13TeV/Triboson%20search as a template, which is where those number come from.  I've set the width to 3% now Wkk and Wr decays.  Also moved to python for sample card generation

sihyunjeon:
Hello, You should change the name of your directories.
R0.5->R0p5, otherwise this will create a bug when making the samples

wpmccormack:
Hi @sihyunjeon , It was actually only that one sample where the naming convention was like that.  I removed that, since it's old anyways.  The cards generated by Triboson_gen_card.py have names like Triboson_M2000_Mr400

PR #3001, opened 2021-09-28
updating aQGC VBS hadronic&semilep cards

stalbrec:
Hi,

this PR updates the aQGC VBS hadronic and semileptonic cards  (for the processes relevant for the VBS hadronic analysis) as they'll be used for the Summer20 samples, adapting all relevant changes made already to the hadronic SM cards in #2976 and updating to the new version of the dim8-UFO.
So this PR changes:
 - moving the decay of the V-bosons from the `madspin_card.dat` to the `proc_card.dat` 
 - adapting the dr-cuts that were changed in PR #2976 that were effectively already handled by the mjj cut
 - updating to the new dim8-UFO `quarticCKM21v2.tgz` (untarred `QuarticCKM_5_Aug21`) and adding reweighting points for the added operators `T3` and `T4`


sihyunjeon:
Looks good to me

PR #3002, opened 2021-09-29
JHUGen gammaH + decay cards with anomalous couplings added for UL production

lucassmguerra:


agrohsje:
Did you discuss the pdf strategy within JHUGEN analyses? I see NNPDF 3.1 NNLO e.g. here:
https://github.com/cms-sw/genproductions/blob/master/bin/JHUGen/cards/2017/13TeV/anomalouscouplings/WH_NNPDF31_13TeV/SM.input

agrohsje:
Let me add one piece of info which seems a misunderstanding (or might depend on the L2 convener). I am open to any choice as long as it is justified and not random. GEN provided all relevant arguments as well as corresponding tunes to have any preferred choice of PDF/tune.

gritsan:
Hi all: I was asked to comment on a "private thread," but I can say that nobody involved in preparation of the JHUGen gridpacks was aware about that private thread and we were sure that UL samples would be generated with the same PDFs as they were done before UL. In fact, changing PDFs inside of all the gridpacks is a lot of work, for those gridpacks which are already ready. The choice was settled in open meetings more than a year ago, why re-discuss this now in private threads? One way or another, even if there is an issue, it is not something that should hold this PR for weeks. Lets try to close this ASAP and move forward.

PR #3003, opened 2021-09-30
Fix to the script to take into account slha-based scans with multiple gridpacks

cericeci:
Hi all,

Going through request validation I noticed multiple signal samples with parameter scans (i.e. multiple gridpacks) failing the checking script. Explicitely I tested SUS-RunIIAutumn18FSPremix-00170 with validated properly around ~last June but not anymore today.

It seems that something in #2949  broke the logic for these requests as the code would try to unpack a generic gridpack name that in most cases does not exist before checking for the existance of slha configuration in the fragment.

This is a possible solution by moving some of the logic for scans to an earlier part of the code


efeyazgan:
Hi @cericeci thanks for fix. It looks OK. Merging.

PR #3006, opened 2021-10-07
Add Data Cards For LFV Analysis

Jingyan95:
We submit this PR for our UL sample request.
These samples are generated based on SMEFT dimension-6 operators with emutc or emutu final states. Specifically, there are two categories of signals:
1)Top production,
2)Top decay.
For each category, operators with different Lorenz structures are probed:
i)vector-like,
ii)scalar-like,
iii)tensor-like.
For each Lorenz structure, operators with different final states are probed:
i)emutc,
ii)emutu.
Therefore, there are 2x3x2=12 samples.


agrohsje:
Dear @Jingyan95, if some settings in the default param_card.dat needs to be changed you can handle via customize_card (the main purpose of this cms specific card). So no need to add on top. The reason why I am saying: There is no way for us to properly review the praram_card.dat while it is easy to check a few settings in the customize card.

Jingyan95:
> Dear @Jingyan95, if some settings in the default param_card.dat needs to be changed you can handle via customize_card (the main purpose of this cms specific card). So no need to add on top. The reason why I am saying: There is no way for us to properly review the praram_card.dat while it is easy to check a few settings in the customize card.

Dear @agrohsje , we removed the parameter card from this PR because it is almost the same as the one produced by the UFO except for the order of the content.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/LFV_TopProduction_TopDecay/SMEFTfr_TT_clequ1_emutu_run_card.dat
```diff
@@ -0,0 +1,265 @@
...
+$DEFAULT_PDF_SETS = lhaid
+$DEFAULT_PDF_MEMBERS = reweight_PDF     ! if pdlabel=lhapdf, this is the lhapdf number 
+#*********************************************************************
+# Renormalization and factorization scales                           *
+#*********************************************************************
+ F        = fixed_ren_scale  ! if .true. use fixed ren scale
+ F        = fixed_fac_scale  ! if .true. use fixed fac scale
+ 471  = scale            ! fixed ren scale
+ 471  = dsqrt_q2fact1    ! fixed fact scale for pdf1
+ 471  = dsqrt_q2fact2    ! fixed fact scale for pdf2
```
agrohsje, line 61: what is the motivation to set the scale but use F = fixed_ren_scale. Note that the values are irrelevant in this case. Is this what you have in mind ?

agrohsje:
@Jingyan95 please comment on the scale choice. Once this is done this is ok for me to be merged. @sihyunjeon please check and merge once the last question is answered. thanks so much!

Jingyan95:
> @Jingyan95 please comment on the scale choice. Once this is done this is ok for me to be merged. @sihyunjeon please check and merge once the last question is answered. thanks so much!

Dear @agrohsje and @sihyunjeon , those numbers are put in by mistake. I just made a commit to set them to default values.

PR #3014, opened 2021-10-14
Adding input cards for l+jet ttbar with large top width variations

leggat:


bin/Powheg/production/2017/13TeV/TT_hvq/TT_hdamp_widthx0p55_NNPDF31_NNLO_ljets.input
```diff
@@ -0,0 +1,89 @@
...
+nubound 100000  ! number of bbarra calls to setup norm of upper bounding function
+iymax 1        ! <= 10, normalization of upper bounding function in iunorm X iunorm square in y, log(m2qq)
+
+xupbound 2      ! increase upper bound for radiation generation
+
+pdfreweight 0
+
+storeinfo_rwgt 0    ! store weight information
+withnegweights 1    ! default 0
+
```
agrohsje, line 79: general remark: these settings have no effect. they will be disabled by the script. you would need to change run_pwg.py if you want to change the behavior.

PR #3022, opened 2021-10-26
Commit of the cards for the top mass widths of 0.55 and 1.45 (dilepton)

KathrynColdham:


bin/Powheg/production/2017/13TeV/TT_hvq/TT_hdamp_widthx1p45_NNPDF31_NNLO_dilepton.input
```diff
@@ -0,0 +1,80 @@
...
+
+!semileptonic 1      ! uncomment if you want to filter out only semileptonic events. For example,
+                     ! with topdecaymode 10011 and semileptonic 1 you get only events with one top going
+                     ! to an electron or positron, and the other into any hadron.
+
+! Parameters for the generation of spin correlations in t tbar decays
+tdec/wmass 80.4  ! W mass for top decay
+tdec/wwidth 2.141
+tdec/bmass 4.8
+tdec/twidth  1.5065 ! 1.45  x 1.31 central width
```
agrohsje, line 49: this line is still buggy. shouldn't it be 1.8995 ?

KathrynColdham: Ok I can change that, thanks. What does the 1.8995 represent here? I originally copied this card and changed the width from 1.15 to 1.45: https://github.com/jfernan2/genproductions/blob/dafd0072238753e9b78a783da1e144ef84b7e2df/bin/Powheg/production/2017/13TeV/TT_hvq/TT_hdamp_widthx1p15_NNPDF31_NNLO_dilepton.input

agrohsje: The top width at 172.5 GeV is 1.31 (called central width above). Now, the 1.45 that you indicate above is the relative width, i.e. 1.31 * 1.45 (45% increase) so 1.8995. @wajidalikhan and @GiuliaNegro please confirm that this is the intention and the setting should be changed according to my proposal.

wajidalikhan: @agrohsje  yes, it has to be relative width as you mentioned.Thanks a lot for having a critical look. Really appreciated.

PR #3031, opened 2021-11-05
Card for top quark Baryon Number Violation processes

rgoldouz:
Hi All,

For submitting our sample requests for top quark baryon number violation search, the TOP modeling group asked me to upload my madgraph cards.

Details about the Gen level studies can be found in the following presentation,
https://indico.cern.ch/event/1065050/contributions/4478002/attachments/2292476/3897998/1_StatusR_TOPBNV.pdf

Please let me know if you have any question.

Thanks,
Reza


agrohsje:
Dear @rgoldouz , the PR adds 96 files.  Can you provide a script with templates and just commit the relevant lines. There is no way to review 96 files.

rgoldouz:
Hi @agrohsje ,

Thanks,
What do you mean by  "a script with templates"? 
Run cards are same. Only process and customize cards are different. Since each sample include an specific flavor combination of the  BNV operators, we need to ask for multiple small samples.

Best regards,
Reza

agrohsje:
hi @rgoldouz . the idea is to minimize the number of committed files and the bugs. when you submit just one template of run_card.dat, I can easily check. While it is hard to check 16 or whatever. At the same time, if you process card always has a certain logic  you can provide a skeleton file and insert the process 
like 
for init ubu~b~ cbc~b~ ... 
    for out tmu+ ... 
     do echo " generate process $init > $out" >> proc_card.dat 
Similar solutions might be possible for customize and reweighting. Just to simplify the review and avoid bugs to go unnoticed.

rgoldouz:
Hi @agrohsje ,

I just uploaded the main python files that are used to make all cards. Is it what you expected?
Please let me know you questions/comments.

Thanks,
Reza

rgoldouz:
Hi @agrohsje ,

I know you are busy. Please let me know if any input from my side can make the review easier. After merging the cards, the TOP GEN contact will submit our fullsim requests. 
For this analysis, the main analysis framework is ready. We just need to work on the signal extraction part. So my target is Moriond and I need samples as soon as possible.

Thanks for your help,
Reza

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TopBNV/generateCards_BNV_tMuProduction.py
```diff
@@ -0,0 +1,88 @@
...
+]
+
+scanValues = 20
+
+for num, name in enumerate(processName):
+    os.system('rm -rf BNV_ST_' + name)
+    os.system('mkdir BNV_ST_' + name)
+    customizecards = ''
+    customizecards = customizecards + 'set param_card mass   6  172.5\n'
+    customizecards = customizecards + 'set param_card yukawa 6  172.5\n'
```
agrohsje, line 53: Do you want to change the top width accordingly?

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TopBNV/generateCards_BNV_tMuProduction.py
```diff
@@ -0,0 +1,88 @@
...
+    customizecards = customizecards + 'set param_card mass   25 125.0\n'
+    customizecards = customizecards + 'set dynamical_scale_choice 3\n'
```
agrohsje, line 55: should we move to run_card.dat ? it applies to all or do i misunderstand ?

agrohsje:
Thanks Reza. Changes are much appreciated. I have 2 small remaining comments and cleaned up the run_card. Please check. @sihyunjeon please check after feedback from @reza and merge if ok.

rgoldouz:
Hi @agrohsje , and  @sihyunjeon ,

Thanks for your reply.
Yes, they apply to all  processes. The first three parameters should be changed in param card. I think I need to add a new param card and change these parameters there? right?

The scale choice can  be included in run card. I put it here to remember that this fix is needed for this model if I am using the default run card.

Personally, I prefer to use the default param and run cards and apply  these  changes  on top of them.
Please let me know if It is s.th that should be changed.

Thanks a lot,
Reza

agrohsje:
No, no. Don't worry. Don't add the param_card. You do it exactly right. I am also fine to leave this for the change of the run_card. No objects. Was a bit surprised but see your point. @sihyunjeon  please check and merge.

rgoldouz:
Dear @sihyunjeon ,

I would be very grateful if you could check our run card. @agrohsje has checked the other inputs.
Everything is ready for submitting samples except for this part.

Thanks a lot,
Reza

agrohsje:
we both check all cards. 4 eyes-system. :-)

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TopBNV/BNV_run_card.dat
```diff
@@ -0,0 +1,266 @@
...
+-1.0	=  time_of_flight  ! threshold below which info is not written
+#*********************************************************************
+# Matching - Warning! ickkw > 1 is still beta
+#*********************************************************************
+0	=  ickkw             ! 0 no matching, 1 MLM, 2 CKKW matching
+1	=  highestmult       ! for ickkw=2, highest mult group
+1	=  ktscheme          ! for ickkw=1, 1 Durham kT, 2 Pythia pTE
+1.0	=  alpsfact          ! scale factor for QCD emission vx
+False	=  chcluster         ! cluster only according to channel diag
+False	=  pdfwgt            ! for ickkw=1, perform pdf reweighting
```
swertz, line 76: @agrohsje does changing this flag require reproducing the gridpacks?

agrohsje: pdfwgt will only be relevant in the case of additional jets. reason why i changed was just to prevent a bug in case someone starts from this pr and adds jets. then it matters. at present this has no impact.

swertz:
Hi Reza, just to be sure: we'll be using Pythia for the SM top quark decays in ttbar (not MadSpin - I don't see any MS cards here), right? (i.e. no spin correlations)

rgoldouz:
Hi @swertz ,
Yes, for this new physics search, we are not sensitive to spin correlations. On the other hand, I remember this new physics with baryon number  violation caused some issue when I used madspin. So lets  use pythia for the SM decay of top quarks.

Thanks,
Reza

swertz:
Perfect. Then I will already go ahead with the existing gridpacks to save time; in the unlikely event that @sihyunjeon catches issues with the cards and new gridpacks are needed we can always roll back.

rgoldouz:
Thank a lot @swertz , I will write to convener today to ask for increasing the priority.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TopBNV/BNV_run_card.dat
```diff
@@ -0,0 +1,266 @@
...
+#*********************************************************************
+# Run to generate the grid pack                                      *
+#*********************************************************************
+True	=  gridpack   !True = setting up the grid pack
+#*********************************************************************
+# Number of events and rnd seed                                      *
+# Warning: Do not generate more than 1M events in a single run       *
+# If you want to run Pythia, avoid more than 50k events in a run.    *
+#*********************************************************************
+10000 =  nevents  ! Number of unweighted events requested
```
sihyunjeon, line 32: hi not an error but 10000 events is meaningless for a pilot run, might want to reduce it to 2K

sihyunjeon:
> Hi @swertz , Yes, for this new physics search, we are not sensitive to spin correlations. On the other hand, I remember this new physics with baryon number violation caused some issue when I used madspin. So lets use pythia for the SM decay of top quarks.
> 
> Thanks, Reza

This can be true. Though there is a way to get away with this, but as long as you are fine with spin blind samples should be ok

sihyunjeon:
> Hi @agrohsje , and @sihyunjeon ,
> 
> Thanks for your reply. Yes, they apply to all processes. The first three parameters should be changed in param card. I think I need to add a new param card and change these parameters there? right?
> 
> The scale choice can be included in run card. I put it here to remember that this fix is needed for this model if I am using the default run card.
> 
> Personally, I prefer to use the default param and run cards and apply these changes on top of them. Please let me know if It is s.th that should be changed.
> 
> Thanks a lot, Reza

This can also be true. I see your point and have seen one or two requests with same issues with dyn scale choices

sihyunjeon:
merged :)

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TopBNV/BNV_run_card.dat
sihyunjeon: @rgoldouz Just to comment here, I haven't changed this and merged already since it seems you were in a hurry. But it would help at least a little bit if you just reduce the nevents to some smaller number like 2K when you make the gridpacks

PR #3033, opened 2021-11-10
Higgs dalitz decay cards without m_ll cut in madspin card

RoccoA97:
Hi,

This pull request adds a copy of the folder [[1]](https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Higgs/dalitz), containing the MG5 cards for the Higgs dalitz decay. The main (and only) change is that the `m_ll` phase space cut 0<m_ll<60 GeV in the madspin card has been removed. The motivation and why it is needed are explained in the long discussion in [[2]](https://github.com/SMP-GEN-Coordination/mc-requests/issues/12). Here, I'll give a brief summary.

After the request for the generation of the UL samples for `*HToMuMuG*`, me and @zucchett had to regenerate the gridpacks since the original ones (for the non-UL samples) are not available anymore. The cards for this Higgs dalitz decay process can be found in [[1]](https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Higgs/dalitz).
We have tried to recreate the gridpacks with the latest commit of genproductions (which uses CMSSW_10_6_19 and MG5_aMC_v2.6.5 for the creation of the tarball), but it turns out that these two lines in the madspin card are causing the failure of a correct gridpack generation:
```
set run_card mmll=0
set run_card mmllmax=60
```
Thanks to the help of @sihyunjeon, we have found out that these cards were made for CMSSW_7_1_30 and MG5_aMC_v2.4.2 3-4 years ago. Multiple updates and changes have been done to madspin and madgraph after v2.4.2, so using CMSSW_7_1_30 and MG5_aMC_v2.4.2 would not be ideal for a UL generation.
@sihyunjeon has tried to work on a solution for this bug, however a clean and safe fix could not be found (keeping at the same time the m_ll phase space cut in the madspin card).

In order to go on with the generation of the samples, which are urgently needed for an on-going analysis (H/Z->JPsi+Gamma), we have proposed a temporary solution: comment the two lines in the madspin card that give the problem. We have successfully performed the gridpack generation with this fix, however removing this phase space cut would make necessary a higher amout of events to take into account the loss in efficiency.
From a raw test on the madgraph lhe output, we have verified that we would need about 3-4 times more events, which should not be a big deal for the production system. The original amout of events was about 0.5-1M per Higgs production mode and per each year, for a total of 6-12M events. By this way, we would need a total amount of about 25-50M events.

I hope everything is clear and that this temporary fix is not a critical problem.

Thanks,
Rocco and Alberto

Links:
* [1] : https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Higgs/dalitz
* [2] : https://github.com/SMP-GEN-Coordination/mc-requests/issues/12


bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Higgs/dalitz_no_mll_cut/WH_012j_NLO_HtoElElGamma/WH_012j_NLO_HtoElElGamma_customizecards.dat
```diff
@@ -0,0 +1 @@
+set param_card mass 25 <MASS>
```
agrohsje, line 1: You want to change the width accordingly or ignore as any small value is good enough?

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Higgs/dalitz_no_mll_cut/WH_012j_NLO_HtoElElGamma/WH_012j_NLO_HtoElElGamma_extramodels.dat
```diff
@@ -0,0 +1,4 @@
+HC_NLO_X0_UFO-v1.3.zip
+
+HC_NLO_X0_UFO-lepton_masses_no_lepton_yukawas-v1.3.tar.gz
```
agrohsje, line 3: What is the reason to have both here?

sihyunjeon: import model HC_NLO_X0_UFO-lepton_masses_no_lepton_yukawas --bypass_check
 
 It imports a different model in MadSpin for decays so maybe this is why

RoccoA97:
Hi @agrohsje,

Unfortunately, we are not the original authors of these cards and we do not know exactly the reason for which two models are employed. If you give a look to the proc_card and to the madspin_card, you can see that two different models are employed.
Since these cards where approved in the past, we thought that there was a particular reason for which this configuration is chosen...

Concerning the MASS parameter, we are only interested in the standard case of m_H = 125 GeV.

Hope this is clear.
Thanks,
Rocco and Alberto

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Higgs/dalitz_no_mll_cut/WH_012j_NLO_HtoElElGamma/WH_012j_NLO_HtoElElGamma_extramodels.dat
agrohsje: My point is: The zipped model appears in the extramodel file but no where else. Why?

govoni:
please ask a question in the Madgraph Launchpad about this problem:
https://launchpad.net/mg5amcnlo

andreypz:
Hi all, 
Since original cards were created by me I can answer some questions:
* The reason to have two models, *HC_NLO_X0_UFO-v1.3.zip* and *HC_NLO_X0_UFO-lepton_masses_no_lepton_yukawas-v1.3.tar.gz* is to use one of them in the production of the Higgs boson and the other one in its decay. In the decay we need to restrict the lowest m_gamma* to 2 times the mass of a lepton (muon or electron, depending on the channel). For this reason, a model with lepton-masses was used.
* It's unfortunate that it is no longer possible to have mll<60 restrictions. Without it the process would include H->Zgamma channel, which is usually produced separately. On the other hand, one could use such sample for cross-checking with H->Zgamma.
* Regarding the Higgs width. In the past I though it was sufficient to use a default value, therefore no changes in customize cards were made. But if the default value now is different from the current theory prediction, then I think it's worth to set it as well (if that works of course).

RoccoA97:
Hi @andreypz,

Thank you for your message! For the UL generation we received the suggestion to use more updated versions of mg and cmssw (as explained in the first message of this pr). Thank you anyway!
We have tried to generate the gridpacks without the mll cut and, as you were pointing out, also the H->Zgamma channel is included and for this reason about 3-4 times more events are required to have almost the same statistics after applying 0<mll<60 GeV.
I have another question for you. Have you performed studies on the ttH and bbH production modes for these Dalitz decay samples?

Thanks,
Rocco

andreypz:
Hi, regarding this question:
 > Have you performed studies on the ttH and bbH production modes for these Dalitz decay samples?

No, I have not studied that. I think at the time it was not possible to do (to have a Dalitz decay for ttH and bbH productions in Madgraph).

agrohsje:
I would not bother with the PR and the cuts. Factor 4 here should be no big deal taking into account the total stat. 
However I don't understand the explanation from @andreypz. He says the two models are used for production resp. decay, however in the process card I see : 
generate p p > V h [QCD] .

andreypz:
Indeed, looks like VH processes one extra-model is not needed. 
It seems like a copy-paste mistake from  ggH cards, where both models are used. So, I think the `HC_NLO_X0_UFO-v1.3.zip` can be removed from extramodels.dat files for ZH and WH.

agrohsje:
Thanks @andreypz . @RoccoA97 please update accordingly.

RoccoA97:
Just done, sorry for the delay.
I removed `HC_NLO_X0_UFO-v1.3.zip` from ZH, WH, VBF, but I have left it for the ggH process.

Thanks @andreypz and @agrohsje

agrohsje:
Two remaining questions: What about the decays of the additional V bosons. Will they happen in P8? Do you plan to adjust the Higgs width according to the mass value?

RoccoA97:
Hi @agrohsje,

1) With the current version, the decay of additional V bosons is handled by P8. We can handle it inside mg5 so that the spin correlations are correctly taken into account. I think it is just a matter of adding something like `(V > all all)`. Do you think it makes more sense to add this decay into the proc_card or into the madspin_card?

2) Actually, we are interested only in the SM case of m_H = 125 GeV, since we will use these samples for the analysis. Thus, at first we were not planning to change the width value. To sum up, the important cards for us are the ones with m_H = 125 GeV, for which it is not necessary to change the width value

sihyunjeon:
Hi,
Some notes to your last comments.

> Apparently, at NLO V has to be decayed in MadSpin  
Yes and no : You can do V decays without MadSpin p p > l+ vl h [QCD]. But of course this would be more time consuming and with up to 2 multipartons, I think it would be excessively long.

> Inside madspin, for this kind of decay, it is necessary to set spinmode=none
Yes, but only because you have loop induced higgs, MadSpin necessarily needs "spinmode none". But for usual cases without loop-induced diagrams you don't need to use "spinmode none" : e.g. p p > w+ z [QCD] this doesn't need such spinmode.

> Thus, I think we are forced to let P8 handle the V decay, as it is done with the current version of the cards
So either way, you would lose spin correlations, MadSpin due to "spinmode none" and  Pythia. So it's up to you.

sihyunjeon:
Can you change like below in every run_card?

     -1 = nevt_job! Max number of events per job in event generation.

sihyunjeon:
Other things look fine to me

RoccoA97:
Hi,

I have just updated the cards. Tell me if you see something wrong.
Many thanks!

chw1207:
Hi all, 

I am doing the Higgs Dalitz analysis.
I  also tried to produce the gridpack privately using the cards in dalitz_no_mll_cut.

> Just done, sorry for the delay. I removed `HC_NLO_X0_UFO-v1.3.zip` from ZH, WH, VBF, but I have not modified the extramodels for the ggH process.
> 
> Thanks @andreypz and @agrohsje

However, the error happened as the following, so maybe``` HC_NLO_X0_UFO-v1.3.zip```  cannot be removed from VBF cards.
Everything is fine once I  added it back to the VBF cards
Screen Shot 2021-12-27 at 1 03 04 PM

``` HC_NLO_X0_UFO-v1.3.zip```  can be removed from ZH and WH cards safely. (without any error happened during the gridpack production)

What's more, we would like to combine all 3 lepton favors if we can!

RoccoA97:
Hi @chw1207,

Thanks for noticing that! If it is ok for @agrohsje and @sihyunjeon, I would add `HC_NLO_X0_UFO-v1.3.zip` again for the VBF mode in the extramodels.

Cheers,
Rocco

RoccoA97:
Hi @sihyunjeon, @agrohsje,

I put again `HC_NLO_X0_UFO-v1.3.zip` in the extramodels for VBF, as suggested by @chw1207. Just consider the last commit and not the one with "K short" since I messed with a name of a commit on another repository.

agrohsje:
Hi @RoccoA97 @chw1207, 
@andreypz mentioned production and decay are different so that's why one needs 2 model files. But that is not the point. What happens here is that one just contains the restriction card, so the full UFO info is not available and you need both. 

About the width. I checked that for the different restriction cards the numerical value of the width is always 4.070000e-03 independent of the mass, so no need to adjust. .
Going to merge.

RoccoA97:
Hi @agrohsje,

I lost your previous message.

> What happens here is that one just contains the restriction card, so the full UFO info is not available and you need both.
But I see you still need to change for all, because there is several UFO from HC are missing. The tarball just has the parameters.

What do you mean exactly?Should I add back all the `HC_NLO_X0_UFO-v1.3.zip` in all the extramodels cards?

PR #3045, opened 2021-11-16
WWjj_OS_noTop_qcd uploaded for UL campaign

mlizzo:
Hi @agrohsje , here's the proper PR for the WWjj_OS qcd MG card. Generator cuts are the same as those used in the legacy campaign, see for instance this request:
https://cms-pdmv.cern.ch/mcm/requests?page=0&dataset_name=WWJJToLNuLNu_QCD_noTop_TuneCP5_13TeV-madgraph-pythia8&member_of_campaign=*GS*&shown=127
They have also been applied in its EWK counter-part:
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/VBS/WWjj_OS/noTop/WWjj_OS_noTop_ewk_run_card.dat
Please let me know what do you think about it.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/VBS/WWjj_OS/noTop/WWjj_OS_noTop_qcd_run_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/VBS/WWjj_OS/noTop/WWjj_OS_noTop_qcd_run_card.dat
@@ -1,0 +1,269 @@
+#*********************************************************************
+#                       MadGraph5_aMC@NLO                            *
+#                                                                    *
+#                     run_card.dat MadEvent                          *
+#                                                                    *
+#  This file is used to set the parameters of the run.               *
+#                                                                    *
+#  Some notation/conventions:                                        *
+#                                                                    *
+#   Lines starting with a '# ' are info or comments                  *
+#                                                                    *
+#   mind the format:   value    = variable     ! comment             *
+#*********************************************************************
+#
+#*******************                                                 
+# Running parameters
+#*******************                                                 
+#                                                                    
+#*********************************************************************
+# Tag name for the run (one word)                                    *
+#*********************************************************************
+  tag_1     = run_tag ! name of the run 
+#*********************************************************************
+# Run to generate the grid pack                                      *
+#*********************************************************************
+  .false.     = gridpack  !True = setting up the grid pack
+#*********************************************************************
+# Number of events and rnd seed                                      *
+# Warning: Do not generate more than 1M events in a single run       *
+# If you want to run Pythia, avoid more than 50k events in a run.    *
+#*********************************************************************
+    10000     = nevents ! Number of unweighted events requested 
+      0       = iseed   ! rnd seed (0=assigned automatically=default))
+#*********************************************************************
+# Collider type and energy                                           *
+# lpp: 0=No PDF, 1=proton, -1=antiproton, 2=photon from proton,      *
+#                                         3=photon from electron     *
+#*********************************************************************
+        1     = lpp1    ! beam 1 type 
+        1     = lpp2    ! beam 2 type
+     6500     = ebeam1  ! beam 1 total energy in GeV
+     6500     = ebeam2  ! beam 2 total energy in GeV
+#*********************************************************************
+# Beam polarization from -100 (left-handed) to 100 (right-handed)    *
+#*********************************************************************
+        0     = polbeam1 ! beam polarization for beam 1
+        0     = polbeam2 ! beam polarization for beam 2
+#*********************************************************************
+# PDF CHOICE: this automatically fixes also alpha_s and its evol.    *
+#*********************************************************************
+'lhapdf'    = pdlabel     ! PDF set
+ $DEFAULT_PDF_SETS = lhaid
+ $DEFAULT_PDF_MEMBERS = reweight_PDF     ! if pdlabel=lhapdf, this is the lhapdf number
+#*********************************************************************
+# Renormalization and factorization scales                           *
+#*********************************************************************
+ F        = fixed_ren_scale  ! if .true. use fixed ren scale
+ F        = fixed_fac_scale  ! if .true. use fixed fac scale
+ 91.1880  = scale            ! fixed ren scale
+ 91.1880  = dsqrt_q2fact1    ! fixed fact scale for pdf1
+ 91.1880  = dsqrt_q2fact2    ! fixed fact scale for pdf2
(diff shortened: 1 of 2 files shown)


mlizzo:
Hi @agrohsje , just a kind a reminder if you could check this PR whenever you can please :)

bin/MadGraph5_aMCatNLO/cards/production/13TeV/VBS/WWjj_OS/noTop/WWjj_OS_noTop_qcd_run_card.dat
```diff
@@ -0,0 +1,269 @@
...
+ 0.4 = drjj    ! min distance between jets 
+ 0.0 = drbb    ! min distance between b's 
+ 0.4 = drll    ! min distance between leptons 
+ 0.0 = draa    ! min distance between gammas 
+ 0.0 = drbj    ! min distance between b and jet 
+ 0.0 = draj    ! min distance between gamma and jet 
+ 0.4 = drjl    ! min distance between jet and lepton 
+ 0.0 = drab    ! min distance between gamma and b 
+ 0.0 = drbl    ! min distance between b and lepton 
+ 0.0 = dral    ! min distance between gamma and lepton
```
sihyunjeon, line 155: to be safe, you should remove all the cuts here and turn off "    T  = cut_decays    ! Cut decay products "
Your phase space it restricted a lot

sihyunjeon:
> Hi @agrohsje , here's the proper PR for the WWjj_OS qcd MG card. Generator cuts are the same as those used in the legacy campaign, see for instance this request: https://cms-pdmv.cern.ch/mcm/requests?page=0&dataset_name=WWJJToLNuLNu_QCD_noTop_TuneCP5_13TeV-madgraph-pythia8&member_of_campaign=*GS*&shown=127 They have also been applied in its EWK counter-part: https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/VBS/WWjj_OS/noTop/WWjj_OS_noTop_ewk_run_card.dat Please let me know what do you think about it.

Ahhhh... I now see this... Are the analyzers fine with such cuts?

mlizzo:
Hi @sihyunjeon , I made a quick check on the VBS WpWmJJ sample we are currently using in the analysis and I looked at gen-level DR variables between leptons and jets. Indeed we already require highly isolated leptons and I don't think loosening these cuts will have sizeable effects:
https://mlizzo.web.cern.ch/mlizzo/VBS/DR_GenLevel/
By the way I can remove those selections in the run card (both EWK and QCD) in order to generate events in a more inclusive phase space, you think it's better to do so?

mlizzo:
Hi @sihyunjeon @agrohsje I'll prepare two gridpacks w an w/o those cuts respectively, to check the difference in the cross section. I'll keep you posted.

mlizzo:
Hi @sihyunjeon @agrohsje , I have made alternative gridpacks by setting the cut_decays option to false instead of true:
EWK W+W- jj:
1)	
	T  = cut_decays    ! Cut decay products
	Cross-section :   0.09283 +- 0.0001569 pb
2)
	F  = cut_decays    ! Cut decay products
	Cross-section :   0.09752 +- 0.0001646 pb

QCD W+W- jj:
1)
	T  = cut_decays    ! Cut decay products	
	Cross-section :   2.197 +- 0.004426 pb
2)
	F  = cut_decays    ! Cut decay products
	Cross-section :   2.328 +- 0.004042 pb

The difference is of the order of 5% for both samples, thus I don't think this setting will have any sizeable impact on a real analysis, where typically isolation requirements on leptons are much more stringent. If you think this could be reasonable please merge the PR

sihyunjeon:
@agrohsje what would you prefer?

bin/MadGraph5_aMCatNLO/cards/production/13TeV/VBS/WWjj_OS/noTop/WWjj_OS_noTop_qcd_run_card.dat
agrohsje: I would suggest to set all of them to 0.01. 0.4 could still be tight although the  effect looks small.

agrohsje:
I would set dR values to 0.01.

mlizzo:
Hi @agrohsje , for both samples you mean? The EWK card was already approved [1] and thus I have created gridpacks and McM requests with those settings some weeks ago. I can update it of course, it was just to inform you. Let me know what to do please, I can add another commit to this PR to modify the EWK card as well. 

[1] https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/VBS/WWjj_OS/noTop/WWjj_OS_noTop_ewk_run_card.dat

agrohsje:
for the pr, i would indeed prefer that we sync both nad remove the dr cuts. i always assumed that the dr cut in the analysis is larger than 0.4 so hence i thought it is ok. however, if this is only explicitly true via the mjj cut, i think it might be better to remove. 
about the gridpacks: what is the cut on drj in the analysis. i mean we talk about 6% but then of course less because also in the sample with no cut you will remove some events by the detector level dr cut. 
should we leave current cards with dr cuts, but make a pr and remove in the future ?

mlizzo:
I don't know whether I have fully grasped what you mean, but indeed in the analysis we have more stringent cuts: no lepton can be within the jet cone (DR = 0.4) and there are tight isolation requirements on leptons themselves. I thought that was a pretty standard definition, I'm not sure if this answer your question.

agrohsje:
What I am talking about is dR(jet, jet). Is it 0.4 in the analysis?

mlizzo:
Being a VBS topology, in the analysis we require deta(j,j) > 2.5 and this sets a lower limit on the dR(j.j) variable, which is far greater than 0.4

mlizzo:
Hi @agrohsje , how shall I proceed then?

agrohsje:
Hi @mlizzo . Was discussing with @sihyunjeon general procedure for VV samples and UL. As many are buggy we need to invalidate. We should start from a consistent setup. You wanted to produce these now for SMP?

mlizzo:
Hi @agrohsje , thanks for your answer. I was preparing these requests for the HWW group (as background samples) but of course they are needed in SMP too. Do you consider them as "buggy"? What are the common requirements to have a "consistent setup"?

agrohsje:
Sorry. Ignore my comment. I wanted to send in the context of another PR. 
@sihyunjeon  please merge if you are ok with the PR.

PR #3046, opened 2021-11-17
Inclusion of FCNC cards for UL gridpacks for both TT & ST processes

vrbouza:
- JIRA ticket: https://its.cern.ch/jira/projects/CMSTOPMCREQ/issues/CMSTOPMCREQ-2

Cards for various FCNC top pair and single top processes. They've been centralised as much as possible.


vrbouza:
Latest commits add also the cards for the additional TT FCNC samples that were found with the same issues as the first ones (as discussed through Mattermost). Also automatised.

agrohsje:
Hi @vrbouza . Can you comment on the structure of the PR. You still have 65 files. Are all run_cards.dat different? Why do you need to submit the param_cards. Settings in param_card should be replace via the customize card.

vrbouza:
> Hi @vrbouza . Can you comment on the structure of the PR. You still have 65 files. Are all run_cards.dat different? Why do you need to submit the param_cards. Settings in param_card should be replace via the customize card.

Hi @agrohsje! For the run_cards.dat under the "refcards" folder I took exactly the cards/files that were before in the 2017 approved cards. I've checked various of them and they are different. I'm not sure whether we would gain much from putting the content into the automatisation script.

For the param_cards, they are also different, but in a smaller way: I just automatised them, so now there are fewer files.

vrbouza:
Dear @agrohsje and @sihyunjeon , I've added a new commit that automatises even more these cards. There are 16 files now only. I hope now this PR would be easier to review.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ST_FCNC/TA_Tleptonic_kappa_act_LO/TA_Tleptonic_kappa_act_LO_proc_card.dat
```diff
@@ -0,0 +1,12 @@
+# Model
+import model TopFCNC_UFO-kappa_act_ProdConventions_no_c_mass --modelname
+ 
+# Multiparticles
+define l+  = e+ mu+ ta+
+define l-  = e- mu- ta-
+
+generate    p p > t  a, (t  > w+ b,  w+ > l+ vl ) NP=1 @10
+add process p p > t~ a, (t~ > w- b~, w- > l- vl~) NP=1 @20
```
sihyunjeon, line 9: The new physics term will come from the vertex "p p > t a" right?
Then I think "p p > t a NP=1, (t  > w+ b,  w+ > l+ vl )" should be correct? @agrohsje

sihyunjeon:
@vrbouza Hello Victor, reminder

vrbouza:
Good morning @sihyunjeon . What should I address? I don't see anything asked to me, there is one question you directed to @agrohsje (that I wouldn't know exactly how to answer, to be sure). I was waiting for the approval of the PR.

sihyunjeon:
Ugh sorry my bad :(. Yes I should've pinged Alexander instead @agrohsje

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ST_FCNC/TA_Tleptonic_kappa_act_LO/TA_Tleptonic_kappa_act_LO_proc_card.dat
agrohsje: Moving NP=1 to t a makes it explicit. With the given model there is no impact.

agrohsje:
replied to the position of np.
let me ask: why are param_cards added instead of customize card only?

vrbouza:
> replied to the position of np. let me ask: why are param_cards added instead of customize card only?

Honestly, I don't know, it seems that indeed they contain the same information, I don't know if both of them are needed (both are present in the current 2017 cards.

agrohsje:
When you call the UFO it comes with a default param_card.dat respectively some restriction card, i.e. an alternative param card that allows e.g. switching between 4fs and 5fs or some eft couplings included or not. On top, we allow users to adjust single parameters. Say if the param_card is generally fine but you like mtop=172.5, yt=172.5 and the corresponding top width, this can be done in the customize card via 
set param_card mass 6 172.5 
So in principle param cards are not needed at all as them come with the model.

vrbouza:
> When you call the UFO it comes with a default param_card.dat respectively some restriction card, i.e. an alternative param card that allows e.g. switching between 4fs and 5fs or some eft couplings included or not. On top, we allow users to adjust single parameters. Say if the param_card is generally fine but you like mtop=172.5, yt=172.5 and the corresponding top width, this can be done in the customize card via set param_card mass 6 172.5 So in principle param cards are not needed at all as them come with the model.

Thanks for the explanation! Then, I can remove all param_cards as it is a redundancy, no? (just to be sure...)

vrbouza:
After discussing in Mattermost and following @agrohsje suggestion, I've removed the customizecards. If there is anything else to improve or modify in this PR to get the approval, please tell me.

agrohsje:
Hi @vrbouza . I have a few more comments on this complex pr. 
1.) You raise the cuts for jets, leptons and photons in your script. Is the cut on the photon well below the detector level cut? You disable cut_decay so the lepton cut values have no impact, or do I misread? 
2.) In the customize card you are changing the top mass, usually that means that also the width will be different from the default UFO values. You might considering adding set param_card DECAY 6 auto to update according to the mass change. 
3.) Do you want to fix the NP setting and make clear where NP appears in production and decay as proposed by Sihyun? 
Cheers, Alexander.

vrbouza:
Hi @agrohsje .
Concerning (1): reviewing the cards, cut_decay is indeed in False for both ST and TT samples. I don't know if this invalidates the cuts that are applied as you say (I picked up the cards present in the genprods. repo, in the 2017 folder). However, if the cuts are invalidated then it doesn't matter, no?

My last commit should address also (2) and (3), modifying the remaining customizecards as well as modifying the proc_card where @sihyunjeon mentioned, hopefully correctly. I changed 
```
generate    p p > t  a, (t  > w+ b,  w+ > l+ vl ) NP=1 @10
add process p p > t~ a, (t~ > w- b~, w- > l- vl~) NP=1 @20
```
to
```
generate    p p > t  a NP=1, (t  > w+ b,  w+ > l+ vl )
add process p p > t~ a NP=1, (t~ > w- b~, w- > l- vl~)
```
.

agrohsje:
Thanks. And the photon cut? You know what that will be? Just to be sure we are well below.

vrbouza:
Hi @agrohsje , as I'm quite ignorant concerning photons (I don't use them directly in my analysis) I've asked the EGamma ID subPOG, and they've answered this:

> Dear Victor,
> We reconstruct photons down to 10 GeV (our energy regression has a cut at 10 GeV) and we provide recipes for IDs and SFs for photons with |eta| < 2.5,  but photons with  2.5 < |eta| < 3.0 are also reconstructed.
> 
> For gen production, however, you should keep some buffer, like instead of  pT> 10 GeV and |eta| < 3,  you can generate photons with pT> 8 GeV and |eta| < 3.1.
> 
> Let us know if you need anything else,
> Swagata and Linda

Therefore, the acceptance cuts should be pT>~8GeV and |eta|~<3.1, following their suggestions. However, if I review the run_card of the TA sample, it seems that the only one I see is of |eta|<5 (so we'd be covered geometrically), energetically, I'd say it's pT>0 GeV, no? Maybe I missed something...

agrohsje:
Sorry @vrbouza . It seems I made a mistake with the photon cuts. They are all loose. I thought you imposed tight cuts. Maybe I had another PR open and looked at a wrong file. Don't know. My fault, sorry for that. 
@sihyunjeon can you check the PR?

sihyunjeon:
looks fine merged

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ST_FCNC/TA_Tleptonic_kappa_act_LO/TA_Tleptonic_kappa_act_LO_run_card.dat
```diff
@@ -0,0 +1,265 @@
...
+# Matching - Warning! ickkw > 1 is still beta
+#*********************************************************************
+0	=  ickkw             ! 0 no matching, 1 MLM, 2 CKKW matching
+1	=  highestmult       ! for ickkw=2, highest mult group
+1	=  ktscheme          ! for ickkw=1, 1 Durham kT, 2 Pythia pTE
+1.0	=  alpsfact          ! scale factor for QCD emission vx
+False	=  chcluster         ! cluster only according to channel diag
+4	=  asrwgtflavor      ! highest quark flavor for a_s reweight
+True	=  clusinfo          ! include clustering tag in output
+3.0	=  lhe_version        ! Change the way clustering information pass to shower.
```
sihyunjeon, line 78: Oh shoot, before you generate the gridpacks can you add a line in your run cards and then start making the gridpacks?

False = pdfwgt

I think it's False by default but just to make sure

vrbouza: Yes, sure. Thanks!

PR #3057, opened 2021-11-26
Generate cards for DYJetsToLL_M-XXtoYY

gourangakole:
Script to generate cards for DYJetsToLL_M-XXtoYY_TuneCP5_13TeV-amcatnloFXFX-pythia8 missing samples for UL
as it was recently reported https://pdmv-pages.web.cern.ch/main_bkg_ul/


SanghyunKo:
Thanks @gourangakole, I ran the script and didn't spot any odds. Looks good to me.
BTW, M-3000toInf is not requested in the first place if I see the link correctly, right?

gourangakole:
Thanks @SanghyunKo for quick review....right, not M-3000toInf bin in first place.

agrohsje:
Should we fix the path structure? It is now blabla_lowmass/*massbinned* which is a bit confusing. 
Please confirm that it is ok for the request to use the new PDG scheme. @SanghyunKo @menglu21 showed that impact is small but still i like to double-check.

agrohsje:
we should still improve a bit for run 3. but let me merge.

PR #3062, opened 2021-12-02
Adding MCHM cards

c-bautista:
- Adding cards for ttHH production in the MCHM5 and the MCHM14 models for specific points in the parameter space (C2 and D7, respectively)
- Adding cards for double production of T1(lightest top partner) with subsequent 2-body decays into tZ or tH, for a specific point in the parameter space(C9) in the MCHM5.
- Adding cards for double production of T1(lightest top partner) with one of the resonances decaying into tZ or tH and the other into WWt (3 body decay), for a specific point in the parameter space(C9) in the MCHM5.
- Adding cards for double production of X53(resonance with electric charge Q=5/3) with one of the resonances decaying into Wt and the other into WZt or WHt for a specific point in the parameter space(C9) in the MCHM5.


agrohsje:
Dear @c-bautista , where can I find the models? You should have extramodel cards added and provide the UFO for our repo (Sihyun or myself can copy it) in case it is not yet there. Why do you need bwcutoff 15000. The setting is recommended to stay around 15, 20, 30 but not 15000. How did you validate the settings in customize card. It is impossible to review and check for us.

c-bautista:
Dear @agrohsje, as explained in a message we sent you, the high value of bwcutoff is particularly important when decaying the resonances since we need to include off-shell contributions. I ran these processes in a local machine with the parameters in customize_cards and they run normally. We sent the UFO's in the message we sent you. Do you please tell me what should I put in the extramodel card?
Dear

c-bautista:
Dear @sihyunjeon, our request here and as presented at UPSG and Future Higgs Physics WG meetings is presently related to the Snowmass 2022 studies thus to HL-LHC. This is why we have defined E(beam)=7TeV.

For further information, there is, on another hand, the ongoing ttHH4b non resonant analysis with Run 2 data that will be extended to Run 3, within the ttHbb WG in the Higgs Physics Analysis Group. We are going to discuss, early next year, within the ttHbb WG the production of the corresponding simulation MCHM samples for this other case.

bin/MadGraph5_aMCatNLO/cards/double_prod_T1_3body_decay_MCHM5_C9/double_prod_T1_3body_decay_MCHM5_C9_run_card.dat
```diff
@@ -86,7 +86,13 @@ $DEFAULT_PDF_MEMBERS = reweight_PDF      ! if pdlabel=lhapdf, this is the lhapdf
...
   .false.	=  auto_ptj_mjj   ! Automatic setting of ptj and mjj
 #**********************************************************
-#                                                                    
+#
+#*********************************************************************
+# Phase-Space Optim (advanced)
+#*********************************************************************
+   1 =  hard_survey ! force to have better estimate of the integral at survey for difficult mode like interference          
+#*********************************************************************
+#
```
agrohsje, line 95: This option doesn't exist in 2.6.5. Where did you find it?

c-bautista: I use a more recent version, 2.9.6 in which this parameter is used to overcome issues MadGraph encounters when generating the events for processes with difficult mode like interference. This happens in the processes: "double_prod_T1_3body_decay_MCHM5_C9" and "double_prod_X53_3body_decay_MCHM5_C9". Would it be possible to use this more recent 2.9.6 version at least for these two cases?

Saptaparna: I'm afraid that master branch is still set to 2.6.5 and the plan is to update it to 3.3.0 for Run III and the patches are currently being tested. @agrohsje should we include this as a pLHE request then?

c-bautista: As replied to Sapta by email two days ago, there is no problem for us provide the LHE files for the 3 resonant processes on double production of T1 and X53 including or not 3 body decay within the MCHM5_C9 scenario. We thus wait for the answer from @agrohsje.

agrohsje:
It is very hard to review this PR. 
1.) If you are using standalone code, you cannot use the CMS settings and logic. PDFDEFAULT blabla will not work.
2.) I see that the model is 4FS but it is not indicated in your run card. Settings point to 5FS.
3.) It is hard to judge the interplay between the parameters you are changing. What is the impact on the width of particles etc. Is this correctly reflected in the scan? 
4.) Can you repeat the argument for bwcutoff?  @sihyunjeon what do you think? 
To be honest, I tend to just let you produce the events you have in mind. I cannot judge the settings. I raised a few concerns, but even once they are solved it is hard to decide. So maybe @Saptaparna and @GurpreetSinghChahal can just say how they want to proceed in terms of MC request.

asavoy829:
We are going to reply once Carlos is available in his time zone. Staying tuned.

c-bautista:
Dear @agrohsje

Thanks for your careful review and please find here below our reply to your different remarks or questions.

1) The model was produced and extensively tested (MadGraph and MadAnalysis as well as Pythia + simplified Delphes version for test) in our local machine.

The parameters to specify the model in the "...customizecards.dat" were tested in this way. Now, the specification of PDFDEFAULT... in the "...runcard.dat" was not tested, since, as you mentioned, it is specific to CMS and thus not available in our local machine.

However, we already used a similar "..runcard.dat" for the MCHM5_C2 case and the gridpack was successfully generated. So, we think there should not be a problem here as well.

2) I don't understand this remark. Where do you find 4FS instead of 5FS? We set 
                                             5	=  maxjetflavor     ! Maximum jet pdg code
in the "..runcard.dat"

3) I am not sure what you mean by "scan" but, as said in the first point, we tested these parameters extensively. The widths specified in "..customizecards.dat" are already the ones modified and computed by MadGraph.
So, we are confident that the overall framework including the parameters are certified.

4) For the scenario with resonances decaying into three body decays (MCHM5_C9), we found that intermediate particles does not follow the narrow width approximation, so we need to include their off-shell contributions. The parameter bwcutoff is usually small since it assumed the narrow width approximation, but we found that for this case we need to have a high value to include all contributions correctly.

We would like to come back now to the exchange we had before Christmas related to the need of the "hard_survey" parameter for generating these MCHM5_C9 three body decay events.
Thus, as this parameter is not available in the 2.6.5 version currently used in CMS, we would like to follow Sapta's proposal (22 days ago) and provide directly the LHE files, to proceed on with the full simulation. So, thanks in advance to you, @Saptaparna, @GurpreetSinghChahal to confirm you agree.

Carlos and Aurore

agrohsje:
about 4FS vs 5FS. If you are doing:
import model MCHM14_UFO
display  multiparticles p
you will see
p = g u c d s u~ c~ d~ s~
so the b is not part of the proton. It is treated as massive meaning 4FS. 

i have no further comments. @sihyunjeon might quickly confirm. 
as this doesn't fit 2.6.5 i would not merge to master. 
if @sihyunjeon gives green light, i would suggest you go ahead and produce lhe files as proposed by @Saptaparna

asavoy829:
@agrohsje 
Many thanks for your clarification and looking forward to hear back from @sihyunjeon concerning the green light to produce the LHE files.

asavoy829:
Dear  @sihyunjeon many thanks in advance to send us your reply in order we go ahead with producing the LHE files for MCHM5_C9 case. Aurore and Carlos

sihyunjeon:
bwcutoff is a very delicate thing to touch. Although you say the event generation is possible, there is absolutely 0 guarantee if it's giving you correct physics in the end. Runs like normal doesn't mean it's safe to use. I am even thinking of forcing the bwcutoff setting to 15 as MadGraph authors themselves say it's not a wise choice to give a big value for it.

Some counterpoints to the statement below
> The parameter bwcutoff is usually small since it assumed the narrow width approximation, but we found that for this case we need to have a high value to include all contributions correctly.
bwcutoff ~ 20 is not small, this is just a normal value and this does not necessarily assume NWA (e.g. p p > e+ e- will give you full spectrum of DY process with bwcutoff=15). Your syntax, that splits intermediate particle production and its decay (e.g. p p > X Y, X > A B ==> "X" here) is the one which assumes/forces NWA. So I won't say "correct" here because you are forcing NWA with the process syntax and smearing out NWA effect with bwcutoff enhancement.

I'll say OK for now but this should be treated with extreme care when doing the analysis (especially with data). Should we just close this merge @agrohsje

Saptaparna:
One consequence of setting the BWcutoff to large values is that one gets fewer events than requested as a result of issues with phase space sampling. @c-bautista the you request 10k (just an example) be generated, does Madgraph actually generate 10k events? In my experience, while BWcutoff > 15 \Gamma can seem to be displaying the correct behavior, Madgraph is still integrating with the NWA and this issue with phase space sampling can show up in myriad ways.

asavoy829:
@sihyunjeon @Saptaparna @agrohsje Dear All, thanks for your comments, we will get back to you in a message later today.Aurore and Carlos

c-bautista:
Dear @sihyunjeon, @agrohsje, and  @Saptaparna:

I ran two kind of processes. 
1) generate p p > T T, T> wwt, T > ht
    add process p p > X X, X > wht, X > wt
     with bwcutoff= 10000

and 

2) p p > wwtth, which contains the processes in 1) and many additional diagrams and with a bwcutoff=15

And obtained basically the same value for the cross sections (differing by ~1%). Moreover I compared the kinematical distributions and they agreed pretty well. This means that the processes in 1) are by far the dominant ones.

So, we think it is safe to put a high bwcutoff value for running the production process of wwtth; besides running it following 1) allows the separation of the two different production processes we are interested in generating. Also the generation and simulation will be achieved in a faster way.

sihyunjeon:
Hi, we never know what kinematic distributions should really be checked to guarantee the safeness. If you checked the pTs only, etas might have differences, if you checked both pTs and etas, phis might have difference.
Moreover, the reason why I still don't think it's good for real data analysis with actual data (not the projection studies in SNOWMASS; for SNOWMASS, I have no objections) is because the bwcutoff recommendations come from the authors of MadGraph, that it's not safe to use such large values for event generation. Something that authors do not really approve is always error prone and can lead to unexpected behaviors in the end.

So to conclude here, I have no objections for SNOWMASS analysis. But for real analysis that you or somebody might do with Run-II or Run-III data this should be treated with caution and you might have to rethink your MC strategy.

sihyunjeon:
@agrohsje merge please if you think it's ok

agrohsje:
Did you fix the flavor scheme settings for private LHE production? As settings do not fit 2.6.5 I will close this PR.

asavoy829:
@agrohsje thanks for your message and for taking care. The LHE are being completed. And for information the MCHM5_C2 and MCHM14_D7 samples have been successfully processed and analysed.

PR #3063, opened 2021-12-05
Minor fixes in gridpack_generation.sh

ktht:
- Temporarily disable bash feature that treats undefined variables as  errors when sourcing `cmsset_default.sh`, otherwise the script complains about [this line](https://github.com/cms-sw/cms-common/blob/be997745c00b3f046795520d349ad52d95ca87f0/cmsset_default.sh#L35) and won't proceed any further
- Not only show stderr on screen but also save it to the log file


agrohsje:
Let me merge for the time being. Ideally the script cmsset_default.sh would use something like:
if [ -z ${CMS_PATH:-''} ]; then 
instead of -f .

PR #3076, opened 2021-12-17
Fix LHE writer patch and add possibility of longer queues for steps 2 and 3

covarell:
@agrohsje 

One of the fixes is the failing patch you observed in bb4l.

The second one is a feature needed for complicated POWHEG-packs like VHbb.

Still working on the fix for @Wilsker ' s case, as it seems that not all the process/subprocess directory structure work in the same way and I would not like to apply too many process-dependent patches


agrohsje:
Thanks @covarell . Do you suggest to check and merge or wait. Let me know!

bbilin:

PR #3085, opened 2022-01-17
dipole recoil check for VBF for aMCNLO and powheg

efeyazgan:


efeyazgan:
@agrohsje @sihyunjeon please take a look at this for vbf local/global recoil check. thanks.

PR #3091, opened 2022-01-19
Add MG5 EWKZ2Jets_ZToQQ

jennetd:
I am working on an EWK Z sample with hadronic Z decay following the example here for Z(ll): 
https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ewkzlljj_5f_LO

I plan to submit similar samples for EWK W+ and W- after this one has been reviewed.  Could you please advise if you have any suggested changes? Thanks.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/EWKZ2Jets_ZToQQ/.log
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/EWKZ2Jets_ZToQQ/.log
@@ -1,0 +1,9 @@
+Starting job on  Tue Jan 18 14:07:20 CST 2022
+Running on  Linux cmslpc160.fnal.gov 3.10.0-1160.45.1.el7.x86_64 #1 SMP Tue Oct 12 08:40:46 CDT 2021 x86_64 x86_64 x86_64 GNU/Linux
+System release  Scientific Linux release 7.9 (Nitrogen)
+name: EWKZ2Jets_ZToQQ_TuneCP5_13TeV-madgraph-pythia8/
+carddir: ewkzjjjj_5f_LO
+queue: local
+scram_arch: slc7_amd64_gcc700
+cmssw_version: CMSSW_10_6_19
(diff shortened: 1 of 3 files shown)


bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/EWKZ2Jets_ZToQQ/.log
```diff
@@ -0,0 +1,9 @@
+Starting job on  Tue Jan 18 14:07:20 CST 2022
```
agrohsje, line 1: I assume the .log was committed by mistake?

jennetd: Yes. I will remove it

agrohsje:
Looks already quite good. Question: do you use drjj in the analysis, or mjj cuts? I see them at 0.4 and 10 GeV?

jennetd:
> Looks already quite good. Question: do you use drjj in the analysis, or mjj cuts? I see them at 0.4 and 10 GeV?

We apply an mjj cut and delta eta cut, but would like this sample to be useful for another analysis as well. I think we can use the minimum cuts that you recommend for these vars.

jennetd:
@agrohsje Do you recommend reducing these cut values or leaving as is?

agrohsje:
What is your mjj cut at analysis level?

jennetd:
Hi Alexander, the min mjj is 250 GeV. However, in the resolved analysis this cut is not applied to the Z candidate. Is there a way to apply the mjj cut to only the VBF jets? Thanks.

agrohsje:
Setting cut_decays to false should leave the particles from the decay out so that the mjj cut is only applied to the VBF jets. I would suggest setting jet dr jj to 0.1 and mjj to 200. @sihyunjeon  what do you think? Once agreed this is ok to merge from my side.

sihyunjeon:
Hello, as Alexander said if you turn off "cut_decays" then the jets from Z will not be applied with cuts.
So as Alexander suggested, I edited drjj to 0.1 and mjj to 200 GeV and turned off cut_decays. Will merge.

jennetd:
Great, thank you very much!

PR #3092, opened 2022-01-22
updates for mg33x

sihyunjeon:
a. MadGraph link to www area
b. MadGraph 33x uses python3 so UFO models need to be converted to python3 as well


sihyunjeon:
@SanghyunKo @agrohsje note!

PR #3094, opened 2022-01-27
JHUGen VBF, WH, ZH + decay cards with anomalous couplings added for UL production

lucassmguerra:


agrohsje:
Hi @lucassmguerra . Why did you invert the order of the arguments for SM_NNPDF31_lo_as_0130.input ? 
Rest looks good. I assume you discussed with the HIG people the  LO PDF. Andrei said it was at some point discussed and agreed but I cannot judge myself who wanted what, so I am asking.

lucassmguerra:
Ok, I updated the card. Let me know if I should change anything else.

PR #3095, opened 2022-01-28
Added cards for tWZ boosted samples and for tWZ "resolved" DR2 sample

michele-mormile:


agrohsje:
Please don't submit tarballs to genproductions. We only store the input cards for gridpack production. In case you are covering multiple setups please provide a script and a template. Please clean up accordingly. Thanks!

michele-mormile:
Sorry for that! I reverted the commit

agrohsje:
Do you mind submitting templates and a script? E.g. you just submit 4 or 5 cards and add 
SUBTOPDECAY 
SUBITR
to your cards and replace the SUBTOPDECAY in a script with either leptonic top or hadronic, and SUBITR with the corresponding DS/DR overlap removal. This helps a lot for reviewing and spotting bugs. It is hard to go through 48 files.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/tWZ/template_run_card.dat
```diff
@@ -65,8 +65,7 @@
 # PDF choice: this automatically fixes also alpha_s(MZ) and its evol.  *
 #***********************************************************************
   lhapdf	= pdlabel ! PDF set
-  $DEFAULT_PDF_SETS = lhaid
-
+  $$DEFAULT_PDF_SETS = lhaid
```
sihyunjeon, line 68: Is there a reason to change this with two $s? It might not work. (same for $$DEFAULT_PDF_MEMBERS)

https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/Utilities/gridpack_helpers.sh

michele-mormile: The python template class, which I use in the script, uses $ as placeholder. Two $s escaped and are replaced with just one $, so it should be fine after running the script

sihyunjeon: oh ok i see

sihyunjeon:
Looks fine to me @agrohsje

agrohsje:
Really nice PR. Thanks for this @michele-mormile . Templates look good to me. Maybe a warning, when you run your script, quickly check that all parameters are properly replaced.

PR #3096, opened 2022-02-02
MG5 EWKW_2Jets_WToQQ

jennetd:
Add cards for EWK W+ and W- production following the template used for Z at:https://github.com/cms-sw/genproductions/pull/3091


sihyunjeon:
What will the analysis level cut will be to extract W (and Z from the previous pull request you made)?

jennetd:
For the boosted analysis, we fit the soft drop mass of the AK8 jet above 40 GeV. The resolved analysis will fit a dijet mass from 70 GeV above.

sihyunjeon:
merging

PR #3100, opened 2022-02-12
NMSSM_XYH_YTobb_HToWWTo2QLNu cards on 13_02_2022

Soumyatifr:
We are adding new cards for NMSSM_XToYH, where Y decays to bb and H decays to WW, followed by W->QQ' and W->LNu. We request you to kindly go through the cards and send us your feedback.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_YTobb_HToWWTo2QLNu/NMSSM_XYH_YTobb_HToWWTo2QLNu_MX_mMX_MY_mMY_customizecards.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_YTobb_HToWWTo2QLNu/NMSSM_XYH_YTobb_HToWWTo2QLNu_MX_mMX_MY_mMY_customizecards.dat
@@ -1,0 +1,6 @@
+set param_card mass 45 mMX
+set param_card mass 35 mMY
+set param_card mass 25 125
+set param_card DECAY 45 1.00000e-03
+set param_card DECAY 35 1.00000e-03
+set param_card DECAY 25 4.00000e-03
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_YTobb_HToWWTo2QLNu/NMSSM_XYH_YTobb_HToWWTo2QLNu_MX_mMX_MY_mMY_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_YTobb_HToWWTo2QLNu/NMSSM_XYH_YTobb_HToWWTo2QLNu_MX_mMX_MY_mMY_proc_card.dat
@@ -1,0 +1,7 @@
+import model NMSSMHET_UFO --modelname
+define p = g u c d s b u~ c~ d~ s~ b~
+define j = p
+define lep = e+ mu+ tau+ e- mu- tau-
+define nut = ve vm vt ve~ vm~ vt~
+generate g g > h03 , (h03 > h02 h01, h02 > b b~, h01 >  j j lep nut)
+output NMSSM_XYH_YTobb_HToWWTo2QLNu_MX_mMX_MY_mMY -nojpeg
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_YTobb_HToWWTo2QLNu/NMSSM_XYH_YTobb_HToWWTo2QLNu_MX_mMX_MY_mMY_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_YTobb_HToWWTo2QLNu/NMSSM_XYH_YTobb_HToWWTo2QLNu_MX_mMX_MY_mMY_proc_card.dat
@@ -1,5 +1,5 @@
 import model NMSSMHET_UFO --modelname
-define p = g u c d s b u~ c~ d~ s~ b~
+define p = g u c d s u~ c~ d~ s~
 define j = p
 define lep = e+ mu+ tau+ e- mu- tau-
 define nut = ve vm vt ve~ vm~ vt~
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_YTobb_HToWWTo2QLNu/NMSSM_XYH_YTobb_HToWWTo2QLNu_MX_mMX_MY_mMY_run_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_YTobb_HToWWTo2QLNu/NMSSM_XYH_YTobb_HToWWTo2QLNu_MX_mMX_MY_mMY_run_card.dat
@@ -110,8 +110,8 @@ $DEFAULT_PDF_MEMBERS = reweight_PDF     ! if pdlabel=lhapdf, this is the lhapdf 
 #*********************************************************************
 # Minimum and maximum pt's (for max, -1 means no cut)                *
 #*********************************************************************
- 20.0  = ptj       ! minimum pt for the jets 
- 20.0  = ptb       ! minimum pt for the b 
+ 15.0  = ptj       ! minimum pt for the jets 
+ 15.0  = ptb       ! minimum pt for the b 
  0.0  = pta       ! minimum pt for the photons 
  10.0  = ptl       ! minimum pt for the charged leptons 
  0.0  = misset    ! minimum missing Et (sum of neutrino's momenta)
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_YTobb_HToWWTo2QLNu/NMSSM_XYH_YTobb_HToWWTo2QLNu_MX_mMX_MY_mMY_run_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_YTobb_HToWWTo2QLNu/NMSSM_XYH_YTobb_HToWWTo2QLNu_MX_mMX_MY_mMY_run_card.dat
@@ -1,0 +1,271 @@
+#*********************************************************************
+#                       MadGraph5_aMC@NLO                            *
+#                                                                    *
+#                     run_card.dat MadEvent                          *
+#                                                                    *
+#  This file is used to set the parameters of the run.               *
+#                                                                    *
+#  Some notation/conventions:                                        *
+#                                                                    *
+#   Lines starting with a '# ' are info or comments                  *
+#                                                                    *
+#   mind the format:   value    = variable     ! comment             *
+#*********************************************************************
+#
+#*******************                                                 
+# Running parameters
+#*******************                                                 
+#                                                                    
+#*********************************************************************
+# Tag name for the run (one word)                                    *
+#*********************************************************************
+  tag_1     = run_tag ! name of the run 
+#*********************************************************************
+# Number of events and rnd seed                                      *
+# Warning: Do not generate more than 1M events in a single run       *
+# If you want to run Pythia, avoid more than 50k events in a run.    *
+#*********************************************************************
+  5000 = nevents ! Number of unweighted events requested 
+  0   = iseed   ! rnd seed (0=assigned automatically=default))
+#*********************************************************************
+# Collider type and energy                                           *
+# lpp: 0=No PDF, 1=proton, -1=antiproton, 2=photon from proton,      *
+#                                         3=photon from electron     *
+#*********************************************************************
+     1        = lpp1    ! beam 1 type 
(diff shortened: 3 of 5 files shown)


Saptaparna:
Hi @Soumyatifr, 

2 questions:
-  I assume you want the b-quark to be massive, correct? Then this line in the run card is inconsistent:
 `5 = maxjetflavor    ! Maximum jet pdg code`
- was the run card made with MG5 2.6.5 or some other version of MG5?

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_YTobb_HToWWTo2QLNu/NMSSM_XYH_YTobb_HToWWTo2QLNu_MX_mMX_MY_mMY_proc_card.dat
```diff
@@ -0,0 +1,7 @@
+import model NMSSMHET_UFO --modelname
+define p = g u c d s b u~ c~ d~ s~ b~
+define j = p
```
agrohsje, line 3: The model is per se 4FS. It makes no sense to just add the b in the definition of the p. Please add a restriction card with massless b if you want to use 5FS.

Soumyatifr:
> Hi @Soumyatifr,
> 
> 2 questions:
> 
> * Do you want the b-quark to be massive?
Yes, for that we changed our card 5 flv to 4 flv scheme. 
> * was the run card made with MG5 2.6.5 or some other version of MG5?
Yes we used MG5 2.6.5 
MG5_version

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_YTobb_HToWWTo2QLNu/NMSSM_XYH_YTobb_HToWWTo2QLNu_MX_mMX_MY_mMY_run_card.dat
```diff
@@ -0,0 +1,271 @@
...
+  True  = cut_decays    ! Cut decay products 
+#*********************************************************************
+# Standard Cuts                                                      *
+#*********************************************************************
+# Minimum and maximum pt's (for max, -1 means no cut)                *
+#*********************************************************************
+ 20.0  = ptj       ! minimum pt for the jets 
+ 0.0  = ptb       ! minimum pt for the b 
+ 0.0  = pta       ! minimum pt for the photons 
+ 0.0  = ptl       ! minimum pt for the charged leptons
```
agrohsje, line 116: is it on purpose to cut on jet in decay but not the others? if not intended i suggest to set the cut to 0.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_YTobb_HToWWTo2QLNu/NMSSM_XYH_YTobb_HToWWTo2QLNu_MX_mMX_MY_mMY_customizecards.dat
```diff
@@ -0,0 +1,6 @@
+set param_card mass 45 mMX
+set param_card mass 35 mMY
+set param_card mass 25 125
+set param_card DECAY 45 1.00000e-03
+set param_card DECAY 35 1.00000e-03
+set param_card DECAY 25 4.00000e-03
```
agrohsje, line 6: do you have input  cross sections? what is the motivation behind the width choices? 4 e-3 vs 1e-3?

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_YTobb_HToWWTo2QLNu/NMSSM_XYH_YTobb_HToWWTo2QLNu_MX_mMX_MY_mMY_run_card.dat
Soumyatifr: We always use jet pt cut of 30 GeV at the detector level, so 20 GeV cut in run card should be fine. Thresholds on lepton pt in offline selection is ~20 GeV, so we can also put a cut of 10 GeV here. For b quark jets, since we target to capture b quark pairs using AK8 jets (i.e. boosted topology), we didn't put any cut on individual b quark jets, but practically we can put a cut of 20 GeV here too. We don't put any cut on photons since we don't have those in our event topology.

Soumyatifr: These are on purpose. We always use |eta|<2.5 on leptons at the reconstruction level, hence a loose cut of 3 here. For jets, we relaxed the cut more since we might use jets with |eta| < 3 in the analysis, but this can also be tightened (say, to 4).

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_YTobb_HToWWTo2QLNu/NMSSM_XYH_YTobb_HToWWTo2QLNu_MX_mMX_MY_mMY_run_card.dat
```diff
@@ -0,0 +1,271 @@
...
+ -1.0   = ebmax ! maximum E for the b
+ -1.0   = eamax ! maximum E for the photons
+ -1.0   = elmax ! maximum E for the charged leptons
+#*********************************************************************
+# Maximum and minimum absolute rapidity (for max, -1 means no cut)   *
+#*********************************************************************
+  5.0 = etaj    ! max rap for the jets 
+  -1.0  = etab    ! max rap for the b
+  -1.0  = etaa    ! max rap for the photons 
+  3.0  = etal    ! max rap for the charged leptons
```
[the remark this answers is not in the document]
Soumyatifr: These are on purpose. We always use |eta|<2.5 on leptons at the reconstruction level, hence a loose cut of 3 here. For jets, we relaxed the cut more since we might use jets with |eta| < 3 in the analysis, but this can also be tightened (say, to 4).

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_YTobb_HToWWTo2QLNu/NMSSM_XYH_YTobb_HToWWTo2QLNu_MX_mMX_MY_mMY_run_card.dat
Soumyatifr: Width for the SM-like Higgs boson (pdg Id: 25) is  kept to the SM value of Higgs boson width, which is 4 MeV. For other heavy Higgs bosons  (pdg Id: 35 and 45), they are taken to be of narrow width; however the  choice of 1 MeV is arbitrary, but this is followed in similar X->YH analyses (ex: https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_bbbb/Template/customizecards.dat, https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_WWWW/Template/customizecards.dat).  Therefore, we would also like to keep it 1 MeV so that it's easier to combine the results with the analyses pointed above.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_YTobb_HToWWTo2QLNu/NMSSM_XYH_YTobb_HToWWTo2QLNu_MX_mMX_MY_mMY_customizecards.dat
Soumyatifr: Width for the SM-like Higgs boson (pdg Id: 25) is kept to the SM value of Higgs boson width, which is 4 MeV. For other heavy Higgs bosons (pdg Id: 35 and 45), they are taken to be of narrow width; however the choice of 1 MeV is arbitrary, but this is followed in similar X->YH analyses (ex: https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_bbbb/Template/customizecards.dat, https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_WWWW/Template/customizecards.dat). Therefore, we would also like to keep it 1 MeV so that it's easier to combine the results with the analyses pointed above.

sihyunjeon: How do you get the cross sections for 1.e-3 width case? The cross sections you get from MadGraph isn't reliable if you don't use physics motivated widths.

Soumyatifr: Indeed cross sections obtained from Madgraph are unphysical. So, we only use the shapes of kinematic distributions from the signal samples generated using Madgraph and to get an upper limit on the signal cross section at 95% CL. The bounds on model parameters need to be obtained from dedicated theory calculations (ex. those provided by LHC Higgs working group). This approach is also followed in the approved analyses, for example, HIG-20-014 ((X->YH.>2b2tau), B2G-21-003 (X->YH.>4b),.

Soumyatifr:
Hi @agrohsje , @sihyunjeon do you have any further comments to approve our cards ?

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_YTobb_HToWWTo2QLNu/NMSSM_XYH_YTobb_HToWWTo2QLNu_MX_mMX_MY_mMY_customizecards.dat
agrohsje: Thanks.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_YTobb_HToWWTo2QLNu/NMSSM_XYH_YTobb_HToWWTo2QLNu_MX_mMX_MY_mMY_run_card.dat
```diff
@@ -0,0 +1,271 @@
...
+# (note that etmiss/ptll/ptheavy/ht/sorted cuts always apply)
+#*********************************************************************
+  True  = cut_decays    ! Cut decay products 
+#*********************************************************************
+# Standard Cuts                                                      *
+#*********************************************************************
+# Minimum and maximum pt's (for max, -1 means no cut)                *
+#*********************************************************************
+ 20.0  = ptj       ! minimum pt for the jets 
+ 20.0  = ptb       ! minimum pt for the b
```
agrohsje, line 114: I would further reduce to 15 if your online jet cut is 25. Just to increase the margin a bit more. We talk about partons -> detector level jets.

agrohsje:
I propose the quark cuts to 15, otherwise this looks good. Once @Soumyatifr provides feedback, @sihyunjeon please check and merge. Thanks.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_YTobb_HToWWTo2QLNu/NMSSM_XYH_YTobb_HToWWTo2QLNu_MX_mMX_MY_mMY_run_card.dat
Soumyatifr: @agrohsje and @sihyunjeon 
Thanks for your comment, I reduced the minimum values of ptj and pyb to 15GeV from 20GeV.

agrohsje:
ok for me. @sihyunjeon should confirm.

PR #3107, opened 2022-02-17
Gkk g r g ww

gqlcms:
These cards are for B2G analysis's signal samples:
1. Process is p p > GKK > R g, R > v v (v=w+ w- z)
2. Cards for different GKK's mass and Radion's mass can be generated using 
python Res1ToRes2GluToGluVV_M1_R_gen_card.py


menglu21:
Hi @gqlcms , 
i have some questions on your PR, 
1. will the decay of w/z be handled by pythia, why not madspin for the decay
2. could you make the name of the directory consistent with the process, you use "Res1ToRes2GluToGluWW" and "Res1ToRes2GluToGluVV..." but the cards seem to be the VV case
Cheers, Meng

gqlcms:
Hi @menglu21,
Thank you very much for your comments! 
1. I have updated the Res1ToRes2GluToGluWW to Res1ToRes2GluToGluVV. Now the directory's name are consitant with the process.
2. We followed WKK analysis's process's setting. The Ws' decay are handled by pythia. I think this setting for the decay is fine.
Thank you!

menglu21:
Hi @gqlcms , 
thanks for the update, i went through your cards again and have some further questions/comments:
1. the model seems a 4f model, could you please modify the corresponding parameters in the run card to make them consistent with each other
2. i'm not sure about the "15000  = bwcutoff", maybe others could help to confirm. @sihyunjeon @agrohsje @Saptaparna @GurpreetSinghChahal

sihyunjeon:
i am strongly against bw=15 for any cases

gqlcms:
Hi @sihyunjeon,

Thanks for your comment! I did some LHE level study. The difference between  bw=15000 and  bw=15 is small. 
https://test-qiguo.web.cern.ch/test-qiguo/gKK/sample_production/LHEStudy/LHEStudy.pdf
I think we can switch to 15 = bwcutoff if you agree. Thank you!

sihyunjeon:
Ah ok sorry your syntax is p p > GKK > R g not p p > R g, yes this makes sense but I would advise you to change.

sihyunjeon:
I changed it myself @gqlcms you seemed to agree on this change.

Saptaparna:
Looks good! Merging.

PR #3109, opened 2022-02-21
Upload cards for ALP model for WZG analysis

Senphy:
Dear GEN experts,

1. The model is from https://arxiv.org/abs/1805.06538, which is an extended Axion-Like-Particle (ALP) model which includes couplings with gauge bosons besides photon.
2. This model is also used by [SMP-17-013](https://cms.cern.ch/iCMS/analysisadmin/cadilines?line=SMP-17-013).
3. For proc card, with QED>6 requirement, all SM processes are discarded. Since the decay width given by the auto computing is pretty small (~e-08), the interference effect could be ignored.
4. For run card, we remove most of the deltaR cut and loose the pt cut. The gray line in the figure is the scheme we choose, which shows good agreement with the results given by the theoretic paper. Also this scheme is more inclusive compared with the purple line, which is using the default Madgraph cut. WZG_ALP_xs
5. For the parameter, `fa`, which is the floating parameter that directly affects the couplings between ALP and bosons, is chosen to use 1TeV to get a larger cross section and is set in the model. The decay width of the ALP (9000005/p1 in the model) is set to AUTO. Several mass points will be set to scan from 50GeV to 400GeV. For brevity, the cards uploaded here only contain 100GeV mass point as an example.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ALP_lvlla_4f_LO/ALP_lvlla_4f_LO_customizecards.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ALP_lvlla_4f_LO/ALP_lvlla_4f_LO_customizecards.dat
@@ -1,0 +1,2 @@
+set param_card mass 9000005 1.000000e+02
+set param_card decay 9000005 AUTO
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ALP_lvlla_4f_LO/ALP_lvlla_4f_LO_extramodels.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ALP_lvlla_4f_LO/ALP_lvlla_4f_LO_extramodels.dat
@@ -1,0 +1,1 @@
+Relaxion_Full_UFO.tar.gz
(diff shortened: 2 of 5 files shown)


menglu21:
Hi @Senphy ,
I'm fine with the current cards generally, just two minor comments,

1. since those cards are used in the WZA analysis, would it be reasonable to make the name consistent with WZAToLNuLLA_4f_NLO, e.g., WZAToLNuLLA_4f_LO_ALP, it may be easy for people to track
2. maybe you could add the script for generating cards of other mass point

Cheers, Meng

Senphy:
Hi @menglu21 ,

Thanks for your comments,

For 1), WZAToLNuLLA_4f_NLO is renamed as LLWA_WToLNu in McM, which means that the W has an on shell requirement and is decayed with Madspin. For this PR, the proc scheme has no on-shell approximation so I think they should be separated. Anyway, you can set a proper name for it if needed.

For 2), I have pushed the script for this generation named `generateCards.py`

menglu21:
thanks, merged

sihyunjeon:
I was looking at this but it has something missing

#3112

tvami:
hi @Senphy @menglu21 I dont see the `Relaxion_Full_UFO.tar.gz` uploaded anywhere, I also looked at 
per @DickyChant 's suggestion, but I dont see it.

Senphy:
Hi @tvami , For better documentation, I changed the name from `Relaxion_Full_UFO.tar.gz` to `ALP_Photophobic_UFO.tar.gz`. Sorry for the inconvenience.

tvami:
Great, thanks @Senphy !

PR #3081, opened 2021-12-24
dummy decay patch for mg33x

shimashimarin:
This patch fixes the problem of using a dummy decay madspin card, details are in this discussion:
 
https://answers.launchpad.net/mg5amcnlo/+question/699822

PR #3080, opened 2021-12-23
add madspin grid mode patch

DickyChant:
This patch is a temporary solution for MadSpin grid running. 

More details can be found in [https://bugs.launchpad.net/mg5amcnlo/+bug/1955614](https://bugs.launchpad.net/mg5amcnlo/+bug/1955614)

PR #3111, opened 2022-02-23
Fix a crash for processes with tagged photon

SanghyunKo:
There is a bug in mg5 3.3.1 that it crashes when trying to generate processes with tagged photon, e.g. `p p > a a [QCD]` (https://bugs.launchpad.net/mg5amcnlo/+bug/1960832). The patch will be part of the official mg5 in 3.3.2, but we'll need this until the release of the new mg5 version (along with #3080 & #3081).


SanghyunKo:
I've created a PR instead of a direct commit to make ourselves aware of this issue @sihyunjeon @shimashimarin

sihyunjeon:
something resolved :))

PR #3112, opened 2022-02-24
Revert "Upload cards for ALP model for WZG analysis"

sihyunjeon:
Reverts cms-sw/genproductions#3109


sihyunjeon:
@menglu21 @Senphy Let me rever this. I don't see madspin here

sihyunjeon:
although in the description you wrote you will use MadSpin to decay W

sihyunjeon:
@Senphy can you make a new pull request?

sihyunjeon:
For 1), WZAToLNuLLA_4f_NLO is renamed as LLWA_WToLNu in McM, which means that the W has an on shell requirement and is decayed with Madspin.

Sorry I misread this message from you a little bit. Let me revert and remerge but you need to remove something in run_card.dat. I will re-revert this and merge with the fix.

Senphy:
Never mind. Sorry for not making it clear. Thanks for your check.

PR #3116, opened 2022-02-25
Early 13.6TeV cards for POGs

sihyunjeon:
Added cards for processes in 13.6 TeV
DY LO 01234j
DY NLO 012j
TTbar NLO 012j

Follow ups from discussion in MccM [1]

[1] https://docs.google.com/document/d/15ATgf2hexWZ3IY4pXrYYXBegVKu2GvPB4W4bdBGbBvo/edit

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/DY01234j_LO_5f/DY01234j_LO_5f_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/DY01234j_LO_5f/DY01234j_LO_5f_proc_card.dat
@@ -1,0 +1,16 @@
+import model sm-ckm_no_b_mass
+
+define p = u d c s b u~ d~ c~ s~ b~ g
+define j = p
+define ell+ = e+ mu+ ta+
+define ell- = e- mu- ta-
+define nu = ve vm vt
+define nubar = ve~ vm~ vt~
+
+generate p p > ell+ ell- @0
+add process p p > ell+ ell- j @1
+add process p p > ell+ ell- j j @2
+add process p p > ell+ ell- j j j @3
+add process p p > ell+ ell- j j j j @4
+
+output DY01234j_LO_5f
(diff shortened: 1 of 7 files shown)


sihyunjeon:
@agrohsje @menglu21
Just to provide you some context, several POGs asked if they can request 13.6TeV samples for DY process (added TTbar here without particular reasons). So I thought it would be a good opportunity to check the PdmV developments for Run3 sample preparations ONLY for gridpacks (the rest - fragment, validation, ... are up to the POGs).
So nothing serious! I think I just used the old cards that were used for DY and TTbar (and didn't really care too much about the customizations with PDG values here).

@jordan-martins @GurpreetSinghChahal @Saptaparna

sihyunjeon:
Change the model files to diagCKM matrix cases

Saptaparna:
Is there a ttbar@NLO with 3 jets sample? I'm fine with ttbar@NLO with 2 jets but to capture the maximal complexity, it might be worth to also test ttbar@NLO with 3 jets.

menglu21:
Hi @sihyunjeon , 
just one small comment on the run card, "store_rwgt_info" should be added for the weight.

menglu21:
> @menglu21 Ugh ok thanks! It actually is steered here but yeah better to have it in the cards
> 
> https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/Utilities/gridpack_helpers.sh#L67

i didn't notice this, so things changed a lot after i left MC contact :-)

sihyunjeon:
Again tested NLO3j but based on the process rate, looks like >20days even with 30 cores. I think it's not really safe to push things in the system where reproducibility is limited too much. Let me merge this.

agrohsje:
> > @menglu21 Ugh ok thanks! It actually is steered here but yeah better to have it in the cards
> > https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/Utilities/gridpack_helpers.sh#L67
> 
> i didn't notice this, so things changed a lot after i left MC contact :-) Then i'm fine with the cards

Hihi. I was so tired of checking this in MCM requests that I made sure people cannot escape. :-D

PR #3117, opened 2022-03-01
Change dynamic scale option

Jingyan95:
Dear @sihyunjeon and @agrohsje,
The default dynamic scale of madgraph appears to be buggy when running with UFO: SMEFT_2l2q_UFO_unitary. The solution (switching to option 1) discussed in [1] solves this problem.
[1]https://cms-talk.web.cern.ch/t/gridpack-generation/6561?u=jingyan


agrohsje:
Hi @Jingyan95. Thanks for fixing after our thread.

PR #3123, opened 2022-03-09
Generic NMSSM X->YH cards

portalesHEP:
Cards to be used for X->YH analyses. A small script to generate cards for any (X,Y) masses is included.

@acarvalh FYI

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XToYH/NMSSM_XToYH_MX_500_MY_300_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XToYH/NMSSM_XToYH_MX_500_MY_300_proc_card.dat
@@ -5,18 +5,12 @@ set loop_color_flows False
 set gauge unitary
 set complex_mass_scheme False
 set max_npoint_for_channel 0
-import model sm
 define p = g u c d s u~ c~ d~ s~
 define j = g u c d s u~ c~ d~ s~
 define l+ = e+ mu+
 define l- = e- mu-
 define vl = ve vm vt
 define vl~ = ve~ vm~ vt~
-set group_subprocesses Auto
-set ignore_six_quark_processes False
-set loop_optimized_output True
-set gauge unitary
-set complex_mass_scheme False
 import model NMSSMHET_UFO --modelname
 generate g g > h03 , (h03 > h02 h01)
 output NMSSM_XToYH_MX_500_MY_300 -nojpeg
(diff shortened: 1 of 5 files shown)


bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XToYH/NMSSM_XToYH_MX_500_MY_300_proc_card.dat
```diff
@@ -0,0 +1,22 @@
...
+define j = g u c d s u~ c~ d~ s~
+define l+ = e+ mu+
+define l- = e- mu-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+set group_subprocesses Auto
+set ignore_six_quark_processes False
+set loop_optimized_output True
+set gauge unitary
+set complex_mass_scheme False
```
menglu21, line 19: Hi, 
could you please remove those lines, some of them are duplicate.

portalesHEP: Hi, the lines are now removed

menglu21:
> Hi @menglu21 , do you think is now ready to merge or we need to adapt further?

Hi, thanks for the updates. i'm satisfied with the cards. Just one further question, i noticed that there are "NMSSM_XYH_others", in which the decay are specified. So in this pr, you are planning to generate the "h02 h01" and then decay them in pythia.

acarvalh:
Thank you for reviewing

Yes, as this time we are doing a big request with several final states that
solution was simpler

The pr with the decay fragments is here


https://github.com/cms-sw/genproductions/pull/3125

All the procedure was tested

menglu21:
ok. i sign off to this pr. Next person reviewing it without other comment could merge.

menglu21:
Hi @acarvalh , though this pr was already merged, but after some offline discussion with @sihyunjeon  and others, i may overlook some parameter setting in your pr. it seems you set the decay width for those resonance and they are fixed for all different mass, this could make the final xs incorrect. I checked some previous cards and found they used the same setting, but i guess it's not the correct setting. Could you please use "auto" for them in the customized cards and make a new PR, sorry for the inconvenience.

acarvalh:
Hi @menglu21 thank you for the comment but this is intentional.

we do not calculate the highly model dependent cross section from this UFO model. We set it to a value to ensure we will always have a narrow resonance

menglu21:
Hi @acarvalh thanks, i see, so if the xs is not derived from here, i'm fine with the setup.

PR #3043, opened 2021-11-16
EFT samples of ttZ ZZ and WZ

denschwarz:
Hi all,

for the submission of my gridpacks I was asked to upload my MadGraph cards for you to review them.
Please let my know if there are any problems.

Thank you and cheers,
Dennis

PR #3124, opened 2022-03-09
EFT samples of ttZ ZZ and WZ: qcut=30 for ttZ

denschwarz:
Changing the qcut value for ttZ to 30 in run card


sihyunjeon:
Follow up from #3043  merging

PR #3127, opened 2022-03-11
fix path for new version JHUGen and LO MG5 reweighting configuration

DickyChant:
Two updates for correcting small path issue of genproduction scripts:

1. After updating JHUGen the directory structure for running Powheg+JHUGen changes
2. Configuration path for MG5 LO reweighting is incorrect, which will lead bug for SLC6 preUL 2016 production.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/Utilities/gridpack_helpers.sh
+++ b/bin/MadGraph5_aMCatNLO/Utilities/gridpack_helpers.sh
@@ -96,7 +96,7 @@ prepare_reweight () {
 	mkdir -p madevent/Events/pilotrun
         cp $WORKDIR/unweighted_events.lhe.gz madevent/Events/pilotrun
         cd madevent
-        config=./madevent/Cards/me5_configuration.txt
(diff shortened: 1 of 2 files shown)


bin/MadGraph5_aMCatNLO/Utilities/gridpack_helpers.sh
```diff
@@ -96,7 +96,7 @@ prepare_reweight () {
 	mkdir -p madevent/Events/pilotrun
         cp $WORKDIR/unweighted_events.lhe.gz madevent/Events/pilotrun
         cd madevent
-        config=./madevent/Cards/me5_configuration.txt
+        config=./Cards/me5_configuration.txt # already in madevent dir
```
agrohsje, line 99: To avoid such mistakes, wouldn't it make sense to give full path here?

DickyChant: Valid point, will modify the commit (given that the Powheg+JHUGen bug is fixed)

PR #3136, opened 2022-04-04
ttZ/ZZ/WZ EFT cards: remove # symbol

denschwarz:
From the MC contact, I was informed that the comment in the reweight.dat causes an issue with NanoAOD productions. Since we want the exact version of the cards in this repo that will be used to submit the sample request, please see these small changes.

Best,
Dennis


sihyunjeon:
This is trivial so i'll just merge

PR #3140, opened 2022-04-08
first commit for powheg run3 prep

agrohsje:
WIP. Please ignore. Will merge and handle myself.

### DIFF
--- a/bin/Powheg/Templates/createTarBall_template.sh
+++ b/bin/Powheg/Templates/createTarBall_template.sh
@@ -11,8 +11,6 @@ pwd
 
 rm -f $$WORKDIR/$$folderName'_'$$process'.tgz'
 
-cp -p $$WORKDIR/run_pwg.py $$WORKDIR/$$folderName
-
 if [ -e $$WORKDIR/$$folderName/pwg-0001-stat.dat ]; then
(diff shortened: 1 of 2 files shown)


sihyunjeon:
Long long long long discussion -
Keeping logs here
1. First hand over a. powheg scripts, b. shower builder to Jordan and Justinas so that they can work next week (by Tuesday).
2. Alexander should move the template-like files for powheg to follow what we have for madgraph - We also discussed about the names of directories but I think it's better to keep the structure at least for now for Justinas (it's basically changing the strings from one to another so we should not really stress Justinas with not-too-unnecessary stuffs)
3. (longer term) I should work on physics parameter setting scripts that works for both generators (giving a dictionary file that stores the SM parameters)

sihyunjeon:
Forgot about possible dataset names that we should publicly discuss later on
e.g. DY4Jets_0Jet__madgraphMLM-pythia8 (or DY4JetsTo2L?)
e.g. TT2JetsTo2BLplusNu2Q__amcatnloFXFX-pythia8 (to indicate it's top that decays leptonically, not antitop)

1. Indicate how many jets we produce (or at least what we assume to be a complete set of sample)
2. Decaying particles in alphabetic order with multiplicity in numbers
3. Having two underscores __ not _ so that we can replace this with beamE and tune settings

agrohsje:
The code still needs more work in terms of checks. Will merge for the time being to run central testing.

PR #3142, opened 2022-04-15
Generic EOY NMSSM X->YH cards

portalesHEP:
This PR adds generic NMSSM X->YH cards, similar to the ones from PR https://github.com/cms-sw/genproductions/pull/3123, but with the `run_card` matching that of the gridpacks used for the EOY production of X->YH->bbbb/bbgamgam (using the cards [here](https://github.com/cms-sw/genproductions/tree/mg242legacy/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_bbbb) and [here](https://github.com/cms-sw/genproductions/tree/mg242legacy/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_bbgg) including the H and Y decays). Similarly to the PR mentioned before, these cards are required for the harmonization effort for the different X->YH resonant analyses. 

Thanks in advance for the review.

@acarvalh FYI

edit: the `run_card` is supposed to match that of the EOY gridpacks, not the `proc_card` as previously stated.


menglu21:
Hi @portalesHEP , why not just use the run card of PR#3123, with only changing the pdf to 292000. The current run card miss some parameter, e.g., the event_norm, and it will be then set to "sum" during the generation, while in PR3123 it's average. 
I don't have comment for other parts in this pr.

portalesHEP:
Hi @menglu21, thanks for the comment. As now stated in my first comment, the idea was to have a set of card directly matching the previously used EOY ones, that did however include the H and Y decays. If you think that this does not require to be added as new cards in the repository before generating the gridpacks and pushing them to cvmfs, then this PR can be closed. Let me know what you think is best.

menglu21:
ok, so the only difference of this PR with the existing samples are the decay now are handled by pythia, and those samples using MG+pythia will be produced and comparing with previous samples (only MG).

portalesHEP:
The new samples would rather complete the existing ones. The plan would be to only produce the missing mass points for the analyses needing them, for the ongoing combination effort.

menglu21:
ok, i have no further comments. @Saptaparna @GurpreetSinghChahal @mlizzo please have a check on those cards

mlizzo:
Hi @menglu21 @portalesHEP , it's fine to me as well

PR #3146, opened 2022-04-20
WWJ interface updated for POWHEG BOX RES

mlizzo:
Dear @covarell , @agrohsje 

I tried to update the WWJ Powheg process to the latest MiNNLO calculation available in POWHEG BOX RES [1]. I did some manipulation in runGetSource_template.sh file because some settings didn't allow the code compilation out of the box. I also removed the link to the patch that was needed to reweight the old WWJ to MiNLO accuracy, which is not necessary anymore. The code relies on an external loop provider, and the corresponding grids are downloaded and included in the tarball by the run_pwg_condor.py script. Could you please kindly review (and test) this PR? It should work fine but please let me know if you encounter any issue. I'll be glad to give you more details about the parameters I used to run this process, I've been in contact with the authors while testing it.

[1] https://arxiv.org/pdf/2103.12077.pdf

### DIFF
--- a/bin/Powheg/Templates/createTarBall_template.sh
+++ b/bin/Powheg/Templates/createTarBall_template.sh
@@ -72,10 +72,15 @@ fi
 sed -i 's/pwggrid.dat ]]/pwggrid.dat ]] || [ -e $${WORKDIR}\/pwggrid-0001.dat ]/g' runcmsgrid.sh
 
 if [ "$$process" = "WWJ" ]; then
-   cp -p $${WORKDIR}/$$folderName/POWHEG-BOX/$$process/testrun-nnlops/binvalues-WW.top .
-   cp -r $${WORKDIR}/$$folderName/POWHEG-BOX/$$process/testrun-nnlops/WW_MATRIX .
-   cp -r $${WORKDIR}/$$folderName/POWHEG-BOX/$$process/testrun-nnlops/WW_MINLO .
-   keepTop='1'
+#   cp -p $${WORKDIR}/$$folderName/POWHEG-BOX/$$process/testrun-nnlops/binvalues-WW.top .
+#   cp -r $${WORKDIR}/$$folderName/POWHEG-BOX/$$process/testrun-nnlops/WW_MATRIX .
+#   cp -r $${WORKDIR}/$$folderName/POWHEG-BOX/$$process/testrun-nnlops/WW_MINLO .
+#   keepTop='1'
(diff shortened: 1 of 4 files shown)


covarell:
I have a few comments:
0) minor - do not put # but remove lines, otherwise it is very difficult to follow the "track changes" mode

1) https://github.com/cms-sw/genproductions/pull/3146/files#diff-a946031d265db406992c699bd5f0960280050183f2643c5e902085667686c173R132
These lines appear to be removed but they create the gridpack for all other processes than WWJ, don't they?

2) https://github.com/cms-sw/genproductions/pull/3146/files#diff-083c2ecbee2516fb3aaf4cb519eea4bbeb462666352937c4562fb5f9ed11d662R207 
This is taken from a personal webpage (what if Marius changes institution?  :-) ) maybe we should copy it in a local CMS aream if really needed?

covarell:
3) https://github.com/cms-sw/genproductions/pull/3146/files#diff-306a635579daba4fb10220a64e87ff0b29416fe9cfa2acf8bb8abd5391f8b7cdR308

would it be a problem to update POWHEG-RES for all processes, not just WWJ. We usually follow this policy

mlizzo:
Dear @covarell 

I've removed unused lines and I've fixed the createTarball script. 2) I agree, is there a specific CMS area where we can copy those files? They are actually three in total and each corresponds to a different way to compile and run the code. I hard-coded this information, I've used authors's recommended setting, I don't think one will need to modify it for future uses. 3) I haven't tested other processes that are supposed to be run with POWHEG-RES, if there are some that need to be cross-checked I can do that. In case they are ready to go out-of-the box, then I guess I can just set a unique POWHEG-RES version for all of them.

covarell:
Thanks. We always assume that POWHEG updates are backword-compatible, otherwise book-keeping becomes unmanageable. This is not always the case, but we fix later based on user complaints...

mlizzo:
Hi @covarell , now there's a single PowhegBOXRES version for all processess that are supposed to be run with it, assuming they work the same way as before

covarell:
+1

mlizzo:
Hi @GurpreetSinghChahal @Saptaparna , could you please review this pending PR? @covarell has already checked it as far as I understood

mlizzo:
Hi @agrohsje @covarell , I've solved some merge conflicts due to https://github.com/cms-sw/genproductions/pull/3179 that has been approved while this was still pending. Please have a look at it, if you don't have further comments I'm going to merge it next week.

Saptaparna:
@mlizzo I am happy to merge this, if you would like. Any tests we should perform? Thanks!

mlizzo:
Hi @Saptaparna thanks! The code should be fine and not affect other Powheg processes, I used it to create a WWJ gridpack. I think it's ready to be merged :)

PR #3152, opened 2022-04-25
Add JHUGen+Powheg cards for HWW4q

StephenChao:
Add the JHUGen+Powheg cards for HWW decays to 4q final state, following the examples here: https://github.com/cms-sw/genproductions/tree/master/bin/Powheg/examples/gg_H_quark-mass-effects_withJHUGen_NNPDF30_13TeV, but change the NNPDF30 to NNPDF31 while using the pdfset=325300. 
Add the "WriteFailedEvents=2" for JHUGen input to satisfy the request_fragment_check.py.


menglu21:
i don't have further comments, @Saptaparna @GurpreetSinghChahal @mlizzo please have a check.

mlizzo:
Hi @StephenChao , I have some questions actually. As far as I can see from JHUGen manual , the PChannel parameter "is only meaningful in the spin-2 case and for ttH and bbH production mechanisms". Also, since you're only using JHUGen for the W bosons decay, I think that setting is not needed at all. I would remove everything but DecayMode1=5 DecayMode2=5 WriteFailedEvents=2, since all other parameters seem unnecessary to me. Please @menglu21 tell me if this sounds reasonable to you as well

menglu21:
Hi @mlizzo you are right. I didn't touch those options since they will be ignored due to default or nor considered, either is fine to me, remove them or keep them there.

mlizzo:
Hi @menglu21 , I would ask @StephenChao to remove them for the sake of clarity, sorry for being nitpicker

StephenChao:
Hi @mlizzo @menglu21  , no problem, thanks for your comments, I will remove the other options but " DecayMode1=5 DecayMode2=5 WriteFailedEvents=2".

PR #3157, opened 2022-04-27
diquark Suu->ut->u l+ nu b production, https://arxiv.org/pdf/1912.131…

iashvili:
MD5 cards for diquark Suu->ut->u Wb->u l+nu b production according to the model by Bogdan Dobrescu:  https://arxiv.org/pdf/1912.13155.pdf

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/diquark_Suu_ut/diquark_Suu_ut_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/diquark_Suu_ut/diquark_Suu_ut_proc_card.dat
@@ -1,0 +1,11 @@
+import model diquarkut2021_UFO
+define l+ = e+ mu+ ta+ 
+define l- = e- mu- ta-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+define lept = l+ l- vl vl~
+
+generate p p > u t, (t > w+ b, w+ > l+ vl)
+
+output diquark_Suu_ut -nojpeg
+
--- a/bin/MadGraph5_aMCatNLO/cards/diquark_Suu_ut/diquark_Suu_ut_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/diquark_Suu_ut/diquark_Suu_ut_proc_card.dat
@@ -3,7 +3,6 @@ define l+ = e+ mu+ ta+
 define l- = e- mu- ta-
 define vl = ve vm vt
 define vl~ = ve~ vm~ vt~
-define lept = l+ l- vl vl~
 
 generate p p > u t, (t > w+ b, w+ > l+ vl)
 
--- a/bin/MadGraph5_aMCatNLO/cards/diquark_Suu_ut/diquark_Suu_ut_run_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/diquark_Suu_ut/diquark_Suu_ut_run_card.dat
@@ -1,0 +1,270 @@
+#*********************************************************************
+#                       MadGraph5_aMC@NLO                            *
+#                                                                    *
+#                     run_card.dat MadEvent                          *
+#                                                                    *
+#  This file is used to set the parameters of the run.               *
+#                                                                    *
+#  Some notation/conventions:                                        *
+#                                                                    *
+#   Lines starting with a '# ' are info or comments                  *
+#                                                                    *
+#   mind the format:   value    = variable     ! comment             *
+#*********************************************************************
+#
+#*******************                                                 
+# Running parameters
+#*******************                                                 
+#                                                                    
+#*********************************************************************
+# Tag name for the run (one word)                                    *
+#*********************************************************************
+  tag_1     = run_tag ! name of the run 
+#*********************************************************************
+# Run to generate the grid pack                                      *
+#*********************************************************************
+  .false.     = gridpack  !True = setting up the grid pack
+#*********************************************************************
+# Number of events and rnd seed                                      *
+# Warning: Do not generate more than 1M events in a single run       *
+# If you want to run Pythia, avoid more than 50k events in a run.    *
+#*********************************************************************
+  250 = nevents ! Number of unweighted events requested 
+      0       = iseed   ! rnd seed (0=assigned automatically=default))
+#*********************************************************************
+# Collider type and energy                                           *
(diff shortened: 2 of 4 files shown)


bin/MadGraph5_aMCatNLO/cards/diquark_Suu_ut/diquark_Suu_ut_run_card.dat
```diff
@@ -0,0 +1,270 @@
...
+#*********************************************************************
+# Matching - Warning! ickkw > 1 is still beta
+#*********************************************************************
+ 0        = ickkw            ! 0 no matching, 1 MLM, 2 CKKW matching
+ 1        = highestmult      ! for ickkw=2, highest mult group
+ 1        = ktscheme         ! for ickkw=1, 1 Durham kT, 2 Pythia pTE
+ 1        = alpsfact         ! scale factor for QCD emission vx
+ F        = chcluster        ! cluster only according to channel diag
+ F        = pdfwgt           ! for ickkw=1, perform pdf reweighting
+ 5        = asrwgtflavor     ! highest quark flavor for a_s reweight
```
menglu21, line 71: line 71, since you are using 4f scheme, please change it to 4.

bin/MadGraph5_aMCatNLO/cards/diquark_Suu_ut/diquark_Suu_ut_run_card.dat
```diff
@@ -0,0 +1,270 @@
...
+# Automatic ptj and mjj cuts if xqcut > 0
+# (turn off for VBF and single top processes)
+#**********************************************************
+   T  = auto_ptj_mjj  ! Automatic setting of ptj and mjj
+#**********************************************************
+#                                                                    
+#**********************************
+# BW cutoff (M+/-bwcutoff*Gamma)
+#**********************************
+  15000  = bwcutoff      ! (M+/-bwcutoff*Gamma)
```
menglu21, line 87: generally a very large number for bwcutoff is not recommended, i suggest to use the default 15 here

bin/MadGraph5_aMCatNLO/cards/diquark_Suu_ut/diquark_Suu_ut_run_card.dat
iashvili: fixed

bin/MadGraph5_aMCatNLO/cards/diquark_Suu_ut/diquark_Suu_ut_run_card.dat
iashvili: fixed

bin/MadGraph5_aMCatNLO/cards/diquark_Suu_ut/diquark_Suu_ut_proc_card.dat
```diff
@@ -0,0 +1,11 @@
+import model diquarkut2021_UFO
+define l+ = e+ mu+ ta+ 
+define l- = e- mu- ta-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+define lept = l+ l- vl vl~
```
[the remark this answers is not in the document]
iashvili: lept line removed

menglu21:
i don't have further comment. @Saptaparna @GurpreetSinghChahal @mlizzo please have a check and feel free to merge.

sihyunjeon:
Hello just short comments (still having fun looking into genproductions :|
1. Can you synchronize your ptb to your ptj? (b pt cut to be 10 as you required for j pt). The impact should be small but as you turned on cut_decays to True, having ptb and ptj separately is a bit awkward.
2. Just one naive question. Is p p > ubar tbar not possible with your model card? I am wondering if you are thinking of doing single 'positive charge' lepton analysis only.

mlizzo:
Hi @sihyunjeon , good catch about point 1! On point 2, as far as I can see in the diquarkut2021 UFO model, the Suu scalar can only decay to tt/tu/uu final states. From the paper @iashvili has linked, it seems to me that the ubar tbar final state involves right-handed quarks, please correct me if I'm wrong.

agrohsje:
Wouldn't it make sense to turn off cut_decay? It will only affect the top decay products which by the involved masses are anyway not small. So there is little inefficiency while it is always easier in the case of external normalization if the sample is fully  inclusive.

sihyunjeon:
> Wouldn't it make sense to turn off cut_decay? It will only affect the top decay products which by the involved masses are anyway not small. So there is little inefficiency while it is always easier in the case of external normalization if the sample is fully inclusive.

either way looks fine

Saptaparna:
I agree with @agrohsje. I looked at the cards yesterday and they seem very close to the default MG5 setup, so setting cut_decay=False is much easier at this stage than worrying about the exact value of the GEN-level cuts.

sihyunjeon:
@Saptaparna Do you mean setting `cut_decay=False` and keeping the rest of the setup as it is now? Or turning of ptj cut as well?

Saptaparna:
Keep everything the same and just use `cut_decay=False`.

sihyunjeon:
@Saptaparna Ummm I don't think that's what alexander meant then.
a. `ptj=10` & `ptb=0` & `cut_decay=False`
b. `ptj=10` & `ptb=0` & `cut_decay=True` 

a. and b. practically doesn't have any difference. it would make sense to turn off `ptj` and set `cut_decay=False` if I understood @agrohsje intention correctly

Saptaparna:
Ok, while @agrohsje clarifies, I wanted to make sure that 9936661 is sufficiently large and does not have potential conflicts with SUSY particles or anything like that.

agrohsje:
Yes. I was proposing to turn off ptj and set cut_decay=False. That should make almost no difference in terms of efficiency but allows easy scaling to inclusive cross section.

Saptaparna:
@iashvili do you already have a preferred range of parameters that you will be scanning? I suspect that you will be looking at vary high values of Suu based on a cursory glance of the paper (https://arxiv.org/pdf/1912.13155.pdf).

mlizzo:
It's fine for me, @menglu21 @Saptaparna @GurpreetSinghChahal if you don't have further comments, we can merge it.

menglu21:
ok, merged.

PR #3158, opened 2022-04-29
Heavy N single lepton channel

sihyunjeon:
templates, DY, W are added in this commit


menglu21:
Hi @sihyunjeon nice pr, i don't have comment, @Saptaparna @GurpreetSinghChahal  @mlizzo please have a check and feel free to merge.

mlizzo:
Hi @sihyunjeon , nice PR indeed! I only have a small comment, which might be driven by my ignorance: is the b quark mass explicitly set only in the W customizecards.dat file because you're including the z/h->bb channel? It does not appear in the corresponding customizecards.dat template as far as I can see, so I guess you added that line manually. Maybe you can let the writeCards script add it for you. Please tell me if this makes sense to you, otherwise the PR can be merged already.

sihyunjeon:
Hi @mlizzo actually the b yukawa didn't really do anything in the end. I was switching back and forth between the 4f and 5f version of this model and in the end decided to use 4f so I didn't need to steer ymb in the customize card at all. I just removed it as ymb is defined in UFO model already.

PR #3160, opened 2022-05-02
added cards for UL WW semileptonic high mass

lenzip:
This is a set of cards for the UL production of High mass samples for the semileptonic H->WW analysis. @mlizzo should be aware that this request is coming


mlizzo:
Hi @menglu21 , this request is fine to me, the PDF set has been updated for UL and the other parameters are standard Powheg settings. Please review it and, if it's ok to you as well, merge it.

menglu21:
Hi @mlizzo @lenzip is JHU card needed for VBF_H

mlizzo:
Hi @menglu21 , it's needed for both VBF and ggH. I've now noticed that the very same JHUGen card is already in the JHUGen folder [1], so I guess that commit can be removed.
[1] https://github.com/cms-sw/genproductions/blob/master/bin/JHUGen/cards/decay/WWlnuqq_withtaus_reweightdecay_CPS.input

PR #3161, opened 2022-05-02
cards for WprimeToTauNu

valesark:
Cards for the SSM W' and NUGIM interpretation.


menglu21:
hi @valesark , could you please use templates cards and then a script to implement all the scanning parameters, actually it's kind of impossible for us to review ~1k cards

valesark:
Hi @menglu21, 
Thank you for your response. I created the pull request for Mr. Si Hyun Jeon to have a look at the cards, since I had a question regarding the generator fragment names. 
But for future, do I understand it correctly that I should upload the scripts that produce these cards? Since there are two different scripts, one calculating the width depending on the mass and coupling and second one generating the cards.

menglu21:
yes, exactly, actually my browser can't load all the cards, so please close this pr and use templates cards, and script to create all those cards.

PR #3164, opened 2022-05-03
AToZHTott cards

srudra3:
Cards for AToZHTott  UL signal sample production.


menglu21:
Hi @srudra3 is this pr for UL? if yes, please put them in the 2017/13TeV directory

menglu21:
Hi @srudra3 is this pr for UL? if yes, please put them in the 2017/13TeV directory

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/AZHToLLtt/AZHToLLtt_customizecards.dat
```diff
@@ -0,0 +1,11 @@
...
+set param_card frblock 2 0.99  #sinbma
+set param_card mass 25 125.0
+set param_card mass 35 HMASS
+set param_card mass 36 AMASS
+set param_card mass 37 AMASS
+set param_card mass 23 9.118760e+01
+set param_card mass 24 80.379
+set param_card width 35 Auto
+set param_card width 23 2.4952
+set param_card width 37 Auto
```
menglu21, line 11: since w/z properties here are the same with those in the default model parameter card, line 7/8/10 could be removed. "set param_card width 36 Auto" should be added.

menglu21, line 11: since w/z properties here are the same with those in the default model parameter card, line 7/8/10 could be removed. "set param_card width 36 Auto" should be added.

menglu21 (commented):
the directory "AZHToLLtt_mA900_mH350" could be removed if it's just for an example output from the copy.sh

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/AZHToLLtt/AZHToLLtt_customizecards.dat
srudra3: By setting width 36 to Auto, the gen level event record for that particle is missing. Is that expected? Do you recommend this setting?

menglu21: you mean GEN level or LHE level, if you mean in the lhe file, the off-shell particles will not appear there. and is it in the lhe when without "width 36 to Auto"

menglu21: it should be in the GEN level also. btw, i didn't find "frblock" in the model, could you please tell me your setup and i could check if i reproduce your results

srudra3: I fixed the model, "frblock" should now be available in the new model. Are you referring to the large widths you see in LHE output for 35 and 36?  If yes, I see it too. What do you suggest?

menglu21: i tried with the new model, and use m_36=1000, m_35=600, (frblock values are 1.0 and 0.99), by setting the decay mode of 35/36/37 to "Auto", now i got "DECAY  36   1.390593e+02", "DECAY  37   1.412480e+02", "DECAY  35   7.387922e+01" (please do a test confirm what i got), i think those should be valid for NWA, so the "Auto" could be used. btw, with the new model, you need to set "p = p b b~" since h3 only couples to "b and top quark". (NWA is valid under LO, so you may need to remove "[QCD]" in the syntax)

srudra3: Thanks, I can confirm that I see the same widths. When I try to define p = p b b~, I see a warning "WARNING: Process: g/u/c/d/s/u~/c~/d~/s~/b/b~ g/u/c/d/s/u~/c~/d~/s~/b/b~ > h3> h2 z WEIGHTED=4 [ all = QCD ] can have real emission processes which are not finite.
To avoid this, please use multiparticles when generating the process and be sure to include all the following particles in the multiparticle definition:
 g, c, s, u, d, d~, u~, s~, c~ " and then it crashes with "Command "launch -n pilotrun" interrupted with error:
KeyError : 'parton_shower'"

menglu21: Hi @srudra3 , i tested using higher MG version and it gave me different message, so please ignore my previous comment on the "p = p b b~". As for the width, if you don't care about the reported cross section, you can leave the width there because the kinematics will be almost the same. if you need the cross section, you may consider following:
1. the previous width is i got from LO, it seems width of 36/37 don't change much, so i think you can still use "auto" for them.
2. as for 35, since you decay it using madspin, the width will be recalculated anyway, so maybe you don't need to care about its width

srudra3: Do I understand correctly that I don't need to redefine p = p b b ~?

I still have the issue of 'A' not showing up in the GEN level  for some mass points when I use 'Auto' width computation. I don't understand this.

menglu21: yes, you can use the default "p" definition. As for the "A" existence in LHE, i did a test (mA=1000TeV) and found only width smaller than 100GeV, the lhe include "A". So it means if width of A is larger than 10% of its mass, it will not show up in the LHE even if it's onshell. in "auto" case, the width is around 140GeV, so it's the later case. I don't know such limitation in MG before, maybe @agrohsje @sihyunjeon could comment here.

srudra3: I looked at two mass points, For mA=1000 GeV and mH=600 GeV, the widths I see in LHE are 139GeV for A and 73.4 GeV for H.  I see 'A'  in the LHE output but it is missing in the GEN event record.

For mA=2100 GeV and mH=2000 GeV, the widths I see in LHE are 181.6 GeV for A and 437 GeV for H. I again see A in LHE output and this time I also see it in the GEN level. But the GEN level distribution of mA has almost twice the width as in LHE.  

I hope this further improves the clarity of the issue.

menglu21: Hi, could you please put your test cards in some place and then i can try with exact your inputs, since i couldn't reproduce what you saw.

srudra3: I put the cards in my lxplus public area "/afs/cern.ch/user/s/srudrabh/public/GENValidationAZH" along with the output LHE files for the two mass points. 

The pythia fragment I used to produce GEN root files is "AZH_fragmentUL2017.py". The GEN level rootfiles can be found there too. Please let me know if you need anything else.

menglu21: i checked the /afs/cern.ch/user/s/srudrabh/public/GENValidationAZH/cmsgrid_final_mA1000mH600.lhe, the 'A' is not in the lhe event list, all the event records are like:
```
<event>
 4      0 +2.6613500e-06 1.04945200e+03 7.81860800e-03 8.14458500e-02
       21 -1    0    0  502  501 +0.0000000000e+00 +0.0000000000e+00 +3.8336892424e+02 3.8336892424e+02 0.0000000000e+00 0.0000e+00 1.0000e+00
       21 -1    0    0  501  502 -0.0000000000e+00 -0.0000000000e+00 -7.1820414879e+02 7.1820414879e+02 0.0000000000e+00 0.0000e+00 1.0000e+00
       35  1    1    2    0    0 +2.2584479271e+02 -2.0860334798e+02 -3.8595747664e+02 7.7684258441e+02 6.0000000000e+02 0.0000e+00 0.0000e+00
       23  1    1    2    0    0 -2.2584479271e+02 +2.0860334798e+02 +5.1122252082e+01 3.2473048862e+02 9.1187600000e+01 0.0000e+00 0.0000e+00
<mgrwt>
...
```
this is what i said particle 36 is not in the event, there are only 4 particles, 2 are incoming partons and the other 2 are outgoing "H" and Z. so did i understand you wrongly

mschrode: Hi @menglu21 , @agrohsje , @sihyunjeon, thanks a lot for your help with these samples. Do you have any further insights on why the A boson should be missing in the event record?

menglu21: Hi, one possible solution is to set the width manually, e.g., if the actual width of A is wA, which corresponds to a mass mA. if wA is larger than 10% of mA, then you set the width to some other value, say wA', which is smaller than 10% of mA. You should be able to have "A" in the lhe file if it's onshell. But you need to correct the MG cross section by a factor wA'/wA (if you care about the cross section). 
i will ask the MG author of such "10% threshold", to be honest i didn't notice it before. In the meantime i also suggest you to have a test using different widths and check the distribution and also the reported cross section.

srudra3: In principle, we don't care about the actual cross-section value coming out of MG. Does this mean I have to make separate cards for each of the mass points?

menglu21: Hi, if you don't not use the cross section i think you can set the width to a small value. and you can check https://answers.launchpad.net/mg5amcnlo/+question/702105, the MG author confirmed that the particle will not show up in LHE even if it's on-shell but its width is larger than 10%*mass.

srudra3: When I set the widths to arbitrarily small values, I see a warning "Critical: BR >1 for particle 35". Can this be safely ignored since we don't use the cross-sections?

srudra3: Hi @menglu21, I have updated the cards with widths at 3% of their masses. Please let me know if this is ok.

menglu21: Hi @srudra3 i think it's fine to set the width to a small value since you don't use the XS. yes, you can implement the width in the script also.

srudra3: Thank you! Please let me know if the cards look ok or if you suggest any changes

menglu21: the cards look fine to me. and will you update your copy.sh to update the 0.03*mass to a single number

srudra3: I have an updated copy.sh to include the widths of A and H. Please let me know if there's anything else I need to do. Thank you!

menglu21:
thanks, the cards look fine to me now, @Saptaparna @GurpreetSinghChahal @mlizzo please have a check and feel free to merge.

PR #3165, opened 2022-05-03
Template cards and scripts to implement all the scanning parameters

valesark:
Hello, 
Here are the template cards and all the necessary scripts to create similar directories with changed parameters (masses and couplings) for the W'-> tau + nu decay.


bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/WprimeToTauNu_LO/customize_cards.py
```diff
@@ -0,0 +1,42 @@
...
+#### Set dir with template cards to adjust ####
+template_dir = "WprimeToTauNu_M-6000_kR-0.01_LO_NNPDF31nnlo/"
+#### Loop through all masses ####
+for mass in masses:
+#### Loop through all couplings ####
+    for coupling in kR: 
+        print(WprimeWidth(mass,coupling))
+        directory = "WprimeToTauNu_M-"+str(mass)+"_kR-"+str(coupling)+"_LO_NNPDF31nnlo" # Set the output name
+        if os.path.exists(directory): # Check if directiory exists already
+            shutil.rmtree(directory)  # Delete if it already exists
```
mlizzo, line 22: These lines would remove all files stored in your template directory when reaching the 6000 GeV mass point and kR = 0.01

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/WprimeToTauNu_LO/WprimeToTauNu_M-6000_kR-0.01_LO_NNPDF31nnlo/WprimeToTauNu_M-6000_kR-0.01_LO_NNPDF31nnlo_run_card.dat
```diff
@@ -0,0 +1,284 @@
...
+#
+#*******************************                                                 
+# Parton level cuts definition *
+#*******************************                                     
+#                                                                    
+#
+#*********************************************************************
+# BW cutoff (M+/-bwcutoff*Gamma) ! Define on/off-shell for "$" and decay  
+#*********************************************************************
+  10000000000.  = bwcutoff      ! (M+/-bwcutoff*Gamma)
```
mlizzo, line 102: Why is bwcutoff set to such a high value? @menglu21 this looks dangerous to me, can you also comment on that?

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/WprimeToTauNu_LO/customize_cards.py
valesark: Yes, sorry, locally I used a template called in another way. That is why there was no conflict. They are unnecessary for this case and can be deleted.

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/WprimeToTauNu_LO/WprimeToTauNu_M-6000_kR-0.01_LO_NNPDF31nnlo/WprimeToTauNu_M-6000_kR-0.01_LO_NNPDF31nnlo_run_card.dat
valesark: Hi @mlizzo , 
The bwcutoff value is set to such a high number because one misses an offshell production when there are large masses and small couplings. For example, mass 5 TeV and kR 0.0001, i.e. the resulting width is very tiny. So, if I have a mass of 5 TeV and a width of 1e-6 and I want to have a distribution to 0, then I need Gamma*bwcutoff to be 5000, so 5000/1e-06 already results in a high number and I took a bit higher number just to be safe.
I checked it for cases with larger width, and this bwcutoff value seems to work fine for those cases as well.

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/WprimeToTauNu_LO/customize_cards.py
valesark: I removed these lines from the code. Thank you for noticing @mlizzo .

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/WprimeToTauNu_LO/WprimeToTauNu_M-6000_kR-0.01_LO_NNPDF31nnlo/WprimeToTauNu_M-6000_kR-0.01_LO_NNPDF31nnlo_run_card.dat
sihyunjeon: + it's disapproved by the authors themselves

valesark: @sihyunjeon Then I guess we could change it to the highest value that is possible. Which value do you recommend to use that would not go into conflict with the shower step?

sihyunjeon: it's much smaller value 15-30. default is 15

valesark: I do not think that the value 15 can work for my case. Could not I at least use bwcutoff = 15000? Like it is also mentioned here: 

mlizzo:
@valesark please also change output and card names, because this will create issue when the gridpack is generated: we cannot have special characters, I would suggest to replace 0.01 with 0p01

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/WprimeToTauNu_LO/WprimeToTauNu_M-6000_kR-0.01_LO_NNPDF31nnlo/WprimeToTauNu_M-6000_kR-0.01_LO_NNPDF31nnlo_run_card.dat
menglu21: i think the twiki page is saying the "equivalent" thing, because for LO process "p p > e+ vl", the only diagram is through w boson propagator, which could be onshell and offshell. so if you use syntax "p p > w+ > e+ vl", and set the bwcutoff be a large number, then the offshell region is added automatically.

sihyunjeon: Hi as Meng said,
`generate p p > wp+ > ta+ vt` would already give offshell effect (although this syntax is also not so nice - it's better than super large bwcutoff).
You only miss the offshell contribution when you do `p p > wp+, wp+ > ta+ vt` not `p p > wp+ > ta+ vt`

menglu21: actually, since the width is so small, i don't think the bwcutoff will even have effect on the result. even though the very offshell region is now becoming onshell by using very large bwcutoff, the probability for a event having such offshell event doesn't change.

valesark: Hi @menglu21 and @sihyunjeon, 
Thank you for your answers.
I tested the results with this syntax `generate p p > wp+ > ta+ vt ` for bwcutoff 15 and bwcutoff 15000. As I mentioned before, for large masses and small couplings, when the width is tiny, using bwcutoff of 15000 is more or less okay, whereas when I use the value of 15, I miss the offshell production.. But I guess if I do so, then it also goes into conflict with the showering step?
As I understand, there are two possibilities: 
1. Stick to the syntax that I have `generate p p > wp+ > ta+ vt ` and `generate p p > wp+ > ta- vt~` , using bwcutoff 15.
2. Switch to the syntax `p p > ta+ vt NP==2` and `p p > ta- vt~ NP==2` and use bwcutoff 15000?
Sorry for asking too many questions. I just want to figure out, which option is the best, so that I start producing cards and gridpacks with correct settings this time.

menglu21: Hi @valesark , i had some offline discussion with Sihyun, sorry i didn't make it clear, actually for your case, these two syntaxs are the same since there is only s-channel. and also value of bwcutoff won't have effect.

valesark: Hi @menglu21,
Thank you for your explanations. 
Okay, then I will leave the syntax like it is and change the bwcutoff value to 15.

sihyunjeon: @valesark 

If you look at https://twiki.cern.ch/twiki/bin/view/CMS/MGSyntaxPitfall, under ` Avoid using ">"s in series without ","s `
it will show you that bwcutoff 15 or 15000 will give equivalent physics (yellow black lines are identical), you would not see much difference in your cross section whether you set it to 15 or 15000 if the studies we did earlier was correct.

Can you compare cross sections between 15 and 15000?

valesark: @sihyunjeon ,  I checked it and the difference is small, indeed. I will create new cards now.

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/WprimeToTauNu_LO/WprimeToTauNu_M-600_kR-0p0001_LO_NNPDF31nnlo/WprimeToTauNu_M-600_kR-0p0001_LO_NNPDF31nnlo_run_card.dat
```diff
@@ -0,0 +1,284 @@
...
+     6800.0     = ebeam2  ! beam 2 total energy in GeV
+#*********************************************************************
+# Beam polarization from -100 (left-handed) to 100 (right-handed)    *
+#*********************************************************************
+     0.0     = polbeam1 ! beam polarization for beam 1
+     0.0     = polbeam2 ! beam polarization for beam 2
+#*********************************************************************
+# PDF CHOICE: this automatically fixes also alpha_s and its evol.    *
+#*********************************************************************
+     nn23lo1    = pdlabel     ! PDF set
```
sihyunjeon, line 47: Hello this should be changed to `lhapdf = pdlabel` actually

@menglu21 @mlizzo

valesark: This setting was set by MG automatically. If needed, of course I can change it as well.

sihyunjeon: so that means it will give you madgraph default pdfs :) you should change it

valesark: Done! Thank you.

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/WprimeToTauNu_LO/WprimeToTauNu_M-600_kR-0p0001_LO_NNPDF31nnlo/WprimeToTauNu_M-600_kR-0p0001_LO_NNPDF31nnlo_run_card.dat
```diff
@@ -0,0 +1,284 @@
...
+#**************************************                                  
+#
+0.5 1 2 = sys_scalefact  # factorization/renormalization scale factor
+None = sys_alpsfact  # alpha_s emission scale factors
+auto = sys_matchscale # variation of merging scale
+# PDF sets and number of members (0 or none for all members).
+NNPDF23_lo_as_0130_qed = sys_pdf # list of pdf sets
+# MSTW2008nlo68cl.LHgrid 1  = sys_pdf
+#
+
```
menglu21, line 283: line 272 to 283 could be removed

menglu21 (commented):
others look fine to me

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/WprimeToTauNu_LO/WprimeToTauNu_M-600_kR-0p0001_LO_NNPDF31nnlo/WprimeToTauNu_M-600_kR-0p0001_LO_NNPDF31nnlo_run_card.dat
valesark: Hi @menglu21,
Thank you. It's done, I removed these lines. Should I proceed to creation of new gridpacks with these cards or is there something else that needs to be edited?

menglu21: Hi @valesark thanks, i don't have further comment, let's see whether others are fine with them, if not we can merge and you can start to produce gps. @Saptaparna @GurpreetSinghChahal @mlizzo please have a check and feel free to merge

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/WprimeToTauNu_LO/WprimeToTauNu_M-600_kR-0p0001_LO_NNPDF31nnlo/WprimeToTauNu_M-600_kR-0p0001_LO_NNPDF31nnlo_run_card.dat
```diff
@@ -0,0 +1,273 @@
...
+ -1.0   = ebmax ! maximum E for the b
+ -1.0   = eamax ! maximum E for the photons
+ -1.0   = elmax ! maximum E for the charged leptons
+#*********************************************************************
+# Maximum and minimum absolute rapidity (for max, -1 means no cut)   *
+#*********************************************************************
+  5.0 = etaj    ! max rap for the jets 
+  -1.0  = etab    ! max rap for the b
+ 2.5  = etaa    ! max rap for the photons 
+ 2.5  = etal    ! max rap for the charged leptons
```
menglu21, line 140: hi, could you please change "2.5  = etal" to "3.0  = etal"

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/WprimeToTauNu_LO/WprimeToTauNu_M-600_kR-0p0001_LO_NNPDF31nnlo/WprimeToTauNu_M-600_kR-0p0001_LO_NNPDF31nnlo_run_card.dat
```diff
@@ -0,0 +1,273 @@
...
+#
+#*********************************************************************
+# BW cutoff (M+/-bwcutoff*Gamma) ! Define on/off-shell for "$" and decay  
+#*********************************************************************
+  15.  = bwcutoff      ! (M+/-bwcutoff*Gamma)
+#*********************************************************************
+# Apply pt/E/eta/dr/mij/kt_durham cuts on decay products or not
+# (note that etmiss/ptll/ptheavy/ht/sorted cuts always apply)
+#*********************************************************************
+   False  = cut_decays    ! Cut decay products
```
menglu21, line 107: and line 107, use True = cut_decays

mlizzo:
Hi @valesark , I'm fine with the new cards, if no other questions will be raised within today I'm going to merge your PR

PR #3175, opened 2022-05-06
Adding cards for bStar to tW for CASE analysis

wpmccormack:
Hello again!  It's been a while since we put in a PR for the CASE analysis - just as a reminder, we're using a variety of dijet samples with different pronginess topologies.  There were some issues with the process that we had originally chosen as our 3+2 signal, so this is a new process.  It's basically b + g -> b* -> t W.  We scan the mass of the b* and of the W.  Decays of the W and t will be handled by pythia.  The run card here is basically copied from this triboson PR: https://github.com/cms-sw/genproductions/pull/2996, though I bumped up the number of quarks from 4 to 5, since we need bottom quarks in the pdf.  Tagging @OzAmram


bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/bstar_tW_Left_CASE/bstar_tW_MB_MW/bstar_tW_MB_MW_proc_card.dat
```diff
@@ -0,0 +1,18 @@
...
+define l- = e- mu- ta-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+add process p p > t w- QAD=2 QCD=0 QED=0
+add process p p > t~ w+ QAD=2 QED=0 QCD=0
+add process p p > t~ w+ QAD=2 QCD=0 QED=0
+add process p p > t w- QAD=2 QED=0 QCD=0
+add process p p > t~ w+ QAD=2 QED=0 QCD=0
+add process p p > t w- QAD=2 QED=0 QED=0
+add process p p > t~ w+ QAD=2 QED=0 QCD=0
```
menglu21, line 18: you need firstly "generate p p > t w- ...." and then add process. and there are duplicate process here, you should only have two processes, t w-, t~ w+

menglu21 (commented):
there are two customized cards and proc cards, could you please remove one of them

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/bstar_tW_Left_CASE/bstar_tW_MB_MW/bstar_tW_MB_MW_proc_card.dat
wpmccormack: Hi @menglu21 , yep that's a silly mistake.  Originally each of these lines did a different decay mode, but we're going to do the decays with pythia I believe.  I added in the generate line

wpmccormack:
> there are two customized cards and proc cards, could you please remove one of them

Hi @menglu21 , I removed those files.  Forgot that I had git added everything in a directory

menglu21:
Hi @wpmccormack i'm fine with those input cards except some concerns on your bstar_tW_gen_card.py, 
1. you are varying w mass, but i guess it won't change. because in the model, the w mass is an internal parameter which is calculated from cmath.sqrt(MZ**2/2. + cmath.sqrt(MZ**4/4. - (aEW*cmath.pi*MZ**2)/(Gf*cmath.sqrt(2)))). you can have a check on it.
2. when you varying the mass the Bp, you set the width of all mass point to be 0.03*MBp, do you have some pheno reference of such setting, otherwise i would suggest to set "DECAY 1005 Auto"

menglu21:
Hi @wpmccormack , you can find the model here https://cms-project-generators.web.cern.ch/cms-project-generators/Bprime_no_b_mass.tar.gz, and you can find the formula inside Bprime_no_b_mass/parameters.py. you can do a test, e.g., you change the w mass manual but you will find that MG warns you it will the default value for model consistence.

menglu21:
Hi @wpmccormack i'm fine with the current cards setting, as for the physical side, personally i think it should be ok since there is no SM contribution in your process so maybe we could play tricks on the those parameters. btw, could you please point me to some CASE analysis stuff (just out of curiosity) @Saptaparna @GurpreetSinghChahal @mlizzo please take a look at those cards.

mlizzo:
I'm fine with this request as well

menglu21:
Hi @wpmccormack thanks for the link. i've been thinking about the process again, since decay of top and W will be handled by pythia in which the coupling between W boson and quark/lepton are still SM values (and also the top decay), and those values are corresponding to the SM EW section, so i'm still concerned about such inconsistency. I'm wondering whether you know other models including e.g., w'/z', which make you can manipulate on them other on SM W/Z. sorry for so many comments...

wpmccormack:
Hi @menglu21 , no no thank you for your careful consideration here!  So I ran some of the bstar events with modified W mass through our pythia8 hadronizer, and I think everything is ok.  For example, let's consider this event:
<event>
 5      1 +1.1046300e-03 2.01252900e+03 7.81860800e-03 8.06303100e-02
       21 -1    0    0  501  502 +0.0000000000e+00 +0.0000000000e+00 +1.3311909965e+03 1.3311909965e+03 0.0000000000e+00 0.0000e+00 -1.0000e+00
        5 -1    0    0  502    0 -0.0000000000e+00 -0.0000000000e+00 -7.6064856595e+02 7.6064856595e+02 0.0000000000e+00 0.0000e+00 -1.0000e+00
     1005  2    1    2  501    0 +0.0000000000e+00 +0.0000000000e+00 +5.7054243059e+02 2.0918395625e+03 2.0125292768e+03 0.0000e+00 0.0000e+00
        6  1    3    3  501    0 +3.0627022635e+02 +3.6804705725e+02 +1.1981675553e+03 1.3020161552e+03 1.7430000000e+02 0.0000e+00 -1.0000e+00
      -24  1    3    3    0    0 -3.0627022635e+02 -3.6804705725e+02 -6.2762512467e+02 7.8982340730e+02 2.5448569573e+01 0.0000e+00 0.0000e+00
<mgrwt>

When I check the gen particles after hadronization, I see (excluding extraneous particles)

pt 0.0   eta 29256.0     pdgid 2212      mass 0.9382699728012085         status 4
pt 0.0   eta -29256.0    pdgid 2212      mass 0.9382699728012085         status 4
pt 0.0   eta 26381.38671875      pdgid 21        mass 0.0        status 21      mother pdgid  21
pt 0.0   eta -23039.01171875     pdgid 5         mass 0.0        status 21     mother pdgid 21
pt 0.0   eta 26098.373046875     pdgid 1005      mass 2025.861083984375          status 22   mother pdgid  21
pt 1007.25732421875      eta 1.4161577224731445          pdgid 6         mass 174.3000030517578          status 22  mother pdgid  1005
pt 986.365478515625      eta 1.1426903009414673          pdgid -24       mass 25.448570251464844         status 22  mother pdgid  1005
pt 574.315673828125      eta 1.3072952032089233          pdgid 24        mass 83.00288391113281          status 22   mother pdgid  6
pt 429.2762451171875     eta 1.5425357818603516          pdgid 5         mass 4.800000190734863          status 23   mother pdgid  6
pt 486.69281005859375    eta 1.1431909799575806          pdgid 15        mass 1.7768199443817139         status 2  mother pdgid  -24
pt 495.174072265625      eta 1.1389080286026     pdgid -16       mass 2.018548002524767e-05      status 1   mother pdgid  -24

You can see that the W from the top decay is on shell as normal, but the W from the LHE has the adjusted mass.  (As you can see, this had a leptonic W decay, which isn't exactly what we're after; I was lazily using a general hadronizer, not the only-jets one that we will actually employ.)

I kind of think of this as if we are feeding a very off-shell W boson into the hadronizer.  Effectively, this is just a vector particle with a mass of our choosing and the same couplings as a W boson.   I don't think pythia is aware of the parameters that we've put into Madgraph.

I did try to devise some alternate model, and I'm not sure if there's anything that would not involve writing our own Madgraph model, which I think is unnecessary for CASE's purposes.  It's tricky, because to get a 3+2 topology, you probably want your secondary resonances to be 1 fermion and 1 boson, which means that you need your main resonance (we're looking at s-channel processes) to be a fermion.  This is quite tricky to accomplish.  It's also possible to have boson as the main mediator with two fermions in the final state, but to get a 3+2 topology, you mostly have to resort to just enforcing hard gluon emissions, which isn't exactly what we're thinking of when we think of pronginess.

Please let me know if this addresses your concerns or if you would like to see additional studies.

menglu21:
Hi @wpmccormack , thanks for the check, i also did some check on the LHE output with changing the Gf and mz, the EW coupling strength is the same with sm one, so what you said should be correct, the EW coupling in LHE level and pythia level are both SM values (at least for the vertices you are interested in). I'm fine with those cards now, but i would like to involve more experts here in case they have comments. @Saptaparna @GurpreetSinghChahal @agrohsje @sihyunjeon @mlizzo , i'll merge this pr if there is no further comment in the next few days.

wpmccormack:
Hi Meng, great, thank you for all your help with this!

PR #3176, opened 2022-05-06
Added New Input Cards for WGammaToJJGamma and ZGammaToJJGamma

akobert:


menglu21:
Hi @akobert , the cards look fine to me, i don't have comment on them. just one concern, i'm not sure whether the w plus 2jet @NLO could be produced successfully or not, if not, you can move back to 1jet @NLO. @Saptaparna @GurpreetSinghChahal @mlizzo please have a look on those cards.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/ZGammaToJJGamma/ZGammaToJJGamma_proc_card.dat
```diff
@@ -0,0 +1,9 @@
+#set low_mem_multicore_nlo_generation True
+#import model loop_sm-ckm_no_b_mass
```
mlizzo, line 2: Please remove unused lines

bin/MadGraph5_aMCatNLO/cards/production/13TeV/WGammaToJJGamma/WGammaToJJGamma_proc_card.dat
```diff
@@ -0,0 +1,8 @@
+#set low_mem_multicore_nlo_generation True
+#import model loop_sm-ckm_no_b_mass
```
mlizzo, line 2: Please remove unused lines

mlizzo:
Cards look fine to me as well, waiting for @akobert 's feedback on the question raised by @menglu21 about NLO W+2jet process

akobert:
@menglu21 is correct about the NLO +2jet process. When using the NLO FXFX matching with 2 extra jets the gridpack production crashes, but it runs fine with 1. I have updated the files.

akobert:
I did want to ask you guys about raising the minimum photon pT to 50 GeV for ZGamma and WGamma. This is because my analysis has a requirement of a 110 GeV photon, so a minimum photon pT requirement at generator level would speed up production and make sure more events are useable for my analyisis?

mlizzo:
Hi @akobert , my only concern about raising the photon pT threshold to 50 GeV would be about the usage of this sample by other analyses that might need a looser selection. Do you have an estimate of how much the cross section would change? Maybe you can run a quick test with ptg=15 and ptg=50. @menglu21 what do you think about it?

DryRun:
Hi @mlizzo @menglu21 - thanks for the feedback! I had a couple questions:

- Keeping the photon pT cut at 15 GeV is fine if it enables another analysis to use the sample. Do you have one in mind, though? Maybe I'm not being imaginative enough, but it's hard to trigger on hadronic W/Z plus a <50 GeV photon.
- Regarding the photon isolation radius parameter: staying consistent with the leptonic sample is good, but I'm just curious where the value of R0_gamma=0.05 came from? It seems quite small.

menglu21:
Hi all, 
These two samples maybe useful for the EW study of WG/ZG (hadronic channel), and this page shows the pt_gamma distribution in EW WG analysis, https://cms.cern.ch/iCMS/jsp/openfile.jsp?tp=draft&files=AN2020_103_v10.pdf#page=32, so pta=50 will reduce the cross section significantly. As for the R0_gamma, it's used for the smooth isolation, the reference is https://arxiv.org/pdf/hep-ph/9801442.pdf (equa 3.3), small R0_gamma means loose constraint on the photon and other partons. I don't remember the default dR cut on dR(photon, other) for NLO events, if the R0_gamma is smaller than that value, it means the smooth isolation is turned off.
As for the photon pt selection, i think it can depend on the events you want, since your analysis requiring photon with such a high pt, i'm fine that we can add high LHE-level cut to make you have enough statistics.

menglu21:
> @menglu21 is correct about the NLO +2jet process. When using the NLO FXFX matching with 2 extra jets the gridpack production crashes, but it runs fine with 1. I have updated the files.

just out of curiosity, it's just stuck or raise error during the production

akobert:
> > @menglu21 is correct about the NLO +2jet process. When using the NLO FXFX matching with 2 extra jets the gridpack production crashes, but it runs fine with 1. I have updated the files.
> 
> just out of curiosity, it's just stuck or raise error during the production

While running on lxplus the job runs out of memory, sends an error, then crashes.

menglu21:
Hi @DryRun , sorry, i misunderstood the "trigger" thing, may i ask which photon trigger you plan to use, since i only know some diphoton triggers are not prescaled used by H->gg analysis (also in https://ram1123.github.io/display_TriggerPreScale/2017/, i didn't find a unprescaled single photon trigger). As for the R0Gamma, i only checked it would affect the overall cross section long time ago, but no check on the distribution, do you mind if you could perform some checks on the choice of R0Gamma. btw, maybe @sihyunjeon @agrohsje know more on this parameter.

mlizzo:
Hi, it seems there are some requests that have been produced with this setting in the past. For instance this a NLO ZG sample (the Z boson decays into leptons in this case though) where R0Gamma = 0.05 and ptgmin=9:
McM request: https://cms-pdmv.cern.ch/mcm/requests?prepid=EXO-RunIIFall18wmLHEGS-01720
gridpack: /cvmfs/cms.cern.ch/phys_generator/gridpacks/2017/13TeV/madgraph/V5_2.6.0/ZGToLLG_01J_5f_lowMLL_lowGPt_TuneCP5_13TeV-amcatnloFXFX-pythia8/ZG_NLO_201718_rwgt_slc6_amd64_gcc630_CMSSW_9_3_8_tarball.tar.xz
I honestly can't say how low one can go with this parameter, but apparently there is not a unique possible choice (although 0.05 seems pretty small to me as well)

DryRun:
- Regarding the photon trigger, the lowest unprescaled single photon triggers in Run 2 were something like `HLT_Photon(175-200) ` (all years) and `HLT_Photon(100-120)EB_TightID_TightIso` (2018 only; adds a barrel requirement in exchange for a lower threshold). 
- Regarding `R0gamma`, I think we should be fine with 0.05, staying consistent with the related datasets. The analysis isolation cuts will be strictly tighter.

menglu21:
Hi @akobert @DryRun thanks for the test, i think 50 is fine (maybe higher value is also acceptable due to the trigger)

akobert:
It is worth mentioning that in my test of ZGamma at ptgmin of 15 I got the following error from pythia

 PYTHIA Error in SimpleSpaceShower::pT2nearThreshold: stuck in loop

I did not get this error from WGamma in either ptgmin value or from the ZGamma at ptgmin of 50

menglu21:
Hi @akobert did you finally get the output, i remember Pythia will bypass the error and go to the next event.

akobert:
It finishes running but the reported cross section is 2e5 pb for ZGamma with the error. This is 3 orders of magnitude higher than the results of the other tests, and this error occurs in ZGamma with ptgmin of 15.

Saptaparna:
- Why do you set `set BW_cut 40`? 
- @menglu21 can you please remind me if the W+2jets issue is handled by this intervention (https://answers.launchpad.net/mg5amcnlo/+question/701195) from Olivier?

menglu21:
> It finishes running but the reported cross section is 2e5 pb for ZGamma with the error. This is 3 orders of magnitude higher than the results of the other tests, and this error occurs in ZGamma with ptgmin of 15.

could you please share me the gridpack and fragment, i can have a test too

menglu21:
> * Why do you set `set BW_cut 40`?
> 
>     * @menglu21 can you please remind me if the W+2jets issue (in general V+2jets) is handled by this intervention (https://answers.launchpad.net/mg5amcnlo/+question/701195) from Olivier? In this case 2 jets will indeed be hard to produce given the photon.
Hi @Saptaparna , yes, it's solved now by the fix from Olivier.

akobert:
Also I was wondering if anyone knows where I can find some information about the MC&I test code output.
1.  Is the madgraph cross section after matching?
2.  Can the cross sections for 0 and 1 extra jet be found separately?
3.  Is the pythia cross section before or after the filter (the 75 GeV filter you had me add in the pythia card)?
4.  Can we see the cross sections at each stage (before matching, after matching, before filter, and after filter)?

An example of the log files for the MC&I test code where I am searching for this information is in
/afs/cern.ch/user/a/akobert/exo-m-csample-requests-2-1/TestDir/logs/condor_WGammaToJJGamma_NLO_FXFX_TuneCP5_13TeV-amcatnloFXFX-pythia8.stdout

Saptaparna:
Hi @akobert you can find some instructions on running the genXsection analysis here: https://twiki.cern.ch/twiki/bin/view/CMS/GeneratorsHATSatLPC2021#Using_GenXSecAnalyzer

- Is the madgraph cross section after matching?
   - I would think that these cross sections are before matching
- Can the cross sections for 0 and 1 extra jet be found separately?
  - Yes, the genXsection analyzer will print the 0 and 1 jet cross sections separately 
- Is the pythia cross section before or after the filter (the 75 GeV filter you had me add in the pythia card)?
  - Currently I don't think the filter efficiency will be printed 
- Can we see the cross sections at each stage (before matching, after matching, before filter, and after filter)?
  - Yes, for before and after matching

menglu21:
Hi @akobert , i did a test using both 15 GeV and 50GeV gridpacks, the fragments i used are under /afs/cern.ch/user/m/melu/public/ZJJGamma, both of the tests are fine. the command i used for cmsDriver is 
`cmsDriver.py Configuration/GenProduction/python/ZGammaToJJGamma_TuneCP5_13TeV_pythia8_15GeV.py --python_filename bb.py --eventcontent RAWSIM,LHE --customise Configuration/DataProcessing/Utils.addMonitoring --datatier GEN,LHE --fileout file:HIG-RunIISummer20UL16wmLHEGEN-00779.root --conditions 106X_mcRun2_asymptotic_v13 --beamspot Realistic25ns13TeV2016Collision --customise_commands process.source.numberEventsInLuminosityBlock="cms.untracked.uint32(100)" --step LHE,GEN --geometry DB:Extended --era Run2_2016 --no_exec --mc -n 150`

i only add the "externalLHEProducer" part in the fragment basing on yours. btw, i would suggest you to add "isPromptFinalState()" in your genfilter.

as for your questions, besides what Sapta has answered, i have some follow-up, if you check the output of the cmsRun, you will find the following message, it contains all the answers to your question.

Overall cross-section summary 
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Process		xsec_before [pb]		passed	nposw	nnegw	tried	nposw	nnegw 	xsec_match [pb]			accepted [%]	 event_eff [%]
0		5.888e+01 +/- 1.132e-01		42	38	4	55	50	5	4.448e+01 +/- 4.124e+00		75.6 +/- 7.0	76.4 +/- 5.7
1		5.068e+01 +/- 1.376e-01		39	31	8	95	68	27	2.843e+01 +/- 6.198e+00		56.1 +/- 12.2	41.1 +/- 5.0
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
Total		1.096e+02 +/- 1.781e-01		81	69	12	150	118	32	7.261e+01 +/- 8.009e+00		66.3 +/- 7.3	54.0 +/- 4.1
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Before matching: total cross section = 1.096e+02 +- 1.781e-01 pb
After matching: total cross section = 7.261e+01 +- 8.009e+00 pb
Matching efficiency = 0.5 +/- 0.0   [TO BE USED IN MCM]
Filter efficiency (taking into account weights)= (11883.5) / (11883.5) = 1.000e+00 +- 0.000e+00
Filter efficiency (event-level)= (81) / (81) = 1.000e+00 +- 0.000e+00    [TO BE USED IN MCM]

After filter: final cross section = 7.261e+01 +- 8.009e+00 pb
After filter: final fraction of events with negative weights = 1.481e-01 +- 5.843e-03
After filter: final equivalent lumi for 1M events (1/fb) = 6.820e+00 +- 5.529e-01

akobert:
@Saptaparna Thanks, I will try the tool. Do you know if the reported cross section is xsec or xsec * Branching Ratio?

Where Branching ratio is BR for W or Z to qq

akobert:
In the genXsecAnalyzer what is the difference between the event efficiency and the acceptance %.

The event efficiency seems to match the matching efficiency, but the acceptance % seems to be used to calculate the final cross section after matching

menglu21:
> In the genXsecAnalyzer what is the difference between the event efficiency and the acceptance %?
> 
> The event efficiency seems to match the matching efficiency, but the acceptance % seems to be used to calculate the final cross section after matching

yes, you are correct.

akobert:
@menglu21 But what is acceptance percentage? What is used to calculate this value other than matching efficiency?

menglu21:
> @menglu21 But what is acceptance percentage? What is used to calculate this value other than matching efficiency?

when the process using FxFx matching, there will be negative weight events, so the matching efficiency will not equal to the acceptance. the acceptance take the weight into account, e.g., this line:
1 5.068e+01 +/- 1.376e-01 39 31 8 95 68 27 2.843e+01 +/- 6.198e+00 56.1 +/- 12.2 41.1 +/- 5.0
the acceptance 56.1% is calculated from (31-8)/(68-27).

akobert:
I compiled all the cross sections at before and after matching, filtering etc. in a spreadsheet. Do the numbers for the cross sections look okay?

https://docs.google.com/spreadsheets/d/13-Hz24elF7DZ9Da6bwYUWuSfJpOwpoJFOHKpnNTPMlk/edit?usp=sharing

menglu21:
Hi @akobert i think the results are fine.

PR #3178, opened 2022-05-12
MiNNLO + Powheg EW + Horace + HTCondor DAG

mseidel42:
* Scripts and cards for MiNNLO production
* Powheg job submission via DAG to get good grids for complicated processes
* Fetch Powheg from SVN instead of tarball (make optional?)
* Powheg EW
* Horace

Probably many conflicts but let's try to get this merged someday...


mseidel42:
It seems some commits (that I pulled from master long time ago) disappeared from there again

agrohsje:
Hi Markus. I deleted some of the files like run_pwg.py as they used the old batch system and just caused confusion but were otherwise identical to the proper ones (like run_pwg_condor.py for run_pwg.py). Is that what you mean?

mseidel42:
There were some spurious commits from other people, I'll make a clean PR

PR #3180, opened 2022-05-16
Madgraph cards for X to YH samples

ammitra:
Madgraph cards for X to YH samples, with Y decaying to WW and H to bb. Contained is also ` generate_grid.py` script for building the repositories for different `(mX, mY)` generator mass points from the templates.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_WWbb/Template/customizecards.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_WWbb/Template/customizecards.dat
@@ -1,0 +1,6 @@
+set param_card mass 45 TEMPLATEMH03
+set param_card mass 35 TEMPLATEMH02
+set param_card mass 25 125
+set param_card DECAY 45 1.00000e-03
+set param_card DECAY 35 1.00000e-03
+set param_card DECAY 25 4.00000e-03
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_WWbb/Template/proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_WWbb/Template/proc_card.dat
@@ -1,0 +1,22 @@
+set group_subprocesses Auto
+set ignore_six_quark_processes False
+set loop_optimized_output True
+set loop_color_flows False
+set gauge unitary
+set complex_mass_scheme False
+set max_npoint_for_channel 0
+import model sm
+define p = g u c d s u~ c~ d~ s~
+define j = g u c d s u~ c~ d~ s~
+define l+ = e+ mu+
+define l- = e- mu-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+set group_subprocesses Auto
+set ignore_six_quark_processes False
+set loop_optimized_output True
+set gauge unitary
+set complex_mass_scheme False
+import model NMSSMHET_UFO --modelname
+generate g g > h03, (h03 > h01 h02, h01 > b b~, (h02 > W+ W-, W+ > j j, W- > j j))
+output NMSSM_XYH_WWbb_MX_TEMPLATEMH03_MY_TEMPLATEMH02 -nojpeg
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_WWbb/Template/proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_WWbb/Template/proc_card.dat
@@ -1,22 +1,3 @@
-set group_subprocesses Auto
-set ignore_six_quark_processes False
-set loop_optimized_output True
-set loop_color_flows False
-set gauge unitary
-set complex_mass_scheme False
-set max_npoint_for_channel 0
-import model sm
-define p = g u c d s u~ c~ d~ s~
-define j = g u c d s u~ c~ d~ s~
-define l+ = e+ mu+
-define l- = e- mu-
-define vl = ve vm vt
-define vl~ = ve~ vm~ vt~
-set group_subprocesses Auto
-set ignore_six_quark_processes False
-set loop_optimized_output True
-set gauge unitary
-set complex_mass_scheme False
 import model NMSSMHET_UFO --modelname
 generate g g > h03, (h03 > h01 h02, h01 > b b~, (h02 > W+ W-, W+ > j j, W- > j j))
 output NMSSM_XYH_WWbb_MX_TEMPLATEMH03_MY_TEMPLATEMH02 -nojpeg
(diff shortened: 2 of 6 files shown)


bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_WWbb/Template/proc_card.dat
```diff
@@ -0,0 +1,22 @@
...
+define j = g u c d s u~ c~ d~ s~
+define l+ = e+ mu+
+define l- = e- mu-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+set group_subprocesses Auto
+set ignore_six_quark_processes False
+set loop_optimized_output True
+set gauge unitary
+set complex_mass_scheme False
```
mlizzo, line 19: Hi @ammitra , I think none of these lines is actually necessary and can be safely removed, as indeed it has been done in your reference commit https://github.com/cms-sw/genproductions/commit/39c661dde790c6ace22994a67d56aa7ad2ae586a

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_WWbb/Template/customizecards.dat
```diff
@@ -0,0 +1,6 @@
+set param_card mass 45 TEMPLATEMH03
+set param_card mass 35 TEMPLATEMH02
+set param_card mass 25 125
+set param_card DECAY 45 1.00000e-03
+set param_card DECAY 35 1.00000e-03
+set param_card DECAY 25 4.00000e-03
```
mlizzo, line 6: The higgs boson mass is already set to 125 GeV by default in the model, I don't think it's needed (and the width should be automatically computed). Why is the h02 and h03 decay width always set to 0.001? Shouldn't this depend by your mass hypothesis? Is the BR computation/xsec taken from some reference theory paper?

ammitra:
Thank you, @mlizzo - I have safely removed the unnecessary lines in `proc_card.dat`.

As for the Higgs and two scalars (X,Y), I set the widths based on the [NMSSM_XYH_bbb](https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_bbbb/Template/customizecards.dat) and [NMSSM_XYH_WWWW](https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_WWWW/Template/customizecards.dat) `customizecards.dat` templates, which keeps their widths constant across a variety of h02 and h03 masses

ammitra:
Hi @mlizzo Thank you for cross-checking and confirming. I do not have a theory paper to reference, but confirmed with a collaborator on the `X->YH->4b` analysis that the decay widths were the same across their samples as well.

sihyunjeon:
> Hi @mlizzo Thank you for cross-checking and confirming. I do not have a theory paper to reference, but confirmed with a collaborator on the `X->YH->4b` analysis that the decay widths were the same across their samples as well.

Hi @ammitra Let me interrupt a bit. This is not really about keeping consistency. The question is the cross sections go totally wrong once you force the width by hand like this in madgraph. How do you set the limits in the analysis? Or are you only interested in the kinematics?

ammitra:
Hi @sihyunjeon that's correct, we're only interested in the kinematics. Whatever results we obtain will be compared with the maximally-allowed xsecs calculated by theorists.

mlizzo:
Ok now I understand, then the request is fine to me, please @menglu21 @Saptaparna @GurpreetSinghChahal have a look at it

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_WWbb/Template/proc_card.dat
```diff
@@ -0,0 +1,3 @@
+import model NMSSMHET_UFO --modelname
+generate g g > h03, (h03 > h01 h02, h01 > b b~, (h02 > W+ W-, W+ > j j, W- > j j))
```
menglu21, line 2: h02 seems not in the model, it should be h2

mlizzo: Hi @menglu21 , I actually see it in the model, if you look at particles.py you'll find this block:
```
h02 = Particle(pdg_code = 35,
                       name = 'h02',
                       antiname = 'h02',
                       spin = 1,
                       color = 1,
                       mass = Param.MH02,
                       width = Param.WH02,
                       texname = 'h02',
                       antitexname = 'h02',
                       charge = 0,
                       GhostNumber = 0,
                       Y = 0)
```
Also, I tried to generate some events and the h02 particle (pdg_Id = 35) is correctly included in the LHE file

menglu21: interesting, because i tried locally and MG asked me to use h2. so please ignore my comment if it goes correctly in your sides.

mlizzo: @ammitra can you confirm that you're also able to produce a h02 resonance with your cards and model? If yes I guess we can merge the PR if there are no further comments

ammitra: Hi @mlizzo I can confirm that I'm able to produce the h02 resonance. I tested it out with a gridpack for the `(1300, 200)` mass point and it completed successfully, with a similar `1.99e+19 +- 1.286e+24 pb` xsec to yours.

acarvalh:
Hi, these area already covered in this repository, and we are trying to make all analyses consistent on simulation, [here for production](https://github.com/cms-sw/genproductions/pull/3123) and [here for decay](https://github.com/cms-sw/genproductions/tree/master/genfragments/ThirteenTeV/Higgs/HY). If you want FH decays of the WW it is just add the new decay fragment there (now it has only SL and DL). Also, we are trying to make that all the X>YH anamyses follow the same mass grid for a future combination. Please contact me.

mlizzo:
Hi @acarvalh , I'm not sure I got what you mean, are you suggesting to move the decay chain to the pythia fragment as you did for the other channels? So in that case there wouldn't be need to merge any additional madgraph cards, just another fragment in the folder you pointed. I think it makes, but please tell me if I understood correctly

acarvalh:
Hi @mlizzo , yes, that is what I suggest. 

I also suggest you in generation to follow the mass points that the other analyses are already following, in that case you can also re-use the already uploaded and tested gridpacks (that is why I suggested to contact me offline, as both those info's do not concern genproductions PR).

menglu21:
Hi all, is there a decision on how to proceed,  i think @acarvalh comment makes sense, what do you think about it @ammitra

acarvalh:
hi @ammitra, we already settled that offline. Just from the organizational point of view, I think this PR can be closed

PR #3185, opened 2022-05-27
ChargedHiggsToCB_UL cards

jshin96:
Uploaded cards for UL production of ChargedHiggsToCB process at LO level.


bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ChargedHiggsToCB/ChargedHiggsToCB_LO_UL/ChargedHiggsToCB_M120_run_card.dat
```diff
@@ -0,0 +1,282 @@
...
+ -1.0   = ebmax ! maximum E for the b
+ -1.0   = eamax ! maximum E for the photons
+ -1.0   = elmax ! maximum E for the charged leptons
+#*********************************************************************
+# Maximum and minimum absolute rapidity (for max, -1 means no cut)   *
+#*********************************************************************
+  5.0 = etaj    ! max rap for the jets 
+  -1.0  = etab    ! max rap for the b
+ 2.5  = etaa    ! max rap for the photons 
+ 2.5  = etal    ! max rap for the charged leptons
```
menglu21, line 146: please use etal = 3.0 since 2.5 is very close to the analyze level cut

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ChargedHiggsToCB/ChargedHiggsToCB_LO_UL/ChargedHiggsToCB_M120_run_card.dat
```diff
@@ -0,0 +1,282 @@
...
+ 2.5  = etaa    ! max rap for the photons 
+ 2.5  = etal    ! max rap for the charged leptons 
+ 0.0  = etajmin ! min rap for the jets
+ 0.0  = etabmin ! min rap for the b
+ 0.0  = etaamin ! min rap for the photons
+ 0.0  = etalmin ! main rap for the charged leptons
+#*********************************************************************
+# Minimum and maximum DeltaR distance                                *
+#*********************************************************************
+ 0.4 = drjj    ! min distance between jets
```
menglu21, line 154: have you check the typical dr between jets, especially the drjj from h+ decay, if the h+ is boosted, you may loose events with this drjj

menglu21:
Hi @jshin96 , since you are varying the mass of chargeHiggs, you should also changing the width correspondingly otherwise the cross section would be wrong, so please change "DECAY  37   auto" in the parameter card. btw, is this card used, [ChargedHiggsToCB/Card_Gen.sh], maybe you can remove it if not

menglu21:
Hi @jshin96 are you setting the limits on the couplings, and will you use the cross section reported by MG, if yes, i assume the constraint will be overestimated if the width is smaller than the value corresponding to the used couplings. if the cross section is not used in your analysis, i think it's fine.

jshin96:
Hello, Thank you for your comment @menglu21. For this analysis, we are assuming t -> H+ b -> c b b~ 100%, so the cross section is not used. The only cross section calculation needed for signal events is pp -> tt~.

menglu21:
> Hello, Thank you for your comment @menglu21. For this analysis, we are assuming t -> H+ b -> c b b~ 100%, so the cross section is not used. The only cross section calculation needed for signal events is pp -> tt~.

ok, thanks for the clarification, then those cards look fine to me. @Saptaparna @GurpreetSinghChahal   @mlizzo please have a check and feel free to merge.

PR #3186, opened 2022-05-27
Continuous integration

mseidel42:
### Description

* Runs on OpenStack VM (see README for setup)
  * Can have a chat how to set it up for cms-sw/genproductions
* Checks that Powheg hvq gridpack can be created, using CC7 docker image
* Checks pwg-stat file against a saved one
* Other processes / MadGraph can be added but should be light-weight and fast

### Screenshots
image

image

image

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -1,0 +1,52 @@
+# README
+
+## Get VM
+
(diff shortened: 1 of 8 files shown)


mseidel42:
Added test matrix for multiple CMSSW versions + retry on cvmfs fail

image

PR #3194, opened 2022-06-09
[mg29x] Fix madspin LO crash + reweighting f2py compile errors

SanghyunKo:
The PR consists of two fixes:
1. As of mg29x the LO gridpack generation does not generate any event when `True = gridpack`, this leads to a crash at the Madspin step. This can be fixed by generating a small number of events as before.
2. `f2py` compiler option had been changed to `f2py3` after migrating to the python3, this can fix a crash during the compile step of the reweighting on mg29x & mg33x.

Follow up of https://cms-talk.web.cern.ch/t/making-gridpacks-using-mgv3/9512


sihyunjeon:
thanks again

PR #3197, opened 2022-06-09
HT binned DY samples datacards for UL

cericeci:
Several items going on on this PR:

- For the samples with the mll > 50 GeV and 4 < mll < 50 GeV requests this includes the changes that were included in the corresponding requests in the UL samples. Given there were differences between what is in the gridpacks and what was in the repository I think it is best to keep it synchronized here.

- For the "extra low mass" samples (1 < mll < 4 GeV) these cards are new and expected to be used for a new low mass extension targeted by compressed lepton analysis.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/DYJets_HT_LO_MLM/DYJets_HT_mll50/DYJets_HT-100to200/DYJets_HT-100to200_run_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/DYJets_HT_LO_MLM/DYJets_HT_mll50/DYJets_HT-100to200/DYJets_HT-100to200_run_card.dat
@@ -69,7 +69,7 @@ $DEFAULT_PDF_MEMBERS = reweight_PDF     ! if pdlabel=lhapdf, this is the lhapdf 
  1        = ktscheme         ! for ickkw=1, 1 Durham kT, 2 Pythia pTE
  1        = alpsfact         ! scale factor for QCD emission vx
  F        = chcluster        ! cluster only according to channel diag
- T        = pdfwgt           ! for ickkw=1, perform pdf reweighting
+ F        = pdfwgt           ! for ickkw=1, perform pdf reweighting
  5        = asrwgtflavor     ! highest quark flavor for a_s reweight
  T        = clusinfo         ! include clustering tag in output
  3.0      = lhe_version       ! Change the way clustering information pass to shower.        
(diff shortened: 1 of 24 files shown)


cericeci:
Hi @sihyunjeon,
Thanks for the quick answer. I'm a bit surprised as these requests come from a group that did use this setup for publishing in preUL (they are now doing an extension in UL). I'll contact them to see if they noticed a significant mismodelling/they can apply additional cuts to restrict this phase space (they are looking at soft leptons down to mll >= 1 so I understand Z/gstar MC that represents this would be needed). 

In the meantime: what kind of handle did you end up using for the EXO requests you mentioned on the hypernews thread? My experience with very low mass virtual Z in powheg is that it tends to explode as well (i.e. https://cms-talk.web.cern.ch/t/incosistency-of-powheg-wz-samples/8345) so I'm not too sure of it working in this situation...

sihyunjeon:
> Hi @sihyunjeon, Thanks for the quick answer. I'm a bit surprised as these requests come from a group that did use this setup for publishing in preUL (they are now doing an extension in UL). I'll contact them to see if they noticed a significant mismodelling/they can apply additional cuts to restrict this phase space (they are looking at soft leptons down to mll >= 1 so I understand Z/gstar MC that represents this would be needed).
> 

Possibly for super large jet pt cuts, the behavior becomes similar? Can you check for the old preUL sample for lowmass DY, how auto_ptj_mjj was set in the run_card.dat?

> In the meantime: what kind of handle did you end up using for the EXO requests you mentioned on the hypernews thread? My experience with very low mass virtual Z in powheg is that it tends to explode as well (i.e. https://cms-talk.web.cern.ch/t/incosistency-of-powheg-wz-samples/8345) so I'm not too sure of it working in this situation...

Yes this was what I had in mind with powheg. The analyzers wanted it for data driven background validation purpose so detailed/precise modelling was not needed - ended up doing DY+1j only in the end without jet merging

cericeci:
Checking the old request ([SUS-RunIIFall18wmLHEGS-00119](https://cms-pdmv.cern.ch/mcm/requests?prepid=SUS-RunIIFall18wmLHEGS-00119)) the gridpack has auto_ptj_mjj set to false.

sihyunjeon:
ah... this is what i suspected. maybe jet pt above the weird cutoff value might be ok...? but i am not sure at all how this should be handled. Can you tell us the CADI number so that we can see the agreement between mc and data?

cericeci:
The published analysis (which used the preUL extralowmass DY samples) is SUS-18-004, the AN is 2019/264. I have gone through it and didn't see plots of HT/jet pt (although it seems to be used for event selection due to some of the triggers if I understood correctly). I pinged the analysis privately through mail, see if they have some control plots where the level of agreement is easier to see...

menglu21:
Hi @cericeci maybe we can close this PR since those DY process events will be handled centrally, we can create a script simialr with https://github.com/cms-sw/genproductions/tree/master/bin/GenValidation/updated_validation_cards/dy_mlm_mll50, to keep all the parameters consistent.

PR #3202, opened 2022-06-21
Update mg29x run_card parameters

SanghyunKo:
Updating LO `run_card.dat` reference parameters for mg29x (no change for NLO templates).
I'll merge it this week unless there is any comment.

FYI @menglu21 @Saptaparna @sihyunjeon @shimashimarin

### DIFF
--- a/bin/GenValidation/updated_validation_cards/dy_mlm_lowmass/inc/dyellell01234j_mll10to50_5f_LO_MLM_run_card.dat
+++ b/bin/GenValidation/updated_validation_cards/dy_mlm_lowmass/inc/dyellell01234j_mll10to50_5f_LO_MLM_run_card.dat
@@ -25,6 +25,21 @@
 #*********************************************************************
   True	= gridpack !True = setting up the grid pack
 #*********************************************************************
+# Phase-Space Optimization strategy (basic options)
+#*********************************************************************
+   0  = nhel          ! using helicities importance sampling or not.
+                             ! 0: sum over helicity, 1: importance sampling
+   1  = sde_strategy  ! default integration strategy (hep-ph/2021.xxxxx)
+                             ! 1 is old strategy (using amp square)
+                             ! 2 is new strategy (using only the denominator)
+# To see advanced option for Phase-Space optimization: type "update psoptim"
+#*********************************************************************
+# Generation bias, check the wiki page below for more information:   *
+#  'cp3.irmp.ucl.ac.be/projects/madgraph/wiki/LOEventGenerationBias' *
+#*********************************************************************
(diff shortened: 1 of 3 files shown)


SanghyunKo:
@sihyunjeon yeap, I've checked all warning messages that say the parameters weren't set, plus also checked the auto-generated `run_card.dat` by the Madgraph

menglu21:
Hi @SanghyunKo it seems some lines are duplicate, e.g., there are two "nhel" lines. i didn't check all those cards, maybe you can start from the default MG29X cards and modify the parameters w.r.t previous cards.

SanghyunKo:
@menglu21 I've rewritten the run cards based on the auto-generated ones from Madgraph. One interesting observation is that for the WJets MLM case, Madgraph tends to omit parameters like `drll` or `mmll`. Then how do we supposed to control the invariant mass of W or opening angle btn l & vl? @sihyunjeon

menglu21:
thanks @SanghyunKo , for W process, there should be no drll or mmll, right? btw, i think we don't apply neutrino cut in MG as far as i know.

sihyunjeon:
MadGraph got really clever (== a lot of hardcoding inside) and only shows the cuts that are relevant. `mmll` -> we don't have 2 leptons for W so MadGraph thinks "it's not worth showing this so let's hide it".

We usually did not control W mass, but if one has to, `mmnl` can be used for WJets process. `mmll` only applies to leptons I guess @SanghyunKo confused?

sihyunjeon:
This is good to go in for 'GEN purposes'.

On a side note though, do we have plans to provide the templates @menglu21 @Saptaparna @GurpreetSinghChahal, case-by-case for the normal users? My only fear is (not about this thread itself this is just good) people will come and see and think "these are the latest validation cards, I would copy paste these" which omits some hidden parameters and we will run into bunch of troubles. https://cms-talk.web.cern.ch/t/madgraph-amcatnlo-default-versions-for-run3-production/10973
@agrohsje and I think it would be useful to provide something for normal users (not for the run3 preparations that Jordan and we are doing), to show which run card need to be used case-by-case. (considering combinations of amcatnlo/madgraph 4f/5f with/without merging)

SanghyunKo:
@sihyunjeon ah right, I completely forgot `mmnl` since this was untouched...

About the showcase of run_cards, I think the best way would be to execute `output` command and check the run_card parameters directly, but this could be painful for complex processes. Maybe there is a way to make a script to retrieve auto-generated run_card only without generating every subprocess:

```python
import madgraph.various.banner as banner_mod
card = banner_mod.RunCardLO()
card.write('/tmp/edit_card/Cards/run_card.dat')
```

where I borrowed above from Magraph's test code - but I'm not sure how to propagate model & process command line info to the `banner.py`.

PR #3204, opened 2022-06-22
added back b mass from lukas that was removed by mistake

agrohsje:


agrohsje:
this is a fix to a fix from sebastien. further corrections might follow in preparation for run 3.

PR #3207, opened 2022-06-23
Add pwgmg5

danielwinterbottom:
Fixing bug that prevents condor jobs being submitted when splitting into multiple jobs.
Modifications related to the X0jj process:
- Updating some input parameters for the SM-like CP scenario 
- Adding input cards for two other CP scenarios
- Adding code to reweight the lhe files to other CP scenarios
- Removing most of the PDF variations to speed up the time taken to generate the events and excluding some additional files from the tarball to reduce the size of the gridpacks
- Adding instructions for producing the gridpacks


danielwinterbottom:
Hello, I opened this PR almost 3 weeks ago but there hasn't been any movement yet - is it possible for someone to check and merge this? We are hoping to make a MC request for this process very soon but would like someone to check the cards etc to make sure everything is in order before we make the request

sihyunjeon:
I don't merge things myself now but can you replace `forX0jj=$forX0jj` with same scripting convention as the minnlo one? It's actually not so secure to give individual processes with some arguments one by one for the python scripts (otherwise it would blow up at some point) so I am not sure how this should be dealt in the future

danielwinterbottom:
Hi, do you mean to switch the flag based on an option specified in the parameters card like what is done for the minlo option here: https://github.com/cms-sw/genproductions/blob/fff73b386f70d6582120478b2d4f22792bb0fbd3/bin/Powheg/Templates/runGetSource_template.sh#L36 ?
If so I can do a similar by checking the input card for this parameter which is unique for this process:
https://github.com/cms-sw/genproductions/blob/fff73b386f70d6582120478b2d4f22792bb0fbd3/bin/Powheg/examples/V2/X0jj_13TeV/X0jj_SM.input#L106

sihyunjeon:
Yes I think then at least if somebody tries to clean up the code at some point, it might be easier

danielwinterbottom:
Hello, would it be possible for someone to check and merge this PR? It was opened almost 6 weeks ago now and we really need this to be merged so that we can proceed with our MC requests

sihyunjeon:
@menglu21 @mlizzo @Saptaparna @GurpreetSinghChahal

Saptaparna:
Hi @danielwinterbottom apologies for the delay. The files look correct to me. @mseidel42 sorry to ping you but I think you are more familiar with this than me. If you have a moment, could you please take a quick look?

bin/Powheg/run_pwg_condor.py
```diff
@@ -77,9 +77,8 @@ def prepareCondorScript( tag, i, folderName, queue, SCALE = '0', njobs = 0, runI
    f.write('periodic_remove         = JobStatus == 5  \n')
    f.write('WhenToTransferOutput    = ON_EXIT_OR_EVICT \n')
    f.write('transfer_output_files   = "" \n')
-   if njobs > 0:
-       f.write('queue '+str(njobs)+'\n')
- 
+   #if njobs > 0:
+   #    f.write('queue '+str(njobs)+'\n')
```
mseidel42, line 81: Hi, I don't know about this physics process but the queue argument should not be removed from the condor config.

danielwinterbottom: Hi, the reason I removed these lines is that they appear to cause a bug with the jobs not getting submitted to the batch using HTCondor, which is related to this commit: https://github.com/cms-sw/genproductions/commit/03ef366e8c7837aa0a6e93f8d7dc710d4ef029f1#diff-306a635579daba4fb10220a64e87ff0b29416fe9cfa2acf8bb8abd5391f8b7cd. 

The bug seems to happen because the "-queue' option is given twice when njobs>0 - once in the condor config (due to the lines you highlighted above), and once when submitting the jobs e.g this line: https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/run_pwg_condor.py#L204

I think the bug should affect anyone trying to use the script with numJobs>0 but I wasn't sure how you wanted to deal with it in the end. So I just left these commented lines in the script. I can remove the comments if you like but then others will run into the same issue in the future. Perhaps there is another way you prefer to fix this - please let me know how you want me to proceed anyway

mseidel42: Hi Daniel, ok, could you remove the occurrence in L204 then, please?
```
        runCommand ('condor_submit ' + condorfile + ' -queue '+ str(njobs))
 ```

We need the submit files to be complete so that we can use HTCondor DAGs to chain the multiple stages together in
https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/run_pwg_parallel_condor.py

danielwinterbottom: Hi, sure I can do this but then I think I should also remove the "-queue" for every occurrence of condor_submit, even the ones that weren't directly causing my issues since they will have similar issues if someone runs with other options. Shall I do this?

danielwinterbottom: or rather only change the other occurrences where njobs is passed to prepareCondorScript, since the default value is 0 in which case the queue option doesn't get added

mseidel42: Probably it's easiest to just give the number of jobs to the condor script `prepareCondorScript` everytime? I.e. just pass `njobs=1` when there was a `-queue 1` afterwards.

danielwinterbottom: OK I have modified the script as you suggested and uncommented those 2 lines. Please let me know if I need to do anything further

mseidel42:
Looks ok to me, should work with the DAG now :+1: 
Your SM config has only moderate changes, so that should be fine. How is the uncertainty on the total cross section after combining all runs?
Let's ping the Powheg contacts @covarell and @agrohsje to comment if they ese any issue.

danielwinterbottom:
If you mean the number that gets printed in the "pwg-st3-*-stat.dat" files it is:
total (btilde+remnants) cross section in pb   40.218268305171087      +-   8.1645587676957898E-002
So about 0.2% relative uncertainty

mseidel42:
> If you mean the number that gets printed in the "pwg-st3-*-stat.dat" files it is: total (btilde+remnants) cross section in pb 40.218268305171087 +- 8.1645587676957898E-002 So about 0.2% relative uncertainty

Ok, nice! :+1:

Saptaparna:
Ok, am merging then! Thank you so much @mseidel42 🌟! Thanks @danielwinterbottom for the quick replies!

danielwinterbottom:
Thank you!

PR #3208, opened 2022-06-24
switched to running in one step due to a bug, updated release to 12_2

agrohsje:


mseidel42:
Ok, thanks for telling me! I freed up some space but we should move this to a GEN VM at some point

agrohsje:
Thanks Markus. Let me mention that I updated the Powheg version in a previous PR. So cool to have this CI checks!

sihyunjeon:
makes me work on madgraph CI 😖

PR #3210, opened 2022-07-01
XYH_YToHH production cards

srosenzweig09:


acarvalh:
Hi, could you remove the "6b" of the file names? Those are up to HHH, that could be confusing in the future

srosenzweig09:
Hi, Xanda,

I’ve removed the 6b from all of the file names (grid packs and card names) and I’ve also removed all of the mass points except for one. It looks like the PR should automatically be updated but please let me know if you don’t see the changes and I will make a new PR. Thanks!

Cheers,
Suzanne

acarvalh:
Hi! Yes it is updated automatically. But I still see a 6b in the parent folder

`bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XYH_YToHH_6b/UL_3H/NMSSM_XYH_YToHH_MX_1200_MY_500/NMSSM_XYH_YToHH_MX_1200_MY_500_proc_card.dat`

acarvalh:
Just as a comment to the main reviewers, those are the generic cards for the pp > X > YH > HHH searches, there are already two analysis teams waiting to it to be merged to start central production

menglu21:
Hi, sorry for the late reply, i have several questions below:
1. do you need the cross section of those samples, otherwise we need to be careful on the DECAY parameter
2. here you only create cards for one mass point, will you need others
3. the syntax you are using now should be fine, but i'm not sure what you mean by "make the decay chain in pythia decay fragment too long", you mean that the long fragment would make the production slower or they just look long

acarvalh:
Hi, thank you for the review!

1) As this is for a model independent resonance search, the cross section that should be put is 1pb. 

2) In the PP>X>HY version of cards that I mentioned only one mass was committed, but a script to generate all masses was also uploaded [here](https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XToYH/getAllMassPoints_YH.py). There are more than 100 mass points. Would a similar script suffice? To commit all replicates would be a waste of space, and I think git repos have a memory cap...

3) Just look long, timing was not tested. Also it was not directly tested if the cascade Y decay to HH from pythia8 would have the same kinematics as  Madgraph,, so we decided to do the gridpacks up to HHH

menglu21:
> Hi, thank you for the review!
> 
>     1. As this is for a model independent resonance search, the cross section that should be put is 1pb.
> 
>     2. In the PP>X>HY version of cards that I mentioned only one mass was committed, but a script to generate all masses was also uploaded [here](https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/NMSSM_XToYH/getAllMassPoints_YH.py). There are more than 100 mass points. Would a similar script suffice? To commit all replicates would be a waste of space, and I think git repos have a memory cap...
> 
>     3. Just look long, timing was not tested. Also it was not directly tested if the cascade Y decay to HH from pythia8 would have the same kinematics as  Madgraph,, so we decided to do the gridpacks up to HHH

Hi, 
1. ok, then those syntax and setup should be fine
2. yes, one mass point with a script is good
3. do you plan to generate all the mass point? i would like to recommend to check several mass points firstly, several in the extreme mass region and several in the middle range, they may be enough to draw a conclusion on the comparison.

menglu21:
Hi @srosenzweig09 it's not recommended to modify the gridpack_generation.sh for a specific need except for bug fix. I would like to suggest create script for other mass points and that's it.

srosenzweig09:
Hi, Meng, I fixed the gridpack_generation.sh file. I had commented out some lines so git status wasn't called each time. It was taking up a lot of space and going back to check was difficult. I'll go ahead and push the changes again.

srosenzweig09:
We are working on validation and everything looks good except for the gen weights. We've reached out to the MC conveners and are waiting for their advice. Thanks!

PR #3211, opened 2022-07-06
Add option to specify local UFO files in extramodels.dat

nsmith-:
Based on whether or not it starts with a `/` for an absolute path.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
+++ b/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
@@ -253,8 +253,14 @@ make_gridpack () {
         do
           #get needed BSM model
           if [[ $model = *[!\ ]* ]]; then
-            echo "Loading extra model $model"
-            wget --no-check-certificate https://cms-project-generators.web.cern.ch/cms-project-generators/$model	
+            if [[ $model == "/"* ]]; then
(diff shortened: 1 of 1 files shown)


Saptaparna:
Looks fine to me.

sihyunjeon:
@nsmith- Hi, would it not be better to use `-e` in shell script to check if the path exists and then copy the model file, instead of checking if it starts with "/"? for secure coding

nsmith-:
It's a good idea to check the path exists yes, I'll add that. But I think its nice to check for both because otherwise someone might be trying to indicate a local file but misspells the path name and it would try to download that path from the web area, leading to a more cryptic error.

nsmith-:
I think this is ready to merge

sihyunjeon:
@menglu21 @mlizzo

mlizzo:
PR is fine to me, @menglu21 if you agree I'm going to merge it

menglu21:
hi all, sorry for the late reply. personally i would prefer not allow local UFO models, since here you also comment that "the local model should be uploaded to cms-project-generators.web.cern.ch", we know that sometimes people will forget to upload it and it may get harder to trackback. we can always help to put the model to the central place. let me know your opinion.

nsmith-:
Ok I just learned one can also put the model in `addons/models` if I want a custom one without uploading... so doesn't that defeat the point of your last statement? Or am I missing something

PR #3215, opened 2022-07-11
Added MiNLO + powheg ggH + JJ input cards

vslokenb:


menglu21:
Hi, just a minor question on the higgs width, how the width is chosen

vslokenb:
> Hi, just a minor question on the higgs width, how the width is chosen

Hello, the Higgs width was set by the WW group that provided the initial gridpacks. It should now be changed to reflect the accepted Higgs width.

PR #3216, opened 2022-07-11
I would like to update one of my coupling constants so that we can stay within the NWA

farrah-simpson:
This is just a Minor change to a coupling constant.


farrah-simpson:
I would just like to follow up on this @menglu21 This is just a minor change to one of my coupling constants and I want to ensure my local cards match the ones in this repo. Thank you!

menglu21:
just a minor change, merged directly.

PR #3219, opened 2022-07-19
add cards

jennetd:
I have created cards for NLO WZTo4Q analogous to those for WWTo4Q_4f_TuneCP5_13TeV-amcatnloFXFX-pythia8 and ZZTo4Q_5f_TuneCP5_13TeV-amcatnloFXFX-pythia8. I have been able to generate a gridpack and generate 5000 test events, and all looks fine. Please let me know if these cards are ok for you. Thanks!

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/VV/WZTo4Q_4f_amcatnloFxFx/WZTo4Q_4f_amcatnloFxFx_madspin_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/VV/WZTo4Q_4f_amcatnloFxFx/WZTo4Q_4f_amcatnloFxFx_madspin_card.dat
@@ -1,0 +1,16 @@
+#directory for gridpack mode
+set ms_dir ./madspingrid
+
+#initialization parameters
+set Nevents_for_max_weigth 250 # number of events for the estimate of the max. weight
+set max_weight_ps_point 400  # number of PS to estimate the maximum for each event
+
+set max_running_process 1
+
+decay w+ > j j
+decay w- > j j
+decay z > j j
+#set onlyhelicity True
+#dummy decay
+
+launch
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/VV/WZTo4Q_4f_amcatnloFxFx/WZTo4Q_4f_amcatnloFxFx_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/VV/WZTo4Q_4f_amcatnloFxFx/WZTo4Q_4f_amcatnloFxFx_proc_card.dat
@@ -1,0 +1,13 @@
+import model loop_sm
+
+define ell+ = e+ mu+ ta+
+define ell- = e- mu- ta-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+
+generate p p > w+ z [QCD] @0
+add process p p > w+ z j [QCD] @1
+generate p p > w- z [QCD] @2
+add process p p > w- z j [QCD] @3
+
+output WZTo4Q_4f_amcatnloFxFx -nojpeg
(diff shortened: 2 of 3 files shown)


bin/MadGraph5_aMCatNLO/cards/production/13TeV/VV/WZTo4Q_4f_amcatnloFxFx/WZTo4Q_4f_amcatnloFxFx_proc_card.dat
```diff
@@ -0,0 +1,13 @@
+import model loop_sm
+
+define ell+ = e+ mu+ ta+
+define ell- = e- mu- ta-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+
+generate p p > w+ z [QCD] @0
+add process p p > w+ z j [QCD] @1
+generate p p > w- z [QCD] @2
```
menglu21, line 10: Hi, please change line 10 to "add process p p > w- z [QCD] @2"

jennetd:
@menglu21 Thanks, I have added the fix.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/VV/WZTo4Q_4f_amcatnloFxFx/WZTo4Q_4f_amcatnloFxFx_madspin_card.dat
```diff
@@ -0,0 +1,16 @@
...
+
+#initialization parameters
+set Nevents_for_max_weigth 250 # number of events for the estimate of the max. weight
+set max_weight_ps_point 400  # number of PS to estimate the maximum for each event
+
+set max_running_process 1
+
+decay w+ > j j
+decay w- > j j
+decay z > j j
```
menglu21, line 12: i didn't notice line12 last time, the 4f usually means the parton flavor, if you use z > j j, you will exclude z > b b~, please make sure this is what you want

jennetd: Thanks, this is actually not what I want - I want to include bb. I copied the madspin card from the existing WZToLNu2Q sample here: https://github.com/jennetd/genproductions/blob/wzto4q/bin/MadGraph5_aMCatNLO/cards/production/13TeV/VV/WZToLNu2Q_4f_amcatnloFxFx/WZToLNu2Q_4f_amcatnloFxFx_madspin_card.dat

Does this also not include Z(bb) ?

menglu21: i think so, in madspin card you can define a new type of jet, e.g., jb = j b b~, and then decay w+ > jb jb, ... z > jb jb

jennetd: Ok. In order to keep consistent with the WZToLNu2Q sample, I will keep this as 4f. I may add a dedicated z(bb) card later on for both of these processes. How does that sound?

mlizzo: Hi @jennetd , I think this is reasonable if we want to keep consistency with other VV 4f cards. Just for my understanding, @sihyunjeon do you know why Z->bb decays were never included in such samples so far? 
https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/VV

sihyunjeon: @mlizzo no reason honestly. nobody objected - this was the reason.
I am not sure what is the best way but it would be better to make 4F "Z > b b" decay mode separately at least to make a complete set of MC samples

menglu21:
ok, i will merge this PR if no other comments. @jennetd please add Z>bb cards later.

PR #3222, opened 2022-07-26
added WplusLNu_Wminus2q JHUGEN cards

fmanteca:
WplusLNu_Wminus2q JHUGEN cards for UL HWW high mass samples request:

- Version with the 2l filter will contain events with 3 leptons

- Version without the l2 filter will mostly contain 1-lepton events


menglu21:
the cards look fine to me, just one question on the name, won't it be WmnusLNu_Wplus2q also?

fmanteca:
Hi Meng,

You are right, RandomizeVVdecays is set to 1 by default.

I have just updated the name of the files. I hope it is clearer now.

menglu21:
Hi, thanks for the clarification, could you please further change the name to "WWTolnu2q", i think it's more suitable and clear.

menglu21:
thanks, @Saptaparna @GurpreetSinghChahal @mlizzo please have a check.

PR #3234, opened 2022-08-18
Gen fragment for J/Psi -> e+e- at COM=13.6TeV for Run3 in 2022

arsahasransu:
Gen fragment for the process J/Psi -> e+ e- at sqrt(s) = 13 TeV. Based on the suggestion by Giulia Sorrentino.

### DIFF
--- a/genfragments/ThirteenPointSixTeV/JpsiEE_Pt_2_30_13p6TeV_TuneCP5_pythia8_cfi.py
+++ b/genfragments/ThirteenPointSixTeV/JpsiEE_Pt_2_30_13p6TeV_TuneCP5_pythia8_cfi.py
@@ -1,0 +1,53 @@
+import FWCore.ParameterSet.Config as cms
+from Configuration.Generator.Pythia8CommonSettings_cfi import *
+from Configuration.Generator.MCTunes2017.PythiaCP5Settings_cfi import *
+
+generator = cms.EDFilter("Pythia8ConcurrentGeneratorFilter",
+                         pythiaPylistVerbosity = cms.untracked.int32(0),
+                         #filterEfficiency = cms.untracked.double(0.003),
+                         pythiaHepMCVerbosity = cms.untracked.bool(False),
+                         crossSection = cms.untracked.double(1742000.0),
+                         comEnergy = cms.double(13600.0),
+                         maxEventsToPrint = cms.untracked.int32(10),
+                         PythiaParameters = cms.PSet(
+        pythia8CommonSettingsBlock,
+        pythia8CP5SettingsBlock,
+        processParameters = cms.vstring(
+            'Charmonium:all = on',                       # Quarkonia, MSEL=61, including feed-down as well
+            '443:onMode = off',                          # Turn off J/psi decays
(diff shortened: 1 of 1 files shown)


genfragments/ThirteenPointSixTeV/JpsiEE_Pt_2_30_13p6TeV_TuneCP5_pythia8_cfi.py
```diff
@@ -0,0 +1,53 @@
+import FWCore.ParameterSet.Config as cms
+from Configuration.Generator.Pythia8CommonSettings_cfi import *
+from Configuration.Generator.MCTunes2017.PythiaCP5Settings_cfi import *
+
+generator = cms.EDFilter("Pythia8GeneratorFilter",
```
menglu21, line 5: please use concurrent mode, "Pythia8ConcurrentGeneratorFilter"

genfragments/ThirteenPointSixTeV/JpsiEE_Pt_2_30_13p6TeV_TuneCP5_pythia8_cfi.py
```diff
@@ -0,0 +1,53 @@
...
+                         pythiaPylistVerbosity = cms.untracked.int32(0),
+                         #filterEfficiency = cms.untracked.double(0.003),
+                         pythiaHepMCVerbosity = cms.untracked.bool(False),
+                         crossSection = cms.untracked.double(1742000.0),
```
Saptaparna, line 9: Is this cross section correct and updated to 13.6 TeV?

genfragments/ThirteenPointSixTeV/JpsiEE_Pt_2_30_13p6TeV_TuneCP5_pythia8_cfi.py
arsahasransu: I have made the changes and squashed the commit. Can you please tell me if this will create a big difference in simulation because I was already running a private MC production? I am still at the GEN stage. I can redo the GEN step now instead of realizing that there is a significant change later.

genfragments/ThirteenPointSixTeV/JpsiEE_Pt_2_30_13p6TeV_TuneCP5_pythia8_cfi.py
arsahasransu: The cross-section is not updated to 13.6 TeV. There are two reasons for this. 

Firstly, I tried to look for cross-section values online and didn't find any reasonable theoretical estimate for this. In the following paper https://arxiv.org/pdf/1006.2750.pdf, the estimate is O(100 nb). Within theoretical uncertainty, there isn't an appreciable change in the production cross-section from 13 TeV to 13.6 TeV.

Secondly, the event weight can be normalized later on. 

If there is a better/more elegant/efficient solution to this, please suggest me, and I will change the value accordingly.

arsahasransu:
@Saptaparna @menglu21 Any updates in the following gen fragment, please?

PR #3237, opened 2022-08-29
New h2hw signal samples

kchrisucy:
Dear all,

My analysis focuses on the search for a charged Higgs boson through its decay to a neutral Higgs(h/H) boson and a W, where subsequently, h/H->ττ.  
The samples that are currently available are for a few charged Higgs boson masses and just 2 neutral Higgs boson masses.
My pull request is related to the need of scanning both masses in a wider range and also explore an additional final state for the neutral Higgs (h/H -> bb).

Best regards,
Konstantinos Christoforou


attikis:
@Saptaparna @menglu21  can you please review this pending request?

Thank you,
Alexandros & Ho-Fung for HIG MC

Saptaparna:
Hi, can you please remake this PR to include only one set of cards and some script to indicate what you want to vary. That will make our lives a lot easier.

kchrisucy:
Hi and thanks for you reply.  I don't think it's easy to include only one set of cards, because for each mH0, the parameters of the model change (i.e. sibma and tanb) and I also have different set of cards for different decay channels for H0 ( ττ or bb), as was done in the past

menglu21:
Hi @kchrisucy i have a check on those cards and they look fine to me (simple check and comparison with existing charge Higgs cards in the same repository)

kchrisucy:
hi @menglu21 and thanks for your reply and efforts.  Sure, I'll make the PR right away

PR #3250, opened 2022-09-07
Wz semilep 5f

anmehta:
Adding cards for review targeting the WZ production at NLO with 5flavor scheme using leptonic (hadronic) decays for the W(Z) bosons


menglu21:
Hi @anmehta i just noticed we already have the cards for semi-WZ, https://github.com/cms-sw/genproductions/tree/mg265UL/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/WZTo1L1Nu2Q01j_5f_NLO_FXFX, they look the same with your PR

anmehta:
Hi @menglu21, indeed the proc & run cards are essentially the same. However, I'm not sure about setting BW_cutoff to 40 in the madspin card?  Should not we use 15?  If it overwrites the value set in run card then we need to change it, right?

PR #3251, opened 2022-09-08
ggH4l decay with Anomalous Photon Couplings

jdavis36:
Cards for ggH production with powheg for analysis targeting Higgs photon couplings.

### DIFF
--- a/bin/Powheg/production/2017/13TeV/Higgs/gg_H_quark-mass-effects_JHUGen_AnomalousCouplings_NNPDF31_13TeV/gg_H_quark-mass-effects_NNPDF31_13TeV.input
+++ b/bin/Powheg/production/2017/13TeV/Higgs/gg_H_quark-mass-effects_JHUGen_AnomalousCouplings_NNPDF31_13TeV/gg_H_quark-mass-effects_NNPDF31_13TeV.input
@@ -6,8 +6,8 @@ ih2   1           ! hadron 2 (1 for protons, -1 for antiprotons)
 ebeam1 6500     ! energy of beam 1
 ebeam2 6500    ! energy of beam 2
 
-lhans1 306000       ! pdf set for hadron 1 (LHA numbering)
-lhans2 306000       ! pdf set for hadron 2 (LHA numbering)
+lhans1 325300       ! pdf set for hadron 1 (LHA numbering)
+lhans2 325300       ! pdf set for hadron 2 (LHA numbering)
 
 ! Parameters to allow or not the use of stored data
 use-old-grid    1 ! if 1 use old grid if file pwggrids.dat is present (<> 1 regenerate)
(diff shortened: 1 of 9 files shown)


bin/Powheg/production/2017/13TeV/Higgs/gg_H_quark-mass-effects_JHUGen_AnomalousCouplings_NNPDF31_13TeV/gg_H_quark-mass-effects_NNPDF31_13TeV.input
```diff
@@ -0,0 +1,71 @@
+numevts NEVENTS
+ih1   1           ! hadron 1 (1 for protons, -1 for antiprotons)
+ih2   1           ! hadron 2 (1 for protons, -1 for antiprotons)
+#ndns1 131         ! pdf set for hadron 1 (mlm numbering)
+#ndns2 131         ! pdf set for hadron 2 (mlm numbering)
+ebeam1 6500     ! energy of beam 1
+ebeam2 6500    ! energy of beam 2
+
+lhans1 306000       ! pdf set for hadron 1 (LHA numbering)
+lhans2 306000       ! pdf set for hadron 2 (LHA numbering)
```
menglu21, line 10: could you please change the pdf to "325300"

menglu21 (commented):
a general comment is that the current master branch is MG299, maybe you can make a PR to mg265UL (run2 UL branch) also

bin/Powheg/production/2017/13TeV/Higgs/gg_H_quark-mass-effects_JHUGen_AnomalousCouplings_NNPDF31_13TeV/gg_H_quark-mass-effects_NNPDF31_13TeV.input
jdavis36: Ok I will change the pdf set here and then in the meantime i will make another PR to the branch you specified

menglu21:
look fine to me, @Saptaparna @mlizzo please have a check

mlizzo:
Hi @jdavis36 , there is actually a ggH powheg card in genproductions already, please check it out:
https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/production/2017/13TeV/Higgs/gg_H_quark-mass-effects_NNPDF31_13TeV/gg_H_quark-mass-effects_NNPDF31_13TeV_M125.input
the only relevant difference I noticed is "ew = 0" in your commit from "ew 1" in the one I've pointed to you, which includes EW corrections. Also, it seems to me that your card was directly extracted from a tarball, since those bottom lines are usually added when Powheg is run. Would it sound reasonable to use the already available ggH card? Unless there are specific reasons to have something different of course

jdavis36:
Hi @mlizzo the reason I am not implementing EW corrections is because these samples will be combined for use with samples such as these:

https://cms-pdmv.cern.ch/mcm/edit?db_name=requests&prepid=HIG-RunIIFall18wmLHEGS-00311
https://cms-pdmv.cern.ch/mcm/requests?prepid=HIG-RunIISummer20UL18wmLHEGEN-00043

Both of these samples use EW = 0 in the powheg cards which is why I wanted to generate something analogous to these. I think it may be improper from an analysis standpoint to have EW corrections in one sample and have them turned off in another, since we do not have plans to redo all the other samples with EW corrections. I'm also sure that at some point these powheg cards were added here as well, but I just cannot find them. Could we just use the already existing cards for EW = 0 if they exist? If they do not exist, I would still like to submit cards with EW = 0.

mlizzo:
Hi @jdavis36 , thanks for the clarification, I think it's fine then. If there are no other comments, I'll merge it soon.

PR #3252, opened 2022-09-09
Update runConfigs.py to include 5 TeV Zprime

jotadram6:
Doing it to this branch, as requested.


menglu21:
Hi, it seems you have made a huge number of modification, please switch to mg265UL do the modification and push then

PR #3263, opened 2022-09-20
Add cards for Higgs+2jet CP gridpack production for Run 3

danielwinterbottom:
- This is adding the equivalent cards used for the Run2 production for Run3. The energy is changed to 13.6 TeV but all other options in the cards are kept the same. PR for Run2 production is this one: https://github.com/cms-sw/genproductions/pull/3207
- After producing the Run3 gridpacks, there was an issue producing events with runcmsgrid.sh. Only 1 event is produced even when asking for more than 1 and then it gives the error "pwhg_main error: Did not produce expected number of events". I traced this issue back to the modification made to runGetSource_template.sh in this commit: https://github.com/cms-sw/genproductions/commit/0f6bf5947251db98672c6e4ccef6a12ad0b58331#diff-083c2ecbee2516fb3aaf4cb519eea4bbeb462666352937c4562fb5f9ed11d662. After I added back the sed commands it seems to work. I have therefore reverted this change in the PR, but as a non-expert I don't know if there is some other reason to remove these lines, and perhaps a different fix is needed..?

### DIFF
--- a/bin/Powheg/Templates/createTarBall_template.sh
+++ b/bin/Powheg/Templates/createTarBall_template.sh
@@ -131,7 +131,7 @@ if [ $$keepTop == '1' ]; then
     echo 'Packing...' $${WORKDIR}'/'$${process}'_'$${SCRAM_ARCH}'_'$${CMSSW_VERSION}'_'$${folderName}'.tgz'
     tar zcf $${WORKDIR}'/'$${process}'_'$${SCRAM_ARCH}'_'$${CMSSW_VERSION}'_'$${folderName}'.tgz' * --exclude=POWHEG-BOX --exclude=powhegbox*.tar.gz --exclude=*.lhe --exclude=run_*.sh --exclude=*temp --exclude=pwgbtlupb-*.dat --exclude=pwgrmupb-*.dat --exclude=run_*.out --exclude=run_*.err --exclude=run_*.log --exclude=minlo-run --exclude=dynnlo* $$exclude_extra
 else
-  if [ $$process == 'WWJ']; then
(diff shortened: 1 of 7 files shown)


menglu21:
Hi @danielwinterbottom i'm fine with those cards, maybe @mlizzo can comment on the modification

mlizzo:
Hi @danielwinterbottom , I removed those lines because they seemed to be problematic for the compilation of the WW MiNNLO Powheg sample, which I tried to implement in genproductions some time ago. I honestly thought they were harmless for other processes, since they should be related to Powheg analysis internal tools, which I don't think they're actually used. If those (missing) lines are causing issues than it's fine to me to add them back, I must say I'm a bit surprised but I don't have a full understanding of what Powheg runs behind the scenes. Out of curiosity, did this happen for other samples as well?

danielwinterbottom:
Hi @mlizzo, I am also unsure about exactly what these lines do, I was only able to identify this as the problematic change by iteratively going back through the commit history to find the point where the gridpack production stops working. I didn't try it for other samples before now but I just tried it quickly for the VBF_H process and this seems to not be affected by the removal of the lines, so I'm not sure why it affects specifically the X0jj process or if other processes will have similar problems or not. 
If these lines are problematic for the WW MiNLO process how do you suggest modifying the code? E.g should I add some option so the lines only get removed for WW?

mlizzo:
Hi @danielwinterbottom , thanks for the quick check! I must say that's weird, but my knowledge is quite limited.  By the way, yes, I think your suggestion is reasonable, could you kindly implement that? I'm sorry I've caused some trouble with that commit, but this was really unexpected to me.

PR #3268, opened 2022-09-30
filter efficiency from fragment and from generator field consistency …

efeyazgan:
filter efficiency from fragment and from generator field consistency check dropped per request by @sihyunjeon , @menglu21 and Sitian Qian.


efeyazgan:
Also see the follow up issue: https://github.com/cms-sw/genproductions/issues/3269

PR #3270, opened 2022-09-30
Adding LFV etau and mutau signals 

Jingyan95:
PR https://github.com/cms-sw/genproductions/pull/3006 added e+mu LFV signal
This pull request adds e+tau and mu+tau signal as well as an inclusive signal that covers all three combinations
Other than the lepton flavor composition, we didn't change any of the set-up


menglu21:
cards look fine to me, but i have a question on the run card, you are using "F  = cut_decays", it's what you want?

Jingyan95:
> cards look fine to me, but i have a question on the run card, you are using "F = cut_decays", it's what you want?

@menglu21 , thank you for reviewing our PR,
Yes, this is option is consistent with we did for TOP-19-006/TOP-22-005/TOP-22-011, and we intend to keep it.

PR #3273, opened 2022-10-04
Updated JHUGen version in gridpack scripts to 7.5.2

lucassmguerra:


gritsan (commented):
I just wanted to add that we would like to move to v7.5.2 of JHUGen for all Run3 simulation, so this PR is needed to make it happen.

PR #3281, opened 2022-10-12
add option for HZJ_ew to decay to all leptons

jieunyoo:
Would like to add this option so that we can get the decay to all leptons, instead of just one. Thanks!


menglu21:
looks fine to me

agrohsje:
Hi @jieunyoo. You still need to apply the patch in the code once you checked out your process directory. Did you check in the code that this is the only relevant part where inclusive option is commented out by the authors. I think there are more spots in the code, no?

jieunyoo:
Hi. No, I didn't check out other spots. I will do that. Could you please let me know what you mean by "apply the patch in the code." After uncommenting everything related to the inclusive option, do a "make" perhaps? Thank you!

agrohsje:
I would suggest to make this a dedicated 9th patch that is only applied for the process you are doing: 
https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/run_pwg_condor.py#L295
https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/Utilities/helpers.py
Note also that the pathes in the patch will not work. So you need to adjust to the proper pathes. Check the other patches in the above example.

jieunyoo:
Hello, 

- I uncommented all of the mentions of this Vdecay mode == 11 in boosted/POWHEG-BOX/HZJ_ew. The new version of my patch reflects that.
- I defined patch9 in helpers.py
- I added patch9 in the condor script

I am not sure if patch9 is correctly defined in helpers, e.g. whether or not I need to again compile Powheg. Thank you!

```
$ grep -nr Vdecmod.eq.11 init_couplings.f
81:          elseif (Vdecmod.eq.11) then
146:      elseif (Vdecmod.eq.11) then

$grep -nr vdecaymode.eq.11
Born.f:64:c$$$     $     .or. vdecaymode.eq.11 .or. vdecaymode.eq.12) then
Born.f:220:c$$$     $     .or. vdecaymode.eq.11 .or. vdecaymode.eq.12) then
real.f:67:               .or. vdecaymode.eq.11 .or. vdecaymode.eq.12) then
virtual.f:186:c$$$     $     .or. vdecaymode.eq.11 .or. vdecaymode.eq.12) then
```
@cmantill

jieunyoo:
I did see the comment about the paths. I added cd POWHEG-BOX/HZJ_ew in helpers.py I am not sure if there are other paths that need to be fixed.

PR #3284, opened 2022-10-17
Add higgs to tau tau gamma cards

brallmond:
Initial PR for higgs to tau tau gamma cards, requires new model (emailed to conveners) 

So far I've only included cards for the main signal process p p > H > ta+ ta- a [QCD]
I've also adjusted the run card to use beam energies of 6800 GeV instead of the default 6500
I've set the default number of events to 100 in the run card since I anticipate needing to test and update the cards after their initial integration


bin/MadGraph5_aMCatNLO/cards/pp_H_TTG_012j_5f_NLO_MidDM/pp_H_TTG_012j_5f_NLO_MidDM_param_card.dat
```diff
@@ -0,0 +1,90 @@
...
+DECAY  4 0.000000e+00 # c : 0.0 
+DECAY  5 0.000000e+00 # b : 0.0 
+DECAY  11 0.000000e+00 # e- : 0.0 
+DECAY  12 0.000000e+00 # ve : 0.0 
+DECAY  13 0.000000e+00 # mu- : 0.0 
+DECAY  14 0.000000e+00 # vm : 0.0 
+DECAY  15 0.000000e+00 # ta- : 0.0 
+DECAY  16 0.000000e+00 # vt : 0.0 
+DECAY  21 0.000000e+00 # g : 0.0 
+DECAY  22 0.000000e+00 # a : 0.0
```
menglu21, line 90: instead of using parameter card, please use customized card, in which only the modified values are included comparing with the default parameter card in the model.

bin/MadGraph5_aMCatNLO/cards/pp_H_TTG_012j_5f_NLO_MidDM/pp_H_TTG_012j_5f_NLO_MidDM_proc_card.dat
```diff
@@ -0,0 +1,29 @@
...
+#set max_t_for_channel 99
+#set zerowidth_tchannel True
+#import model sm
+#define p = g u c d s u~ c~ d~ s~
+#define j = g u c d s u~ c~ d~ s~
+#define l+ = e+ mu+
+#define l- = e- mu-
+#define vl = ve vm vt
+#define vl~ = ve~ vm~ vt~
+#set auto_convert_model T
```
menglu21, line 19: please remove those commented lines

bin/MadGraph5_aMCatNLO/cards/pp_H_TTG_012j_5f_NLO_MidDM/pp_H_TTG_012j_5f_NLO_MidDM_proc_card.dat
```diff
@@ -0,0 +1,29 @@
...
+
+define p = p b b~
+define j = j b b~
+
+import model model DipoleMomentMid_UFO
```
menglu21, line 24: line 24 should be put before line 21 and 22

bin/MadGraph5_aMCatNLO/cards/pp_H_TTG_012j_5f_NLO_MidDM/pp_H_TTG_012j_5f_NLO_MidDM_run_card.dat
```diff
@@ -0,0 +1,166 @@
...
+#                                                                    
+#*********************************************************************
+# Tag name for the run (one word)                                    *
+#*********************************************************************
+  tag_1     = run_tag ! name of the run 
+#*********************************************************************
+# Number of events and rnd seed                                      *
+# Warning: Do not generate more than 1M events in a single run       *
+#*********************************************************************
+  100 = nevents ! Number of unweighted events requested
```
menglu21, line 26: change "100 = nevents" to "5000 = nevents"

bin/MadGraph5_aMCatNLO/cards/pp_H_TTG_012j_5f_NLO_MidDM/pp_H_TTG_012j_5f_NLO_MidDM_run_card.dat
```diff
@@ -0,0 +1,166 @@
...
+     1        = lpp2    ! beam 2 type
+     6800.0     = ebeam1  ! beam 1 total energy in GeV
+     6800.0     = ebeam2  ! beam 2 total energy in GeV
+# To see polarised beam options: type "update beam_pol"
+
+#*********************************************************************
+# PDF CHOICE: this automatically fixes also alpha_s and its evol.    *
+#*********************************************************************
+     nn23lo1    = pdlabel     ! PDF set                                     
+     230000    = lhaid     ! if pdlabel=lhapdf, this is the lhapdf number
```
menglu21, line 43: use following three lines instead for pdf,
lhapdf    = pdlabel   ! PDF set
$DEFAULT_PDF_MEMBERS = reweight_PDF     ! reweight to get PDF uncertainty

bin/MadGraph5_aMCatNLO/cards/pp_H_TTG_012j_5f_NLO_MidDM/pp_H_TTG_012j_5f_NLO_MidDM_run_card.dat
```diff
@@ -0,0 +1,166 @@
...
+#
+#*********************************************************************
+# Store info for systematics studies                                 *
+# WARNING: Do not use for interference type of computation           *
+#*********************************************************************
+   True  = use_syst      ! Enable systematics studies
+#
+systematics = systematics_program ! none, systematics [python], SysCalc [depreceted, C++]
+['--mur=0.5,1,2', '--muf=0.5,1,2', '--pdf=errorset'] = systematics_arguments ! see: https://cp3.irmp.ucl.ac.be/projects/madgraph/wiki/Systematics#Systematicspythonmodule
+# Syscalc is deprecated but to see the associate options type'update syscalc'
```
menglu21, line 166: the run card seems to be not correct in many cases, please refer to https://github.com/cms-sw/genproductions/blob/master/bin/GenValidation/updated_validation_cards/dy_fxfx_mll50/dy012jfxfx_inc/dyellell012j_5f_NLO_FXFX_run_card.dat and change accordingly

menglu21 (commented):
besides all the comments above, please push this PR to https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/13p6TeV

bin/MadGraph5_aMCatNLO/cards/pp_H_TTG_012j_5f_NLO_MidDM/pp_H_TTG_012j_5f_NLO_MidDM_proc_card.dat
brallmond: I found out from a post on launchpad that I needed to be using madspin and have since included the above model there. However, if I set maxjetflavor = 5 in the run card, do I need to set p = p b b~ and j = j b b~ in the proc card as well?

bin/MadGraph5_aMCatNLO/cards/pp_H_TTG_012j_5f_NLO_MidDM/pp_H_TTG_012j_5f_NLO_MidDM_run_card.dat
menglu21: there are many customised setups, e.g., the pdfs, so it's usually recommended to follow the genrepo to set up the card, especially the run card.

bin/MadGraph5_aMCatNLO/cards/pp_H_TTG_012j_5f_NLO_MidDM/pp_H_TTG_012j_5f_NLO_MidDM_proc_card.dat
menglu21: yes, you need to import the model first (btw, it should "import model DipoleMomentMid_UFO"), and then define the flavor

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/pp_H_TTG_012j_5f_NLO_MidDM/pp_H_TTG_012j_5f_NLO_MidDM_run_card.dat
```diff
@@ -0,0 +1,172 @@
...
+#
+#*********************************************************************
+# Store info for systematics studies                                 *
+# WARNING: Do not use for interference type of computation           *
+#*********************************************************************
+   True  = use_syst      ! Enable systematics studies
+#
+systematics = systematics_program ! none, systematics [python], SysCalc [depreceted, C++]
+['--mur=0.5,1,2', '--muf=0.5,1,2', '--pdf=errorset'] = systematics_arguments ! see: https://cp3.irmp.ucl.ac.be/projects/madgraph/wiki/Systematics#Systematicspythonmodule
+# Syscalc is deprecated but to see the associate options type'update syscalc'
```
menglu21, line 172: line 170-172 can be removed

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/pp_H_TTG_012j_5f_NLO_MidDM/pp_H_TTG_012j_5f_NLO_MidDM_proc_card.dat
```diff
@@ -0,0 +1,3 @@
+generate p p > h [QCD]
```
menglu21, line 1: you need first import the model, and define the 'p' and 'j' to five flavor

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/pp_H_TTG_012j_5f_NLO_MidDM/pp_H_TTG_012j_5f_NLO_MidDM_madspin_decay.dat
```diff
@@ -0,0 +1,28 @@
...
+set spinmode none          # Use one of the madspin special mode
+set max_weight_ps_point 400  # number of PS to estimate the maximum for each event
+
+#import model DipoleMomentMid_UFO models/DipoleMomentMid_UFO/param_test.dat --bypass_check
+#import model DipoleMomentMid_UFO models/DipoleMomentMid_UFO/param_signal.dat --bypass_check
+import model DipoleMomentMid_UFO ./param_signal.dat --bypass_check #unsure what to specify param_signal.dat relative to
+
+decay h > ta+ ta- a
+
+launch
```
menglu21, line 28: please change the card to following (one question, i never use the syntax like you, where did you find the reference):

set ms_dir ./madspingrid

#initialization parameters
set Nevents_for_max_weight 250 # number of events for the estimate of the max. weight
set max_weight_ps_point 400  # number of PS to estimate the maximum for each event

set max_running_process 1

decay h > ta+ ta- a

launch

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/pp_H_TTG_012j_5f_NLO_MidDM/pp_H_TTG_012j_5f_NLO_MidDM_run_card.dat
```diff
@@ -0,0 +1,172 @@
...
+     6800.0     = ebeam1  ! beam 1 total energy in GeV
+     6800.0     = ebeam2  ! beam 2 total energy in GeV
+# To see polarised beam options: type "update beam_pol"
+
+#*********************************************************************
+# PDF CHOICE: this automatically fixes also alpha_s and its evol.    *
+#*********************************************************************
+     #nn23lo1    = pdlabel     ! PDF set                                     
+     #230000    = lhaid     ! if pdlabel=lhapdf, this is the lhapdf number
+
```
menglu21, line 45: remove those commented lines

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/pp_H_TTG_012j_5f_NLO_MidDM/pp_H_TTG_012j_5f_NLO_MidDM_run_card.dat
```diff
@@ -0,0 +1,172 @@
...
+ 0.0  = etajmin ! min rap for the jets
+ 0.0  = etaamin ! min rap for the photons
+ 0.0  = etalmin ! main rap for the charged leptons
+ {} = eta_min_pdg ! rap cut for other particles (use pdg code). Applied on particle and anti-particle
+ {} = eta_max_pdg ! rap cut for other particles (syntax e.g. {6: 2.5, 23: 5})
+#*********************************************************************
+# Minimum and maximum DeltaR distance                                *
+#*********************************************************************
+ 0.4 = drll    ! min distance between leptons 
+ 0.4 = dral    ! min distance between gamma and lepton
```
menglu21, line 124: you want these two cuts or they are there by default. if madspin is used for decay, these cuts won't be applied

menglu21 (commented):
the model is not needed here, i already put it in the central pool.

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/pp_H_TTG_012j_5f_NLO_MidDM/pp_H_TTG_012j_5f_NLO_MidDM_madspin_decay.dat
brallmond: Between the card at this link and the other comment saying to import the model before I define protons and jets, I'm confused where I should be importing my model. Per this post on launchpad, I'm currently importing the model in the madspin card and not in the proc card: https://answers.launchpad.net/mg5amcnlo/+question/703508

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/pp_H_TTG_012j_5f_NLO_MidDM/pp_H_TTG_012j_5f_NLO_MidDM_run_card.dat
brallmond: Since i am using madspin, it's okay to leave these cuts in? they are there by default. Would it be better to directly copy a run card from an existing set of NLO cards ?

bin/MadGraph5_aMCatNLO/cards/pp_H_TTG_012j_5f_NLO_MidDM/pp_H_TTG_012j_5f_NLO_MidDM_run_card.dat
brallmond: I was following this twiki which seems to say I should just upload my own fragments, https://twiki.cern.ch/twiki/bin/view/CMS/GitRepositoryForGenProduction 
Sorry for my confusion, so it is recommended to use an existing run card from an existing set of run cards that is most similar to the process I'd like to generate? Probably the HtoEleEleGamma would be good, since you suggested its madspin in the other comment?

agrohsje:
Hi @brallmond let's have the discussion about the technical details here.  
First comment. In your process card you import no explicit model. The default is 4FS. So when doing p = p b b~ you are adding massive b's to the proton which should be avoided (also this is a loop induced process). I would do 
import model loop_sm-no_b_mass. 
Second, there is a mismatch between the param_card from your madspin UFO and the proc card. Who told you about the settings in the customize card? Did you test the setup standalone? What is the purpose of the new UFO in the decay, i.e. your physics goal? Can one harmonize the param cards?

brallmond:
> Hi @brallmond let's have the discussion about the technical details here. First comment. In your process card you import no explicit model. The default is 4FS. So when doing p = p b b~ you are adding massive b's to the proton which should be avoided (also this is a loop induced process). I would do import model loop_sm-no_b_mass. Second, there is a mismatch between the param_card from your madspin UFO and the proc card. Who told you about the settings in the customize card? Did you test the setup standalone? What is the purpose of the new UFO in the decay, i.e. your physics goal? Can one harmonize the param cards?

Thank you for the comments! I'll include the import model statement in the proc card.

My physics goal is to include the diagram h -> ta+ ta- gamma where all four particles interact by a single vertex (see Fig. 1 of this paper https://arxiv.org/pdf/1810.09570.pdf). These diagrams can't be generated with default MG, so we use the imported model, which was written by the authors of the referenced paper. 

Right now, I am trying to generate a signal sample using the gridpack framework. I have generated a signal lhe and root file in a standalone environment, but only at the gen level. After I am able to generate a simple signal sample in the gridpack framework, I anticipate making more complicated samples with jets in the final state, and accounting for tau decays to other leptons. 

The settings in the customize card are a product of my standalone testing. Originally, I was importing the model with a restricted param card where masses were set to zero and yukawa couplings were set to zero, except for those of the particles used in the madspin card. Without doing this, I was getting errors of "parameter Yukawa: 4 is not zero" as described here (https://answers.launchpad.net/mg5amcnlo/+question/703526). I was getting these errors for every particle that wasn't part of the madspin decay. 

I found including the customizecard.dat gave the same behavior (prevented errors of that kind). I would like to harmonize the param cards, but I'm not sure how to do this when the p p > h in the proc card and the h > ta+ ta- a in the madspin card use different parameters (i.e. higgs should couple to everything in p p collision, but only to taus in the decay).

brallmond:
Hello, I'm following up on the comments I posted the other day. Is what I describe above the correct way to approach using MadGraph?

Also, as I mentioned in the CMStalk post, my gridpack generation is not able to converge. Even after including the no-b-mass model as you suggested, I find many print outs such as the one below and the gridpack never finishes being generated. Can you help me understand and overcome this problem? 

If it is simpler to discuss than to type, we can meet over Zoom? 
Or maybe we could meet in person if you are based at CERN?

"""
INFO: fail to reach target 1200 
store_events
  === Results Summary for run: GridRun_166257119 tag: tag_1 ===

     Width :   0 +- 0 GeV
     Nb of events :  11
"""

agrohsje:
Dear @brallmond, I would need to sit down and check myself to understand how to solve. Unfortunately, my agenda is totally filled up. Maybe someone from MEFG L3 or the MG5 team can check @mlizzo @DickyChant ?

DickyChant:
Hi @brallmond Could you share us with some log files? Anyhow, based on `INFO: fail to reach target 1200` it looks like a MadSpin issue. To me it looks like that you are simulating too strict phase space which is hard for MadSpin to have good sampling efficiency.

brallmond:
> Hi @brallmond Could you share us with some log files? Anyhow, based on `INFO: fail to reach target 1200` it looks like a MadSpin issue. To me it looks like that you are simulating too strict phase space which is hard for MadSpin to have good sampling efficiency.

Hello, thank you so much for taking the time to look into this! Here is a log file that I just generated with the cards as they appear in the most recent push to this PR, which includes the no-mass-b model.

Most interesting are these lines towards the bottom of the log, which shows 30 minutes to decay 100 events

```
INFO: fail to reach target 7200 
store_events
  === Results Summary for run: GridRun_199004711 tag: tag_1 ===

     Width :   0 +- 0 GeV
     Nb of events :  168
 
decay_events -from_cards
DONE
write ./events.lhe.gz
decaying event number 100 [1844.92604637146 s]
```



[HToTauTauGamma_NLO_5f.log](https://github.com/cms-sw/genproductions/files/10024103/HToTauTauGamma_NLO_5f.log)

DickyChant:
I think there are several interesting aspects of this request: First of all, I am not sure if the customized card is really working for **modifying** the parameters for the madspin running (I dont think so as I terminated the gridpack generation and checked the working directory, it seemed that the parameter cards didn't receive modification. I think the madspin running may crashes because of that which is what I have observed from standalone MG29x running. Even if the parameter settings are fine (I am referring to the customized card where you unset the fermion masses, just a reminder that you should not put `set yukawa 23/24 0` here as for weak gauge bosons it is not yukawa coupling, it has no effect here because there is no such parameter), regarding the madspin decay I think there is another issue that the "dipole moment" introduced here (via a h-a-l-l vertex whose coupling defined as GC_15 in the model parameters) seems to be too small to ensure good sampling efficiency. I did a standalone test with 1. unset udsg+e+mu masses/yukawa couplings and 2. 1000x larger GC_15 and the processes run smoothly (however the width will be not physical). Regarding this I also have a question on the externally provided Higgs width, is that value from the theoretical literature you have shared?

brallmond:
Hi Sitian,

I used your instructions to make some local runs. First I modified the customization file like you mentioned,
then I changed the coupling GC_15, multiplying it by 1000. This ran fine standalone. I repeated this except 
multiplying by 100x, 10x, and 1x (i.e. no change). All ran fine standalone. I did this with the 5f and 4f 
schemes and didn't run into issues standalone.  

Since this worked standalone for the 1x case (i.e. no change), I tried again to submit a gridpack with
the changed customization file but I ran into the same problem (lack of convergence with MadSPIN).

I also ran a test where the customization file isn't included at all and I have the same convergence issues (log is attached).
[HToTauTauGamma_NLO_5f.log](https://github.com/cms-sw/genproductions/files/10050707/HToTauTauGamma_NLO_5f.log)

The standalone version of madgraph I am using is 2.9.12, and the gridpack uses 2.9.9. 

I think it would be worthwhile to try running the on the grid the 1000x GC_15 case, so I sent that version
of the model to Meng Lu so it can be uploaded (it will of course be deleted after we figure out this convergence issue).

I'll also try downloading v2.9.9 of MadGraph and seeing if I can replicate the convergence issues standalone...

What do you think of this approach? Did you run into convergence issues for the 1x case?

Aside: 
You mentioned previously that the customization file wasn't being used, but you can find lines like
the one below in the log that I shared in a previous comment.

In the log files, this message appears for each modified parameter. So it is reading my customization file.
>[1;34mWARNING: Note that this parameter seems to be ignore by MG.
MG will use instead the expression: : 0.0
You need to match this expression for external program (such pythia). [0m
[1mINFO: modify param_card information BLOCK mass with id (13,) set to 0.0 [0m

DickyChant:
Hi @brallmond

I highly appreciate your efforts. Here are several remarks:

0. The log for gridpack production itself includes only the standard out of the running, so by default it doesn't include failure.
1. My suggestion on the customized card is more physics related. I have clarified that it has no effect for your case. 
2. If you read the madspin related part in your log file, there is a line:
>
>[1;34mWARNING: No param_card defined for the new model. We will use the default one but this might completely wrong. 
>
Which means if you import a model **different** from the UFO model for process generation, MadGraph will **not** reuse the parameter card. 
3. Following the last bullet, customized cards are trying to input commands that controls the running of MadGraph. However, in gridpack generation, the whole process is factorized into several isolated MadGraph running, and customized cards will not take care of the MadSpin step.

Regarding the usage of customized cards, I never intended to claim it is **not used**, I am just saying the line setting *yukawa* coupling for pdgid==23 and pdg==24 is not used. In fact, if you read your log, there are lines right below the line you've posted:
>
>[1;34mWARNING: invalid set command param_card yukawa 24 0.0 [0m
>

brallmond:
Hi all (@menglu21 @DickyChant), thanks very much for your comments and help with this! 

In the past two weeks we were finally able to make a gridpack with the 1x (standard) version of our model, and we verified that the output events we get from the gridpack look the same as those generated by a local (standalone, no gridpack) run. 

We would like to go forward with official production using the cards in their current state (just updated the extramodels card to use the 1x model's name). Is this possible? What remains to be done/verified and how can I make an official request?

Aside: 
In case it's helpful to include in a twiki somewhere or if someone runs into an issue similar to mine in the future... 
When comparing between local and gridpack versions, I used the exact same cards. This is fine IF you remember to move/rename the directory "madspingrid" when running locally (standalone, no gridpack). If you don't do this, madgraph will simply use whatever is in that directory, which was the 100x version of the model in my case. This went unnoticed for two weeks until my advisor caught it. After we fixed this, everything agreed and made perfect sense.

brallmond:
Hello and happy new year! I am pinging this PR again, please see the previous comment about next steps for officially integrating these cards. @menglu21 

Best,
Braden

menglu21:
Hi @brallmond have you updated the model in https://cms-project-generators.web.cern.ch/cms-project-generators/, i did a local test and it still gave "parameter yukawa: 4 is not at zero"

brallmond:
Hi @menglu21 sorry for the delay in my response to this. 
I'm not sure why you're encountering that error as that model worked for me the last time I used it.
However, I don't think it matters now because we realized the model we were using was missing important interference terms, so we are now waiting for our theory contacts to help resolve this issue.
Again, sorry for the delay in this message. I was waiting to hear back from our contact, and by the time we did I had forgotten about this message... I will reach out to you again when we have something more solid to report

PR #3289, opened 2022-10-26
add more to clean up tool

agrohsje:
this is a follow up on the discussion with david lange and o&c. 
i tested the additional rm on a lo and nlo gridpack. both still worked after removal and the lhe didn't change after removal.


agrohsje:
hi @sihyunjeon, this is the pr i mentioned in our chat.

sihyunjeon:
Let's start working ! :)

PR #3291, opened 2022-11-01
First Run 3 HZZ SM Cards Powheg plus JHUGen

jdavis36:
Currently we will only be using cards for MH 124 - 126. I will update this again once we start high mass production (Just modifying the python script with more mass points). 
These samples will be used for early Run3 SM Higgs analysis.


mlizzo:
Hi @jdavis36 , cards look ok to me, I have one naive question though: why is the `masswindow` parameter set to just 1? Could you please motivate this choice? I've noticed that in the corresponding Run2 template card [1] that parameter was set to 9999d0, but I guess you don't need that much to go off-shell. 

[1] https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/production/2017/13TeV/Higgs/gg_H_ZZ_quark-mass-effects_NNPDF31_13TeV/gg_H_ZZ_quark-mass-effects_NNPDF31_13TeV_template.input

jdavis36:
Hi @mlizzo. Thanks for noticing this. I don't know why that did not continue over to the new cards. I will change this and update the PR.

Dominic-Stafford:
I had a look though, and the only think I noticed is these new cards use lhapdf set 325300, rather than 306000, which was used in the previous cards. Has the recommendation of which set to use changed for run 3?

mlizzo:
Hi @Dominic-Stafford , if I remember correctly, the recommended PDF sets for Run3 are the same as those for UL, i.e. 325300 for 5f processes and 325500 for 4f processes. I think old cards were made for pre-UL samples, that's why are different I guess

Dominic-Stafford:
OK, thanks. Looks good to me, then!

PR #3301, opened 2022-11-11
Run3pdf el8

agrohsje:
I kept the UL PDF's, but didn't make a proper flag to call it from outside. people can for the time being just change inside make_rwl.py. The logic should be extended and at the same time fordy and forxojj cleaned up. made a todo comment in the code. the pdfs should agree with the choice for mg5 4FS and 5FS. LO cards are dropped. I kept the powheg split between hessian and mc replicas. please check that the ranges are ok and that no pdf is missing.


Saptaparna:
Merging now. Might need another round of check just in case important PDF sets are missing.

PR #3303, opened 2022-11-15
Added Higgs_13p6TeV signal madgraph card

chuwang1:
Dear expert,
We have produced some Higgs analysis madgraph cards at 13.6TeV for RunIII , which included:
- ggF cards
- VBF cards
- ttH cards
- VH cards
For these cards, we only changed the beam energy in the run_card.
Please review them, if there are any issues, please tell us.
Thanks in advance!
Chu Wang


menglu21:
Hi Chu, why there are two directories for each process, e.g., there are ggh012j_5f_NLO_FXFX_ and ggh012j_5f_NLO_FXFX_125, while the former one includes the later one, and the proc_card and run_card are exactly the same for them. And it's also the case for the other processes.

chuwang1:
> 

Hello, Meng
The one with suffix 125 (eg. ggh012j_5f_NLO_FXFX_125) are 125GeV card, you can see the mass has been set to 125GeV in cards. 
And the one without suffix 125 (eg. ggh012j_5f_NLO_FXFX_) are templates, which can be read by the shell scripts (eg. ggh012j_5f_NLO_FXFX.sh ). Then the script will produce multiple mass points.


Chu Wang

menglu21:
Hi Chu, 
yes, my question is that you have the mass point 125 in the repo without suffix, maybe you don't need the one with 125 suffix. They are exactly the same, is that right

Cheers, Meng

chuwang1:
> Hi Chu, yes, my question is that you have the mass point 125 in the repo without suffix, maybe you don't need the one with 125 suffix. They are exactly the same, is that right
> 
> Cheers, Meng

Hello, Meng,
Yes, you are right, I will remove the directories with 125 suffix.
Bests!
Chu Wang

chuwang1:
> Hi Chu, yes, my question is that you have the mass point 125 in the repo without suffix, maybe you don't need the one with 125 suffix. They are exactly the same, is that right
> 
> Cheers, Meng
Yes, you are right.

chuwang1:
> Hi Chu, yes, my question is that you have the mass point 125 in the repo without suffix, maybe you don't need the one with 125 suffix. They are exactly the same, is that right
> 
> Cheers, Meng

Hello, Meng,
I have removed the directories with suffix 125.
Please review them again.
Thanks!
Chu Wang

PR #3304, opened 2022-11-17
Run3 VBF & GGF processes

Soumyatifr:
This is for the Run3 production of the POWHEG VBF and HJ MinLO processes, only the CoM has been changed to 6800 from 6500 TeV. The PDFs are kept to the UL one. Please verify the cards and send your comments.


menglu21:
Hi, those cards seem to be the same with UL ones except the energy as mentioned. I just have one question on the "storeinfo_rwgt", it's set to be 0, will it be useful to switch it to 1 for potential use in the future. @agrohsje

PR #3306, opened 2022-11-23
ALPs signal samples for ZZ and ZH with all hadronic final states

anmehta:
This PR means to get the generator cards reviewed for the non-resonant ALP signal samples  for ZZ (linear EFT) and ZH (chiral EFT) productions where both Z & H decay hadronically.


bin/MadGraph5_aMCatNLO/cards/ZHhad_ALP_chiral/ZHhad_ALP_chiral_proc_card.dat
```diff
@@ -0,0 +1,14 @@
...
+define lep+ = e+ mu+ ta+
+define lep- = e- mu- ta-
+define vw = w+ w-
+
+generate g g > z h, (z > q q , h > vw > lep+ lep- vl vl~)
+add process g g > z h, (z > q q , h > a  lep+ lep-)
+add process g g > z h, (z > q q , h > lep+ lep-  lep+ lep-)
+add process g g > z h, (z > q q , h > z > lep+ lep- vl vl~)
+add process g g > z h, (z > q q , h > z > lep+ lep- q q)
+add process g g > z h, (z > q q , h >  q q)
```
menglu21, line 12: have you tried with the syntax? i think the parentheses should be removed

bin/MadGraph5_aMCatNLO/cards/ZZhad_ALP_linear/ZZhad_ALP_linear_proc_card.dat
```diff
@@ -0,0 +1,6 @@
+import model ALP_linear_UFO
+define q = u c d s b u~ c~ d~ s~ b~
+define l+ = e+ mu+ ta+
+define l- = e- mu- ta-
+generate g g > z z, z > q q, z > q q
```
menglu21, line 5: it should be enough to use "generate g g > z z, z > q q"

bin/MadGraph5_aMCatNLO/cards/ZHhad_ALP_chiral/ZHhad_ALP_chiral_proc_card.dat
anmehta: Hi @menglu21 , yes we have already produced and tested the gridpacks :)

bin/MadGraph5_aMCatNLO/cards/ZZhad_ALP_linear/ZZhad_ALP_linear_proc_card.dat
anmehta: Hi @menglu21, indeed, this works. I will recreate the PR. thanks!

PR #3308, opened 2022-11-29
GammaGamma to tau tau inclusive cards

michael-pitt:
Dear expert,

I've produced gammagamma->tautau inclusive production madgraph cards at 13TeV, which included BSM weights, and are used for g-2 measurement of the tau lepton.

For inelastic photons. the _MSHT20qed_nnlo_inelastic_ set is used:
```
lhapdf = pdlabel   ! PDF set
26500 = lhaid        ! if pdlabel=lhapdf, this is the lhapdf number
```

For elastic photons hadronic survival probabilities with charge forma factors were used based on [gamma-UPC](https://arxiv.org/abs/2207.03012) model, and implemented using:
```
2      =  lpp1         ! beam 1 type - photon from proton
chff  = pdlabel   ! charge form factors
```

Please review the cards, and if there are any issues, please let me know.

Thank you in advance!
Michael


bin/MadGraph5_aMCatNLO/cards/production/13TeV/GGToTauTau_Inclusive_M_50GeV_TuneCP5_madgraphLO_reweight/GGToTauTau_Inclusive_M_50GeV_TuneCP5_madgraphLO_reweight_run_card.dat
```diff
@@ -0,0 +1,174 @@
...
+  0.0	= drbl ! min distance between b and lepton 
+  -1.0	= drbbmax ! max distance between b's
+  -1.0	= drbjmax ! max distance between b and jet
+  -1.0	= drjlmax ! max distance between jet and lepton
+  -1.0	= drblmax ! max distance between b and lepton
+#*********************************************************************
+# Minimum and maximum invariant mass for pairs                       *
+#*********************************************************************
+  0.0	= mmbb ! min invariant mass of a b pair 
+  50.0	= mmll ! min invariant mass of leptons
```
menglu21, line 136: what's the reason to require mmll=50 here

menglu21:
Hi @michael-pitt another minor question is that have you had any expectation on the pdf uncertainty, since the current pdf list of UL doesn't include many qed pdf

michael-pitt:
Hi @menglu21 

I've changed the cards. Previously we were using pdf-based photon fluxes, which are incorrect if one uses proton tagging or rapidity gaps to identify photon-induced processes.
A recent MadGraph implementation of the correction factors was introduced in [2207.03012](https://arxiv.org/abs/2207.03012) and will become available in MG3.5.0 for elastic case only (photons from protons using `lpp=2`).

I prepared in advance cards in 4 folders:
Elastic tautau production with a mass cut of 50 or 300 GeV (two folders)
Elastic-Inelastic and Inelastic-Elastic with mtautau>50 GeV cut (two folders)

The Elastic-Inelastic need to be debugged (fails during generate_events pilotrun), but the first two cards are now used in the analysis.

The cards can run with older MadGraph releases if one switches off the corrections by changing `pdlabel` from `edff` to `none` in the run_card.dat.

Regards,
Michael

michael-pitt:
This is a followup on the error with Inelastic cards. I uploaded a card for fully inelastic gammagamma->tautau production (both photons taken from pdf). Unfortunately, the following command works only with `mg265UL` but not with the latest branches (tested with `mg27x` and `mg33x`), complaining about LHAPDF:
```
./gridpack_generation.sh GGToTauTau_Inelastic_M_50GeV_TuneCP5_madgraphLO_reweight cards/production/13TeV/GammaGammaToTauTau/GGToTauTau_Inelastic_M_50GeV_TuneCP5_madgraphLO_reweight/
```

menglu21:
Hi @michael-pitt so currently only fully inelastic cards have issue

bin/MadGraph5_aMCatNLO/cards/production/13TeV/GGToTauTau_Inclusive_M_50GeV_TuneCP5_madgraphLO_reweight/GGToTauTau_Inclusive_M_50GeV_TuneCP5_madgraphLO_reweight_run_card.dat
michael-pitt: Keep similar generator cut as in standard DY samples (for example https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/dyellell012j_5f_NLO_FXFX/dyellell012j_5f_NLO_FXFX_run_card.dat#L130)

sihyunjeon:
you can't have `tau > all all` in madspin since SMEFT model does not treat mesons as a stable particle. you have to let pythia do the rest (or tauola).

also `drll 0.4` is a bit tight as your analysis level will likely give the same cut for the leptons. for better (or realistic) modelling, it would be better to set this to 0

bbilin:
We keep this one open, until we manage to get MG3.5 available soon.

Saptaparna:
Since the analysts are giving a talk right now, I was wondering if the MG3.5 or MG3.X branch is ready to use and have the analysts tried updating to it?

michael-pitt:
Hi @bbilin can we reopen this PR?
All these datacards used to produce gridpacks for [SMP-23-005](https://cms.cern.ch/iCMS/analysisadmin/cadilines?line=SMP-23-005&tp=an&id=2726&ancode=SMP-23-005) and [SMP-24-019](https://cms.cern.ch/iCMS/analysisadmin/cadilines?line=SMP-24-019&tp=an&id=2921&ancode=SMP-24-019). 

The https://github.com/cms-sw/cmssw/pull/47183 should be used with these generators to allow realistic pT smearing (which is not supported by Madgraph, and is not available in Pythia)

What is still missing is the recent EFT model recommended by the EFT Forum https://arxiv.org/abs/2411.02483 and available at https://github.com/gdurieux/aqgc to be uploaded to https://cms-project-generators.web.cern.ch/cms-project-generators/

michael-pitt:
The request migrated to https://gitlab.cern.ch/cms-gen/genproductions_cards/-/merge_requests/8

PR #3310, opened 2022-11-30
Add Run 3 13p6TeV Higgs cards

alintulu:
I would like to review the following Run 3 13p6TeV Powheg input cards for Higgs production. The cards are exactly the same as for UL, with the exception of the beam energy which has been changed from 6500 to 6800.


menglu21:
cards look fine to me. I'll merge if there is no other comment in one or two days.

PR #3316, opened 2022-12-13
Cards for t-ch. 4FS Powheg+Pythia8 for 5 TeV for rereco (non-UL)

vrbouza:
Here are the cards for the rereco 5 TeV production of t-channel with Powheg + Pythia8. They are almost identical to rereco/UL's 13 TeV, but adjusting the parameters to those in other 5 TeV rereco (mainly quark masses, CKM elements, and PDF's) for coherence.

### DIFF
--- a/bin/Powheg/production/2017/5TeV/st_tch_4f_ckm_13TeV/st_tch_4f_ckm_antitop_13TeV.input
+++ b/bin/Powheg/production/2017/5TeV/st_tch_4f_ckm_13TeV/st_tch_4f_ckm_antitop_13TeV.input
@@ -27,8 +27,8 @@ ebeam2 2510
 
 
 ! To be set only if using LHA pdfs
-lhans1 306900
-lhans2 306900
+lhans1 325500
+lhans2 325500
 
 ! Parameters to allow or not the use of stored data
 use-old-grid    1 ! if 1 use old grid if file pwggrids.dat is present (<> 1 regenerate)
--- a/bin/Powheg/production/2017/5TeV/st_tch_4f_ckm_13TeV/st_tch_4f_ckm_antitop_13TeV.input
+++ b/bin/Powheg/production/2017/5TeV/st_tch_4f_ckm_13TeV/st_tch_4f_ckm_antitop_13TeV.input
@@ -1,0 +1,74 @@
+! ST-tchannel inputs
+
+#smartsig       1 
+withdamp       1
+hdamp 237.8775 ! 1.379*mtop
+
+withnegweights 0   ! will generate also events with negative-weight 
+
+renscfact  1d0    ! (default 1d0) ren scale factor: muren  = muref * renscfact 
+facscfact  1d0    ! (default 1d0) fac scale factor: mufact = muref * facscfact 
+
+#bornonly   1     ! (default 0) if 1 do Born only
+#testplots  1     ! (default 0, do not) do NLO and PWHG distributions
+
+iseed    SEED    ! initialize random number sequence 
+#rand1     0      ! initialize random number sequence 
+#rand2     0      ! initialize random number sequence 
+#manyseeds 1   
+
+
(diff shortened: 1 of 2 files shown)


bin/Powheg/production/2017/5TeV/st_tch_4f_ckm_13TeV/st_tch_4f_ckm_antitop_13TeV.input
```diff
@@ -0,0 +1,74 @@
+! ST-tchannel inputs
+
+#smartsig       1 
+withdamp       1
+hdamp 237.8775 ! 1.379*mtop
+
+withnegweights 0   ! will generate also events with negative-weight
```
menglu21, line 7: have you checked that the negative weighted events are not important, usually we keep the negative weighted event, i.e., use "withnegweights 1" (same comment for top card)
maybe @agrohsje can also comment here since i see he updated 13TeV st card

bin/Powheg/production/2017/5TeV/st_tch_4f_ckm_13TeV/st_tch_4f_ckm_antitop_13TeV.input
```diff
@@ -0,0 +1,74 @@
...
+ih1   1           ! hadron 1 (1 for protons, -1 for antiprotons)
+ih2   1           ! hadron 2 (1 for protons, -1 for antiprotons)
+
+ebeam1 2510
+ebeam2 2510
+
+
+! To be set only if using LHA pdfs
+lhans1 325500
+lhans2 325500
```
menglu21, line 31: i checked current 5tev cards and they are typically using PDF 306000 (same comment for top card)

bin/Powheg/production/2017/5TeV/st_tch_4f_ckm_13TeV/st_tch_4f_ckm_antitop_13TeV.input
vrbouza: I did not, but I do not see why we should not as you mention. As I said, this comes from UL, but I'll change it in 5 TeV if you prefer.

bin/Powheg/production/2017/5TeV/st_tch_4f_ckm_13TeV/st_tch_4f_ckm_antitop_13TeV.input
vrbouza: Yes, but the other processes are 5FS ones, whereas this one is 4FS. Thus, I searched for the NNPDF3.1 4FS analogous and it seemed to be 325500, thus I put that value.

vrbouza: Hi again @menglu21 : to try to speed up things, in case you see everything OK, I've started producing the gridpacks (in singularity). However, when I use the 4FS PDFs, 325500, there are errors when running steps 1, 2, and 3 of the gridpack production. In particular,

`terminate called after throwing an instance of 'LHAPDF::UserError'
  what():  PDF NNPDF31_nnlo_as_0124/1600 is out of the member range of set NNPDF31_nnlo_as_0124`

Thus, combined with this text in the Twiki where it says that CMSSW_10_6_* releases have _proper PDFs for UltraLegacy campaigns_ , and given that 325500 (the set I was using which is if I'm not correct NNPDF3.1 hessian for 4FS) is the recommended UL dataset, makes me think that maybe only some NNPDF3.1 PDF sets are present in rereco releases slc6-based. Could that be the case?

vrbouza:
After reviewing with authors, in the light of https://cms-talk.web.cern.ch/t/error-while-creating-gridpack-for-5-tev-rereco-sample , they decided to ask for 5FS samples. I'll open a new PR with the cards now.

PR #3318, opened 2022-12-14
Suu to chi chi fully hadronic MC data cards

emcannaert:
Request to create a new folder in cms-sw/genproductions/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ containing MC data cards for the Suu -> chi chi fully hadronic model. These files are located in the SuuToChiChi_FullyHadronic folder at this location and are to be used to make MC request gridpacks. The extramodels file necessary for this has already been uploaded to the extramodels webpage.


menglu21:
Hi @emcannaert is it true that the only difference among those cards are mass of particles, if so i think only the template cards and the script should be enough for this PR. Another question is that you use same decay width for different mass, have you tried using "AUTO" mode for decay width? if set the decay width manually you will get incorrect cross section, or you don't care about the cross section and only the kenematic shape will be used.

emcannaert:
Hi, I will create an updated version with just the template cards (the 8TeV Suu and 3 TeV chi cards) and the script. In regards to the decay width, the model writer has told us that the cross sections calculated by MG are not correct, so we use these for just the shape and scale them manually.

PR #3323, opened 2022-12-16
add powheg cards for HZJ_EW PowhegRes version, boosted

jieunyoo:
- HZJ_EW is currently only available for the leptonic mode
-  also no "inclusive" leptonic mode available
- Reference: https://arxiv.org/abs/1706.03522


menglu21:
cards look fine to me. @Saptaparna @mlizzo please have a check

mlizzo:
Hi @jieunyoo , cards look good to me as well, have you tried to make gridpacks out of them? I'm asking because I see a lot of comments across the cards that I'm not 100% sure the tarball creation script is able to digest. If you've already checked never mind, I'll wait for your feedback and then merge your PR

jieunyoo:
Hi @mlizzo These are for a new Powheg version. These cards are based on an example card, but have added the parameter bornsuppfactH. I am hoping for feedback from Powheg contacts @agrohsje  perhaps after vacation because I could not get .lhe files created from this. I was able to make a gridpack but no .lhe files

mlizzo:
Hi @jieunyoo , sorry I meant a valid gridpack capable of producing an lhe file. The tarball creation script indeed should work fine, but I'm not sure the runcmsgrid.sh script is able to correctly parse your card, sorry for my confusing message. You could try to eliminate all unnecesary/commented lines from your input card, create the tarball again (just it, you don't have to run all powheg steps from scratch) and then launch the runcmsgrid.sh script and see if it works. It happened to me something similar once, I think it's worth to give it a shot

bbilin:
@jieunyoo any news on this PR? We are applying a 1 year old policy to close older PRs. Yours is at the border.

jieunyoo:
Hi. Yes, you can close this pull request. HZJ_EW only is for Z->leptonic, but we decided to change to Z->hadronic due to lack of signal events unfortunately. A Run 3 analysis may later benefit from this; they can re-open I guess. Thanks.

PR #3325, opened 2022-12-19
WWJTo2L2Nu_MiNNLOPS card added for UL campaign

sv3048:


Saptaparna:
@covarell and @agrohsje:  could you please take a look, if possible. Thanks!

mlizzo:
Hi, this is the input card I tested for generating the MiNNLO WWJ process, following authors' instructions. Other than some specific MiNNLO settings, it is the same as this one https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/production/2017/13TeV/WWJTo2L2Nu_NNLOPS/WWJTo2L2Nu_NNLOPS_13TeV.input

covarell:
+1 from me

mlizzo:
Hi @agrohsje , actually the gridpack was already produced with these settings - after weeks of trials and errors to get it done. I don't know if it's worth it to redo it from scratch just to change this parameter, as I don't think we need it determined with uttermost precision in diboson analyses. I guess the m(WW) distribution is quite more relevant there, but please correct me if I'm wrong

agrohsje:
Let me merge so it is clear that it is approved @sv3048 .

mlizzo:
Thank you!

PR #3340, opened 2023-01-10
Alps mg265 ul

anmehta:
This PR means to get  generator cards reviewed for the non-resonant ALP signal samples for ZZ & WW (linear EFT) productions where both Z & W decay hadronically.


menglu21:
cards look good to me. @Saptaparna @mlizzo please have a look

PR #3341, opened 2023-01-10
Alps mg299

anmehta:
This PR means to get generator cards reviewed for the non-resonant ALP signal samples for ZH (chiral EFT) production with hadronic decays for Z and inclusively decaying Higgs.


bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ZHhad_ALP_chiral/ZHhad_ALP_chiral_proc_card.dat~
```diff
@@ -0,0 +1,20 @@
...
+
+
+generate g g > z h, z > q q , h > vw  > lep+ vl lep- vl~
+add process g g > z h, z > q q , h > a  lep+ lep-
+add process g g > z h, z > q q , h > lep+ lep-  lep+ lep-
+add process g g > z h, z > q q , h > z > lep+ lep- vl vl~
+add process g g > z h, z > q q , h > z > lep+ lep- q q
+add process g g > z h, z > q q , h >  q q
+output ZHhad_ALP_chiral -nojpeg 
+
```
menglu21, line 20: this card can be removed

menglu21 (commented):
btw, what's the difference between these two processes

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ZHhad_ALP_chiral/ZHhad_ALP_chiral_proc_card.dat~
anmehta: thanks and sorry about that and the  duplicate cards. Should be fine now.

anmehta: Hi @menglu21, sorry I forgot to update it here. But while testing the gridpacks I realized the fix to overcome the confusion b/w two MEs does not work. Madevent gives a zero cross section and hence no events are saved post the hadronization step. That said, I'm gonna use MG to generate the process and then decay higgs with pythia8. I have not tried it yet though and if you have any suggestions please let me know. But for now, I guess we should not push cards to the repo.

PR #3342, opened 2023-01-10
Fix two issues

chuwang1:
Dear expert,
I made this MR to update two issues:
There are several files named Virt_full_cHHH*grid.
These files will be copied into work directory.
But in latest powheg, these files have different names in ggHH model.
So I have updated the file names (removed _cHHH_ from file names).
Bests!
Chu Wang


covarell:
OK for me

Saptaparna:
Thanks @covarell! +1

PR #3346, opened 2023-01-16
Pythia fragment for DiBoson Backgoround Samples

jdavis36:
pythia fragment for ZZ_TuneCP5-pythia8 for Run3 analysis

### DIFF
--- a/genfragments/ThirteenPointSixTeV/ZZ/ZZToAny_TuneCP5_13p6TeV_pythia8_cfi.py
+++ b/genfragments/ThirteenPointSixTeV/ZZ/ZZToAny_TuneCP5_13p6TeV_pythia8_cfi.py
@@ -1,0 +1,37 @@
+import FWCore.ParameterSet.Config as cms
+
+from Configuration.Generator.Pythia8CommonSettings_cfi import *
+from Configuration.Generator.MCTunes2017.PythiaCP5Settings_cfi import *
+from Configuration.Generator.PSweightsPythia.PythiaPSweightsSettings_cfi import *
+
+generator = cms.EDFilter("Pythia8GeneratorFilter",
+	maxEventsToPrint = cms.untracked.int32(1),
+	pythiaPylistVerbosity = cms.untracked.int32(1),
+	filterEfficiency = cms.untracked.double(1.0),
+	pythiaHepMCVerbosity = cms.untracked.bool(False),
+	comEnergy = cms.double(13600.0),
+
+	crossSection = cms.untracked.double(12.90),
+	
+	PythiaParameters = cms.PSet(
+            pythia8CommonSettingsBlock,
+            pythia8CP5SettingsBlock,
+            pythia8PSweightsSettingsBlock,
+	        processParameters = cms.vstring(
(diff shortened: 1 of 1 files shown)


genfragments/ThirteenPointSixTeV/ZZ/ZZToAny_TuneCP5_13p6TeV_pythia8_cfi.py
```diff
@@ -0,0 +1,37 @@
+import FWCore.ParameterSet.Config as cms
+
+from Configuration.Generator.Pythia8CommonSettings_cfi import *
+from Configuration.Generator.MCTunes2017.PythiaCP5Settings_cfi import *
```
menglu21, line 4: from Configuration.Generator.MCTunesRun3ECM13p6TeV.PythiaCP5Settings_cfi import *

genfragments/ThirteenPointSixTeV/ZZ/ZZToAny_TuneCP5_13p6TeV_pythia8_cfi.py
```diff
@@ -0,0 +1,37 @@
...
+from Configuration.Generator.PSweightsPythia.PythiaPSweightsSettings_cfi import *
+
+generator = cms.EDFilter("Pythia8GeneratorFilter",
```
menglu21, line 7: generator = cms.EDFilter("Pythia8ConcurrentGeneratorFilter",

menglu21 (commented):
there is already samples for Run3, [BTV-Run3Summer22EEGS-00015](https://cms-pdmv.cern.ch/mcm/requests?prepid=BTV-Run3Summer22EEGS-00015)

jdavis36:
> BTW, do you really want to generate Pythia-only diboson samples? They usually don't provide a good description of the data -- they are LO to begin with plus have other issues like not modeling off shell decays.

I think that it is used at least for HZZ analysis to model some qqZZ background. We also separately generate other Di-Boson samples to use in conjunction with the pythia samples.

jdavis36:
> there is already samples for Run3, [BTV-Run3Summer22EEGS-00015](https://cms-pdmv.cern.ch/mcm/requests?prepid=BTV-Run3Summer22EEGS-00015)

genfragments/ThirteenPointSixTeV/ZZ/ZZToAny_TuneCP5_13p6TeV_pythia8_cfi.py
jdavis36: Hi, do you know why BTV did this sample? I was requested by Higgs Gen group to produce these samples. It is fine if they did it, I just want to know if the sample I was requested to make is the same as this one.

jdavis36:
Just Putting this here, I may have misinterpreted what sample was requested of me to generate. HZZ needs specifically 4L decays from this process. I will verify that this is not being produced and then I will change the data cards.

Saptaparna:
I am happy to merge this as long as you know that these are LO samples and we prefer to use NLO exclusive decays of diboson samples when possible.

Saptaparna:
or should I close this now?

PR #3347, opened 2023-01-16
Added cards for legacy rereco (non-UL) t-channel 5FS at 5 TeV.

vrbouza:
This is a new PR after talking about the previous 4FS analogous card [1] in CMS Talk and with authors. In this case, these are cards for the t-ch. in 5FS with analogous settings as similar samples of the same campaign and energy for coherence.

[1]: https://github.com/cms-sw/genproductions/pull/3316


menglu21:
cards look fine to me, @Saptaparna @mlizzo @Dominic-Stafford please have a look

Saptaparna:
Looks fine to me.

PR #3348, opened 2023-01-17
cards for VBF HH production at 13.6 TeV

fabio-mon:
Cards copied from the corresponding Run 2 folders (bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/VBF_HH/) with the only change in the beam energy which is set to 6800.


Saptaparna:
@fabio-mon are the setups identical except for the variations in the couplings?

fabio-mon:
> @fabio-mon are the setups identical except for the variations in the couplings?

yes

Saptaparna:
@fabio-mon can you please clean up the PR then? So, we can only check one set of cards.

fabio-mon:
The cards corresponding to the different couplings are now replaced by a single template card + a script to edit it

PR #3353, opened 2023-01-30
Increase AGAIN maximum pdf weights

covarell:
Run2 UL weights exceed again the maximum number of allowed weights


agrohsje:
Shouldn't we reduce the weights @menglu21 @Saptaparna instead of further increasing? The PDF weights slow down production a lot and I am pretty sure majority is not used at all.

covarell:
Agreed, but this is for Run2UL where "in principle" many gridpacks should have already been produced (I'm wondering how...)

covarell:
Run3 weights are already within the limits

Saptaparna:
I am merging this, so we can at least proceed with production.

PR #3357, opened 2023-02-07
Update to MG5 v2.9.13

shimashimarin:
Hello,

The LTS version of MG5 was updated from v2.9.9 to v2.9.13. The latest version should only have bug fixes. I removed useless patches and made the rest compatible with v2.9.13.
I validated these updates using the w0j_fxfx and the Gridpack was generated successfully. Everything looks fine on my side. Please consider updating to this version.

Best regards,
Jie Xiao


sihyunjeon:
https://answers.launchpad.net/mg5amcnlo/+question/704952

@shimashimarin hey, can we include this patch as well?

shimashimarin:
> 

I added a patch for this issue. But I see version v2.9.14 has been released 😂. May need to check with the author to confirm that it is a stable release.

sihyunjeon:
@shimashimarin That should be, but I think it's not a problem to keep this and then move on to higher releases like 2.9.15 or 16 later on since you already cleared out some of the patches

sihyunjeon:
@Saptaparna @menglu21

menglu21:
sounds reasonable to me, will merge if no more comment

PR #3360, opened 2023-02-17
add cards for tWgamma LO run2 production

beatrizlopes:
Cards for requesting production of a tWgamma sample at LO, for the run2 ttgamma differential measurement (TOP-23-002).

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/tWa_LO/tWa_LO_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/tWa_LO/tWa_LO_proc_card.dat
@@ -1,0 +1,16 @@
+import model sm-ckm_no_b_mass
+#switch to non-diagonal ckm matrix if relevant
+#import model loop_sm-no_b_mass
+
+#include b quark in proton and jet definition for consistent 5 flavour scheme treatment
+define p = p b b~
+define j = j b b~
+define part =  ve vm vt e- mu- ve~ vm~ vt~ e+ mu+ ta- ta+
+generate p p > t w- a, t > b part part, w- > part part @1
+add process p p > t~ w+ a, t~ > b~ part part, w+ > part part @2
+add process p p > t w-, t > b part part a, w- > part part  @3
+add process p p > t~ w+, t~ > b~ part part a, w+ > part part  @4
+add process p p > t w-, t > b part part, w- > part part a @5
+add process p p > t~ w+, t~ > b~ part part, w+ > part part a @6
+
+output tWa_LO -nojpeg
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/tWa_LO/tWa_LO_run_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/tWa_LO/tWa_LO_run_card.dat
@@ -1,0 +1,267 @@
+#*********************************************************************
+#                       MadGraph5_aMC@NLO                            *
+#                                                                    *
+#                     run_card.dat MadEvent                          *
+#                                                                    *
+#  This file is used to set the parameters of the run.               *
+#                                                                    *
+#  Some notation/conventions:                                        *
+#                                                                    *
+#   Lines starting with a '# ' are info or comments                  *
+#                                                                    *
+#   mind the format:   value    = variable     ! comment             *
+#*********************************************************************
+#
+#*******************                                                 
+# Running parameters
+#*******************                                                 
+#                                                                    
+#*********************************************************************
+# Tag name for the run (one word)                                    *
+#*********************************************************************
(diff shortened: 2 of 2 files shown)


bin/MadGraph5_aMCatNLO/cards/production/13TeV/tWa_LO/tWa_LO_proc_card.dat
```diff
@@ -0,0 +1,16 @@
...
+define j = j b b~
+define part =  ve vm vt e- mu- ve~ vm~ vt~ e+ mu+ ta- ta+
+generate p p > t w- a, t > b part part, w- > part part @1
+add process p p > t~ w+ a, t~ > b~ part part, w+ > part part @2
+add process p p > t w-, t > b part part a, w- > part part  @3
+add process p p > t~ w+, t~ > b~ part part a, w+ > part part  @4
+add process p p > t w-, t > b part part, w- > part part a @5
+add process p p > t~ w+, t~ > b~ part part, w+ > part part a @6
+
+output tWa_LO -nojpeg
```
menglu21, line 16: I think there would be overlap between your syntax, could you please try using 
"generate p p > t > w- w+ b a
add process p p > t~ > w- w+ b~ a"
and check the cross section with your current one

menglu21 (commented):
except the common inline, could you please push this PR to directory 2017/13TeV instead of 13TeV

bin/MadGraph5_aMCatNLO/cards/production/13TeV/tWa_LO/tWa_LO_proc_card.dat
beatrizlopes: Could you please clarify what cross section check you would like me to make? The syntax you suggest doesn't have the decays so I can't compare it directly to mine (and because of the photons it's not just multiplying by the branching ratio)

menglu21: please ignore my first comment, it still miss some diagram. maybe you can try "generate p p > t > l+ l- b a vl vl~ / z a", i got the cross sections are:
case1. your syntax: 0.035pb (with "True  = cut_decays" and all the cuts as default)
case2. p > t > l+ l- b a vl vl~ / z a, 0.0243pb
but i found case2 has 144 diagrams and case1 only has 87 diagrams, maybe you can check case2 to see all the diagrams are of your interest.

beatrizlopes: No problem. I tried but I don't get the same result that you mention. For example, when I run with my syntax it yields a cross section of 0.080pb and not 0.035pb. I was discussing with my group and we're not sure we understand the issue. Could you maybe give an example of a diagram that could be missing or overlapping in my case?

beatrizlopes: Dear Meng Lu,

Sorry for the late reply. We have performed a number of checks and indeed the last syntax you suggested seems to contain all the diagrams we need. We found that there seems to be an overlap in my original syntax, particularly in two cases:
- when the photon comes from the top quark. This diagram is accounted for twice: one in the twa process, and another in tw when the top decays into a photon and other particles.
- when the photon comes from the w (from the associated production, not the w from top decay). Same thing, it is accounted in the twa process and then again when the w decays to photon, lepton and neutrino.

So I will update my pull request to your suggested syntax. Let me know if it's ok to approve it this way or if you need us to perform any more checks.

beatrizlopes: Ok, done! Please let me know if everything is ok now

beatrizlopes: Just one more thing: it seems that with your suggested syntax with l and vl I only get diagrams with electrons and muons. But we would like also taus to be included? Do I have to include it explicitly, like define the leptons before generating the process?

menglu21: Hi, thanks for the update, for the tau case, the default definition of "l" only includes electron and muon, so you can add additional lines "lep+ = e+ mu+ ta+" and "lep- = e- mu- ta-" after you importing the model in the proc card, the "vl" definition already includes vl_ta so it's fine.

beatrizlopes: Ok, now it should be correct then! Let me know. And thanks a lot!

PR #3362, opened 2023-02-21
ttH cards for Higgs signal sample (Run3)

neharawal14:
Cards for production of ttH signal samples for Run3


menglu21:
I don't see ZZ information in the card, H decay mode is set to -1,could you please comment

neharawal14:
It was set to -1 following the consistency with  cards used in Run2. And also we submitted cards for other channels ggH, VBFH, WH, ZH in which the decay mode was set to-1 and they are approved also. 
Naively it looks to me that there is a need to change decay mode to ZZ. But could you explain any reasoning for it to be -1 and how it affects generation.

PR #3363, opened 2023-02-22
Add Higgs self-coupling modifications for GF HHH

mstamenk:
Updated Madgraph cards for GF HHH signal with different values for BSM scenarios.

As there are multiple interference effects between the various scenarios, at least 9 samples are needed. In addition, samples with very large values are added for cross-checks. 

The cards are exactly the same as the ones approved in: https://github.com/cms-sw/genproductions/pull/3198 

The only different is with respect to the customizedcards.dat where the value is changed. All the comments mention in the previous pull request have been ported to this one too.

In addition, provided Run 3 cards for 13.6 TeV

Tagging previous reviewers (apologies if I'm not tagging the right people):
@menglu21 @Saptaparna @GurpreetSinghChahal @mlizzo


menglu21:
Hi @mstamenk could you use script to generate different coupling cards, it will make our work easier. btw, in the customized card, you have -1.000.000, it should be -1.000

mstamenk:
Hi @menglu21 , many thanks for the prompt answer and for catching these issues. 

I've added a python script to generate the cards from the main GF_HHH_c3_0_d4_0. I've also fixed the issue you mentioned. 

Let me know if there's anything else! And thanks again

menglu21:
thanks for the update, I think you can remove most of the cards and only keep one template cards for the script.

mstamenk:
Hi @menglu21 ,

I've removed the repositories, thanks a lot

menglu21:
look good to me, @Saptaparna @mlizzo @Dominic-Stafford please have a check

PR #3367, opened 2023-03-01
bpbj_LO_M2200_100k_narrow_Single

DaLu0:
Request to merge cards to mg265UL branch. They are already available in the master branch


menglu21:
Hi, it seems you miss the madspin card, and set the width manually. another difference is the syntax in proc_card. do you want them to be exactly the same

DaLu0:
Hi, thanks for your help.
I already added the madspin card.
Yes, the width is set in the customize card and the correct syntax is this new one. Actually we would like to have this syntax in the master branch also.
Best,
Dalia

DaLu0:
Dear @menglu21 just a kind reminder.
Is there anything more to be done?
Cheers,
Dalia

PR #3371, opened 2023-03-06
fix merge script

michael-pitt:
This fix allows generation of > 5000 events in one job.

Before the fix, if the number of generated events exceeded 5000, then the simulation was split into Nevents / 5000 jobs, and output lhe files were merged using [Utilities/scripts/merge.pl](https://github.com/michael-pitt/genproductions/blob/master/Utilities/scripts/merge.pl) script, which unproperly retrieves the `lhe_version`, assuming it was stored in `MGGenerationInfo` block. The `MGGenerationInfo` block contains lines with `Number of events` and `Integrated weight`, while the version of the LHE file is stored in `LesHouchesEvents` block.

The fix allows the generation of a large number of events which can be later interfaced with external event filtering (e.g. [TaggedProtonHepMCFilter](https://github.com/cms-sw/cmssw/blob/master/GeneratorInterface/Core/interface/TaggedProtonHepMCFilter.h) or [EmbeddingHepMCFilter](https://github.com/cms-sw/cmssw/blob/master/GeneratorInterface/Core/interface/EmbeddingHepMCFilter.h)).

### DIFF
--- a/Utilities/scripts/merge.pl
+++ b/Utilities/scripts/merge.pl
@@ -60,11 +60,13 @@ foreach $infile (@ARGV) {
     if ($gzbytes == -1) { die("Error reading from file $infile\n"); }
     if ($gzbytes == 0)  { last; }
 
-    # Extract <MGGenerationInfo> information
+    # Extract <MGGenerationInfo> information and lhe_version
     if ($initblock == 0) {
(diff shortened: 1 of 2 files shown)


menglu21:
Hi @michael-pitt thanks for the fix, it take me some time to go back to the script and it does fix the pattern match. @Saptaparna @mlizzo @Dominic-Stafford please have a check and feel free to merge.

PR #3380, opened 2023-03-14
A temporary fix patch for reweight

shimashimarin:
A temporary fix patch to avoid errors when `rwgt_name` is not set in `reweight_card.dat`. The cause of this problem is [this update](https://github.com/mg5amcnlo/mg5amcnlo/commit/4a4979df52bf769ff191c2685d83773eed22c49e).


Saptaparna:
@shimashimarin do we need any additional validation for this?

shimashimarin:
This PR is now aligned with the official fix [here](https://bugs.launchpad.net/mg5amcnlo/+bug/2011584). I validated it with one [thq_4f_ckm_LO_ctcvcp_MH](https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Higgs/thq_4f_ckm_LO_ctcvcp_MH) process.

PR #3390, opened 2023-03-29
POWHEG fixes for CMSSW12 and 13 (processes with GoSam libraries and multiple fullgrid files)

covarell:
Cumulative fixes for:

- Processes using GoSam which needs special compilation flags for gfortran >= 10.3
- POWHEG-RES processes that produce multiple "fullgrid" files at stage 3

@agrohsje


Saptaparna:
They look good to me! @agrohsje unless you have any additional comments, I will merge the PR soon.

PR #3391, opened 2023-03-29
Updated energy and PDFs for Run 3

marlow511:
Updated energy and PDFs for Run 3 for ggH to ZGamma


menglu21:
there is a similar file already exist in https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/production/Run3/13p6TeV/Higgs/gg_H_quark-mass-effects_with_JHUGen_NNPDF31_13p6TeV/gg_H_quark-mass-effects_NNPDF31_13p6TeV_M125.input, could you please check it's what you need. if yes, please close this PR

PR #3402, opened 2023-04-12
ZZG Madgraph NLO cards for Run3

jaebak:
Dear GEN experts,

Here are MG cards for Run3 ZZG.
Please let me know if there are any comments.

Best regards,
Jaebak


menglu21:
Hi @jaebak, do I understand correctly that you are interested in our final status and will use pythia for Z decay. btw, I don't think it's necessary to include the customized card.

jaebak:
Hi @menglu21,

Thanks for having a look at the cards. I am working on the Run-3 H to Z(ll)gamma analysis and ZZG would be one of the backgrounds we are considering. So we would like to produce some events for ZZG.

For the customized cards, I am following the recommendation from the below link.
https://cms-talk.web.cern.ch/t/run-3-madgraph-modifications-needed-to-change-run-2-into-run-3/22402/3?u=jaebak

PR #3405, opened 2023-04-19
Adding missing Run3 H production cards with H undecayed

maravin:
Adding missing Run3 H production cards for the standard Higgs and processes (ggH, W-H, W+H, and HZ). Plan is to use for the future H->tau tau requests.


maravin:
Ping

maravin:
Ping

menglu21:
cards look fine to me, @Saptaparna @mlizzo @Dominic-Stafford please have a check

PR #3412, opened 2023-04-26
Cards for vector-like B -> tW in UL

jmhogan:
This PR adds two folders of cards for vector-like B quarks decaying to tW. 
"narrowLO" uses the narrow-width approximation, and "tenpctLO" has a finite width/mass of 0.1
Compared to earlier generation, both folders use an updated UFO with the author's suggested MG commands.


menglu21:
Hi @jmhogan so all the mass points in the NWA will have width 5 GeV, is it correct

jmhogan:
@menglu21 That's right, we're leaving alone the theorist's width setting for everything in NWA.

menglu21:
cards look fine to me, @Saptaparna @mlizzo @Dominic-Stafford please have a check

PR #3413, opened 2023-05-01
Add the thw and thq cards of run3 

XuanhaoZhang:


bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/Higgs/thq_4f_ckm_LO_ctcvcp_MH/thq_4f_ckm_LO_ctcvcp_MH125_customizecards.dat
```diff
@@ -0,0 +1,4 @@
+#put card customizations here (change top and higgs mass for example)
+set param_card mass 6 172.5
+set param_card mass 25 125
+set param_card yukawa 6 -172.5
```
menglu21, line 4: you set the coupling to be negative, is it a typo

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/Higgs/thq_4f_ckm_LO_ctcvcp_MH/thq_4f_ckm_LO_ctcvcp_MH125_run_card.dat
```diff
@@ -0,0 +1,268 @@
...
+#                                                                    
+#**********************************
+# BW cutoff (M+/-bwcutoff*Gamma)
+#**********************************
+  15.0	=  bwcutoff       ! (M+/-bwcutoff*Gamma)
+#**********************************************************
+# Apply pt/E/eta/dr/mij cuts on decay products or not
+# (note that etmiss/ptll/ptheavy/ht/sorted cuts always apply)
+#**********************************************************
+  False	=  cut_decays     ! Cut decay products
```
menglu21, line 94: do you want the following cuts been applied on lepton/jet from Top/W decay, if yes, please use "True	=  cut_decays" here

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/Higgs/thw_5f_ckm_LO_ctcvcp_MH/thw_5f_ckm_LO_ctcvcp_MH125_run_card.dat
```diff
@@ -0,0 +1,268 @@
...
+  $DEFAULT_PDF_SETS = lhaid
+  $DEFAULT_PDF_MEMBERS = reweight_PDF     ! if pdlabel=lhapdf, this is the lhapdf number
+#*********************************************************************
+# Renormalization and factorization scales                           *
+#*********************************************************************
+  True	=  fixed_ren_scale   ! if .true. use fixed ren scale
+  True	=  fixed_fac_scale   ! if .true. use fixed fac scale
+  40	=  scale             ! fixed ren scale
+  40	=  dsqrt_q2fact1     ! fixed fact scale for pdf1
+  40	=  dsqrt_q2fact2     ! fixed fact scale for pdf2
```
menglu21, line 61: what's the reason to use different ren/fac scale than Z mass value

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/Higgs/thw_5f_ckm_LO_ctcvcp_MH/thw_5f_ckm_LO_ctcvcp_MH125_run_card.dat
```diff
@@ -0,0 +1,268 @@
...
+#                                                                    
+#**********************************
+# BW cutoff (M+/-bwcutoff*Gamma)
+#**********************************
+  15.0	=  bwcutoff       ! (M+/-bwcutoff*Gamma)
+#**********************************************************
+# Apply pt/E/eta/dr/mij cuts on decay products or not
+# (note that etmiss/ptll/ptheavy/ht/sorted cuts always apply)
+#**********************************************************
+  False	=  cut_decays     ! Cut decay products
```
menglu21, line 94: same comment here

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/Higgs/thq_4f_ckm_LO_ctcvcp_MH/thq_4f_ckm_LO_ctcvcp_MH125_customizecards.dat
XuanhaoZhang: for this one,it was used in the same way in Run2 if I understand correctly: https://github.com/cms-sw/genproductions/blob/8d97961cda6fffccfa09cc245b715ce6f2b43f6f/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/Higgs/thq_4f_ckm_LO_ctcvcp_MH/thq_4f_ckm_LO_ctcvcp_MH_customizecards.dat, Should we set it positive?

menglu21: I'm not sure about it, it's better to confirm with the analyzer or the person who made the initial PR for the motivation of this setup

XuanhaoZhang:
Hi Meng, analyzers are wondering wether it will be better to set <True    =  fixed_ren_scale > to false?

menglu21:
> Hi Meng, analyzers are wondering wether it will be better to set <True = fixed_ren_scale > to false?

yes, when F = fixed_ren_scale/fixed_fac_scale" is used, the scale is the dynamic one which corresponds to the final state kinematics

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/Higgs/thq_4f_ckm_LO_ctcvcp_MH/thq_4f_ckm_LO_ctcvcp_MH125_customizecards.dat
menglu21: @sihyunjeon I checked the model HC_NLO_X0_UFO-v1.3.zip, it's 
ymt = Parameter(name = 'ymt',
                nature = 'external',
                type = 'real',
                value = 164.5,
                texname = '\\text{ymt}',
                lhablock = 'YUKAWA',
                lhacode = [ 6 ])
also I don't understand the meaning of a negative yukawa.

hftsoi:
hi @menglu21 just a kind reminder, could you please take a look and review? this request is urgently needed, thanks!

FloMau:
Dear @menglu21, @hftsoi, 

if I understand correctly, the cards merged in [#3475 ](https://github.com/cms-sw/genproductions/pull/3475) still have the negative top Yukawa couplings (which drastically change the physics by e.g. increasing the cross-section by an order of magnitude) as nominal value and still have the fixed scales of 40GeV for tHW, where I asked an expert from theory who clarified that this setting should be avoided. 
We updated the cards two months ago, but it seems to be stuck here. 

Is there a reason why the cards with the other settings were merged now or was it accidentally assumed that they are identical to these ones? 

Best, 
Florian

hftsoi:
so after clarifying with @FloMau, the MG cards here are for SM simulation, which is correct for their analysis and should be kept in this PR. The merged cards in #3475 appear to belong to another analysis which looks for BSM searches.

menglu21:
I am a bit confused now, it seems the cards here are the same with cards in https://github.com/cms-sw/genproductions/pull/3475/, are they asked by different person. I merged the cards in another PR cuz @hftsoi does find the reason (from analysis point of view) to set the negative Yukawa coupling. could you please double check with each other

lviliani:
We are cleaning the old PRs.
This one seems obsolete, so if we don't receive any objections we will **close** it later today.
Please react if instead this PR is still needed.

bbilin:

PR #3414, opened 2023-05-04
Fix mandatory parameters in powheg.input

covarell:
- fixes for unnumbered ubound files for vRES
- fixes wrong logic in mandatory powheg.input parameters

@agrohsje

### DIFF
--- a/bin/Powheg/Templates/createTarBall_template.sh
+++ b/bin/Powheg/Templates/createTarBall_template.sh
@@ -20,44 +20,51 @@ fi
 
 FULLGRIDRM=`ls $${WORKDIR}/$${folderName} | grep fullgrid-rm | head -n 1`
 FULLGRIDBTL=`ls $${WORKDIR}/$${folderName} | grep fullgrid-btl | head -n 1`
+UBOUND=`ls $${WORKDIR}/$${folderName} | grep ubound | head -n 1`
+PWGSTAT=`ls $${WORKDIR}/$${folderName} | grep st3-stat | head -n 1`
 FULLGRIDRM=$${FULLGRIDRM%?}
 FULLGRIDBTL=$${FULLGRIDBTL%?}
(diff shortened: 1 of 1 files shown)


agrohsje:
Thanks Roberto. I will ping the authors to understand the difference between ubound in v2 and vres. Will put you in cc.

covarell:
@agrohsje do you have further comments on this, besides checking the exact working mode of ubound? You mentioned this while talking, but I do not see them here

covarell:
Thanks @agrohsje , could you please approve also the backport to the powhegUL branch?

#3415

mseidel42:
That's weird, it works fine for us in 12_3_1.

But we get crashes when Powheg does not have rwgt information in subsequent reweighting runs (for some things that cannot be done on-the-fly)

covarell:
I suggest you open a PR where this is just re-instated for minnlo?

mseidel42:
Hi Roberto,

Digging in the Powheg code I see that for MiNNLO we got "saved" by the fact that Powheg stores #rwgt also when an rwl_file is present in the config, regardless of the storeinfo_rwgt setting.
Which makes me wonder even more why it helped with anybody's workflow to remove storeinfo_rwgt in the first place?

My problem is now with Powheg EW, where I need to run Powheg without weights, run the Photos script for the QED shower (which discards any weights but keeps #rwgt), and then add the weights afterwards.
I can work around that but I'm really sceptical that setting `storeinfo_rwgt 0` by default had any effect in the first place for normal workflows. Usually the xml check fails due to bad chars or double dashes like lines `--------` or arrows `-->` in the input file.

Best,
Markus

mseidel42:
Hi Roberto,

Can you tell me which process / config cards were affected by this?

If they used a rwl_file this patch did not do anything, and the error was probably unrelated.

Best,
Markus

PR #3426, opened 2023-05-16
ttW and ttZ cards for Run3

Cvico:
Dear experts,

As discussed between TMG and GEN, ttW and ttZ have been dropped from the common MC background list, and thus these must be dealt with from TMG side. I've uploaded the cards (adapted to Run3) to start gridpack generation. These gridpacks will be used to produce central ttW and ttZ samples with the old FxFx procedure. 

The only change between Run2 and Run3 is the center of mass energy in the run cards. 

Best regards and thank you in advance for your feedback,
Carlos Vico

PR #3431, opened 2023-05-17
ttW NLO cards for Run3

Cvico:
Hi, 

In [PR3426](https://github.com/cms-sw/genproductions/pull/3426) we agreed to split the PR into parts to make the review easier. I've prepared this PR with only the ttW NLO cards, I think for Run3 we can drop the LO process. For ttW the NLO corrections are huge so I don't think LO here is very trustful.

Of course if you think the LO should be add, I can add the cards.

I wanted to also ask about a comment from @menglu21 in the previous PR, about categorizing similarly to what's done in single top (where they have single top and single antitop in separate processes). I think for the case of ttW, this would make sense if we separate in ttW+ and ttW-. I've not done the split yet, but I can do it if needed. Maybe this can help reducing the gridpack size as well.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/TTWJets/TTWJetsToLNu_5f_NLO_FXFX/TTWJetsToLNu_5f_NLO_FXFX_run_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/TTWJets/TTWJetsToLNu_5f_NLO_FXFX/TTWJetsToLNu_5f_NLO_FXFX_run_card.dat
@@ -1,0 +1,146 @@
+#***********************************************************************
+#                        MadGraph5_aMC@NLO                             *
+#                                                                      *
+#                      run_card.dat aMC@NLO                            *
+#                                                                      *
+#  This file is used to set the parameters of the run.                 *
+#                                                                      *
+#  Some notation/conventions:                                          *
+#                                                                      *
+#   Lines starting with a hash (#) are info or comments                *
+#                                                                      *
+#   mind the format:   value    = variable     ! comment               *
+#***********************************************************************
+#
+#*******************                                                 
+# Running parameters
+#*******************                                                 
+#
+#***********************************************************************
+# Tag name for the run (one word)                                      *
+#***********************************************************************
+  tag_1     = run_tag ! name of the run 
+#***********************************************************************
+# Number of events (and their normalization) and the required          *
+# (relative) accuracy on the Xsec.                                     *
+# These values are ignored for fixed order runs                        *
+#***********************************************************************
+  1500 = nevents ! Number of unweighted events requested 
+ 0.001 = req_acc ! Required accuracy (-1=auto determined from nevents)
+    20 = nevt_job! Max number of events per job in event generation. 
+                 !  (-1= no split).
+average = event_norm ! Normalize events to sum or average to the X sect.
+#***********************************************************************
+# Number of points per itegration channel (ignored for aMC@NLO runs)   *
+#***********************************************************************
+ 0.01   = req_acc_FO       ! Required accuracy (-1=ignored, and use the 
+ 	                   ! number of points and iter. below)
+# These numbers are ignored except if req_acc_FO is equal to -1
+ 5000   = npoints_FO_grid  ! number of points to setup grids
+ 4      = niters_FO_grid   ! number of iter. to setup grids
+ 10000  = npoints_FO       ! number of points to compute Xsec
+ 6      = niters_FO        ! number of iter. to compute Xsec
+#***********************************************************************
+# Random number seed                                                   *
+#***********************************************************************
+     0    = iseed       ! rnd seed (0=assigned automatically=default))
+#***********************************************************************
+# Collider type and energy                                             *
+#***********************************************************************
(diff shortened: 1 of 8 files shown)


sihyunjeon:
I think ttW+ and ttW- is worth splitting only if the cross sections are a lot different. can you check this? (LO check should be good enough)

Cvico:
From TOP-21-011 (the ttW measurement) the values for ttW+ and ttW- at NLO in the new FxFx framework (these values were provided by Rikkert) are:
ttW+ : 475 +/- 49 fb
ttW- : 247 +/- 26 fb
ttW+ / ttW-:  1.92

sihyunjeon:
I guess it's worth doing it if someone is interested in doing charge split analysis. But how many stats are we going to request for ttW at LHE level? around 1000fb x 50fb-1 x 100 ? 100 to account for statistical abundance (NLO negative weights and events that gets filtered out due to jet merging)

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/TTWJets/TTWJetsToLNu_5f_NLO_FXFX/TTWJetsToLNu_5f_NLO_FXFX_run_card.dat
```diff
@@ -0,0 +1,146 @@
...
+# Renormalization and factorization scales                             *
+# (Default functional form for the non-fixed scales is the sum of      *
+# the transverse masses of all final state particles and partons. This *
+# can be changed in SubProcesses/set_scales.f)                         *
+#***********************************************************************
+ F        = fixed_ren_scale  ! if .true. use fixed ren scale
+ F        = fixed_fac_scale  ! if .true. use fixed fac scale
+ 91.188   = muR_ref_fixed    ! fixed ren reference scale 
+ 91.188   = muF1_ref_fixed   ! fixed fact reference scale for pdf1
+ 91.188   = muF2_ref_fixed   ! fixed fact reference scale for pdf2
```
sihyunjeon, line 75: Regarding "scales" I think TOP samples all fix their scales to 172.5 GeV. Can you double check with TOP MC conveners?

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/TTWJets/TTWJetsToLNu_5f_NLO_FXFX/TTWJetsToLNu_5f_NLO_FXFX_run_card.dat
```diff
@@ -0,0 +1,146 @@
...
+#***********************************************************************
+# Cuts on the jets                                                     *
+# Jet clustering is performed by FastJet.
+# When matching to a parton shower, these generation cuts should be    *
+# considerably softer than the analysis cuts.                          *
+# (more specific cuts can be specified in SubProcesses/cuts.f)         *
+#***********************************************************************
+   1  = jetalgo   ! FastJet jet algorithm (1=kT, 0=C/A, -1=anti-kT)
+ 1.0  = jetradius ! The radius parameter for the jet algorithm
+  20  = ptj       ! Min jet transverse momentum
```
sihyunjeon, line 119: what is the qcut that will be used in pythia?

Cvico:
> I guess it's worth doing it if someone is interested in doing charge split analysis. But how many stats are we going to request for ttW at LHE level? around 1000fb x 50fb-1 x 100 ? 100 to account for statistical abundance (NLO negative weights and events that gets filtered out due to jet merging)

About the split charge analysis, we measured the separate cross sections (+ and -) in Run2 and since you can always categorize events at analysis level, I'm not sure if there's anything to gain from having an splitted sample in that sense... Unless I'm missing something more interesting about the gains of splitting the samples.

About statistics, In 2018 (which is ~60 fb-1, similarly to what is expected in 2022-2023 if I'm not mistaken) we had ~5M events, and in the Run2 analysis the statistics coming from the MC were not worrysome. So I'd say your estimate should be good enough.

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/TTWJets/TTWJetsToLNu_5f_NLO_FXFX/TTWJetsToLNu_5f_NLO_FXFX_run_card.dat
menglu21: @sihyunjeon since "F = fixed_ren_scale/fixed_fac_scale" is used, the scale is the dynamic one which corresponds to the final state kinematics, I think it should be fine.

sihyunjeon: @menglu21 yes so i was wondering if that should be changed as well for consistency, because tt samples and single t samples were all using 172.5 scale fixed by hand.

bbilin:
@Cvico Is this PR sitll needed?

bbilin:
@Cvico Is this PR sitll needed?

PR #3436, opened 2023-05-24
Patch for ttJ_MiNNLO in Powheg.

zimermmane-santos:
Here I add a patch for the compilation of the`ttJ_MiNNLO` in _powheg_. 

- Here is the patch: [bin/Powheg/patches/ttJ_minnlo_compiler.patch](https://github.com/cms-sw/genproductions/compare/master...zimermmane-santos:genproductions:ttJ_MiNNLO-svn4025?expand=1#diff-0fc9d51560f83c0cee0c2814cf8e130ea8414342af3b9ca1fa7e6aeeb3eeb777)
- It is tested under CMSSW_12_4_5.
- [Templates/prepareJob_template.sh](https://github.com/cms-sw/genproductions/compare/master...zimermmane-santos:genproductions:ttJ_MiNNLO-svn4025?expand=1#diff-dc1ab0cc10bdb5e8dd91d134d1935da61d9ea57a11c489d4ab37b903caf70086) is changed to accommodate LHAPDF.6.3.0, which is not the standard pdfset of CMSSW_12_4_5. For that, it was necessary to update the function prepareJob() in [bin/Powheg/run_pwg_condor.py](https://github.com/cms-sw/genproductions/compare/master...zimermmane-santos:genproductions:ttJ_MiNNLO-svn4025?expand=1#diff-306a635579daba4fb10220a64e87ff0b29416fe9cfa2acf8bb8abd5391f8b7cd) to also include `process` as argument.
- [bin/Powheg/Utilities/helpers.py](https://github.com/cms-sw/genproductions/compare/master...zimermmane-santos:genproductions:ttJ_MiNNLO-svn4025?expand=1#diff-1d99f7f5640369b186ec8661a6d108a59bf46ee1eca5ab04e3f346e24dbf242d) Adding the patch and a local installation of chaplin 1.2 plus some minor changes for compatibility with the automation in genproduction.

**Note**: the compilation only works with the latest versions of POWHEG-BOX-V2, which currently is in its revision 4025. Therefore it is mandatory that `python run_pwg_condor.py `is called with `--svn 4025`for the compilation goes successfully.


menglu21:
Hi @zimermmane-santos sorry for the late reply, I have several questions related to this pr.
1. why is lhapdf630 needed for this process, the default in cmssw_12_4 is lhapdf640
2. I don't see the place where the second patch is used, did I overlook something

Cheers, Meng

zimermmane-santos:
Hi @menglu21 , thanks for looking into it.

1. This process is not compatible with lhapdf640. We can only wait for the code authors to fix it.
2. I have added only one patch. Which second patch were you referring to? Maybe you saw ttJ_MiNNLOPS_v2.1_beta1 at some point, but that was a test in a non-public version of the code, so I let that out for the moment.

Let me know if you have further questions.

Cheers,

Andre


> Hi @zimermmane-santos sorry for the late reply, I have several questions related to this pr.
> 
>     1. why is lhapdf630 needed for this process, the default in cmssw_12_4 is lhapdf640
> 
>     2. I don't see the place where the second patch is used, did I overlook something
> 
> 
> Cheers, Meng

menglu21:
Hi @zimermmane-santos I mean these two patches ttJ_minnlo_rwt_topmass_compiler.patch and ttJ_minnlo_compiler.patch

zimermmane-santos:
> Hi @zimermmane-santos I mean these two patches ttJ_minnlo_rwt_topmass_compiler.patch and ttJ_minnlo_compiler.patch

Hi, @menglu21, sorry I have overlooked. You are right the ttJ_minnlo_rwt_topmass_compiler.patch is not used yet. Should I do a new merge request or we are able to do remove it from here?

Cheers,
Andre

menglu21:
Hi @zimermmane-santos I think you can remove it from your repo, it will update here automatically

zimermmane-santos:
Hi, @menglu21. Just did that, let me know if there is something else to do.
Thanks for the feedback :)

zimermmane-santos:
Hi, @menglu21, does that look good now? Do you need more information from my side?

Saptaparna:
I don't see any issues. @covarell Is this fine? Just to make sure the commits to the helper code etc are fine before merging.

Saptaparna:
The code looks fine to me, so I am merging it now. @covarell if you notice anything strange, please let me know and we will revert the PR.

covarell:
Hello, in general I would avoid manipulating pdfs. For example, don't you observe issues with reweighting? The CMS weight list includes NNPDF4 weights, which are not available with lhapdf 6.3. 

It is much better to generate the grid pack in UL conditions (CMSSW 10) and then use it in Run3 workflows.

amsimone:
Hi @covarell,

Unfortunately the code does not compile with LHAPDF versions higher than 6.3. The authors are aware but do not know how to fix the issue, so for now we are forced to move on like this. But why should this give trouble with the PDF weights? Can't we use modern PDFs with LHAPDF6.3?

PR #3438, opened 2023-05-25
New Filter for GGZZ and GGHZZ

jdavis36:
Hi,

I am committing this because there was something incorrect happening in the ggZZ bkg samples. We need to relax the pt and eta cuts in the generation and apply the cuts after hadronization. This more correctly mimics what happens in experiment. Also, I will have to do something similar for the VBF -> ZZ bkg samples. However all that will be added is essentially the four lepton filter at the end. Do I need to make separate cards for them as well?

Best,
Jeff


menglu21:
Hi, it's not needed to make a separate one, anyhow we will check the fragment in the request.

PR #3442, opened 2023-05-25
Fix to  ttbar + tW photoproduction in aMC@NLO

Cvico:
Dear experts,

In the TMG group we are working on a tt and tW production in photoproduction modes at NLO
This PR is just to:
 1. change the value of the `store_rwgt_info` option store_rwgt_info in the run card for each
of the processes to `True`. The reason for this is mainly because for MG29 production that value needs to be set to true, otherwise the McM validation crashes. We need to use MG29 rather than MG26 because the patch provided by the
madgraph experts [1] can not be done in MG26 (the lines that need to be patched are not there). 
 2. I've also fixed a space in the `.patch` file that was giving problems.
 3. Gridpack generation of tW photoproduction crashes for some reason when the incoming particle is a photon (`a p > t W`). It gives an over weight problem that I think may be happening because NWA does not work well in this case. I've read that these kinds of problemas may give crashes in the production later. Since this production is rather tricky, I think it's better if we drop madspin from both tW productions to be consistent. Results in terms of cross section are:
   3.1 **Without madspin**: 5.113e-01 +- 4.1e-03 pb (for `p a > t W`)
   3.2 **With madspin**: 5.113e-01 +- 4.1e-03 pb (for `p a > t W`)

Thank you in advance for your feedback.

[1]: https://answers.launchpad.net/mg5amcnlo/+question/689742


sihyunjeon:
1. just for my education, why do you need to separate pa and ap? won't it be better to merge the two?
2. why not use 325100 so that we keep things consistent with how many orders we have in pdf (also consistency with cp5 tuning)?
3. regarding overweight issues, how frequently does it show up? 1 out of 1000 events is more or less fine (we see it in normal ttbar sample from time to time) Also if you don't use madspin for decays, you are decaying things all with pythia, and there you use all spin correlations

sihyunjeon:
> 3.1 Without madspin: 5.113e-01 +- 4.1e-03 pb (for p a > t W)
> 3.2 With madspin: 5.113e-01 +- 4.1e-03 pb (for p a > t W)

When you use madspin, madgraph does not show you the "cross section x BR". so 3.2 is not really the proper number.

Cvico:
> 1. just for my education, why do you need to separate pa and ap? won't it be better to merge the two?
> 
>     2. why not use 325100 so that we keep things consistent with how many orders we have in pdf (also consistency with cp5 tuning)?
> 
>     3. regarding overweight issues, how frequently does it show up? 1 out of 1000 events is more or less fine (we see it in normal ttbar sample from time to time) Also if you don't use madspin for decays, you are decaying things all with pythia, and there you use all spin correlations

Hi, thanks for the comments:
1. Since this is not p p, one needs to set lpp1 and lpp2 in the run cards to specify the type of beam. So for a p we have a run card with lpp1 = 2, and for p a we have a run card with lpp2 = 2. Analysers also requested to have the samples splitted depending on which direction the incoming photon comes. 
2. This inherits from a comment in the original [PR to genproductions](https://github.com/cms-sw/genproductions/pull/3336#discussion_r1068828913). Appart from that, I don't think there's another specific reason on the choice of PDF.
3. So I think this was happening every 2-4k events while generating 10k if I recall correctly. However, after your comment I tried to generate the gridpack again and saw that this time everything worked well... So it seems like as you say it's a random error. I've never encountered this error before so I didn't know it was common when producing these processes. In that case I can add madspin cards again if you think it's needed.

Cvico:
> > 3.1 Without madspin: 5.113e-01 +- 4.1e-03 pb (for p a > t W)
> > 3.2 With madspin: 5.113e-01 +- 4.1e-03 pb (for p a > t W)
> 
> When you use madspin, madgraph does not show you the "cross section x BR". so 3.2 is not really the proper number.

Apologies, you are right

sihyunjeon:
after chatting - we decided to stick to default scale for all samples

sihyunjeon:
Hi, 

> Since this is not p p, one needs to set lpp1 and lpp2 in the run cards to specify the type of beam. So for a p we have a run card with lpp1 = 2, and for p a we have a run card with lpp2 = 2. Analysers also requested to have the samples splitted depending on which direction the incoming photon comes.

This is not a correct statement. If you use QED PDF set and do "a p > X X" you don't need to change the beam type setting differently. Where did you get this message? I chatted with MadGraph authors themselves on this a while ago.

PR #3450, opened 2023-06-05
Fragments cards used for WHtoWhh and ZHtoZhh processes 

zhiyuanlcern:


zhiyuanlcern:
Hi @menglu21 , could you comment on the fragment card if the W/Z bosons are set to decay inclusively correctly? Thanks!

PR #3452, opened 2023-06-07
customized loop_sm model to generate ggZHH samples with non-SM kf, kv, kvv, k_lambda couplings

yihui-lai:
loop_sm_VHH is adapted from loop_sm model. 4 new external parameters are added corresponding to different coupling modifier .

2 files are modified:
parameters.py, where 4 new parameters are declared 
couplings.py, new parameters (coupling modifiers) are added in the appropriate position


menglu21:
Hi @yihui-lai have you tested with all those additional coupling value equal to 1, you can go back to default loop_sm

yihui-lai:
Hi Meng, yes when these coupling value set to 1, this model goes back to default loop_sm

yihui-lai:
I tried running with standalone madgraph and check the Gen level distributions. It's the same as default loop_sm. Do you suggest to run some events with full CMSSW production chain to make sure ?

menglu21:
then that's fine for me. @Saptaparna @bbilin @mlizzo @Dominic-Stafford please have a check

mlizzo:
Hi, cards look ok to me as well, I'm going to merge this request

PR #3458, opened 2023-06-12
Update the STARlight setup for HIN wmLHE

stahlleiton:
This PR updates the setup to produce tarballs for the STARlight generator to move to the use of wmLHE (instead of doing pLHE which has caused issues in past productions).

The following updates are included:
- Move to the latest version of the STARlight code (v 317, now obtained from the cms-generation website).
- compile the output to LHE converter C++ macro (convert_SL2LHE.cpp) when making the tarball (before was running the macro with ROOT during the LHE production, which was inefficient)
- Synchronize the runcmsgrid.sh to be ran with run_generic_tarball_cvmfs.sh during production (as done with the other generators).
- Add patches to fix compilation issues with the code as described in the manual  (before was directly replacing the files  instead of only applying the diff).
- Add a production input card directory.


menglu21:
Hi @stahlleiton have those updates been validated already, i.e. can you get the grid with this new method and able to get events with wmLHE request

stahlleiton:
I have tested it locally so far (i.e. made the tarball on lxplus and then use it on a  GEN-SIM step with cmsDriver). Not sure how test it on the central McM framework.

menglu21:
great, that's fine.

PR #3461, opened 2023-06-12
Add cards for individual VLQ decays for SuuToChiChi Fully Hadronic (mg265ul version) 

emcannaert:
To get around the problem we were seeing with the branching fractions being off with the full Suu->chi chi -> hadrons process, we decided it would make more sense to generate the subprocesses (Suu -> chi chi > {WbWb, htht, ZtZt, Wbht, WbZt, htZt} individually. This way we can scale to guarantee  branching fractions are what we want, and it ultimately makes it easier for us to do our NN training. The MC these cards will create will be UL16-UL18, so these cards have been created in a fork of the mg265UL branch, and I have made gridpacks from each subprocess and verified that these are working. No other aspect of the cards aside from the process has been changed.


menglu21:
Hi, do I understand correctly that same widths are used for different mass point

emcannaert:
Could you also tell me where I should be putting the gridpacks I'll get from these cards? I see information in this (older)  [twiki ](https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideSubgroupMC#1_1_Using_wmLHE_campaigns) that tells me to create  a directory for these processes in /store/group/phys_generator/cvmfs/gridpacks/slc7_amd64_gcc10/13TeV , but I do not have permission to create a directory here.

menglu21:
Hi, usually the MC contacts in each PAG/POG have permission to upload gridpacks.

emcannaert:
Understood, I have contacted them about this.

emcannaert:
Do you have other comments for this, or can it be merged?

PR #3466, opened 2023-06-15
Adding Run2 and Run3 0-40GeV Diphoton Sherpa Datacards

atsatsos97:
This PR is for adding the datacards of the Run 2 and Run 3 0-40 GeV diphoton background sample generated using Sherpa for official production.

### DIFF
--- a/bin/Sherpa/cards/production/13TeV/higgs/Run.dat_13TeV_gamgam_3j_loop_Mgg0-40
+++ b/bin/Sherpa/cards/production/13TeV/higgs/Run.dat_13TeV_gamgam_3j_loop_Mgg0-40
@@ -1,0 +1,35 @@
+ (run){
+ EVENTS 1000;
+ EVENT_MODE HepMC;
+ ME_SIGNAL_GENERATOR Comix Internal;
+ EVENT_GENERATION_MODE Unweighted;
+ BEAM_1 2212; BEAM_ENERGY_1 6500.;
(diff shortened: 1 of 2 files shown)


menglu21:
Hi @atsatsos97 the cards are already in the central pool, https://github.com/cms-sw/genproductions/blob/master/bin/Sherpa/cards/production/2017/13TeV/higgs/Run.dat_13TeV_gamgam_3j_loop_Mgg0-40 and https://github.com/cms-sw/genproductions/blob/master/bin/Sherpa/cards/production/13p6TeV/Run.dat_13p6TeV_gamgam_3j_loop_Mgg40-80 
I think they are the same

atsatsos97:
Hi @menglu21 the Run 3 link you post is not the same as the one I have - for one thing, the mass points are different.

As for the Run 2 sample, while it is the same, I would argue that keeping it in the 2017 folder gives the wrong impression for what I want from this sample. I would like this sample to be produced with 2018 conditions for my analysis.

menglu21:
ok, please put the run3 one in the same directory with the one I sent. There is no difference between 13TeV and 2017/13TeV (I guess this naming is following MG one). why are you using "NNPDF30_nnlo_as_0118", are you doing UL analysis.

sihyunjeon:
Hi Meng,
NNPDF30 was the tune used for sherpa internally by the authors, so i thought it would be better to use that one instead of NNPDF31.

PR #3470, opened 2023-06-23
Organizing single VLQ MG cards

acarvalh:
Dear 

There would be a lot of upcoming MG card updates to run3. In order to make them more easily foundable I suggest the following organization of folders


menglu21:
make sense, merged

PR #3477, opened 2023-06-29
Fix automated gridpack production for Powheg Box RES

lauridsj:
This PR fixes a problem in gridpacks for Powheg Box version RES which caused  the pre-sampling information, calculated during the gridpack production, to not be read in correctly. The result was that, when producing events, Powheg would hang for several hours while calculating the grid, only to end up with bad statistics since it only ran on one core. 

The cause for the bug is that Powheg vRES (in contrast to V2) only reads in information from previous stages when the option `manyseeds 1` is set. We use this option to compute the pre-sampling (`parallelstage 1-3`), but not for the actual event generation (`parallelstage 4`).

The fix consists of two parts:

a) When creating the final tarball, we make sure that the `manyseeds 1`,  `parallelstage 4` and `xgriditeration 1` options are set. `xgriditeration 1` is only needed because without it, Powheg will complain when `manyseeds 1` is set - its actual value is ignored.

b) As a consequence, the `runcmsgrid.sh` script has to be modified to run Powheg with the `manyseeds 1` option. This involves writing the RNG seed to the `pwgseeds.dat` file in addition to the `powheg.input` file, passing an additional command line argument to select said seed, and finally making sure the produced LHE file is at the correct spot for any post-processing.

I have checked that a working gridpack is produced with the `b_bbar_4l` process as an example. I also checked with the `hvq` process that Powheg Box V2 works the same as before.

In addition to this, an additional minor fix in `run_pwg_condor.py` is included which fixes the job index passed to Powheg during the pre-sampling in stages 2 and 3. For Powheg, indices start at 1 but so far, the first job had index 0, which caused this particular job to crash and thus one independent pre-sampling file less to be produced.

### DIFF
--- a/bin/Powheg/Templates/createTarBall_template.sh
+++ b/bin/Powheg/Templates/createTarBall_template.sh
@@ -43,8 +43,9 @@ ISVRES="false"
 grep -qi "powhegboxRES" $$WORKDIR/$$folderName/VERSION ; test $$? -ne 0  || ISVRES="true"
 # For Powheg vRES, if the gridpack has been produced in several stages (using manyseeds 1),
 # we need to turn on manyseeds & parallelstage also for the final gridpack.
-# we test whether manyseeds has been used by checking if "pwgubound-0001.dat" exists
-if [ $$ISVRES == "true" -a -f "$$WORKDIR/$$folderName/pwgubound-0001.dat" ]; then
+# we test whether manyseeds has been used by checking if there is more than one "pwgubound-XXXX.dat" file exists
+NUMUBOUND=`ls $${WORKDIR}/$${folderName} | egrep 'pwgubound-.+.dat' | wc -l`
+if [ $$ISVRES == "true" -a $$NUMUBOUND -gt 1 ]; then
   echo "Detected Powheg Box RES and a parallel gridpack production. Turning on manyseeds in final gridpack"
   sed -i "s/^.*manyseeds.*/manyseeds 1/g" powheg.input
   sed -i "s/^.*parallelstage.*/parallelstage 4/g" powheg.input
--- a/bin/Powheg/Templates/createTarBall_template.sh
+++ b/bin/Powheg/Templates/createTarBall_template.sh
@@ -39,10 +39,21 @@ sed -i "s/^numevts.*/numevts NEVENTS/g" powheg.input
 sed -i "s/^iseed.*/iseed SEED/g" powheg.input
 grep -q "withnegweights" powheg.input; test $$? -eq 0 || printf "\nwithnegweights 1\n" >> powheg.input
 
-# turn into single run mode
-sed -i "s/^manyseeds.*/#manyseeds 1/g" powheg.input
-sed -i "s/^parallelstage.*/#parallelstage 4/g" powheg.input
-sed -i "s/^xgriditeration/#xgriditeration 1/g" powheg.input
+ISVRES=`grep -i 'powhegboxRES' $$WORKDIR/$$folderName/VERSION`
+# For Powheg vRES, if the gridpack has been produced in several stages (using manyseeds 1),
+# we need to turn on manyseeds & parallelstage also for the final gridpack.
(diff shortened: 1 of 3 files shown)


bin/Powheg/Templates/createTarBall_template.sh
```diff
@@ -39,10 +39,22 @@ sed -i "s/^numevts.*/numevts NEVENTS/g" powheg.input
...
-# turn into single run mode
-sed -i "s/^manyseeds.*/#manyseeds 1/g" powheg.input
-sed -i "s/^parallelstage.*/#parallelstage 4/g" powheg.input
-sed -i "s/^xgriditeration/#xgriditeration 1/g" powheg.input
+ISVRES="false"
+grep -qi "powhegboxRES" $$WORKDIR/$$folderName/VERSION ; test $$? -ne 0  || ISVRES="true"
+# For Powheg vRES, if the gridpack has been produced in several stages (using manyseeds 1),
+# we need to turn on manyseeds & parallelstage also for the final gridpack.
+# we test whether manyseeds has been used by checking if "pwgubound-0001.dat" exists
+if [ $$ISVRES == "true" -a -f "$$WORKDIR/$$folderName/pwgubound-0001.dat" ]; then
```
agrohsje, line 47: we shouldn't require 0001 here but just make sure 00x is there. maybe we can use the logic from this pr: 
https://github.com/cms-sw/genproductions/pull/3415/files

bin/Powheg/run_pwg_condor.py
```diff
@@ -366,7 +366,7 @@ def runEvents(parstage, folderName, EOSfolder, njobs, powInputName, jobtag, proc
           f.write('tar xzf WW_MiNNLO_2loop_grids_reduced1.tar.gz\n')
           f.write('ls\n')
         f.write('cp -p ' + rootfolder + '/' + folderName + '/powheg.input.' + parstage + ' ./powheg.input' + '\n') # copy input file for this stage explicitly, needed by condor dag
-        f.write('echo ' + str (i) + ' | ./pwhg_main \n')
+        f.write('echo ' + str (i+1) + ' | ./pwhg_main \n')
```
agrohsje, line 369: what is the purpose of this modification ?

covarell:
It is really great that there is now a solution for the POWHEG-vRES issue. May I ask you to keep the two branches for now, and make also a backport PR on "powhegUL" for this? 

It is true they mostly differ by the PDFs, but there are a couple of things that needed to be changed for specific processes, to comply with the different compilers in CMSSW10 and CMSSW12. While the Run2UL campaign slows down, this branch will be used less and less, so at some point we can stop backporting. Thanks for your patience.

bin/Powheg/run_pwg_condor.py
lauridsj: See the last part of my initial comment:

> In addition to this, an additional minor fix in run_pwg_condor.py is included which fixes the job index passed to Powheg during the pre-sampling in stages 2 and 3. For Powheg, indices start at 1 but so far, the first job had index 0, which caused this particular job to crash and thus one independent pre-sampling file less to be produced.

It's not really related to the issue of vRES not reading the pre-sampling, but I stumbled upon it while checking the logs of each job during stages 1-3, so I thought I might as well fix it. :)

bin/Powheg/Templates/createTarBall_template.sh
lauridsj: I changed it to check whether more than one file matching `pwgubound-*.dat` exists. As far as I can tell, I cannot simply copy the logic from the PR you linked since that would also catch an `pwgubound.dat` file from a single-process production (without `manyseeds`).

lauridsj:
I made a backport of this PR to the powhegUL branch: #3478

bin/Powheg/run_pwg_condor.py
agrohsje: Excellent!

agrohsje:
Thanks Laurdis. Looks good to me.  @menglu21 @Saptaparna

Saptaparna:
+1

PR #3479, opened 2023-07-01
trim down pdfsets

sihyunjeon:
Removed several pdf sets for reasons

1. 320900 NNPDF31_nnlo_as_0118_nf_4 101 -> gone
This is the same PDF as 325500 but statistically treated differently when estimating PDF uncertainty. And we are taking hessian as the nominal way to estimate the uncertainty. Don't see a reason to keep this.

2. 292400 NNPDF30_nnlo_nf_4_pdfas 103 -> 1
I agree that keeping NNPDF30 could be helpful for some cases if we have to compare prelegacy 2016. But don't see why we should keep all error sets.

3. 13091 CT14nnlo_NF4 1 13191 CT14nlo_NF4 1 13202 CT14lo_NF4 1 -> gone
They are too old to be used.

4. 27870 MSHT20nnlo_as_smallrange_nf4 3 -> 1
Removed error sets as this is a minor version of secondary PDF set (as CMS is using NNPDF as nominal)

5. 316200 NNPDF31_nnlo_as_0118_mc 101 -> gone
Same as 1.

6. 305800 NNPDF31_nlo_hessian_pdfas 103 -> 1
Similar to 2.

7. 331300 NNPDF40_nnlo_pdfas 103 -> gone
Same as 1.

8. 332100 NNPDF40_nnlo_pch_as_01180 101 -> 1
Similar to 4.

9. 14100 CT18ZNNLO 59 -> 1
Similar to 4.


sihyunjeon:
I just tested a very rough case, storing only NNPDF31NNLO and it error sets + various scale choices in LHE files. Number of weights dropped to 150 from 900. And literally LHE file size dropped by 1/6 (tested with 1000 events, 33M -> 6M).

Trimming down PDF sets is not going to be a genuine fix for the I/O, size, nfiles problem but this will at least help the problem to be less occurring, especially for cases where we need to make additional LHE events for later filtering/merging purposes. So this has to be decided quickly.

sihyunjeon:
Also from GEN side, it needs to be decided on which PDF sets are going to be kept in NanoAOD for run3

https://github.com/cms-sw/cmssw/blob/master/PhysicsTools/NanoAOD/python/genWeightsTable_cfi.py

menglu21:
Hi @sihyunjeon we had some discussion with pdf forum people on this also, will work together to make the list more reasonable

sihyunjeon:
@bbilin @menglu21 @joonblee 

hi all, is there any progress with this discussion?

similarly related to powheg
https://github.com/cms-sw/genproductions/pull/3574

PR #3480, opened 2023-07-03
Large width Single Tprime production at LO, T ->tH for Run 2

kpal99:
Dears,
I am Kuldeep Pal, working in NISER, India. I along with my supervisor, Dr. Prolay Kumar Mal are working with B2G group. We would like to produce Large width samples for single tprime production at leading order, T further decays to top and Higgs.

We would like to produce sample using VLQ_v4_4FNS_UFO model for Run 2.
The mass point requested are: 700, 1000, 1500, 2000 GeV
and for each mass point four decay width are request: 5, 10, 20, 30% (written as pct in cards)
and for each set of mass point and decay; three different channels are specified: s-channel (sch), t-channel (tch) and interference between signal and SM backgroud (int).

We have presented our gen-level study in the B2G group, which can be access [here](https://indico.cern.ch/event/1297817/#10-production-of-large-width-v).

Kindly let us know if any changes are required. 
Cheers,


menglu21:
Hi @kpal99 before moving review, could you please trim the PR a bit, e.g., with template cards and script to implement all the different parameters.

kpal99:
Hi @menglu21 I have implemented the suggest changes to another [PR](https://github.com/cms-sw/genproductions/pull/3481).
Kindly proceed to the review and do let me know if anything is required.

PR #3487, opened 2023-07-13
added the script for the Fastjet DJR calculation in python

jobach18:


menglu21:
Hi @jobach18 we have several scripts to get the djr in the same directory, what's new in your script

jobach18:
Hi @menglu21, 
All of them just plot DJR values they get from the generator code (or private code). Since the DJRs are not saved in nanoaod format,  one needs to calculate them themselves. Which is what this script does: It calculates the DJR for a nanoaod output by CMSSW. 
We have found this important in the evaluation of herwig produced samples where we look at the merging. It's a draft PR anyway for now, since the results are not yet validated. 
Best
Joern

sihyunjeon:
@menglu21, markus @mseidel42 and I asked for this so that we can perhaps utilize it for nanogen datatiers as well for different generators (since djr comes from the pythia hook)

menglu21:
I see, do you already have some validation plots between using existing DJR value and using fastjet method. Another comment is that could you make the code more user friendly, e.g., I will not assume everyone has coffea and mplhep is only available in recent CMSSW version

bbilin:

PR #3490, opened 2023-07-18
Increased the threshold for the 5F/4F check in gridpack_generation.sh

Cvico:
Hi,

## Description
We have encountered an issue with some processes for which the 5F scheme is not properly settled
in the run card. The issue comes from the following check:

```
if tail -n 20 $LOGFILE | grep -q -e "^p *=.*b\~.*b" -e "^p *=.*b.*b\~"; then 
  is5FlavorScheme=1
fi
```

At that point, the `gridpack_generation.sh` script reads the last 20 lines of the logfile and searches for that line. It seems that searching for the last 20 lines is not enough for some cases, and because of this the process is considered as 4F. This is solved
by simply increasing the number of lines that are grepped with tail.

## Test bench
This has been tested in particular for ttW Run3 with MG29: with  `tail -n 20 $LOGFILE` the process was considered as 4F, with ` tail -n 999 $LOGFILE` the process is considered 5F.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
+++ b/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
@@ -339,7 +339,7 @@ make_gridpack () {
       fi
 	
       is5FlavorScheme=0
-      if tail -n 20 $LOGFILE | grep -q -e "^p *=.*b\~.*b" -e "^p *=.*b.*b\~"; then 
+      if tail -n 999 $LOGFILE | grep -q -e "^p *=.*b\~.*b" -e "^p *=.*b.*b\~"; then 
         is5FlavorScheme=1
       fi
     


sihyunjeon:
To provide some more context, this is mostly relevant to MG29X (=Run3) samples. We think how madgraph prints out the logs change a little bit and ended up checking the last 20 lines is not really enough.

This affects the physics pretty much since PDF sets and proton definition mismatches.

Carlos is working on some scripts that can be generically used on gridpacks so I think somebody should execute this for all RunIII gridpacks (a little bit more scripting is needed - fetching the gridpacks from all Run3 prepids and running carlos's scripts)

menglu21:
Hi @Cvico @sihyunjeon thanks for the information, do you have one example for the issue using MG29X

Cvico:
> 

Hi @menglu21, you can quickly check this by untaring the run_card of the centrla gridpack we have for ttlnu:

`tar -xvf /cvmfs/cms.cern.ch/phys_generator/gridpacks/PdmV/Run3Summer22/TTLNu-1Jets_amcatnloFXFX-pythia8_slc7_amd64_gcc10_CMSSW_12_4_8_tarball.tar.xz process/Cards/run_card.dat`

If you look there lhaid is set to 4F PDF (325500). If I do the same gridpack with the master branch, but changing
the lines above, I get 5F properly.

sihyunjeon:
@menglu21 Ah interesting! Maybe the first codegen step in cmsconnect makes some difference. But all in all, would it not be better just to check with `-n LARGENUMBER` and for all Run3Summer22 campaign gridpacks (since it's not really possible to figure out if it is lxplus or cmsconnect generated)?

menglu21:
Hi @sihyunjeon yes, the codegen step has independent output and makes the difference. I agree with you, merged.

PR #3491, opened 2023-07-19
adding madspin cards for the intermediary particle decay

gmilella12:


sihyunjeon:
Hi I guess you changed this as it bounced back from the official sample production submissions due to too large gridpacks right? the one you did won't really help since your number of feynman diagrams are still kept (actually increase instead as it's fully inclusive sample)

gmilella12:
> Hi I guess you changed this as it bounced back from the official sample production submissions due to too large gridpacks right? the change you made won't really help since your number of feynman diagrams are still kept (actually increase instead as it's fully inclusive sample)

Hi! 
the problem was not related to the large gridpack size but with the large number of files that the gridpack contains. 
There is a threshold of 15k files and this new madsping configuration leads to a gridpack that contains **6k** files. 
The gridpack size is similar to the one produced with the former configuration (~60 MB)

menglu21:
Hi @sihyunjeon with the initial syntax, the number of files is ~ 800K iirc.

sihyunjeon:
there is something totally fishy with madgraph. madspin is basically the same as madgraph and full_me grid should give same feynman diagrams eh weird ok

gmilella12:
> there is something totally fishy with madgraph. madspin is basically the same as madgraph and full_me grid should give same feynman diagrams eh weird ok

to clarify, we have switch off the ```spinmode``` ([here](https://github.com/cms-sw/genproductions/pull/3491/commits/71a48d4922e1c53ac3ab7591e20e5e206e64769b)) so we don't with top polarization

sihyunjeon:
ah so that makes sense for number of diagrams becoming small.

but is there a reason to keep madspin if that's the case? physics-wise this will not change anything when you use pythia8.
also (@menglu21 , this heavily heavily relies on madspin so in case if we drop concurrent run) it will be computationally super bad/slow.

matt-komm:
@sihyunjeon not sure I understand why it would be computationally worse compared to the full ME simulation; madspin should really speed up the generation and it is used heavily for that purpose in the generation of other (top) samples as well ...

sihyunjeon:
right, it's better than full ME when allocating 1 core.

this is more related to central production using N (2-4) cores at once. madspin is forced to use 1 core so after tttt is made from the first madevent step with N cores, your top decays will be handled with 1 core which will drop your computing efficiency and as this is 4 top that has to be handled, I expect quite huge drop. This statement (using madspin is bad in terms of computing efficiency) is true for ttbar as well but in that case we still want precise modeling for top decays so we live with it. But for your case I don't really see that much benefit compared to using pythia for top decays.

matt-komm:
ah I did not knew that the LHE/GEN step uses more than one core; is there a particular reason? I cannot imagine one can save a lot of RAM this way since there is not a lot of shared memory compared to other steps e.g SIM ...

sihyunjeon:
We run GEN and SIM step together (allocating N cores for both GEN and SIM step from the beginning). So there is not that much save in resources but inevitable. So I would suggest doing the decays pythia instead of spin correlation dropped madspin (again, physics wise they are equivalent).

PR #3507, opened 2023-08-25
Harmonise Run-3 cards for tHq and tHW with previous MR

FloMau:
Harmonise the Run-3 MG cards for tHq and tHW using what we had in the [previous, still open MR](https://github.com/cms-sw/genproductions/pull/3413/).

In particular, the nominal top Yukawa coupling is set to be positive here and the low, fixed scales for tHW are changed to dynamic scales. 

Tagging @menglu21

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/thq_4f_ckm_LO_ctcvcp/thq_4f_ckm_LO_ctcvcp_run_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/thq_4f_ckm_LO_ctcvcp/thq_4f_ckm_LO_ctcvcp_run_card.dat
@@ -60,7 +60,7 @@
   91.188	=  dsqrt_q2fact1     ! fixed fact scale for pdf1
   91.188	=  dsqrt_q2fact2     ! fixed fact scale for pdf2
   -1	=  dynamical_scale_choice  ! Select one of the preselect dynamical choice
-  0.33 =  scalefact         ! scale factor for event-by-event scales
+  1 =  scalefact         ! scale factor for event-by-event scales
 #*********************************************************************
 # Matching - Warning! ickkw > 1 is still beta
 #*********************************************************************
@@ -70,7 +70,7 @@
   1	=  alpsfact          ! scale factor for QCD emission vx
   False	=  chcluster         ! cluster only according to channel diag
   True	=  pdfwgt            ! for ickkw=1, perform pdf reweighting
-  5	=  asrwgtflavor      ! highest quark flavor for a_s reweight
+  4	=  asrwgtflavor      ! highest quark flavor for a_s reweight
   True	=  clusinfo          ! include clustering tag in output
   3.0	=  lhe_version        ! Change the way clustering information pass to shower.        
 #*********************************************************************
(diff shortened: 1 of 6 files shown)


bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/thq_4f_ckm_LO_ctcvcp/thq_4f_ckm_LO_ctcvcp_run_card.dat
```diff
@@ -56,7 +70,7 @@
   1	=  alpsfact          ! scale factor for QCD emission vx
   False	=  chcluster         ! cluster only according to channel diag
   True	=  pdfwgt            ! for ickkw=1, perform pdf reweighting
-  4	=  asrwgtflavor      ! highest quark flavor for a_s reweight
+  5	=  asrwgtflavor      ! highest quark flavor for a_s reweight
```
menglu21, line 73: although this block will not be called due to no matching, it should be 4 in 4 flavor scheme. btw, line 63, "0.33 =  scalefact" is set deliberately, right?

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/thw_5f_ckm_LO_ctcvcp/thw_5f_ckm_LO_ctcvcp_run_card.dat
```diff
@@ -40,8 +54,8 @@
 #*********************************************************************
 # Renormalization and factorization scales                           *
 #*********************************************************************
-  True	=  fixed_ren_scale   ! if .true. use fixed ren scale
-  True	=  fixed_fac_scale   ! if .true. use fixed fac scale
+  False	=  fixed_ren_scale   ! if .true. use fixed ren scale
+  False	=  fixed_fac_scale   ! if .true. use fixed fac scale
```
menglu21, line 58: do you want the default setup, cuz you are using "-1	=  dynamical_scale_choice"

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/thq_4f_ckm_LO_ctcvcp/thq_4f_ckm_LO_ctcvcp_run_card.dat
FloMau: Hi, 

thanks for your check. I updated it to `4	=  asrwgtflavor` and set the `scalefact` to 1. It was by mistake used from the previous version.

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/thw_5f_ckm_LO_ctcvcp/thw_5f_ckm_LO_ctcvcp_run_card.dat
FloMau: I would like to have the default setup, isn't `-1 = dynamical_scale_choice` the default? 
It is also chosen in the tHq run card.

menglu21: -1 is the default setup

PR #3516, opened 2023-09-18
WGToLNuG 13.6 TeV MC Sample

phazarik:
Dear expert,

I prepared MadGraph cards for the WGToLNuG_01J_5f_NLO sample at 13.6 TeV for RunIII production.
Only the beam energy is chnaged in the run_card.
Please take a look and inform us in case of any issues.

Thanks and regards,
Prachurjya Hazarika (EGM MC contact)
Adding @mobassirameen


sihyunjeon:
Hi,

WG is being produced in pt gamma bins (10-100, 100-200, 200-)
If EGM is fine with these I don't think you need to work on this yourself.
Plan to submit this once PTG-200 is finished

https://cms-pdmv-prod.web.cern.ch/gridpack/

menglu21:
close it since the sample is handled centrally

PR #3520, opened 2023-09-22
adding Mass and Width parameters to be used for high mass points in Run3 samples (HZZ)

neharawal14:
These are the list of Mass and Width parameters to be used for Higgs samples at high mass points in Run3 samples (HZZ). 
We will use these parameters in our same input cards (which are already approved).


menglu21:
Hi @neharawal14 could you please add some references for those mass+width

PR #3523, opened 2023-09-27
Run 3 generic Spin-0 and Spin-2 X->HH cards

portalesHEP:
Hello,

This PR contains example cards for the generation of "generic" (narrow) X->HH samples, without decays of the H, to be used for the Run 3 sample production for several B2G analyses (X->HH-> 4b, bbgamgam, bbtautau, bbWW, ...). Simple scripts are included to generate the cards for the fixed list of mass points agreed upon by B2G Diboson & MC&Int conveners.

The cards themselves are effectively copies of the Run 2 cards ([Spin-0](https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/exo_diboson/Spin-0/Radion_hh_narrow/Radion_hh_narrow_M900), [Spin-2](https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/exo_diboson/Spin-2/RSGraviton_hh_narrow/RSGraviton_hh_narrow_M900)), with the CoM energy updated in the "run" cards.

@acarvalh


bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/HHresonant/Spin-0/Radion_hh_narrow_M900_proc_card.dat
```diff
@@ -0,0 +1,8 @@
+set group_subprocesses Auto 
+set ignore_six_quark_processes False 
+set gauge unitary 
+set complex_mass_scheme False 
+import model heft_radion 
+define p = g u c d s u~ c~ d~ s~ 
+generate p p > h2, ( h2 > H H )
```
menglu21, line 7: here you require h2 to be on-shell, this is different from the Run2 setup "generate p p > h2 > H H", it's already discussed?

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/HHresonant/Spin-0/getAllMassPoints.py
```diff
@@ -0,0 +1,51 @@
...
+        tgt_file = outdir + "/" + src_file.replace('M900','M{}'.format(X_m))
+        
+        print('{} -> {}'.format(src_file,tgt_file))
+        
+        src_txt = open(src_file, 'r').read()
+
+        if 'proc_card' in card:
+            tgt_txt = src_txt.replace('900', str(X_m))
+        elif 'customizecards' in card:
+            tgt_txt = src_txt.replace('900', str(X_m))
```
menglu21, line 43: how is the width modified for each mass point? please refer to https://github.com/cms-sw/genproductions/blob/d81b75a6f5c0ea44b6777a04e365c5290689402f/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/b2g_dihiggs/genGridpack_diboson.py

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/HHresonant/Spin-2/BulkGraviton_hh_GF_HH_narrow_M900_proc_card.dat
```diff
@@ -0,0 +1,8 @@
+set group_subprocesses Auto 
+set ignore_six_quark_processes False 
+set gauge unitary 
+set complex_mass_scheme False 
+import model RS_bulk_ktilda 
+define p = g u c d s u~ c~ d~ s~ 
+generate p p > y, ( y > H H )
```
menglu21, line 7: same question here

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/HHresonant/Spin-2/getAllMassPoints.py
```diff
@@ -0,0 +1,51 @@
...
+        tgt_file = outdir + "/" + src_file.replace('M900','M{}'.format(X_m))
+        
+        print('{} -> {}'.format(src_file,tgt_file))
+        
+        src_txt = open(src_file, 'r').read()
+
+        if 'proc_card' in card:
+            tgt_txt = src_txt.replace('900', str(X_m))
+        elif 'customizecards' in card:
+            tgt_txt = src_txt.replace('900', str(X_m))
```
menglu21, line 43: same question on the width

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/HHresonant/Spin-0/Radion_hh_narrow_M900_proc_card.dat
portalesHEP: There may have been different versions of spin-0/2 X->HH gridpacks generated for Run 2. AFAIK the on-shell syntax is the one that was used by the analyses formerly part of the HIG PAG.

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/HHresonant/Spin-0/getAllMassPoints.py
portalesHEP: It is not, the plan here is to consider only narrow resonances. The width would be set to `1.000000e-03 ` for all mass points.

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/HHresonant/Spin-2/BulkGraviton_hh_GF_HH_narrow_M900_proc_card.dat
portalesHEP: see answer above

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/HHresonant/Spin-2/getAllMassPoints.py
portalesHEP: see answer above

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/HHresonant/Spin-0/Radion_hh_narrow_M900_proc_card.dat
portalesHEP: @menglu21 since we are assuming NWA, the two syntax should be equivalent, but if you think `p p > h2 > H H` would be better (maybe for "generality" if one wants to introduce width later on), I'm happy to make the change.

menglu21: ok, It's fine generally, my concern is for the mass point M=250, whether these two syntaxs will be different or not, could you please have a check on that

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/HHresonant/Spin-0/getAllMassPoints.py
menglu21: ok, in the case of very small width corresponding to the mass, the choice of width should not matter

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/HHresonant/Spin-0/Radion_hh_narrow_M900_proc_card.dat
portalesHEP: image
There is indeed some very small difference here (mX histograms for Spin-2 resonances), but for all practical purpose it should be negligible

menglu21: thanks, I'll merge this PR

PR #3526, opened 2023-09-27
Positive/negative tag changed

antoniocotar:


menglu21:
Hi @antoniocotar there is no madspin card in the master branch, please correct it first

antoniocotar:
Hi @menglu21 thanks, I think now I solved the conflicts, do I need to do something more?

PR #3528, opened 2023-09-30
adding cards for VHNobb Run3 Sample

sv3048:
Dear experts !
 
I have prepared madgraph cards for VHTonobb sample. The only change is the energy from 13 to 13.6 TeV. Kindly have a look at it at your convenience.

Thanks and Regards,
Sadhana Verma
adding @amanagrawal1280


menglu21:
Hi @sv3048 where do you define the H->Nonbb, I don't see it in the madspin card. Another issue is that you include too many processes in the card, and use madspin to decay Z/W in all channel. It's no doubt that the final gridpack will have large number of files which is not favored by computing, as well as the large size of gridpack. So I would suggest to separate the process, at least you can separate them to V->qq and V->ll

PR #3531, opened 2023-10-01
Dedicated MG5 background samples for H--> mumu analysis

Soumyatifr:
Two new cards have been included for the background processes dedicatedly prepared for the H--> mu mu analysis for the Run-3 13.6 TeV conditions.  
(1) DY mu mu 0/1/2 jets NLO samples in the mass window of  105-160 GeV region, mostly 125 +/- 20 GeV region of the Higgs mass. 
(2) EWK ll + 2jet LO sample, with the di-lepton mass window of 105-160 GeV region, additionally an invariant mass of the two VBF jets mqq > 120 GeV applied. This sample is an irreducible background for the analysis of VBF H --> mums analysis. 

The validity of the cards have already been checked by producing gridpack and then LHE events and also by preparing the basic validation plots of the di-muon/lepton invariant mass. 

LHE_mmumu

On behalf of the H to mums analysis team, I am requesting to review the cards of the process and approve it for the sample production. 
--
Soumya

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/LLJJ_105-160_EWK_SM_5f_LO/LLJJ_105-160_EWK_SM_5f_LO_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/LLJJ_105-160_EWK_SM_5f_LO/LLJJ_105-160_EWK_SM_5f_LO_proc_card.dat
@@ -1,0 +1,19 @@
+set group_subprocesses Auto
+set ignore_six_quark_processes False
+set loop_color_flows False
+set gauge unitary
+set complex_mass_scheme False
+set max_npoint_for_channel 0
+import model sm-ckm_no_b_mass
+define p = g u c d s u~ c~ d~ s~
+define j = g u c d s u~ c~ d~ s~
+define l+ = e+ mu+
+define l- = e- mu-
+define p = p b b~
+define j = j b b~
+define l+ = e+ mu+ ta+
+define l- = e- mu- ta-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
+generate p p > l+ l- j j / t t~ h QED=4 QCD=0
+output LLJJ_105-160_EWK_SM_5f_LO -nojpeg
--- a/bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/LLJJ_105-160_EWK_SM_5f_LO/LLJJ_105-160_EWK_SM_5f_LO_run_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/LLJJ_105-160_EWK_SM_5f_LO/LLJJ_105-160_EWK_SM_5f_LO_run_card.dat
@@ -1,0 +1,272 @@
+#*********************************************************************
+#                       MadGraph5_aMC@NLO                            *
+#                                                                    *
+#                     run_card.dat MadEvent                          *
+#                                                                    *
+#  This file is used to set the parameters of the run.               *
+#                                                                    *
+#  Some notation/conventions:                                        *
+#                                                                    *
+#   Lines starting with a '# ' are info or comments                  *
+#                                                                    *
+#   mind the format:   value    = variable     ! comment             *
+#*********************************************************************
+#
+#*******************                                                 
+# Running parameters
+#*******************                                                 
+#                                                                    
+#*********************************************************************
+# Tag name for the run (one word)                                    *
+#*********************************************************************
+  tag_1	=  run_tag  ! name of the run 
+#*********************************************************************
+# Run to generate the grid pack                                      *
+#*********************************************************************
+  True	=  gridpack   !True = setting up the grid pack
+#*********************************************************************
+# Number of events and rnd seed                                      *
+# Warning: Do not generate more than 1M events in a single run       *
+# If you want to run Pythia, avoid more than 50k events in a run.    *
+#*********************************************************************
+  5000	=  nevents  ! Number of unweighted events requested 
+  0	=  iseed    ! rnd seed (0=assigned automatically=default))
+#*********************************************************************
+# Collider type and energy                                           *
+# lpp: 0=No PDF, 1=proton, -1=antiproton, 2=photon from proton,      *
+#                                         3=photon from electron     *
+#*********************************************************************
+  1	=  lpp1     ! beam 1 type 
+  1	=  lpp2     ! beam 2 type
+  6800.0	=  ebeam1   ! beam 1 total energy in GeV
+  6800.0	=  ebeam2   ! beam 2 total energy in GeV
+#*********************************************************************
+# Beam polarization from -100 (left-handed) to 100 (right-handed)    *
+#*********************************************************************
+  0.0	=  polbeam1  ! beam polarization for beam 1
+  0.0	=  polbeam2  ! beam polarization for beam 2
+#*********************************************************************
+# PDF CHOICE: this automatically fixes also alpha_s and its evol.    *
+#*********************************************************************
+  'lhapdf'	=  pdlabel      ! PDF set                                     
+  $DEFAULT_PDF_SETS = lhaid
+  $DEFAULT_PDF_MEMBERS  = reweight_PDF
+#*********************************************************************
+# Renormalization and factorization scales                           *
+#*********************************************************************
+  False	= fixed_ren_scale ! if .true. use fixed ren scale
+  False	= fixed_fac_scale ! if .true. use fixed fac scale
+  91.188	= scale ! fixed ren scale
+  91.188	= dsqrt_q2fact1 ! fixed fact scale for pdf1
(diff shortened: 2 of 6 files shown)


menglu21:
Hi @Soumyatifr the EWK samples are already submitted with Mll>50 and Mjj>100 https://cms-pdmv.cern.ch/mcm/requests?prepid=GEN-Run3Summer22EEwmLHEGS-00201&page=0&shown=4398046773375, also the FXFX DY samples are ready, why not directly use the mumu events inside.
For the EWK sample, if you think the official sample doesn't have enough events inyou phase space, could you please extract the inputs cards from /cvmfs/cms.cern.ch/phys_generator/gridpacks/PdmV/Run3Summer22/VBFto2L_MLL-50_madgraph-pythia8_slc7_amd64_gcc10_CMSSW_12_4_8_tarball.tar.xz, and modify the setup correspondingly

scooperstein:
Hi @menglu21 , all - these samples are meant to target the H->mumu region where the dimuon mass pair is near 125 GeV. To give an idea, the Z+jets background (both QCD and EWK) cross section in the region 105 < m(ll) < 160 GeV is 2-3 orders of magnitude less than the inclusive cross sections. Therefore dedicated QCD and EWK Z+jets samples with a generator-level filter on m(ll) from [105, 160] GeV are very important to maximize the MC statistics in the H->mumu analysis. Similar samples were produced for the Run-2 analysis, and made possible the critical new analysis strategy in the VBF channel. We therefore would like to request these samples as Soumya prepared so that we can begin the Run-3 H->mumu effort. Could you please let us know what is different about the VBFto2L sample you linked compared to the "VBFZ" samples we used in Run-2?

Thank you very much,
Stephane

menglu21:
Hi @scooperstein, the syntax used for VBFto2L is "generate p p > ell+ ell- j j QCD=0 QED=4" using 5f, which is equivalent with the syntax here. Regard to the QCD Z+jet, I'm fine to approve those cards, my suggestion is to extract the input cards of gridpack (centrally produced) in official  FXFX DY samples and modify the mass cut correspondingly, in order to be consistent with each other. btw, before the samples are ready, I think you can also use the powheg MLL-binned samples, e.g.,  https://cms-pdmv-prod.web.cern.ch/grasp/samples?dataset_query=DYto2Mu_MLL-50to120_TuneCP5_13p6TeV_powheg-pythia8&campaign=Phase2Spring23*GS,Run3Summer22*GS,Run3Summer23*GS,Run3Winter22*GS,Run3Winter23*GS,RunIISummer20UL16*GEN,RunIISummer20UL16*GENAPV,RunIISummer20UL17*GEN,RunIISummer20UL18*GEN

sihyunjeon:
> (2) EWK ll + 2jet LO sample, with the di-lepton mass window of 105-160 GeV region, additionally an invariant mass of the two VBF jets mqq > 120 GeV applied. This sample is an irreducible background for the analysis of VBF H --> mums analysis.

I can't agree that this is really needed.

For the DY process, cross section already being too large + Z onshell populating statistics limits the statistics for your desired region -> So (1) could be worth it.

We already have mjj > 100 gev and mll > 50gev sample that is statistically abundant IIRC even for offshell Z. Cross section was 8 pb, so 8 pb x 50 fb-1 x 1000 = 0.4M events. Even for 50fb-1 lumi, 0.4M event is enough and we submitted 16M for the whole 22 (including 22EE) so weight per event is already less than 0.1.

https://cms-pdmv.cern.ch/mcm/requests?dataset_name=VBFto2L_MLL-50_TuneCP5_13p6TeV_madgraph-pythia8&page=0&shown=4398046511231

rgerosa:
> Hi @scooperstein, the syntax used for VBFto2L is "generate p p > ell+ ell- j j QCD=0 QED=4" using 5f, which is equivalent with the syntax here. Regard to the QCD Z+jet, I'm fine to approve those cards, my suggestion is to extract the input cards of gridpack (centrally produced) in official FXFX DY samples and modify the mass cut correspondingly, in order to be consistent with each other. btw, before the samples are ready, I think you can also use the powheg MLL-binned samples, e.g., https://cms-pdmv-prod.web.cern.ch/grasp/samples?dataset_query=DYto2Mu_MLL-50to120_TuneCP5_13p6TeV_powheg-pythia8&campaign=Phase2Spring23*GS,Run3Summer22*GS,Run3Summer23*GS,Run3Winter22*GS,Run3Winter23*GS,RunIISummer20UL16*GEN,RunIISummer20UL16*GENAPV,RunIISummer20UL17*GEN,RunIISummer20UL18*GEN

@menglu21 thanks! About using the POWHEG MLL samples, I agree that is an interesting set of alternative events that can be considered to assign uncertainties on the background (if not used to extract the central value of the bkg estimate). However, are these MLL samples from a Powheg NLO fixed-order? If so, after applying 2-jet selections, the description of aMC@NLO FxFx multi-jet merged 0,1,2 jet is expected to be better ...

rgerosa:
> > (2) EWK ll + 2jet LO sample, with the di-lepton mass window of 105-160 GeV region, additionally an invariant mass of the two VBF jets mqq > 120 GeV applied. This sample is an irreducible background for the analysis of VBF H --> mums analysis.
> 
> I can't agree that this is really needed (separately, other than central sample).
> 
> For the DY process, cross section already being too large + Z onshell populating statistics limits the statistics for your desired region -> So (1) could be worth it.
> 
> We already have mjj > 100 gev and mll > 50gev sample that is statistically abundant IIRC even for offshell Z. Cross section was 8 pb, so 8 pb x 50 fb-1 x 1000 = 0.4M events. Even for 50fb-1 lumi, 0.4M event is enough and we submitted 16M for the whole 22 (including 22EE) so weight per event is already less than 0.1.
> 
> https://cms-pdmv.cern.ch/mcm/requests?dataset_name=VBFto2L_MLL-50_TuneCP5_13p6TeV_madgraph-pythia8&page=0&shown=4398046511231

@sihyunjeon has some analyses team in CMS requested EWK-Z samples showed with Herwig angular-ordered shower instead of Pythia dipole recoil? Having an identical samples shower with Herwig will be quite important to assess parton-shower related uncertainties (perhaps also a MG+Vincia sample would be great). Coming back to the original request, having a dedicated Z-EW sample for Hmm would be important to have extra-events for DNN trainings. However, if we cannot obtain a dedicated sample, can we perhaps increase by a factor two the statistics of the current sample via an extension?

sihyunjeon:
Not sure if we have Vincia setup ready. Herwig we can perhaps do but no one yet requested the samples (honestly I don't know how to handle properly and neither do @agrohsje nor @Cvico). If we get inputs, we can do it easily.

More statistics is up to @menglu21 @bbilin

scooperstein:
Hi @menglu21 , @sihyunjeon , thanks for the feedback. Sure no problem, we can clone the existing gridpacks and adjust those directly, as you suggest. Just too make sure we are 100% in line with the existing samples. 

While I realize that the effective statistics of the existing EWK Z sample is already quite good, even off-shell, I think there is a very strong physics case to produce a limited number (say ~1M) events of m(ll) [105, 160] GeV EWK Z MC anyway to give the Run-3 H->mumu authors full leverage to exploit the new simulation-based approach to the VBF H->mumu channel. In the high-S/B region of the VBF analysis from Run-3 there are O(10) data events - indeed this is why switching from the data-driven to the simulation-based background prediction was important. With O(1M) events produced for this sample we could improve the effective statistics of the EWK Z in this channel by more than an order of magnitude, giving potentially more leverage to design the analysis. It seems like a well justified tradeoff to me. If you still do not agree then we are happy to include the HIG PAG L2 conveners to get their opinion.

Thank you for bringing to our attention that no Herwig-showered EWK Z sample has been requested. This is an oversight, as having alternative herwig-showered MC samples for both the EWK Z and VBF H processes is very important for CMS Run-3 VBF analyses in general (not only H->mumu). Can you please also tell us if the VBF H MC requested so far are showered with herwig as well, or just Pythia? For the Pythia sample, can you please confirm that the dipole recoil scheme is used rather than the default parton shower settings? I realize that this discussion on the alternative PS with herwig perhaps goes beyond the scope of this thread, but it is great that you brought this to our attention and this is something that will be very important for us to work together to fix, to make sure that there are not urgent requests later for these samples from the authors.

Best,
Stephane

sihyunjeon:
for VBF H, it has to be asked to HIG MC contacts, central samples don't include higgs processes

regarding VBF Z, is it about 1M (for whole 2022) that you require in that mll 105-160? I am not sure how that O(1M) will give more effective statistics. I just quickly ran and the ratio of cross sections between mll 105-160 and mll 50 is larger than 1/10 (different PDF set was used but the ratio of cross sections should more or less be the same). So I don't see any leveraging with that scale of nevents (unless it's magnitude larger than what you mentioned), you will likely get more than 1M from the samples that have been already produced. how did you get your numbers?

menglu21:
@qyguo could you please comment on the statistics you need for these two samples. The cards are generally fine to me, but I would suggest you to sync the parameters here with the central produced one, as we mentioned above

menglu21:
Close this PR, @qyguo will make two new PRs

PR #3536, opened 2023-10-04
Update for `request_fragment_check.py` script

### LINKED ISSUE #3535: Update for `request_fragment_check.py` script
Related to CERN SSO [migration](https://cern.service-now.com/service-portal?id=outage&n=OTG0072195) for PdmV applications, in particular for McM, it is required to update the fragment check script to use the new McM client to handle requests for this application. This script, managed by the CMS GEN team, is downloaded and embedded into McM validation batch jobs and used by McM users who want to manually run the “Get test” command for a McM request. Before migrating McM to be serviced only by the new SSO, it is required to update this dependency.

### Is your feature related to a problem?

If this script continues using the old SSO, PdmV will not be able to migrate the McM application without breaking the validation component. Furthermore, after the “old” CERN SSO is decommissioned, some sections of the script will break because it will not be able to consume McM's protected resources.

### Describe the solution you'd like

Refactor the script to use the [new](https://github.com/cms-PdmV/mcm_scripts/blob/b2c38b86337ce785b247edb4d5b90e0fec3d6a15/rest.py#L40) McM client and avoid consuming protected resources that require authentication. Use only the McM public API.

## Current behavior

The `request_fragment_check.py` script uses two instances of the [“old”](https://github.com/cms-PdmV/mcm_scripts/blob/f3c3daf5f817b40c0c4e6c317ca540a439a707ce/rest.py#L36) McM client ([mcm](https://github.com/cms-sw/genproductions/blob/master/bin/utils/request_fragment_check.py#L54), [mcm2](https://github.com/cms-sw/genproductions/blob/master/bin/utils/request_fragment_check.py#L55)) to perform queries to that application. `mcm2` requires a session cookie to consume only [this](https://github.com/cms-sw/genproductions/blob/master/bin/utils/request_fragment_check.py#L88) resource, for the remaining operations related to McM only the public API is used via `mcm`. The session cookie is manually requested via the shell execution of `cern-get-sso-cookie`

## Expected behavior

`request_fragment_check.py` script avoids handling any authentication process by itself and uses only the features provided by the McM client. Also, all the data from McM required by this script is retrieved only from the public API.  This deletes the dependency of handling two instances of the client and avoids issues related to authentication like the one described [here](https://github.com/cms-PdmV/mcm_scripts/issues/15).

ggonzr:
Fixes #3535 

### Status

Tested.

This update includes unit tests to verify that the new McM client works properly with the fragment script. Moreover, a little test was carried out in the McM production environment using this version to check that it integrates properly with this application. The artifacts related to the execution of the last validation test in McM production are available at: `/afs/cern.ch/user/p/pdmvserv/public/mcm/gen_script_update/PPD-Run3Summer23GS-00005`

### Is it backward compatible (if not, which system it affect?)

YES

The new version of the McM client supports Python interpreters between versions [2.7, 3.11]. Moreover, the update for the script `request_fragment_check.py` deletes dependencies on external host packages to handle authentication so no backward compatibility issues are expected.

### External dependencies / deployment changes

- McM QA client

This client is available at: `/afs/cern.ch/cms/PPD/PdmV/tools/McM-QA/`. This artifact includes all the changes available until this [commit](https://github.com/cms-PdmV/mcm_scripts/blob/b2c38b86337ce785b247edb4d5b90e0fec3d6a15/rest.py#L40).

### DIFF
--- a/bin/utils/request_fragment_check.py
+++ b/bin/utils/request_fragment_check.py
@@ -48,7 +48,7 @@ if args.develop is False:
        sys.exit()
 	
 # Use no-id as identification mode in order not to use a SSO cookie
-mcm = McM(id=McM.SSO, dev=args.dev, debug=args.debug)
+mcm = McM(id=None, dev=args.dev, debug=args.debug)
 mcm_link = mcm.server
 
 def get_request(prepid):
@@ -78,11 +78,12 @@ def get_ticket(prepid):
     return result
 
 def get_requests_from_datasetname(dn):
-    result = mcm.get('requests', query='dataset_name=%s' % (dn))
-    if not result:
-        return {}
-
-    return result
+    raw_result = mcm._McM__get(
+        "public/restapi/requests/from_dataset_name/%s" % (dn)
+    )
+    if not raw_result:
+        return []
+    return raw_result.get("results", [])
 
 def find_file(dir_path,patt):
     for root, dirs, files in os.walk(dir_path):
--- a/bin/utils/request_fragment_check.py
+++ b/bin/utils/request_fragment_check.py
@@ -9,11 +9,8 @@ import os.path
 import string
 import glob
 from datetime import datetime
-###########Needed to check for ultra-legacy sample consistency check############################################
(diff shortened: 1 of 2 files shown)


efeyazgan:
This looks OK to me. @bbilin @menglu21 please merge if you think this is OK as well.

PR #3537, opened 2023-10-04
Split HH->4V and HH->2T2V fragments into two

ktht:
The proposal is described [here](https://gitlab.cern.ch/cms-b2g/b-2-g-m-csample-requests/-/issues/4#note_7174353) with technical details and analysis available [here](https://cernbox.cern.ch/s/ds0KJ492iTN58Q5). I'll repeat the main points:
* split `HHto4V` into two samples (40% and 60%, respectively):
    * `HHto4Wto2L2Nu4Q` -- exactly two leptons ($e/\mu/\tau$), two neutrinos and four quarks in final state, all coming from W decays;
    * `HHto4Vto3Lplus` -- three or more leptons in final state, including those from Z decays;
* split `HHto2T2V` into two samples (20% and 80%, respectively):
    * `HHto2T2V_FullyLeptonic` -- both vector bosons (W or Z) decay into leptons;
    * `HHto2T2V_SemiHadronic` -- the two vector bosons produce exactly two quarks in final state.

These changes would cover phase space that is targeted by HH multilepton analysis almost perfectly, yet would respectively gain us a factor of ~5~ 3 and 2 in effective event statistics. For example, if we previously requested, say, 1M events for `HHto4V`, then now we'd request ~400k~ 700k events for `HHto4Wto2L2Nu4Q` and another ~600k~ 300k events for `HHto4Vto3Lplus` so that the event weights would be approximately equal in both samples.

I removed a duplicate fragment (`HHto2V2Tau`) and added the following:
* `*_TuneCP5_13p6TeV_powheg-pythia8.py` for nonresonant NLO ggHH production;
* `*_TuneCP5_13p6TeV_madgraph-pythia8.py` for resonant LO ggHH production;
* `*_dipoleRecoilOn-TuneCP5_13p6TeV_madgraph-pythia8.py` for resonant LO qqHH production.


acarvalh:
Hi @ktht 

thank you for that!
Could you add full hadronic V ones and an inclusive one for the options?

For the HHto4Wto2L2Nu4Q: why this is not 2W2V?

genfragments/ThirteenPointSixTeV/Higgs/HH/HHto4Vto3Lplus_TuneCP5_13p6TeV_madgraph-pythia8.py
```diff
@@ -0,0 +1,37 @@
...
+        pythia8CommonSettingsBlock,
+        pythia8CP5SettingsBlock,
+        pythia8PSweightsSettingsBlock,
+        processParameters = cms.vstring(
+            '23:mMin = 0.05',
+            '24:mMin = 0.05',
+            '25:m0 = 125.0',
+            '25:onMode = off',
+            '25:onIfMatch = 24 -24',
+            '25:onIfMatch = 23 23',
```
menglu21, line 22: add
'23:onMode = off',
'23:onIfAny = 11 13 15'
for W and Z should have high filter efficiency

genfragments/ThirteenPointSixTeV/Higgs/HH/HHto4Vto3Lplus_TuneCP5_13p6TeV_powheg-pythia8.py
```diff
@@ -0,0 +1,41 @@
...
+        pythia8PSweightsSettingsBlock,
+        pythia8PowhegEmissionVetoSettingsBlock,
+        processParameters = cms.vstring(
+            'POWHEG:nFinal = 2',
+            '23:mMin = 0.05',
+            '24:mMin = 0.05',
+            '25:m0 = 125.0',
+            '25:onMode = off',
+            '25:onIfMatch = 24 -24',
+            '25:onIfMatch = 23 23',
```
menglu21, line 25: same comment as above

genfragments/ThirteenPointSixTeV/Higgs/HH/HHto4Vto3Lplus_dipoleRecoilOn-TuneCP5_13p6TeV_madgraph-pythia8.py
```diff
@@ -0,0 +1,38 @@
...
+        pythia8CommonSettingsBlock,
+        pythia8CP5SettingsBlock,
+        pythia8PSweightsSettingsBlock,
+        processParameters = cms.vstring(
+            '23:mMin = 0.05',
+            '24:mMin = 0.05',
+            '25:m0 = 125.0',
+            '25:onMode = off',
+            '25:onIfMatch = 24 -24',
+            '25:onIfMatch = 23 23',
```
menglu21, line 22: same comment as above

ktht:
Hi @acarvalh, thank you for the input. I've now commited fragments for the inclusive resonant LO production; the rest was already there.

Your questions reminded me a recent discussion with @tolange about the idea of exploring $2\ell$ OS final states. The fragments I provided earlier were developed before this discussion and hence omit $2\ell$ OS final states. It prompted me to scrutinize the branching ratios more carefully and I found that the BR of $HH\to4W\to2\ell2\nu4j$ was off by a combinatorial factor of 3. After accounting for that mistake in my initial estimate, the gain in effective event statistics would be a factor of 3, not a factor of 5 as I previously claimed. I've corrected my original post and the notebok that I refer to.

In [another notebook](https://cernbox.cern.ch/s/1NxyzV8kAYewuSv) I show that extending to $2\ell$ OS final states would reduce the gain in event statistics from 3 to 2.7. Still, I think it's worth the trouble to apply the filter given that we're limited by signal statistics. I've replaced `HHto4Wto2L2Nu4Q` and `HHto4Vto3Lplus` fragments with a single fragment named `HHto4Vto2Lplus`. There is no point in filtering $HH\to2\tau2V$ events because the $\tau$ leptons alone can produce $2\ell$ OS signature. I thus removed the `HHto2T2V_FullyLeptonic` and `HHto2T2V_SemiHadronic` fragments.

Regarding the fully hadronic samples, I'd ike some input from @anmalara before commiting anything here. It's not possible to design a single set of decay rules that fully complement $HH\to4V\to2\ell+\mathrm{anything}$. As I understand, you're only interested in $HH\to4W\to8Q$? What about the possibility that one of the W bosons decays leptonically instead of hadronically? Here are all decay modes not covered by the `HHto4Vto2Lplus` fragment:

1. $HH\to4W\to8j$;
2. $HH\to4W\to\ell\nu6j$;
3. $HH\to2W2Z$ with $2W\to4j$ and A) $2Z\to4j$ B) $2Z\to2\nu2j$ C) $2Z\to4\nu$;
4. $HH\to2W2Z$ with $2W\to\ell\nu2j$ and A) $2Z\to4j$ B) $2Z\to2\nu2j$ C) $2Z\to4\nu$;
5. $HH\to4Z$ with A) $2Z\to8j$ B) $2Z\to6j2\nu$ C) $2Z\to4j4\nu$ D) $2Z\to2j6\nu$ E) $2Z\to8\nu$.

We could do one of the following to complement `HHto4Vto2Lplus`:
1. just go with the 1st decay mode, i.e, $HH\to4W\to8j$;
2. combine the 1st and 2nd decay modes by requiring $H\to WW$ and at least 6 jets in the final state;
3. combine the 1st, 3rd and 5th decay modes by requiring $W\to jj$ and disallowing $Z\to\ell\ell$;
4. combine the 1st, 2nd, 3.A, 3.B, 4.A, 5.A and 5.B decay modes by requiring at least 6 jets in the final state and disallowing $Z\to\ell\ell$.

@anmalara which one do you prefer?

anmalara:
Hi, after discussing with the other B2G people, we came up with a different proposal, motivated by the number of leptons. we would like to have the following two combinations

1) 1, 3a, 3b, 5a, 5b (>=6j + 0l)
2) 2, 4a, 4b (>=4j + 1l)

We will therefore leave aside the decays with many neutrinos. While the multi-lepton cases should be included in the `HHto4Vto2Lplus` sample.

We also want to make sure that the `HHto4Vto2Lplus` contains the multilepton decays of the `4tau` mode. Could you confirm?

Finally, we think that the `HHto4Vto2Lplus` name might be misleading since it also contains `2V2taus` and `4tau`. 
Thanks,
Andrea

ktht:
The 1st fragment is easy to implement, but I don't think it's possible to construct the 2nd fragment without including 4c or without excluding 4b. You can find my reasoning in [this notebook](https://cernbox.cern.ch/s/1NxyzV8kAYewuSv). Judging from the BRs, I think the price to pay for being more inclusive and accepting 4c isn't that big of a deal, because 4c constitutes less than 1% of total phase space covered by the 2nd fragment. I've now added two additional fragments for the resonant HH production to this PR:
- `HHto4Vto0L6Jplus`, which accepts DMs 1, 3a, 3b, 5a and 5b;
- `HHto4Vto1L4Jplus`, which accepts DMs 2 and 4.

> We also want to make sure that the HHto4Vto2Lplus contains the multilepton decays of the 4tau mode. Could you confirm?

`HHto4Vto2Lplus` is completely separate from $HH\to4\tau$ decay mode.

> Finally, we think that the `HHto4Vto2Lplus` name might be misleading since it also contains `2V2taus` and `4tau`.

I disagree: `HHto4V`, `HHto2T2V` and `HHto4T` clearly correspond to $HH\to4V$, $HH\to2\tau2V$ and $HH\to4\tau$ decay modes, respectively, and are thus exclusive to each other. What we're currently discussing here are the additional generator-level filters that are applied on top of `HHto4V`, which is why I named the new fragments as `HHto4VtoXYZ`, where `XYZ` encodes relevant final state particles that are common to all decay modes enabled by the fragment.

For future reference (@tolange), here are the BRs of every HH multilepton decay fragment:

|       Sample       | BR       |
|:------------------:|----------|
| `HHto4T`           | 0.003934 |
| `HHto2T2V`         | 0.030092 |
| `HHto4V`           | 0.057547 |
| `HHto4Vto2Lplus`   | 0.021443 |
| `HHto4Vto0L6Jplus` | 0.013611 |
| `HHto4Vto1L4Jplus` | 0.022194 |

@acarvalh note that the size of `HHto4Vto2Lplus`, `HHto4Vto0L6Jplus` and `HHto4Vto1L4Jplus` samples would have to respectively exceed 37%, 24% and 39% of the inclusive `HHto4V` sample, or otherwise we'd lose in effective event statistics. Also, since almost no phase space would be left uncovered after applying the filters, there's no benefit in doing so in the first place (other than added flexibility to increase sample statistics relative to other samples) if the resulting MC samples are all used to study the same HH process. However, if the samples are generated to study HH production in different regimes (e.g., B2G covers resonant HH production above 1 TeV while HH multilepton studies HH resonances below 1 TeV), then it's still worthwhile to apply the filters to benefit from increased event stats.

If @anmalara agrees then this PR is ready to be merged.

anmalara:
Thanks for the clarifications. Yes, this makes sense to us and you can go ahead. Thanks for taking care of this.

genfragments/ThirteenPointSixTeV/Higgs/HH/HHto4Vto1L4Jplus_TuneCP5_13p6TeV_madgraph-pythia8.py
```diff
@@ -0,0 +1,41 @@
...
+        pythia8PSweightsSettingsBlock,
+        processParameters = cms.vstring(
+            '23:mMin = 0.05',
+            '24:mMin = 0.05',
+            '25:m0 = 125.0',
+            '25:onMode = off',
+            '25:onIfMatch = 24 -24',
+            '25:onIfMatch = 23 23',
+            '23:onMode = off',
+            '23:onIfAny = 1 2 3 4 5 12 14 16',
```
menglu21, line 25: you will have 1L2J4Nu or 1L6J event

ktht: Yes, that's intended. Although now that I think about it, `HHto4Vto1L4Jplus` is a bit misleading here.. What would you recommend for the name though? Just `HHto4Vto1L`?

ktht: @menglu21 I've now renamed the fragment.

tommasodiotalevi:
Hi @menglu21 I have some doubts regarding the dataset names for this MC samples. 

* Concerning `HHto4Vto0L6Jplus`: the *plus* is not correct in naming, maybe we could use "AtLeast" as new terminology (?). Another option is to operate by negation: in this case we don't have leptons, so the name could be: `HHto4VtoNonL`

* Concerning `HHto4Vto1L`: also here "AtLeast" could be used. Here the negations are quite challenging: we can say `ZToNon2L` but then you should say that "AtLeast" one W is "NonLNu".

* Concerning `HHto4Vto2Lplus`: here only "AtLeast" is a viable option.

Do you have any suggestions?

ktht:
Are there any official guidelines on how to name datasets that specifically address non-exclusive final states? I think the bottom line should be that we don't violate those guidelines and that the name is clearly and universally understood. The only argument in favor of `AtLeast` I can come up with is that there are a handful of Run II samples (`/DYJetsToTauTau_M-50_AtLeastOneEorMuDecay*`) containing the string while there are no samples with `plus` or `Plus` in the name with the same meaning as those two are usually reserved for the electric charge (but then again it's difficult to interpret the current fragment names that way). If you really insist, we can rename `HHto4Vto0L6Jplus` to `HHto4Vto0LAtLeast6J` and `HHto4Vto2Lplus` to `HHto4VtoAtLeast2L`. However, I prefer keeping `HHto4Vto1L` the same since there are no better alternatives. In my opinion, it should be clear from the names that they are all mutually exclusive.

tommasodiotalevi:
Hi @ktht, unfortunately the [only guide](https://cms-pdmv.gitbook.io/project/mccontact/rules-for-run3-dataset-names) does not provide enough information about this. Maybe @menglu21 can comment on that.

ktht:
Thanks for the link, @tommasodiotalevi From that web site I found [this section](https://cms-pdmv.gitbook.io/project/mccontact/rules-for-dataset-names#filter), which basically states that we can be creative when naming datasets with generator-level filters.

agrohsje:
@sihyunjeon and myself are discussing good names since more than a year. the point is that names should follow strict rules so that everyone can find the samples. however, sometimes this is just impossible. so what i would advice is in the case of filters: think about something that is close to the other rules that we have and add the explicit examples to https://cms-pdmv.gitbook.io/project/mccontact/rules-for-dataset-names#filter @menglu21 @bbilin what if we add a table there with explicit filter names and what they are doing ?

menglu21:
Hi @agrohsje this is only the fragment so I didn't require them to strictly following the naming convention. when comes to the request name, they should follow that

agrohsje:
Hi @menglu21. Indeed I am referring to the dataset name. @tommasodiotalevi asked about this on cms talk and I answered here not to spam everyone.

ktht:
In a [related discussion](https://gitlab.cern.ch/cms-hig-mc/mc-hig-requests/-/issues/45#note_7716325) it was pointed out that the lines starting with `ResonanceDecayFilter:filter` are redundant in `HHto4V_*` and `HHto4T_*` cards. I confirm this really to be the case. Since this is such a small change in the code, I can prepare the PR that removes the redundant lines to prevent similar comments about the fragments that may crop up in the future. Or what do you think?

PR #3548, opened 2023-10-21
add mglib_path for precompiled lib

sihyunjeon:
taking madgraph interface in CMSSW to fetch ninja (+oneloop) instead of installing ninja everytime when making gridpacks

### DIFF
--- a/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
+++ b/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
@@ -223,8 +223,11 @@ make_gridpack () {
     #  echo "set output_dependencies internal" >> mgconfigscript
       echo "set lhapdf_py3 $LHAPDFCONFIG" >> mgconfigscript
     #   echo "set ninja $PWD/HEPTools/lib" >> mgconfigscript
+      mglib_path=`scram tool info madgraph5amcatnlo | grep MADGRAPH5AMCATNLO_BASE`
+      mglib_path=${mglib_path#*=}/HEPTools/lib
       if [ -n "$mglib_path" ]; then
         echo "set ninja $mglib_path" >> mgconfigscript
+        echo "install collier" >> mgconfigscript
       fi
 
       if [ "$queue" == "local" ]; then
@@ -709,19 +712,15 @@ export SYSTEM_RELEASE=`cat /etc/redhat-release`
 # set scram_arch 
 if [ -n "$5" ]; then
     scram_arch=${5}
-    mglib_path=""
 else
     if [[ $SYSTEM_RELEASE == *"release 7"* ]]; then 
         scram_arch=slc7_amd64_gcc10
-        mg_buildtag=50a30a0356f42c10763f5933e2773041
     elif [[ $SYSTEM_RELEASE == *"release 8"* ]]; then
         scram_arch=el8_amd64_gcc10
-        mg_buildtag=50a30a0356f42c10763f5933e2773041
     else 
         echo "No default scram_arch for current OS!"
         if [ "${BASH_SOURCE[0]}" != "${0}" ]; then return 1; else exit 1; fi        
     fi
-    mglib_path=/cvmfs/cms.cern.ch/${scram_arch}/external/madgraph5amcatnlo/2.7.3-${mg_buildtag}/HEPTools/lib
 fi
 
 #set cmssw 
--- a/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
+++ b/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
@@ -223,7 +223,10 @@ make_gridpack () {
     #  echo "set output_dependencies internal" >> mgconfigscript
       echo "set lhapdf_py3 $LHAPDFCONFIG" >> mgconfigscript
     #   echo "set ninja $PWD/HEPTools/lib" >> mgconfigscript
-    
+      if [ -n "$mglib_path" ]; then
+        echo "set ninja $mglib_path" >> mgconfigscript
+      fi
+
       if [ "$queue" == "local" ]; then
           echo "set run_mode 2" >> mgconfigscript
       elif [ "$queue" == "pdmv" ]; then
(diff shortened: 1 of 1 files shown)


sihyunjeon:
@DickyChant present for you

DickyChant:
Is there a way that we can systematically include the libs?

e.g. extracting from `scram tool info madgraph5amcatnlo`

sihyunjeon:
> Is there a way that we can systematically include the libs?
> 
> e.g. extracting from `scram tool info madgraph5amcatnlo`

it could be read from `MADGRAPH5AMCATNLO_BASE` but parsing this seemed like out of my shell scripting knowledge...
i can do `scram tool info madgraph5amcatnlo > log`
and make it parse using python scripts if you think it's better.

or... this could be dirty but one can do

```
scram tool info madgraph5amcatnlo > log
grep "scram tool info madgraph5amcatnlo" log > set.sh
source set.sh
```

sth like this

sihyunjeon:
and then set the paths to `$MADGRAPH5AMCATNLO_BASE/blahblah`

DickyChant:
maybe sth like:
```
MG5_PATH=`scram tool info madgraph5amcatnlo | grep MADGRAPH5AMCATNLO_BASE`
MG5_PATH=${MG5_PATH#*=}
echo $MG5_PATH
```
There are more tricks with `awk` and `grep` but for now I guess this should be enough

DickyChant:
Given that `hepforge` seems to have more downtime, shall we move with this direction?

sihyunjeon:
hi sitian, thanks for the snippets

sihyunjeon:
@Dominic-Stafford i think we also need to add collier installation in CMSSW
MadGraph internally has a setting that it looks for ninja and collier libraries altogether,

problem is that there is no collier in CMSSW so if I only set ninja path, madgraph looks for both ninja and collier, and then it just decides to NOT use collier instead of installing it. so now it has "install collier" after setting ninja. ultimately we should fix this with CMSSW

sihyunjeon:
```
WARNING: The collier reduction library could not be found with PATH:/data9/Users/shjeon/test2/genproductions/bin/MadGraph5_aMCatNLO/wplustest_4f_NLO/wplustest_4f_NLO_gridpack/work/MG5_aMC_v2_9_13/HEPTools/lib specified in mg5_configuration.txt. It will not be available. 
```

This is the error message I get

Dominic-Stafford:
> @Dominic-Stafford i think we also need to add collier installation in CMSSW MadGraph internally has a setting that it looks for ninja and collier libraries altogether,
> 
> problem is that there is no collier in CMSSW so if I only set ninja path, madgraph looks for both ninja and collier, and then it just decides to NOT use collier instead of installing it. so now it has "install collier" after setting ninja. ultimately we should fix this with CMSSW

Hi Sihyun. Sure, I can try adding collier to cmsdist so it builds this as an external for CMSSW. One other thing is we should be careful about making sure the relevant CMSSW releases have the right versions of madgraph. Do you want to do this for both run 2 UL and run 3, or just run 3? And am I right in saying we need MG5 2.9.13 in CMSSW master and 12_4, and 2.6.5 in CMSSW_10_6?

sihyunjeon:
@Dominic-Stafford Hi dominic, I leave this decision to @menglu21 and @bbilin

DickyChant:
> > @Dominic-Stafford i think we also need to add collier installation in CMSSW MadGraph internally has a setting that it looks for ninja and collier libraries altogether,
> > problem is that there is no collier in CMSSW so if I only set ninja path, madgraph looks for both ninja and collier, and then it just decides to NOT use collier instead of installing it. so now it has "install collier" after setting ninja. ultimately we should fix this with CMSSW
> 
> Hi Sihyun. Sure, I can try adding collier to cmsdist so it builds this as an external for CMSSW. One other thing is we should be careful about making sure the relevant CMSSW releases have the right versions of madgraph. Do you want to do this for both run 2 UL and run 3, or just run 3? And am I right in saying we need MG5 2.9.13 in CMSSW master and 12_4, and 2.6.5 in CMSSW_10_6? Also, should we use a specific ninja or collier versions, or just the latest ones?

Hi Dominic, maybe I am wrong (my schedule is totally ruined recently), I remembered MG usually have a relatively fixed version choice of those packages. (either always the newest one, or a fixed version per major release). So it is a relatively easier decision to make... 

Shall we try at some moment having those libraries decoupled from MG as an ultimate solution?

sihyunjeon:
@DickyChant They are fixed for every version of madgraph, it doesn't really depend on madgraph version. They are imported from HEPToolsInstaller.py which is CENTRAL. So madgraph 220 and madgraph 352 would share the same install libraries...

DickyChant:
> @DickyChant They are fixed for every version of madgraph, it doesn't really depend on madgraph version. They are imported from HEPToolsInstaller.py which is CENTRAL [link](https://github.com/mg5amcnlo/HEPToolsInstallers). So madgraph 220 and madgraph 352 would share the same install libraries...
> 
> One of the reasons why madgraph can't really support too old versions...

That's what I thought actually because I never see a different version, and I assume they are the newest available version

DickyChant:
> I might've misunderstood but I think/agree that having totally decoupled versions of external libraries will be the safest choice

yeah that's what I meant, in that case we could perhaps have more flexible MG solution as after all apart from those libs we don't have anything to be compiled...

Dominic-Stafford:
From what I can see with hepforge being down, it doesn't look like ninja or collier versions change too much, actually (I guess they're pretty stable, hopefully all the bugs have been found). So I think it would be fine to put collier in CMSSW, if it was often being updated it would a different story but here I think it's fine. So just to check, you're not actually taking MG from CMSSW, just ninja and collier?

DickyChant:
> From what I can see with hepforge being down, it doesn't look like ninja or collier versions change too much, actually (I guess they're pretty stable, hopefully all the bugs have been found). So I think it would be fine to put collier in CMSSW, if it was often being updated it would a different story but here I think it's fine. So just to check, you're not actually taking MG from CMSSW, just ninja and collier?

Indeed, as hepforge seems to be less stable. Our idea is that for MG itself we need patches & upgrades from time to time, but for those heptools (ninja, collier... etc) we don't need them to be installed and compiled every time. It is neither computationally efficient nor safe with nowadays unstable hepforge. 

Currently, a temporary solution is to nicely use those libraries from your wrapped-up MG in CMSSW, but ultimately it might be a good idea to completely disentangle :)

sihyunjeon:
for now, if no objections until monday afternoon, i'll merge. at least having something is better altho it's not everything (i believe backporting will take longer).

also it would not "likely" happen, but i did check with `pp > w [qcd]` process whether gridpack makes the LHE files properly or not, just in case there could be hidden problems.

sihyunjeon:
actually, let me merge this now. hepforge is down again... :)

PR #3551, opened 2023-10-23
adding decay and filter for ttH using JHUGen config

neharawal14:
Previous ttH samples were missing the 4l filter and the decay chain. We want to produce ttH samples with decay  H->ZZ->2l 2X, and then apply a 4l filter. So adding the decay setting in the JHUGen.input

### DIFF
--- a/bin/Powheg/production/Run3/13p6TeV/Higgs/ttH_ZZ_NNPDF31_13TeV/JHUGen.input
+++ b/bin/Powheg/production/Run3/13p6TeV/Higgs/ttH_ZZ_NNPDF31_13TeV/JHUGen.input
@@ -1,0 +1,1 @@
+DecayMode1=8 DecayMode2=9 NLepMin=4 WriteFailedEvents=2


menglu21:
Hi @neharawal14 could you please modify the "ttH_inclusive_ZZ_NNPDF31_13TeV_template.input" to "ttH_ZZto2L2X_NNPDF31_13TeV_template.input"

neharawal14:
Hi @menglu21 ,
As far as I understand we do not need to modify anything in the  "ttH_inclusive_ZZ_NNPDF31_13TeV_template.input" , since the input cards just specify the production and do not describe the decay chain there. The decay should happen according to JHUGen settings, so adding the JHUGen input separately should be enough. 
Please let me know if I am missing something, or understanding it wrong.

menglu21:
then you should use "ttH_NNPDF31_13TeV_template.input" since you don't specify the decay of higgs

neharawal14:
Hi @menglu21 ,
I modified the name of the ttH input card. Could you please merge the PR now. 
Thank you!

neharawal14:
Hi @menglu21 , 
Could you confirm if the cards are fine now to merge ?

neharawal14:
Hi,  

I changed name of all the input cards and ttH folder name. It will be great if it the PR could be merged. 
Thank you!

neharawal14:
Hi @menglu21 , 
Could you confirm if there are more issues left for this PR to be merged ?

Thank you!

neharawal14:
Hi Meng, 

We would also like to add the gluon gluon fusion input card with the flag ew "off" in the cmssw gen repository. 

Since, we decided that we need to create samples with mass points above M=165 GeV with ew "OFF" and not the flag ON (as in previous cards).  Further we want to produce two ew OFF samples at M=125, 124 GeV for comparison with the already produced samples with ew flag ON. 
So I am submitting a template for the input card with ew off.
Let me know if I need to create specific input cards with the mass points, or need to do something else? 

Could you also comment on ttH samples input cards in the PR.

Thank you!

menglu21:
Hi @neharawal14 sorry for the late reply, the current cards look fine to me, I'll merge them. For the other mass point, I agree with you, no need to make PRs for them

PR #3554, opened 2023-11-02
Add SMEFTsim and SMEFTfr cards for top LFV analysis UL

emily-tsai11:
This PR adds data cards used to produce MC samples for charged LFV in top decay and production.


menglu21:
Hi @emily-tsai11 a general comment on the proc_card, how large will be the gridpack and what's the number of file after you untarring the gridpack. if the number of file inside the gridpack is large, i.e., >15K, it's not favored by the computing. Another question is that do you need the (SM) top decay in the syntax, if it's not necessary needed, then you can leave it there and do the top decay through pythia.

emily-tsai11:
Hi @menglu21 
> how large will be the gridpack and what's the number of file after you untarring the gridpack
There are ~36k files in the untarred gridpack because these are inclusive samples. What can I do in this case? 

> do you need the (SM) top decay in the syntax, if it's not necessary needed, then you can leave it there and do the top decay through pythia
The SM top decay syntax is needed because we don't want to include all possible top quark decays. It also helps to make clear what is included in our samples.

menglu21:
Hi @emily-tsai11 I mean e.g. "t~ > w- b~, w- > Dec Dec", this is an inclusive decay, you can remove those SM decay of top and pythia will do the decay.

PR #3556, opened 2023-11-07
Revert "add mglib_path for precompiled lib"

sihyunjeon:
Reverts cms-sw/genproductions#3548


sihyunjeon:
reverting this because ignoring collier installation causes hidden problems

PR #3558, opened 2023-11-08
gammagamma production

smking22:
For Run 2 analysis of diphoton production (NLO)

### DIFF
--- a/bin/Sherpa/cards/production/13TeV/DiphotonNLO/Run.dat_yyNLO
+++ b/bin/Sherpa/cards/production/13TeV/DiphotonNLO/Run.dat_yyNLO
@@ -1,0 +1,60 @@
+(run){
+  EVENTS 1000;
+  EVENT_MODE HepMC;
+
+  % scales, tags for scale variations
+  FSF:=1.; RSF:=1.; QSF:=1.;
+  SCALES STRICT_METS{FSF*MU_F2}{RSF*MU_R2}{QSF*MU_Q2};
+  HEPMC_USE_NAMED_WEIGHTS=1;
+  SCALE_VARIATIONS=0.25,0.25 0.25,1. 1.,0.25 1.,1. 1.,4. 4.,1. 4.,4.;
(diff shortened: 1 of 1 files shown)


sihyunjeon:
there is a sample already IIRC from HGG

ajgilbert:
Hi, the only HGG samples we're aware of are LO multijet, e.g.: https://github.com/cms-sw/genproductions/blob/master/bin/Sherpa/cards/production/13TeV/higgs/Run.dat_13TeV_gamgam_3j_loop_Mgg80-13000

The card is here is for NLO 0,1 jet + LO 2,3 jet merged

sihyunjeon:
bah my bad, you are right

bin/Sherpa/cards/production/13TeV/DiphotonNLO/Run.dat_yyNLO
```diff
@@ -0,0 +1,60 @@
...
+
+  Process 93 93 -> 22 22 93{NJET};
+  Order (*,2);
+  CKKW sqr(QCUT/E_CMS)/(1.0+sqr(QCUT/0.6)/Abs2(p[2]+p[3]));
+  NLO_QCD_Mode MC@NLO {LJET};
+  ME_Generator Amegic {LJET};
+  RS_ME_Generator Comix {LJET};
+  Loop_Generator LOOPGEN {LJET};
+  PSI_ItMin 50000 {3}
+  Integration_Error 0.99 {3}
```
menglu21, line 47: 0.99 should be a typo?

ajgilbert: it's actually intended - we are following the configuration used by ATLAS, and we understand this is what is recommended by their sherpa experts. Our understanding is the `PSI_ItMin` setting is effectively controlling the accuracy here, and indeed we see precisions on the {3,4,5} subprocesses of < 2% from the sherpa logs.

PR #3561, opened 2023-11-16
Adding monoHiggs_ZpBaryonic datacards to mg265UL

sslomte:
The datacards are cloned from pre-legacy. 
1. Can these be used for ultra-legacy (UL)?
2. Which custom PDF choice should be chosen (unsure if the one in meta file is consistent with UL recommendations)?
3. Which tune should be used with the custom PDF in UL?


sslomte:
Hi @menglu21, could you please confirm the PDF choice and Tune to use for UL? We're targeting Zprime mass above 1 TeV and wondering if LO PDF is more appropriate to avoid negative weights issue as per [1]. Upon confirmation, I can produce the gridpacks. Thanks a lot!

[1]: https://twiki.cern.ch/twiki/bin/view/Main/ExoMCInstructions#PDF_Issues

hftsoi:
@menglu21 gentle ping:)

Issue #3269, opened 2022-09-30
Filter efficiency from fragments should be dropped once and for all


efeyazgan:
@menglu21  @sihyunjeon  @Saptaparna

Since Si hyun and Sitian are sure (see  https://cms-talk.web.cern.ch/t/filter-efficiency-check-in-gen-check-script/15409) that the filter efficiency is not used at all I removed the check in the request checking script (see:
https://github.com/cms-sw/genproductions/pull/3268) so this filter should be removed from all fragments. Please do that soon and then I will adapt the script not to cause inconsistency in UL consistency check. However, if it is not removed soon, I think we should revert back the request checking script. 

In any case, I do not take any responsibility for any problems that  this change may cause: https://github.com/cms-sw/genproductions/pull/3268. 

Thanks.

efeyazgan:
Hi Alexander, yes, but following the same reasoning we could have kept the check. If something is used, we should check how it is used. I am sure that check was there for a reason...

sihyunjeon:
Hi so there are several things to think about here.

This creates somewhat unnecessary entropy for the MC contacts because
a. We don't know whether there are any real users for this metadata so we might be putting our efforts to something without real gain.
b. Removing the filter efficiency fragment still works (avoids gen checking script errors) - which means subset of the samples would for sure have broken filter efficiency written in GEN files as metadata. So subset of our samples are already broken.
c. It's not so easy to modify this through python scripts because efficiency from the run log (for filter efficiency) only gets delivered through email. I don't think the run log results are stored somewhere in McM, at least I am not aware. If I am correct, one needs to crop out the values from email boxes and tweak it into the fragments and I would hardly imagine MC contacts doing this. So in the end it would return to b. where MC contacts will just remove the line to avoid the problem, breaking the variables that should not be used.

But as Alexander said, SOME might use this and it MIGHT be not totally useless to store correct values. So allowing default settings (filter efficiency >= 1.0 or <=0.0 which doesn't make sense) is sort of compromise in between.

agrohsje:
Hi Efe,
maybe this was not clear in my message: I think we should have kept the check (and the motivation now is the same as back then: it is stored in EDM and we don't know if someone is using it or not) but just slightly modify it as proposed by Sihyun:
But as Alexander said, SOME might use this and it MIGHT be not totally useless to store correct values. So allowing default settings (filter efficiency >= 1.0 or <=0.0 which doesn't make sense) is sort of compromise in between.
[...]

PR #3572, opened 2023-12-06
Fix the case when filter eff is negative in the fragement

tvami:
Sometimes the filter eff is set to -1 to show that it is not used, currently the script does not let the validation pass.


tvami:
@efeyazgan this extra patch seems to be in line with what was discussed in https://github.com/cms-sw/genproductions/issues/3269 please merge at your earliest convenience

tvami:
I think you can use this https://cms-pdmv-prod.web.cern.ch/mcm/requests?prepid=EXO-Run3Summer22EEwmLHEGS-01460&page=0&shown=127 for testing

PR #3574, opened 2023-12-06
Modify the make_rwl.py file with corrected pdf number

De-Cristo:
Dear experts,

There are several mismatch between PDF and LHAPDF ID in previous `make_rwl.py` file in `/bin/PowHeg` which will cause error in the process of generation.

I have modified the lines under Run3 4FS part according to the updated LHAPDF IDs.

Please check. Thanks @covarell for pointing out!

Best,
Licheng


covarell:
Basically it is a +1 from me.
I am just wondering why the PDF4LHC15 set was added? This looks like a pretty old set.

De-Cristo:
Hi,
In the bbH generation, people from the working group suggesting to use this pds instead of the CMS recommended one for Run2, so we think it might be better to include this even only for the tests.

sihyunjeon:
But this change will affect all the other samples as well. Why not just use pdf4lhc15 for samples of the working group's interest ONLY and leave other samples as it is?

De-Cristo:
Perhaps @danielwinterbottom could provide a clearer explanation of the current status.

danielwinterbottom:
About the PDF sets, the use of PDF4LHC15 is the recommendation of the bbH/bH subgroup of the LHCHWG (see here: https://twiki.cern.ch/twiki/bin/view/LHCPhysics/LHCHWGBBH#Recommendations_for_the_acceptan). 

For the previous Run-2 analysis, we double-checked with the working group conveners if it was really necessary to use this PDF set over the NNPDF sets usually used by CMS and they expressed a strong preference for it. We also checked with the CMS MC validation group if not using NNPDF was OK and they told us the preference of NNPDF in CMS is for ME/PS consistency, but its is only a guidance and if there for good reasons for using a different PDF set GEN would not object unless it was  something clearly incorrect (e.g using LO PDFs for NLO/NNLO MC samples).

covarell:
For me it is OK to proceed. This will affect only 4FS POWHEG processes, which are in very small number.

danielwinterbottom:
Just to add though, this recommendation is only for bbH. I think there should be some option in the code so that the PDF4LHC set is only used for this process and not the other 4FS cases.

sihyunjeon:
@danielwinterbottom If one wants to use different PDF set one can use it with valid reasons, sure. My point is different, "is it necessary to add this to central scripts back again while you can steer it only for samples that you are interested in" was my question. Given that (a) we were planning to minimize the PDF sets (not sure what the progress is, raised last summer https://github.com/cms-sw/genproductions/pull/3479) (b) PDF reweighting is much more expensive for powheg than other generators.

danielwinterbottom:
> @danielwinterbottom If one wants to use different PDF set one can use it with valid reasons, sure. My point is different, "is it necessary to add this to central scripts back again while you can steer it only for samples that you are interested in" was my question. Given that (a) we were planning to minimize the PDF sets (not sure what the progress is, raised last summer #3479) (b) PDF reweighting is much more expensive for powheg than other generators.

I agree, it should only be added for the bbH process.

menglu21:
Hi all, we are still using LHAPDF 6.5.1, in which MSHT20nnlo_nf4 with ID 27810. The ID change happens in LHAPDF 6.5.4 (6.5.3 version still has 27810 for MSHT20nnlo_nf4)

De-Cristo:
Hi, we firstly met this problem during generating the bbH gridpack,
the details and discussion could be fine in this [thread](https://cms-talk.web.cern.ch/t/generation-of-bbh-process-with-powheg-for-run3-mc/31713/9)[1]

[1] https://cms-talk.web.cern.ch/t/generation-of-bbh-process-with-powheg-for-run3-mc/31713/9

menglu21:
I just checked that 12_4_4 is using lhapdf 6.4.0, which still has id 27810 for MSHT20nnlo_nf4, I haven't checked how lhapdf is called in powheg production, maybe someone already know it and can comment.

agrohsje:
Hi Meng, on https://twiki.cern.ch/twiki/bin/viewauth/CMS/PowhegBOXPrecompiled we recommend 
 slc7_amd64_gcc700 and slc7_amd64_gcc820: CMSSW_10_6_X (github branch: powhegUL)
 slc7_amd64_gcc10: CMSSW_12_4_X (github branch: master) 
 12_4_4 is used by the authors. So the mismatch between PDF's cannot happen there if 12_4_4 has 6.4.0. Some info in this thread is not correct.

bbilin:

PR #3576, opened 2023-12-07
Adding genfragments for multiply charged particles - pythia8

RafeyH:
Adding genfragments for multiply charged particles - pythia8 version, to the HSCP repo. 
Charges={1e..16e,20e,24e,30e} 
Masses={100,200,300,400,500,600,700,800,900,1000,1200,1400,1600,1800,2000,2400,2800}GeV


tvami:
This is supposed to come to the EXO MC&I GitLab not under GEN

PR #3577, opened 2023-12-08
lxplus9 setting

sihyunjeon:


sihyunjeon:
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/gridpack_generation.sh#L731-L732

Added this as CMSSW_12_4_8 was not available for rhel9 but i am not sure if this is OK

PR #3578, opened 2023-12-11
Update Starlight generator to v326

stahlleiton:
This PR adds the following updates:
- Update Starlight generator to v326
- Add 4 input cards used for double diffraction, dimuon, dielectron and coherent Jpsi processes.
- Add support for el9 and update CMSSW release.
- Use nproc for number of threads when compiling, instead of using 8 threads.
- Remove invalid characters (i.e. "--") in the input cards before copying it to the LHE file.
- Add cross section to output LHE.
- Add option to define the Woods-Saxon skin depth and radius, and the inelastic n-n cross section in the input card.

It was tested locally for el7, el8 and el9, using the input cards provided in this PR.


sihyunjeon:
Did you check if this works for the actual production chain?

For Madgraph case, el9 gridpacks were not working for the actual production (run3 using el8)

stahlleiton:
yes, I tested locally for el7, el8 and el9 and it did work (tested using CMSSW_13_0_16 both LHE and GEN steps)

sihyunjeon:
Ha interesting. thanks for the feedback. Maybe it is only madgraph being wonky

stahlleiton:
Starlight is a much simpler generator compared to madgraph, so I would not be surprised there are differences.

PR #3579, opened 2023-12-11
adding Narrow Width(NW) and Large Width(LW) AZH to 4b samples gridcar…

zhiyuanlcern:
…ds generation

Based on AZHtoLLtt, with lambda2/3 values calculated with 2HDMC. Checked gridpacks distributions and distributions after PYTHIA decay


menglu21:
you will use pythia for Higgs and Z decay, right?

zhiyuanlcern:
> you will use pythia for Higgs and Z decay, right?

Yes, the same as what was done for the VH->Vhh samples.
Cheers,
Jordan

PR #3582, opened 2023-12-12
Add SuperChic 4.2

stahlleiton:
This PR adds the setup to run vmLHE production using the [SuperChic](https://superchic.hepforge.org/) 4.2 generator, used to simulate photon-photon and photon-nucleus interactions in ultra-peripheral heavy-ion collisions.

The setup was tested locally using 3 input cards (light-by-light, QED dimuon and ChiC production, included in this PR) for scram arch releases el7, el8 and el9.

### DIFF
--- a/bin/Superchic/gen_fragment.py
+++ b/bin/Superchic/gen_fragment.py
@@ -1,0 +1,25 @@
+import FWCore.ParameterSet.Config as cms
+
+externalLHEProducer = cms.EDProducer("ExternalLHEProducer",
+    args = cms.vstring(ADD_PATH_TO_TARBALL),
+    nEvents = cms.untracked.uint32(1),
+    numberOfParameters = cms.uint32(1),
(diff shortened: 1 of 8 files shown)


menglu21:
Hi @stahlleiton same question with UPCGen here

stahlleiton:
> Hi @stahlleiton same question with UPCGen here

The last version of SuperChic was released on 13/03/23 and another version is expected early next year, so seems it gets updates at least once per year. Moreover, APFEL (a PDF evolution library that SuperChic uses), also gets updated roghly once per year.

Regarding the duration, downloading SuperChic and APFEL takes 7 seconds and installing: APFEL takes 77 seconds and SuperChic takes 11 seconds.

menglu21:
thanks, double check on lxplus8, gridpack is obtained and it's able to get lhe events. Merge now

PR #3591, opened 2023-12-22
Genfragments for VLL LLP samples

wonyongc:
Hi,
This includes both the template/generator (makeGenfragments.py) as well as the generated files.


genfragments/ThirteenPointSixTeV/VLL_LLP/VLLS_ele_M450_D1e-14_13p6TeV-madgraph_cfi.py
```diff
@@ -0,0 +1,69 @@
...
+            'UncertaintyBands:nFlavQ = 4',
+            'UncertaintyBands:MPIshowers = on',
+            'UncertaintyBands:overSampleFSR = 10.0',
+            'UncertaintyBands:overSampleISR = 10.0',
+            'UncertaintyBands:FSRpTmin2Fac = 20',
+            'UncertaintyBands:ISRpTmin2Fac = 1'
```
menglu21, line 58: you don't need to write explicitly those lines, instead you just need to import those modules like https://github.com/cms-sw/genproductions/blob/73d598db80c4cefec0df9564c765df6e1b38c539/genfragments/ThirteenPointSixTeV/Higgs/TTHtoNon2B_M-125_TuneCP5_13p6TeV_powheg-pythia8.py. Another thing is that I think only the template and the shell script should be enough, no need to create and upload all the fragments here.

wonyongc: Thank you for the tip! I updated the template and deleted the generated fragments - please excuse the numerous commits to delete the generated fragments - I was on the web UI and deleted them one-by-one.

PR #3609, opened 2024-01-15
Add Run3 ZG nTGC old model cards and new model cards (2 set of cards)

Muty0:
For run3 ZG nTGC searching analysis, use reweighting method, one set is the old model, the other set is the new model (from SJTU theory group).


menglu21:
Hi @Muty0  please remove the .log and tar.gz files in the PR. And also please not use string "old", you can simply remove it

Muty0:
Hi @menglu21 ,
Thank you for the suggestion, I have now removed .log and tar.gz files, and changed the name into ZGtoLLG01j_5f_LO_mlm_nTGC_sjtu (the nTGC model create by SJTU's theory group) and ZGtoLLG01j_5f_LO_mlm_nTGC (the conventional nTGC model CMS and Atlas used in the previous analysis).
Please check and see, I would be glad to fix the cards if there are still any.

PR #3619, opened 2024-01-15
add and test cleanup repo

efeyazgan:


efeyazgan:
Hi again, @menglu21 @bbilin, now this is final. Please review and merge. Thanks.

.github/workflows/main.yml
```diff
@@ -42,10 +44,24 @@ jobs:
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
+      - name: Cleanup build folder
+        run: |
+          ls -la ./
+          rm -rf ./* || true
+          rm -rf ./.??* || true
+          ls -la ./
```
menglu21, line 52: Hi @efeyazgan why the powheg part is also modified

efeyazgan: Hi, because Markus discovered that the space is filling up with these workflows (which was blocking the workflows). So, we added this cleaning both to powheg and request checking script parts.

efeyazgan: @menglu21 @bbilin Any issues remianing for this PR? thanks.

PR #3625, opened 2024-01-27
URGENT: GEN cards of VBFH and VH (SUS HtoAAto2Mu2Tau) for 3 UL samples production

Fengwangdong:
As requested by SUS conveners, we submit our GEN cards of VBF-H and VH to request central production of three UL2018 MC samples for SUS-23-005 pre-approval Q&A.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/HToAATo2Mu2Tau/SUSY_VBFH_HToAATo2Mu2Tau/VBFH_M125_ToAA_M10/SUSY_VBFH_M125_ToAA_M10_To2Mu2Tau_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/HToAATo2Mu2Tau/SUSY_VBFH_HToAATo2Mu2Tau/VBFH_M125_ToAA_M10/SUSY_VBFH_M125_ToAA_M10_To2Mu2Tau_proc_card.dat
@@ -1,0 +1,5 @@
+import model NMSSMHET --modelname
+
+generate p p  > h01 j j $$ W+ W- Z, (h01 > a01 a01, a01 > tau+ tau-, a01 > mu+ mu-)
+
+output SUSY_VBFH_M125_ToAA_M10_To2Mu2Tau -nojpeg
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/HToAATo2Mu2Tau/SUSY_VBFH_HToAATo2Mu2Tau/VBFH_M125_ToAA_M10/SUSY_VBFH_M125_ToAA_M10_To2Mu2Tau_customizecards.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/HToAATo2Mu2Tau/SUSY_VBFH_HToAATo2Mu2Tau/VBFH_M125_ToAA_M10/SUSY_VBFH_M125_ToAA_M10_To2Mu2Tau_customizecards.dat
@@ -1,0 +1,3 @@
+set param_card mass 25 125
+set param_card mass 36 10
+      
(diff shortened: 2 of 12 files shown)


bin/MadGraph5_aMCatNLO/cards/production/13TeV/HToAATo2Mu2Tau/SUSY_VBFH_HToAATo2Mu2Tau/VBFH_M125_ToAA_M10/SUSY_VBFH_M125_ToAA_M10_To2Mu2Tau_proc_card.dat
```diff
@@ -0,0 +1,5 @@
+import model NMSSMHET --modelname
+
+generate p p  > h01 j j $$ W+ W- Z, (h01 > a01 a01, a01 > tau+ tau-, a01 > mu+ mu-)
```
DickyChant, line 3: Is this syntax good for VBF?

sihyunjeon: i think this should be done with QCD suppression. P1_ccx_h01gg I already see this in subprocess directory which definitely does not look like vbf

sihyunjeon: image

also you have things that are chimed in by other bsm particles... not sure if this is intended + how it's interpreted.

image

but definitely invaded by QCD...

DickyChant:
Is VBF topology used in this analysis?

bin/MadGraph5_aMCatNLO/cards/production/13TeV/HToAATo2Mu2Tau/SUSY_VBFH_HToAATo2Mu2Tau/VBFH_M125_ToAA_M10/SUSY_VBFH_M125_ToAA_M10_To2Mu2Tau_proc_card.dat
menglu21: I missed that part, @Fengwangdong if you need pure EW process, please add "QCD==0"

Fengwangdong:
Hi all, thank you for your checks. I also just realized that QCD should be disabled when I want pure VBF topology. I will remake the proc card and recommit for a new pull request.

sihyunjeon:
Hi you should run standalone madgraph and check the diagrams. even if you restrict qcd, it looks like there could be many more diagrams and make sure what you have is what you want.

DickyChant:
In fact it would be rather weird if the run card stays as what it is... No mjj requirement and eta_j cut is just 5 (slightly above 4.7 which is basically RECO acceptance for forward jet) 

Are you really targeting a VBF signature? since VBF signature would be like several percent level v.s. gluon fusion (taking from SM higgs production)

Fengwangdong:
We already produced all samples of ggF. We are just requested one VBF sample to cross check the acceptance for Q&A. The run card content is the same as the one for ggF. We think it is not a big deal due to the fiducial selection in the analysis

DickyChant:
Yeah make sense, though the contribution should be really percent level...

Fengwangdong:
Could you please point to me where to check the generated diagrams? i have not used MG5 for a while a totally forgot...

Fengwangdong:
Well, as far as I know there is a bit syntax difference when people are using different models (eg.  ours is NMSSMHET). I think it makes more sense to check the diagrams in the output dir when I am generating the gridpack. Are they stored anywhere?

DickyChant:
They are stored in the subprocess directory in madevent if you configure it to store

But, what prevents you from checking with standalone running? most time the diagrams are not saved to save storage space.

Fengwangdong:
Thanks, I did not find a way to import the NMSSMHET model for MG5 if I run the interactively. But OK, I figured it out and obtain the diagrams to check now

DickyChant:
Great!

In any case you might want to check the upcoming MG tutorial in few weeks :)

PR #2848, opened 2021-03-02
ggF higgs->aa->2mu2tau, various H/a masses

red1habibullah:
cards for ggF higgs->aa->2mu2tau , various h/a masses.
I noticed that these cards don't contain any specification related to SysCalc, so I wonder if it is safe to not having them and we still can get the systematic uncertainties done. Maybe experts can comment.

PR #3626, opened 2024-01-29
Revision for VBF H (disable QCD) -- SUS HtoAAto2Mu2Tau

Fengwangdong:
Following up the pull request "#3625", I double checked and updated the VBFH proc card. QCD is disabled. The diagrams with offshell BSM particles are anomalous gauge boson couplings that are allowed by the NMSSM model.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/HToAATo2Mu2Tau/SUSY_VBFH_HToAATo2Mu2Tau/VBFH_M125_ToAA_M10/SUSY_VBFH_M125_ToAA_M10_To2Mu2Tau_customizecards.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/HToAATo2Mu2Tau/SUSY_VBFH_HToAATo2Mu2Tau/VBFH_M125_ToAA_M10/SUSY_VBFH_M125_ToAA_M10_To2Mu2Tau_customizecards.dat
@@ -1,3 +1,3 @@
 set param_card mass 25 125
 set param_card mass 36 10
-      
+set param_card DECAY 36 2.495587e-04
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/HToAATo2Mu2Tau/SUSY_VBFH_HToAATo2Mu2Tau/VBFH_M125_ToAA_M10/SUSY_VBFH_M125_ToAA_M10_To2Mu2Tau_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/HToAATo2Mu2Tau/SUSY_VBFH_HToAATo2Mu2Tau/VBFH_M125_ToAA_M10/SUSY_VBFH_M125_ToAA_M10_To2Mu2Tau_proc_card.dat
@@ -1,5 +1,5 @@
 import model NMSSMHET --modelname
 
-generate p p  > h01 j j $$ W+ W- Z, (h01 > a01 a01, a01 > tau+ tau-, a01 > mu+ mu-)
+generate p p  > h01 j j $$ W+ W- Z  QCD=0, (h01 > a01 a01, a01 > tau+ tau-, a01 > mu+ mu-)
 
 output SUSY_VBFH_M125_ToAA_M10_To2Mu2Tau -nojpeg
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/HToAATo2Mu2Tau/SUSY_VBFH_HToAATo2Mu2Tau/VBFH_M125_ToAA_M10/SUSY_VBFH_M125_ToAA_M10_To2Mu2Tau_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/HToAATo2Mu2Tau/SUSY_VBFH_HToAATo2Mu2Tau/VBFH_M125_ToAA_M10/SUSY_VBFH_M125_ToAA_M10_To2Mu2Tau_proc_card.dat
@@ -1,5 +1,5 @@
 import model NMSSMHET --modelname
 
-generate p p  > h01 j j $$ W+ W- Z  QCD=0, (h01 > a01 a01, a01 > tau+ tau-, a01 > mu+ mu-)
+generate p p > h01 j j $$ Z W- W+ h01 / a  HIG=0 QCD=0 , (h01 > a01 a01, a01 > tau+ tau- , a01 > mu+ mu-)
 
 output SUSY_VBFH_M125_ToAA_M10_To2Mu2Tau -nojpeg
--- a/bin/MadGraph5_aMCatNLO/cards/production/13TeV/HToAATo2Mu2Tau/SUSY_VBFH_HToAATo2Mu2Tau/VBFH_M125_ToAA_M10/SUSY_VBFH_M125_ToAA_M10_To2Mu2Tau_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13TeV/HToAATo2Mu2Tau/SUSY_VBFH_HToAATo2Mu2Tau/VBFH_M125_ToAA_M10/SUSY_VBFH_M125_ToAA_M10_To2Mu2Tau_proc_card.dat
@@ -1,5 +1,5 @@
 import model NMSSMHET --modelname
 
-generate p p > h01 j j $$ Z W- W+ h01 / a  HIG=0 QCD=0 , (h01 > a01 a01, a01 > tau+ tau- , a01 > mu+ mu-)
+generate p p > h01 j j $$ Z W- W+ / a  HIG=0 QCD=0 , (h01 > a01 a01, a01 > tau+ tau- , a01 > mu+ mu-)
 
 output SUSY_VBFH_M125_ToAA_M10_To2Mu2Tau -nojpeg


Fengwangdong:
Thanks Meng for the comments. Because tau reconstruction and identification from PFJets are commissioned efficiently within |eta| < 2.5, we don't motivate for VBS process. 

I just found that my current VBF-H gen-cards resulted in a cross section of "5.069e+11 +/- 3.756e+09 pb" from my local test. It does not make physics sense. Please don't merge this pull request yet. I am debugging it and will commit an updated pull request after this problem is solved. In case any of you has a straightforward idea about the solution, please let me know.

menglu21:
Current syntax has opened contribution h03/h02/h01 couplings at the same time, are all these contributions needed in your case.

Fengwangdong:
OK, I disabled those couplings by adding "HIG=0". My local test shows the cross section is "3.940e-07 +- 1.710e-08 pb" which looks reasonable now. Thank you for your suggestions. Please let me know if we may merge this pull request, so that I can ask the Higgs MC contacts for the sample production

sihyunjeon:
Hi,
You are not computing decay widths of the `A` particles. so the cross sections do not mean anything without it...

DickyChant:
out of my curiosity: 
image
is this kind of diagram intended as well?

menglu21:
> OK, I disabled those couplings by adding "HIG=0". My local test shows the cross section is "3.940e-07 +- 1.710e-08 pb" which looks reasonable now. Thank you for your suggestions. Please let me know if we may merge this pull request, so that I can ask the Higgs MC contacts for the sample production

besides the comments from others, one minor comment, since you are using decay chain, you need to use "True = cut_decay" to turn on the cut

Fengwangdong:
> Hi, You are not computing decay widths of the `A` particles. so the cross sections do not mean anything without it since branching ratios are going to be just wrong...

The decay widths of A are specified in the NMSSMHET model packet, in which the "particles.py" has.

Fengwangdong:
> out of my curiosity: image is this kind of diagram intended as well?

This is allowed by the 2HDM+S model. The VBF is not only referred to vector boson (W/Z) fusion, but also to an off-shell pseudo-scalar boson coupling with an offshell vector boson.

DickyChant:
> > out of my curiosity: image is this kind of diagram intended as well?
> 
> This is allowed by the 2HDM+S model. The VBF is not only referred to vector boson (W/Z) fusion, but also to an off-shell pseudo-scalar boson coupling with an offshell vector boson.

OK makes sense if this is the convention for 2HDM studies.

Fengwangdong:
> then BR will be



> > > Hi, You are not computing decay widths of the `A` particles. so the cross sections do not mean anything without it since branching ratios are going to be just wrong...
> > 
> > 
> > The decay widths of A are specified in the NMSSMHET model packet, in which the "particles.py" has.
> 
> Yes model has it but you are not asking madgraph to use the proper total width. Partial width would be correct but not the total width, then BR will be wrong

Could you please point for me where to specify this width? I checked some example cards eg. HAA4b, but I did not see anywhere about it

DickyChant:
> > then BR will be
> 
> > > > Hi, You are not computing decay widths of the `A` particles. so the cross sections do not mean anything without it since branching ratios are going to be just wrong...
> > > 
> > > 
> > > The decay widths of A are specified in the NMSSMHET model packet, in which the "particles.py" has.
> > 
> > 
> > Yes model has it but you are not asking madgraph to use the proper total width. Partial width would be correct but not the total width, then BR will be wrong
> 
> Could you please point for me where to specify this width? I checked some example cards eg. HAA4b, but I did not see anywhere about it

usually set decay auto like this: https://github.com/cms-sw/genproductions/blob/e43c5f10c3dfc19d800aa000600ac8c40a8a891a/bin/MadGraph5_aMCatNLO/cards/diquark_Suu_ut/diquark_Suu_ut_customizecards.dat#L5

Fengwangdong:
> > OK, I disabled those couplings by adding "HIG=0". My local test shows the cross section is "3.940e-07 +- 1.710e-08 pb" which looks reasonable now. Thank you for your suggestions. Please let me know if we may merge this pull request, so that I can ask the Higgs MC contacts for the sample production
> 
> besides the comments from others, one minor comment, since you are using decay chain, you need to use "True = cut_decay" to turn on the cut

Well, this parameter might be tricky especially for boosted topologies. There are ∆R requirements for leptons and photos which may kill substantial signal events (eg. boosted tau pair decays to collimated leptons, and pion0 from tau decays to collimated photons). This may bias our signal efficiency study. I think it is safer to keep it off and let the events be filtered by our fiducial selections later. Because our ggF samples from previous production also had it OFF. It may be better to keep the consistency.

Fengwangdong:
> > > then BR will be
> > 
> > 
> > > > > Hi, You are not computing decay widths of the `A` particles. so the cross sections do not mean anything without it since branching ratios are going to be just wrong...
> > > > 
> > > > 
> > > > The decay widths of A are specified in the NMSSMHET model packet, in which the "particles.py" has.
> > > 
> > > 
> > > Yes model has it but you are not asking madgraph to use the proper total width. Partial width would be correct but not the total width, then BR will be wrong
> > 
> > 
> > Could you please point for me where to specify this width? I checked some example cards eg. HAA4b, but I did not see anywhere about it
> 
> usually set decay auto like this:
> 
> https://github.com/cms-sw/genproductions/blob/e43c5f10c3dfc19d800aa000600ac8c40a8a891a/bin/MadGraph5_aMCatNLO/cards/diquark_Suu_ut/diquark_Suu_ut_customizecards.dat#L5

I tried but the gridpack generator complained:

```> WARNING: invalid set command param_card DECAY 9936661 auto```

**Comment by DickyChant** [comment]
> > > > then BR will be
> > > 
> > > 
> > > > > > Hi, You are not computing decay widths of the `A` particles. so the cross sections do not mean anything without it since branching ratios are going to be just wrong...
> > > > > 
> > > > > 
> > > > > The decay widths of A are specified in the NMSSMHET model packet, in which the "particles.py" has.
> > > > 
> > > > 
> > > > Yes model has it but you are not asking madgraph to use the proper total width. Partial width would be correct but not the total width, then BR will be wrong
> > > 
> > > 
> > > Could you please point for me where to specify this width? I checked some example cards eg. HAA4b, but I did not see anywhere about it
> > 
> > 
> > usually set decay auto like this:
> > https://github.com/cms-sw/genproductions/blob/e43c5f10c3dfc19d800aa000600ac8c40a8a891a/bin/MadGraph5_aMCatNLO/cards/diquark_Suu_ut/diquark_Suu_ut_customizecards.dat#L5
> 
> I tried but the gridpack generator complained:
> 
> `> WARNING: invalid set command param_card DECAY 9936661 auto`

ofc you need to change the particle ID :)

**Comment by Fengwangdong** [comment]
> > > > > then BR will be
> > > > 
> > > > 
> > > > > > > Hi, You are not computing decay widths of the `A` particles. so the cross sections do not mean anything without it since branching ratios are going to be just wrong...
> > > > > > 
> > > > > > 
> > > > > > The decay widths of A are specified in the NMSSMHET model packet, in which the "particles.py" has.
> > > > > 
> > > > > 
> > > > > Yes model has it but you are not asking madgraph to use the proper total width. Partial width would be correct but not the total width, then BR will be wrong
> > > > 
> > > > 
> > > > Could you please point for me where to specify this width? I checked some example cards eg. HAA4b, but I did not see anywhere about it
> > > 
> > > 
> > > usually set decay auto like this:
> > > https://github.com/cms-sw/genproductions/blob/e43c5f10c3dfc19d800aa000600ac8c40a8a891a/bin/MadGraph5_aMCatNLO/cards/diquark_Suu_ut/diquark_Suu_ut_customizecards.dat#L5
> > 
> > 
> > I tried but the gridpack generator complained:
> > `> WARNING: invalid set command param_card DECAY 9936661 auto`
> 
> ofc you need to change the particle ID :)

Thanks. I added two lines in the ```customizecards.py```:
```
set param_card decay 36 auto
set param_card decay 46 auto
```

The 36 and 46 are pdgIDs for ``A``. But the program stuck at this step forever:

```
INFO: get decay diagram for a01 
Vertexlist of this model has not been searched.Automatically run the model.find_vertexlist()
Found 4 stable particles
INFO: get decay diagram for a02 
INFO:     current estimated error: 0.00746220179221 go to 4-body decay: 
100 / 3168: 27s
200 / 3168: 50s
300 / 3168: 73s
400 / 3168: 98s
500 / 3168: 124s
600 / 3168: 139s
700 / 3168: 155s
800 / 3168: 175s
900 / 3168: 195s
1000 / 3168: 219s
1100 / 3168: 233s
1200 / 3168: 244s
1300 / 3168: 265s
1400 / 3168: 284s
1500 / 3168: 308s
1600 / 3168: 327s
1700 / 3168: 346s
1800 / 3168: 373s
1900 / 3168: 393s
2000 / 3168: 411s
2100 / 3168: 433s
2200 / 3168: 460s
2300 / 3168: 477s
2400 / 3168: 494s
2500 / 3168: 511s
2600 / 3168: 530s
2700 / 3168: 543s
2800 / 3168: 567s
2900 / 3168: 585s
3000 / 3168: 605s
3100 / 3168: 619s
100 / 882: 640s
200 / 882: 650s
300 / 882: 658s
400 / 882: 668s
500 / 882: 677s
600 / 882: 690s
700 / 882: 697s
800 / 882: 702s
Pass to numerical integration for computing the widths:
INFO: More info in temporary files:
    /genproductions/bin/MadGraph5_aMCatNLO/SUSY_VBFH_M125_ToAA_M10_To2Mu2Tau/SUSY_VBFH_M125_ToAA_M10_To2Mu2Tau_gridpack/work/processtmp/tmp1N9Jf9/temp_decay/index.html
```

sihyunjeon:
but 46 is a02 not a01. you only have a01 in your process definition. why does this matter?

sihyunjeon:
you also do not touch the mass of a02 so it is the ufo default mass (which i guess it is large given that there are many diagrams being tested), not sure what you want to do with a02 here

Fengwangdong:
Oh right, now it works without ``a02``. But the ``working on Subprocesses`` seems to take much longer time than the case without adding the decay width of ``a01``. 

eg: ``INFO:  Idle: 1567,  Running: 8,  Completed: 65``

There are thousands in idle at the beginning, while before (without decay width line) there were only ~400 idle at the beginning. Is this normal?

sihyunjeon:
> Oh right, now it works without `a02`. But the `working on Subprocesses` seems to take much longer time than the case without adding the decay width of `a01`.
> 
> eg: `INFO: Idle: 1567, Running: 8, Completed: 65`
> 
> There are thousands in idle at the beginning, while before (without decay width line) there were only ~400 idle at the beginning. Is this normal?

could be, especially if your actual total width is hugely different from the model default value. which then i expect the cross sections could be different quite a lot

jingyucms:
Just one comment: 
https://github.com/cms-sw/genproductions/pull/2848/files
This is the request for our ggF sample in which the width of a is set to a very small value as it is much smaller than the experimental resolution anyway

sihyunjeon:
@jingyucms the width here, problem is with the cross section value. Not related to the experimental features at all.

DickyChant:
i think we need to clarify several things before moving forward.

1: I think the signal is used for assuming SM Higgs (M 125) and the analysis will set upper limits on the SM Higgs decay BR while the cross section to be injected comes from the SM ones @Fengwangdong @jingyucms is this the case?
2. If so, cross section might not be the top 1 priority to be checked here. BUT, for this VBF do you feel consistent to include those BSM contribution then? 
3. If not, we should really think about this signal

sihyunjeon:
Maybe this should be clarified

Are you using the cross sections in the result or not?

I assumed you would as you mentioned "cross section is too large i fixed the syntax"

If you dont use it, you just need to hardcode your width to a reasonable value as jingyu did in the PR

sihyunjeon:
Ah wrote the same thing at the same time :) based on the comment discussing "cross section being reasonable or not" i was thinking you actually use it as a theory reference. If this is the case setting proper width matters

jingyucms:
We set limits on the model-independent BR(h->aa->mumutautau) and model-dependent Br(h->aa). We take h cross-section and Br(aa->mumutautau) from theorists. So we don't care about the MG cross-section values. We just need SM VBF production processes and 100% h->mumutautau decay.

Fengwangdong:
Hi all, I was just wondering if this cross section value could result in unreasonable event weight. But indeed we don't use this cross section value computed by MG5 in our analysis. So it does not matter whether we set the `A` width or not. I believe the current commit should be OK for merging.

sihyunjeon:
You should set the width to a reasonable value, or do you know what the width default setting is?

DickyChant:
> Hi all, I was just wondering if this cross section value could result in unreasonable event weight. But indeed we don't use this cross section value computed by MG5 in our analysis. So it does not matter whether we set the `A` width or not. I believe the current commit should be OK for merging.

Are you using genweight for such unweighted sample?

DickyChant:
And for this merge request, do I understand it correctly that you are OK to non SM VBF contribution? Or NOT?

Fengwangdong:
OK thank you for your detailed comments. Let me summarize below:

* We don't need the cross section computed by MG5 in general. We just use the shape of the generated unweighted events, NOT using absolute gen-weight values for normalization. Hence my previous concern about this xsec is not an issue anymore. The width is also not needed.

* We DO need the BSM contributions in the syntax because they are predicted by the NMSSM model.

Please let me know if it's clear now.

sihyunjeon:
Hello again,

You DO NEED the width : at least with experimental justifications as Jingyu mentioned above.

I do not know the model and especially its model default settings, but if your model default value for the width is 100TeV, this doesn't make sense and also would create an issue of the results obviously. You DO HAVE TO CHECK the width of UFO model setting which you are completely neglecting in this PR and the customization card.

DickyChant:
I’d suggest to align with previous request, at least check what was done for width there…

Fengwangdong:
The previous request did not specify the width in the customize card. The default width value is 2.495587e-04.

Fengwangdong:
The width of `A` is added in the customize card

Fengwangdong:
> Hi, I think the width situation is clear now from above comment. one more question, why onshell "h01" in s-channel is excluded.

Hi, because it was already included in the ggF samples that we required before [1]

[1] https://github.com/cms-sw/genproductions/pull/2848/files#diff-9bc412e774b63ed9dc89a24cdc243ebaf5c5f16e3f1dc43a7e75a14fef71050d

menglu21:
Hi @Fengwangdong I would say these should be different things cuz they are two different processes. I would suggest to have a check on that, 
generate p p > h01 j j $$ Z W- W+ / a  HIG=0 QCD=0 -> 1166 diagrams
generate p p > h01 j j $$ Z W- W+ h01 / a  HIG=0 QCD=0 -> 1110 diagrams
I'm not sure how different they would be between these two syntax, so would be safer to check the impact to avoid losing some important effect in the analysis

Fengwangdong:
Hi @menglu21, thanks for pointing it out. The associated h01 (offshell) was disabled [1], which should not matter given its negligible effect in our analysis. It is also OK to enable it [2], which I just committed.

[1] ``generate p p > h01 j j $$ Z W- W+ h01 / a HIG=0 QCD=0``
[2] ``generate p p > h01 j j $$ Z W- W+ / a HIG=0 QCD=0``

PR #3633, opened 2024-02-02
Synchronize mg352 to master

stahlleiton:
Synchronize the branch mg352 with the master without changing MadGraph5 version and patches for v3.5.2.
This PR also add the script for kT smearing of Gamma-UPC+MadGraph LHE events (provided by Gamma-UPC authors).


menglu21:
Hi @stahlleiton do you plan to update the GammaUPC part, right? if yes, can you please update your local repo first and then update that part, currently this PR includes too many updates which is hard to check

stahlleiton:
The idea of this PR was to simply align the commits in master to this branch (which is behind)

PR #3642, opened 2024-02-21
Fixing WWJ process (Run3)

mlizzo:
Dear @menglu21 , @bbilin , @covarell 

This is a PR to fix the WWJ Powheg process, which is taken from the POWHEG-BOX-RES repository. The code needs some external libraries that are now properly added to the LD_LIBRARY_PATH variable in the runcmsgrid script. Additionally, 2-loops amplitude computations are soft-linked from cvmfs when the code is compiled. In order to speed up the event generation, I have drastically reduced the set of alternative PDFs, where only central values are retained. To this extent, an additional flag was added when the make_rwl.py script is executed, similarly to what was done for the X0jj Powheg process. Please have a look at it and let me know if other modifications are needed. When approved, I would like to add a section in the CMS Powheg Twiki page to provide some documentation on how to run this sample, if possible.

Best regards,
Mattia


menglu21:
Hi @mlizzo thanks for the pr. so people use wwj sample are fine with the reduced pdfs, right

mlizzo:
Hi @menglu21 , thanks for your reply. Let me add some people interested to this sample (@fmanteca , I couldn't add Guillelmo but I will let him know), maybe they can comment on this

covarell:
For the time being, it is OK to proceed. Of course, it is unpractical to have a boolean ${forXXX} for each process which requesting custom PDF sets, we should streamline this.
@agrohsje @jshin96

mlizzo:
Hi, I've uploaded a small modification which was necessary to compile the code with the `slc7_amd64_gcc10` architecture. Is there a smart way to automatically grasp the correct cmake version? The one that comes with lxplus is quite old and is not compatible with WWJ code. It would be great if we could have the cmake package linked to cmssw and not as an external code, but I don't know if this is feasible

mlizzo:
@covarell @agrohsje @jshin96 I have modified the tarball creation script, I think that the first part (where some powheg files are copied) was broken (see the last commit). I also added some WWJ-specific parts, I would really appreciate if someone could review this PR (adding @menglu21 and @bbilin as well), thank you in advance

PR #3643, opened 2024-02-21
Created directory for TT_SStt, added cards

dpilipov:


dpilipov:
Created directory for TT_SStt with all the cards for TT->SStt pair production.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TT_SStt/TT_SStt_run_card.dat
```diff
@@ -0,0 +1,255 @@
...
+1        = lpp2    ! beam 2 type
+6500.0     = ebeam1  ! beam 1 total energy in GeV
+6500.0     = ebeam2  ! beam 2 total energy in GeV
+# To see polarised beam options: type "update beam_pol"
+
+#*********************************************************************
+# PDF CHOICE: this automatically fixes also alpha_s and its evol.    *
+#*********************************************************************
+'lhapdf' = pdlabel     ! PDF set
+325500 = lhaid
```
menglu21, line 43: you can use "$DEFAULT_PDF_SETS = lhaid" since the script will catch the flavor scheme and choose the pdfs

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TT_SStt/TT_SStt_run_card.dat
```diff
@@ -0,0 +1,255 @@
...
+#
+#*********************************************************************
+# BW cutoff (M+/-bwcutoff*Gamma) ! Define on/off-shell for "$" and decay
+#*********************************************************************
+  15.0  = bwcutoff      ! (M+/-bwcutoff*Gamma)
+ #*********************************************************************
+ # Apply pt/E/eta/dr/mij/kt_durham cuts on decay products or not
+ # (note that etmiss/ptll/ptheavy/ht/sorted cuts always apply)
+ #*********************************************************************
+   False  = cut_decays    ! Cut decay products
```
menglu21, line 91: since you are using decay chain, you need to change it to "True" if you want the following cuts to be applied on the decay objects

bbilin:
@dpilipov why do you push the param_cards? Is it different from the one generated by the model?

menglu21 (commented):
besides the comments inline, please create a script to implement the different mass point, then you can remove those txt files

dpilipov:
As per your request, added mass points script to proc_card and deleted the unnecessary mass points txt files and edited the run_card.dat.  Customizecards is still used to set the general parameter values, except for the mass points.

Please let me know if there is anything else needed.  Thank you!

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TT_SStt/TT_SStt_run_card.dat
```diff
@@ -0,0 +1,280 @@
...
+# Parameter of the systematics study
+#  will be used by SysCalc (if installed)
+#**************************************                                  
+#
+#0.5 1 2 = sys_scalefact  # factorization/renormalization scale factor
+#0.5 1 2 = sys_alpsfact  # \alpha_s emission scale factors
+#30 50 = sys_matchscale # variation of merging scale
+# PDF sets and number of members (0 or none for all members).
+#CT10nlo.LHgrid = sys_pdf # matching scales
+# MSTW2008nlo68cl.LHgrid 1  = sys_pdf
```
menglu21, line 280: these lines can be removed

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TT_SStt/TT_SStt_proc_card.dat
```diff
@@ -0,0 +1,185 @@
...
+launch
+set mass 6000006 1500
+set mass 6100001 250
+launch
+set mass 6000006 1500
+set mass 6100001 300
+launch
+set mass 6000006 1500
+set mass 6100001 350
+launch
```
menglu21, line 185: for all the mass point, you can make a separate script to implement them instead of putting them together with proc card

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TT_SStt/TT_SStt_customizecards.dat
```diff
@@ -0,0 +1,10 @@
+set param_card decay 6000006 0.1
+set param_card decay 6100001 0.01
```
menglu21, line 2: will you use the cross section from sample production in your analysis, if yes, are those width chosen randomly small

dpilipov:
Dear genproductions, Please see my questions/answers below:Sent from my iPhoneOn Mar 15, 2024, at 7:18 AM, Meng_Lu ***@***.***> wrote:﻿
@menglu21 commented on this pull request.



In bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TT_SStt/TT_SStt_run_card.dat:
> +#
+# Parameter of the systematics study
+#
+#0.5 1 2 = sys_scalefact  # factorization/renormalization scale factor
+#0.5 1 2 = sys_alpsfact  # \alpha_s emission scale factors
+#30 50 = sys_matchscale # variation of merging scale
+#CT10nlo.LHgrid = sys_pdf # matching scales

these lines can be removedDP: will do. 



In bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TT_SStt/TT_SStt_proc_card.dat:
> +set mass 6000006 1500
+set mass 6100001 300
+launch
+set mass 6000006 1500
+set mass 6100001 350
+launch

for all the mass point, you can make a separate script to implement them instead of putting them together with proc cardDP:  can you point me to an example for another search?  I couldn’t find an example in genproductions.  That would be very helpful and much appreciated!



In bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TT_SStt/TT_SStt_customizecards.dat:
> +set param_card decay 6000006 0.1
+set param_card decay 6100001 0.01

will you use the cross section from sample production in your analysis, if yes, are those width chosen randomly smallDP.  These parameter values are based on the narrow width approximation and discussion in Lucas’ paper 1907.05929. I will explain.  6000006 is the VLQ decay width expressed as a percentage of its mass. Hence, it corresponds to roughly the largest width for pair production in this model to also allow for the narrow width approximation.  For large VLQ mass this is a fairly wide decay width. I should also mention that 0.1 (hence 100GeV for a 1000GeV VLQ mass, for example) is only the ‘starting’ point for this model. The final decay width is ultimately determined by many other parameters and ends up being roughly three times this value, in our experience so far. The second is for the new scalar and also expressed as a percentage but relative to the mass of the scalar. It is narrower to correspond to the theoretical value for the diphoton decay channel. (See the paper.)So, to answer your question, we use the cross section for the pair production of the VLQs - which is a function of the VLQ mass - to obtain the Xsec for the sample. (See my last update at the VHF, https://indico.cern.ch/event/1355727/.). Even though we are using the LO version of the model, the resulting Xsec are quite close to the corresponding NLO values discussed in the paper and also in our last update at the VHF. But if you feel we need to change these values, just let us know!

dpilipov:
Sorry - let me reformat my question / response from above so that it is easier to read.  

Question regarding mass points: Can you point me to an example of a mass-points script for another search?  I couldn’t find an example in genproductions.  That would be very helpful and much appreciated!

Response to your question regarding decay widths:  The decay width parameter values (6000006 and 6100001) are based on the narrow width approximation and discussion in Lucas’ paper 1907.05929. I will explain.  

6000006 is the VLQ decay width expressed as a percentage of its mass. Hence, it corresponds to roughly the largest width for pair production in this model to also allow for the narrow width approximation.  For large VLQ mass this is a fairly wide decay width. I should also mention that 0.1 (hence 100GeV for a 1000GeV VLQ mass, for example) is only the ‘starting’ point for this model. The final decay width is ultimately determined by many other parameters and ends up being roughly three times this value, in our experience so far. The second is for the new scalar and also expressed as a percentage but relative to the mass of the scalar. It is narrower to correspond to the theoretical value for the diphoton decay channel. (See the paper.)

So, to answer your question, we use the cross section for the pair production of the VLQs - which is a function of the VLQ mass - to obtain the Xsec for the sample. (See my last update at the VHF, https://indico.cern.ch/event/1355727/, which lists cross sections for several VLQ masses). Even though we are using the LO version of the model, the resulting Xsec are quite close to the corresponding NLO values discussed in the paper and also in our last update at the VHF. 

But if you feel we need to change these values, just let us know!  Looking forward to hearing your recommendations!

sihyunjeon:
For the first question,

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/BulkGraviton_hh_tttt/generateCards.py

Something like this is needed. You are packing up multiple mass parameter settings in gridpack but you should have one gridpack per parameter point

dpilipov:
Added script for generating mass points.  Deleted comment lines in run_card as per gen expert request.

PR #3644, opened 2024-02-26
Modify PDF choice for UL Run2 and add gridpack generation script

sslomte:
Modified the PDF lhaid to 325500 for UL Run2. Added a python script to make datacards from template card, and to generate gridpacks for various mass points.


sihyunjeon:
not sure what readme.md is doing there, can you relocate?

sslomte:
Thanks for the catch. The readme.md and make_datacards_gridpacks.py files are now relocated to cards/monoHiggs_ZpBaryonic/.

PR #3645, opened 2024-02-26
Add DY, TTbar and W MadGraph input cards for PbPb 5.36 TeV

stahlleiton:
Add DY, TTbar and W MadGraph input cards for PbPb 5.36 TeV.

Used 2018 (Run 2) PbPb production cards and changed the beam energy to 5.36 TeV as in Run 3.

@DickyChant 

This PR requires: https://github.com/cms-sw/genproductions/pull/3650


bin/MadGraph5_aMCatNLO/cards/production/PbPb_5p36TeV/dyellell012j_5f_NLO_FXFX/dyellell012j_5f_NLO_FXFX_run_card.dat
```diff
@@ -0,0 +1,150 @@
...
+#***********************************************************************
+    1   = lpp1    ! beam 1 type (0 = no PDF)
+    1   = lpp2    ! beam 2 type (0 = no PDF)
+ 2680   = ebeam1  ! beam 1 energy in GeV
+ 2680   = ebeam2  ! beam 2 energy in GeV
+#***********************************************************************
+# PDF choice: this automatically fixes also alpha_s(MZ) and its evol.  *
+#***********************************************************************
+ lhapdf    = pdlabel   ! PDF set                                     
+ $DEFAULT_nPDF_SETS = lhaid     ! if pdlabel=lhapdf, this is the lhapdf number
```
menglu21, line 58: won't you want to use the same pdf with that in the powheg card

DickyChant: I think we want to steer in the nuclear PDFs as PRed in https://github.com/cms-sw/genproductions/pull/3586

@stahlleiton do we properly define `DEFAULT_nPDF_SETS` somewhere? Should be implemented like this I guess: https://github.com/cms-sw/genproductions/blob/2f1af7387ec2cf1d3b0e218011d8a350efdba5d7/bin/MadGraph5_aMCatNLO/Utilities/gridpack_helpers.sh#L1C1-L32C2

Or just use `DEFAULT_PDF_SETS`? Am I missing some part?

stahlleiton: I was wondering whether to add the PDF explicitly (i.e. 904400) or keep it with the default value.

In my local gridpack generation, I was replacing the pdf list by doing:
cp [bin/MadGraph5_aMCatNLO/cards/production/PbPb_5p36TeV/pdflist_5f_run3.dat](https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/PbPb_5p36TeV/pdflist_5f_run3.dat)  [MetaData/pdflist_5f_run3.dat](https://github.com/cms-sw/genproductions/blob/master/MetaData/pdflist_5f_run3.dat)

So I assume when running submit_gridpack_generation.sh, then the default PDF would be: [904400](https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/PbPb_5p36TeV/pdflist_5f_run3.dat#L1)

DickyChant: I think this is implementation details lol, the bits I posted is the PDF setup which replaces the `DEFAULT_PDF_SETS`.

Do we have the produced gridpack somewhere? We could easily check if the PDF is handled correctly by checking the process directory.

stahlleiton: ok, good point. I think DEFAULT_nPDF_SETS is not defined. I will make a PR that automatically grabs the nPDF (without having to copy it) if using DEFAULT_nPDF_SETS variable in the process card

DickyChant: Fantastic! Should be very helpful!

PR #3652, opened 2024-02-29
Fix for generic X->YH->HHH cards for Run 3

ferencek:
This is a follow-up to https://github.com/cms-sw/genproductions/pull/3646 which had an incomplete requirement on m_Y in the card generation script (m_Y >= 2*m_H requirement was missing).


ferencek:
Hi @menglu21, I made a silly mistake in https://github.com/cms-sw/genproductions/pull/3646 and this PR adds a missing mass requirement as explained above.

PR #3653, opened 2024-03-04
rm getting cross_section from generator field in mcm

efeyazgan:


efeyazgan:
This is done following https://github.com/cms-PdmV/cmsPdmV/pull/1126/files

PR #3661, opened 2024-03-12
dymumu012j_5f_NLO_FXFX_M105to160 

qyguo:
It is one of the PR#3531 
We are requesting one dedicated MG5 background samples for H--> mumu analysis #3531 on behalf of Hmumu team. 
The cards have been synchronized with DYto2L-2Jets_MLL-50_amcatnloFXFX-pythia8.
The differences are the mumu final states and the mass window [105,160].
The distributions are checked by @Soumyatifr. More details could be found under the PR #3531 
We would greatly appreciate it if we could get the approval.


menglu21:
look good to me, @bbilin please have a check

PR #3670, opened 2024-03-25
Minor update to collision energy for Run3 PbPb cards

stahlleiton:
This PR introduces a minor update of the collision energy for all Run 3 PbPb cards, changing it from 5360 GeV to 5362 GeV, to slightly match better the real collision energy used during Run3 data taking (i.e. 13600*82/208 = 5361.54 GeV). It also adds two extra starlight cards of UPC Jpsi to dielectrons.


bbilin:
@stahlleiton Fine for us, but is it really worth to change 2 GeV the sqrt(s) ? Were there any samples/packs produced with the current setup? If not, I am ready to merge this.

stahlleiton:
Not yet, there has been any samples produced since the cmsDriver chains for UPC had not been fully implemented yet

PR #3682, opened 2024-04-09
Add genfragments for STARlight, SuperChic and UPCgen

stahlleiton:
This PR adds the genfragments for official PbPb 5.36 TeV (Run3) production of STARlight, SuperChic and UPCgen generators in the directory: genfragments/PbPb_5p36TeV/ .

The PYTHIA8 tune "pythia8CP5Settings" is excluded from the UPC gen fragments since this is tuned for hadronic events.

@sarteagae @DickyChant @menglu21 @bbilin

### DIFF
--- a/genfragments/PbPb_5p36TeV/Starlight/STARLIGHT_QED_tau-fragment.py
+++ b/genfragments/PbPb_5p36TeV/Starlight/STARLIGHT_QED_tau-fragment.py
@@ -1,0 +1,43 @@
+import FWCore.ParameterSet.Config as cms
+
+externalLHEProducer = cms.EDProducer("ExternalLHEProducer",
+    args = cms.vstring('/cvmfs/cms.cern.ch/phys_generator/gridpacks/RunIII/5p36TeV/starlight/starlight_ditau_el8_amd64_gcc11_CMSSW_13_0_18_HeavyIon_tarball.tgz'),
+    nEvents = cms.untracked.uint32(1),
+    numberOfParameters = cms.uint32(1),
+    outputFile = cms.string('cmsgrid_final.lhe'),
+    scriptName = cms.FileInPath('GeneratorInterface/LHEInterface/data/run_generic_tarball_cvmfs.sh')
+)
+
+#Link to datacards:
+#https://github.com/cms-sw/genproductions/blob/master/bin/Starlight/production/PbPb_5p36TeV/starlight_ditau.in
+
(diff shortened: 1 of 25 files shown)


genfragments/PbPb_5p36TeV/Starlight/STARLIGHT_QED_tau-fragment.py
```diff
@@ -0,0 +1,43 @@
...
+        parameterSets = cms.vstring('pythia8CommonSettings','pythia8PSweightsSettings','skip_hadronization')
+    ),
+    comEnergy = cms.double(5362.0),
+    filterEfficiency = cms.untracked.double(1.0),
+    maxEventsToPrint = cms.untracked.int32(1),
+    pythiaHepMCVerbosity = cms.untracked.bool(False),
+    pythiaPylistVerbosity = cms.untracked.int32(1)
+)
+
+ProductionFilterSequence = cms.Sequence(generator)
```
menglu21, line 43: why tau process doesn't need the filter like electron and muon channels

stahlleiton: I could add a filter. Just wasn't sure how to handle since the tau is not a final product

DickyChant: Interesting point and I might have some hint...

https://github.com/cms-sw/cmssw/blob/6e82cc8aa6fff019b94eb34a5b5297c2eaa62573/GeneratorInterface/Core/interface/EmbeddingHepMCFilter.h#L10 this is the so-called embeddinghepmcfilter that should be used for tau embedding samples? but i dont know if it plays well with tauola

DickyChant: Here exo has an example usage: https://gitlab.cern.ch/cms-exo-mci/EXO-MCsampleRequests/-/blob/master/genFragments/Hadronizer/13p6TeV/GluGluPhiTauMuTauH/GluGluPhiTauMuTauH_M100_Fragment.py?ref_type=heads

DickyChant: Just checked and this doesn't seem to work.. since `tauola` runs the filter being applied... Need another handle then...

PR #3638, opened 2024-02-12
ttH and tHq at NLO with BSM reweighting for BSM CP structure of top Yukawa

FloMau:
Adding cards for ttH and tHq production at NLO, including reweighting to different CP-properties of the top Yukawa coupling. 

Tagging @sihyunjeon and @DickyChant for the error we discussed today:
When I run the gridpack generation step as described [here](https://twiki.cern.ch/twiki/bin/view/CMS/QuickGuideMadGraph5aMCatNLO#Quick_tutorial_on_how_to_produce), for both of the processes it ends up with the same error once it enters the reweighting calculation for the first parameter configuration.

I would greatly appreciate your help to solve this issue, many thanks in advance! Let me know if you need further information. 

Short output containing the error is given here, the full log file is attached.
```
INFO: Running Reweighting
change mode NLO
change rwgt_dir rwgt
launch --rwgt_name=kHtt_0p5_kAtt_1_alpha_45
INFO: detected model: HC_NLO_X0_UFO. Loading...
INFO: generating the square matrix element for reweighting
INFO: generate p p > t x0 b~ j $$ w+ w-   --no_warning=duplicate;define pert_QCD = -4 -3 -2 -1 1 2 3 4 21;add process p p > t x0 b~ j  pert_QCD $$ w+ w-   --no_warning=duplicate;add process p p > t~ x0 b j $$ w+ w-   --no_warning=duplicate;define pert_QCD = -4 -3 -2 -1 1 2 3 4 21;add process p p > t~ x0 b j  pert_QCD $$ w+ w-   --no_warning=duplicate;
INFO: Done 68.59
INFO: generate p p > t x0 b~ j $$ w+ w- [ virt=QCD] ;add process p p > t~ x0 b j $$ w+ w- [ virt=QCD] ;
INFO:  Idle: 0,  Running: 75,  Completed: 0 [  5m 0s  ]
compile OLP rw_mevirt
Do you want to edit a card (press enter to bypass editing)?
/------------------------------------------------------------\
|  1. param : param_card.dat                                 |
|  9. plot  : plot_card.dat                                  |
\------------------------------------------------------------/
 you can also
   - enter the path to a valid card or banner.
   - use the 'set' command to modify a parameter directly.
     The set option works only for param_card and run_card.
     Type 'help set' for more information on this command.
   - call an external program (ASperGE/MadWidth/...).
     Type 'help' for the list of available command
 [0, done, 1, param, 9, plot, enter path]
[...]

PR #3686, opened 2024-04-12
tHq NLO CP cards

FloMau:
NLO tHq cards for CP analyses.
Replaces tHq-part of previous MR https://github.com/cms-sw/genproductions/pull/3638 to circumvent problematic reweighting + MadSpin at NLO. 

Here, seven parameter configurations are chosen in 2D plane of CP-even and -odd top Yukawa coupling modifiers. Semi leptonic and hadronic top decays are separated.

The cards can be generated with `python writeCards.py` from the ones in the `originalCards/` directory.


FloMau:
Dear @menglu21 and @bibilin, 

could you please have a look at these cards and check? We would like to move forward with the sample request since the analysis has a tough timeline. 

Thanks!

menglu21:
Hi @FloMau please only keep the writeCard.py and the template cards

FloMau:
> Hi @FloMau please only keep the writeCard.py and the template cards

Hi @menglu21, sure, I removed the explicit cards. Is the rest OK?

FloMau:
Dear @menglu21 and @bbilin,

could you please have a look at these cards and let us know if we can move forward with the production requests?

Thanks!

PR #3687, opened 2024-04-12
typo in SM ZH coupling

jdavis36:
fix a small typo in the cards. Otherwise it will cause a segfault in jhugen.

### DIFF
--- a/bin/JHUGen/cards/decay/ZZ2l2any_withtaus_filter4l.input
+++ b/bin/JHUGen/cards/decay/ZZ2l2any_withtaus_filter4l.input
@@ -1,1 +1,1 @@
-DecayMode1=8 DecayMode2=9 FilterNLept=4 WriteFailedEvents=2
+DecayMode1=8 DecayMode2=9 NLepMin=4 LHAPDF=NNPDF31_lo_as_0130/NNPDF31_lo_as_0130.info WriteFailedEvents=2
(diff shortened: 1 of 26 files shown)


jdavis36:
I also added the pdf set for the decay cards and changed the syntax for the HZZ 4lepton filters.

bin/JHUGen/cards/decay/ZZ2l2any_withtaus_filter4l.input
```diff
@@ -1 +1 @@
-DecayMode1=8 DecayMode2=9 FilterNLept=4 WriteFailedEvents=2
+DecayMode1=8 DecayMode2=9 NLepMin=4 LHAPDF=NNPDF31_lo_as_0130/NNPDF31_lo_as_0130.info WriteFailedEvents=2
```
menglu21, line 1: As for the parameter, v6.9.8 uses "FilterNLept" and the definition is "For decay mode, reject events that have less than FilterNLept lepton", it's not clear to me whether it's on the lepton or on the charged lepton. In the current version, "NLepMin" is used to specify the number of charged leptons.If you want to use "NLepMin=4", why not just require "DecayMode1=8 DecayMode2=8"

jdavis36: Hi, we do this because these decay cards are used for Z+H samples. Events with Z->2l and H->2l2q may pass 4 lepton selection and we want to account for that.

Issue #3694, opened 2024-04-25
Move user-specific patching after extramodel download, or vice versa, in MadGraph gridpack generation


danbarto:
I have a use case where patching a UFO in the Madgraph5 gridpack generation would be needed. However, user provided patches are only applied *before* the extramodel is downloaded from the CMS model database. Can we either move the patching to after the extramodel is downloaded, or move the extramodel download up? I have tried the former and am happy to open a PR either way.

sihyunjeon:
Or if you have working version already make a PR if you want. I dont see any potential issue right now

PR #3695, opened 2024-04-25
Moving user-specific patching to after extramodel download

danbarto:
Addressing #3694

### DIFF
--- a/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
+++ b/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
@@ -190,11 +190,6 @@ make_gridpack () {
       cd $MGBASEDIRORIG
       cat $PRODHOME/patches/*.patch | patch -p1
       cp -r $PRODHOME/PLUGIN/CMS_CLUSTER/ PLUGIN/ 
-      # Intended for expert use only!
-      if ls $CARDSDIR/${name}*.patch; then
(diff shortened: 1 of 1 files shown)


bbilin:
What is the motivation to change this file? If it is for a specific model, I would suggest modifying it for your own use case.

kdlong:
I don't think there was any particular reason to have it before he model loading in the past, so I don't see any reason not to accept this change (speaking as the one who implemented this ages ago).

bbilin:
> I don't think there was any particular reason to have it before he model loading in the past, so I don't see any reason not to accept this change (speaking as the one who implemented this ages ago).

Thanks @kdlong ! Our only concern is that the patches move after MG has been run in l267. Would there be any cases where this would be a problem? ./bin/mg5_aMC mgconfigscript

danbarto:
Thanks for the comments!
In my opinion there are two motivations to not just upload a new model: this way we are flexible in changing things as we need them, and it is also more transparent to the user what is changed wrt the "official" model release.
As mentioned in the issue, we could also move the model download earlier and leave the patching in the same place.

bbilin:
> Thanks for the comments! In my opinion there are two motivations to not just upload a new model: this way we are flexible in changing things as we need them, and it is also more transparent to the user what is changed wrt the "official" model release. As mentioned in the issue, we could also move the model download earlier and leave the patching in the same place.

Thanks, I would indeed prefer this option.

PR #3700, opened 2024-05-03
Remove testplots parameter from HIN powheg cards

stahlleiton:
This PR removes the testplots parameter from Run 3 PbPb powheg cards

@DickyChant @bbilin


bbilin:
Thanks!

PR #3705, opened 2024-05-08
B2G Run2 UL Fragments update

toa-p:
Added new fragments but several other files seem to be missing. Unable to clone the full repository to my lxplus account. Please advise if possible.


menglu21:
Hi @toa-p it seems you are deleting files, what's the error when you did the clone

PR #2786, opened 2020-12-11
Add WZAToLNuLLA_4f_NLO Process Cards for UL production

Senphy:
Adding cards for WZAToLNuLLA_4f_NLO process

- MG process: p p > w l+ l- a [QCD], w includes w+ w-.
- Madspin process: decay w > l nu, l includes l+ l-, nu includes vl vl~

Thanks a lot!

PR #3707, opened 2024-05-14
Add WZAToLNuLLA_4f_NLO_lowmllsf Process Cards for UL production

Senphy:
Adding cards for WZAToLNuLLA_4f_NLO_lowmllsf process. This is a supplementary sample for #2786 with a different scheme and different mll cuts.

MG process: p p > l nu l+ l- a [QCD], l includes l+ l-. nu contains vl vl~.

Thanks a lot!


menglu21:
Hi @Senphy it seems you are requiring a new sample rather than a supplementary sample, e.g., different syntax, mll>4 instead of 4<mll<10, ptg>10 instead of ptg>15. maybe you can elaborate a bit more on the needs, the setups are generally fine

Senphy:
- For the syntax, In #2786, we are using syntax with p p > w l+ l- a with w decays to l vl with madspin. In this case the FSR contributions from w boson are excluded. In this PR, we are using p p > l vl l+ l- a. The FSR contributions from both W and Z bosons are included.
- We lower the `ptg` cut to `ptg>10` in case we need to do estimations for reco `ptg>15`
- Since with the syntax in this PR, three leptons are in the final state, we reduce the `mll` cut from 1->0, the `mll_sf` cut from 10->4, with respect to #2786

Saptaparna:
Hi @menglu21 thanks to @Senphy for initiating the injection of this sample! From SMP-VV, we fully support the production of this sample!

PR #3710, opened 2024-05-15
Add ttH_HToTT datacards for Run3

sitongan:
Add ttH_HToTT datacards for Run3 for MC request to SUS

### DIFF
--- a/bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/ttH_HToTT/templatecards_run3/ttH_proc_card.dat
+++ b/bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/ttH_HToTT/templatecards_run3/ttH_proc_card.dat
@@ -1,5 +1,5 @@
-# import model sm-no_b_mass
-import model sm
+import model sm-no_b_mass
+#import model sm
 
 define p = p b b~
 define j = j b b~


sitongan:
Hi @menglu21 ,

Template card / script is now used. Please see the UL PR for response to the slha issue.

Many thanks,
Sitong

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/ttH_HToTT/templatecards_run3/ttH_proc_card.dat
```diff
@@ -0,0 +1,11 @@
+# import model sm-no_b_mass
+import model sm
```
menglu21, line 2: why not using directly the no_b_mass model, if we also discussed this in UL PR please add the PR link here

sitongan: Hi @menglu21, it's like this because this is what was used in the Run 2 MC, but we don't think it shouldn't make a difference so we are open to suggestion if you think it's better to change it.

menglu21: please change to the no_b mass model, thanks. same for the Run2 one

sitongan: Sure! Switched to sm-no_b_mass model for both run2UL and run3 now

sitongan: > please change to the no_b mass model, thanks. same for the Run2 one

Hi @menglu21 , the change has been made as requested

PR #3712, opened 2024-05-19
changing path for hepforge libs

sihyunjeon:


menglu21:
Hi @sihyunjeon @Dominic-Stafford is this update validated in gridpack generation, another question is that do you check the tool versions are the most updated version in their official web page

Dominic-Stafford:
Hi @menglu21. Once we have stable stable 10_6 and 12_4 releases with the collier toolfile, I was going to update this merge to point at those releases to have stable collier and ninja from directly inside CMSSW. One downside is we will then be building against fixed versions of those packages, so if new versions are released we'll need to update the versions in CMSSW, then change the CMSSW version we use in gridpack generation to get an up-to-date version. However these packages are pretty stable (new versions are only released every few years), and arguably it's good for us to intentionally decide to update (or not), rather than having the version changing without us noticing

menglu21:
I submitted PR for new release, let's wait for the release and then merge to test

menglu21:
Hi @Dominic-Stafford 12_4_21 is built with this update

sihyunjeon:
hi meng, can we revert this and ask genvalidation to check if the NLO gridpacks run fine with the mglib_path?

Otherwise there could be some users who use master and face issues if not validated

Dominic-Stafford:
Yeah, sorry, I actually need to update some paths in the commits to point to the new CMSSW releases, to get ninja from gosam-contrib rather than inside madgraph and to get collier, and I wanted to test it privately before we merged, so could we revert please? Also is there a 10_6 release available yet?

sihyunjeon:
@menglu21 since dominic confirmed the paths are not properly set yet in the branch i reverted the changes

PR #3716, opened 2024-05-28
POWHEG-RES Change

jshin96:
Modified the patches such that it includes updated loop calculation codes and compile those codes that need separate compilation.

### DIFF
--- a/bin/Powheg/Templates/runGetSource_template.sh
+++ b/bin/Powheg/Templates/runGetSource_template.sh
@@ -175,7 +175,7 @@ fi
 if [ `grep particle_identif pwhg_analysis-dummy.f` = ""]; then
    cp ../pwhg_analysis-dummy.f .
 fi
-if [[ $$process != "WWJ" && $$process != "ZgamJ" && $$process != "ZZJ" && $$process != "Zgam" ]]; then
+if [[ $$process != "WWJ" && $$process != "ZgamJ" && $$process != "ZZJ" && $$process != "Zgam" && $$process != "VV_dec_ew" ]]; then
   sed -i -e "s#PWHGANAL[ \t]*=[ \t]*#\#PWHGANAL=#g" Makefile
   sed -i -e "s#ANALYSIS[ \t]*=[ \t]*#\#ANALYSIS=#g" Makefile
   sed -i -e "s#_\#ANALYSIS*#_ANALYSIS=#g" Makefile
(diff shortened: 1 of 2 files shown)


agrohsje:
We heavily rely on pulling code from external sites. Shouldn't we copy to our gen space and pull from there. @menglu21 @bbilin @DickyChant @sihyunjeon @covarell ?

covarell:
Hi @agrohsje , for the moment only local gridpack making is supported (on lxplus8 or equivalent) so I don't think pulling code from external sites should be a problem. On the other side, we automatically get updates.

menglu21:
@jshin96 all these updates are validated, right. As for using GEN website, maybe we can add another try between the first "wget" and "fail_exit", 
`wget --user NLOX --password LoopsAreCool http://www.hep.fsu.edu/~nlox/downloads/v1.2.1/NLOX_1.2.1.tar.gz || fail_exit \"Failed to get NLOX_1.2.1 tar ball\"\n \ `
to
`{wget --user NLOX --password LoopsAreCool http://www.hep.fsu.edu/~nlox/downloads/v1.2.1/NLOX_1.2.1.tar.gz || wget XXXX GENWEBPLACE} || fail_exit \"Failed to get NLOX_1.2.1 tar ball\"\n \` to enable the wget can always succeed and get the package from official place as much as possible

covarell:
@jshin96 , can you please fix this suggestion from @menglu21 , so we can proceed with this PR?

jshin96:
@menglu21 Sorry for late response, I was bit caught up with attending AEPSHEP. By "XXXX GENWEBPLACE", could you give me an example of where these webplace is?

menglu21:
Hi @jshin96 actually my comment is based on Alex's comment, copy them to /eos/project/c/cmsweb/www/generators/directories/cms-project-generators, and try to download from this path if downloading from their official site is failing

covarell:
Hi again, can we merge this now (I am not sure if the above unsuccessful checks are meaningful or just depend on strange Linux version conflicts)?

agrohsje:
Dear @bbilin @lviliani @covarell @DickyChant @sihyunjeon @jshin96 : What is the status with this PR? Who is supposed to work on it? We urgently need to split code and cards. We really put our MC production at risk if we don't follow up on such essential requests.

bbilin:
The last question was asked by Meng, who was reviewing the pr. Then, we simply missed the last update, as there was no message but just another commit. To me this looks fine and we can merge it.

jshin96:
I apologize, as I thought making a commit automatically sends messages.

PR #3719, opened 2024-06-06
added fragments to genproductions/genfragments/ThirteenPointSixTeV/ f…

amankelttu:
added fragments for bstartobg,bstartobbg,and zprimetobb, for masses 1-9 TeV to the directory: /afs/cern.ch/user/a/amankel/signal_generation/genprods_modified/genproductions/genfragments/ThirteenPointSixTeV/


menglu21:
Hi @amankelttu please remove those files end with "~". another comment is that since only the mass is different, maybe you can use one template card and use script to get all the mass points.

amankelttu:
> Hi @amankelttu please remove those files end with "~". another comment is that since only the mass is different, maybe you can use one template card and use script to get all the mass points.

I have made the requested changes.

PR #3726, opened 2024-06-17
Missing NLO VV final states

jennetd:
These three missing decay channels should complete the set of VV NLO samples with aMC@NLO


bin/MadGraph5_aMCatNLO/cards/production/13TeV/VV/WZTo2Q2L_4f_amcatnloFxFx/WZTo2Q2L_4f_amcatnloFxFx_madspin_card.dat
```diff
@@ -0,0 +1,16 @@
+#directory for gridpack mode
+set ms_dir ./madspingrid
+
+#initialization parameters
+set Nevents_for_max_weight 250 # number of events for the estimate of the max. weight
```
menglu21, line 5: use "Nevents_for_max_weigth" for MG265, same for other cards

jennetd: Done

PR #3737, opened 2024-07-10
Adding Run3 cards for AToZHToLLTTbar

mschrode:
This PR adds cards for Run 3 samples of the 2HDM heavy Higgs signal process A -> ZH with H -> ttbar for various mass points (mA, mH). They follow the Run 2 versions used in B2G-23-006 and B2G-24-002.


mschrode:
Just adding here the link to [the Run 2 cards](https://github.com/cms-sw/genproductions/tree/mg265UL/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/AZHToLLtt) for completeness. Only the beam energy has changed for the Run 3 cards.

mschrode:
Hi, are there any comments to these cards? If not, could this PR be merged such that I can proceed with the sample request? Thanks!

PR #3742, opened 2024-07-21
Add early abortion for systematics module

celia-lo:
This PR implements an early abortion mechanism for `systematics` module of MG5aMC as agreed among xPOG PnR and GEN. 

It should work as intended, as shown in https://gitlab.cern.ch/sqian/systematics/-/pipelines/7794147 

`wplus_LO/NLO_good` jobs are without error case, while `wplus_LO/NLO_buggy` ones are with an error at `systematics` module part.

This PR is to address https://github.com/cms-sw/cmssw/issues/43784 

@DickyChant

### DIFF
--- a/bin/MadGraph5_aMCatNLO/runcmsgrid_LO.sh
+++ b/bin/MadGraph5_aMCatNLO/runcmsgrid_LO.sh
@@ -204,7 +204,10 @@ mv process/$event_file process/madevent/Events/${runlabel}/events.lhe
 pushd process/madevent
 pdfsets="PDF_SETS_REPLACE"
 scalevars="--mur=1,2,0.5 --muf=1,2,0.5 --together=muf,mur,dyn --dyn=-1,1,2,3,4 --alps=0.5,1,2"
-echo "systematics $runlabel --start_id=1001 --pdf=$pdfsets $scalevars" | ./bin/madevent
+if { echo "systematics $runlabel --start_id=1001 --pdf=$pdfsets $scalevars" | ./bin/madevent 2>&1 >&3 3>&- | grep '^' >&2; } 3>&1; then
+    echo "Encounter Error in Running Systematics Module"
+    exit 10086
+fi
 popd
 
 # check lhe output  
--- a/bin/MadGraph5_aMCatNLO/runcmsgrid_NLO.sh
+++ b/bin/MadGraph5_aMCatNLO/runcmsgrid_NLO.sh
@@ -146,7 +146,10 @@ if [ ! -e $LHEWORKDIR/header_for_madspin.txt ]; then
     pdfsets="PDF_SETS_REPLACE"
     scalevars="--mur=1,2,0.5 --muf=1,2,0.5 --together=muf,mur --dyn=-1"
 
-    echo "systematics $runlabel --start_id=1001 --pdf=$pdfsets $scalevars" | ./bin/aMCatNLO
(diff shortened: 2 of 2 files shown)


vlimant:
please go ahead as soon as possible

vlimant:
could you please document somewhere what are the gridpacks that are being updated with this ?

DickyChant:
Not yet as we want to have a test first. As emphasized again and again this issue happens randomly during production so we better have a test first.

We have a plan in mind, just need submit the jobs and execution.

vlimant:
What test do you have in mind ?

As emphasised "again and again", this is dragging on for way too long and causing large number of failures for way too long and need to be addressed ASAP

DickyChant:
> What test do you have in mind ?
> 
> As emphasised "again and again", this is dragging on for way too long and causing large number of failures for way too long and need to be addressed ASAP

A request with read-only gridpack, as PnR reports, have a large number of failure related to this, therefore it should serve as a nice pilot test.

I believe we need to have a test before we decide whether to implement a fix or not, since as a physics collaboration we should prioritize our activities based on physics performance rather than computing issues. The former defined who we are the latter defined nothing but how good we could achieve. Have we heard any loudly complaint about this issue like "our analysis cannot proceed because of missing weights"? I don't recall any.

I could quote tons of brutal decisions that makes us suffer for physics performance (for example, not being able to produce ANY sample for half a year) to just make computers happy. I was responsible for some of the proposed solution for which I was fighting for months, you are welcome.

vlimant:
cool your jets @DickyChant !

vlimant:
so what is the test you have in mind ?

DickyChant:
> so what is the test you have in mind ?

Sorry, was too emotional.

The test we could do is to redo the NanoGEN requests with the readonly gridpack. i.e. this one: https://cms-pdmv-prod.web.cern.ch/mcm/requests?prepid=GEN-Run3Summer22wmLHEGS-00201&page=0&shown=127

NanoGEN request will minimize the pilot running time, as well as the output size so that we don't create a huge sample just for test...

From a PnR report, the nanogen request has a large failure rate! Therefore, we could likely then test out whether this fix helps and how it behaves when the issue happened. Therefore PnR could setup roles to automatically react on those failure jobs.

sihyunjeon:
This PR has issues with indenting if and fi, no? @celia-lo @menglu21 @DickyChant or am i somehow blind

DickyChant:
> This PR has issues with indenting if and fi, no? @celia-lo @menglu21 @DickyChant or am i somehow blind

Yeah if you care about indenting in a shell script?

sihyunjeon:
> > This PR has issues with indenting if and fi, no? @celia-lo @menglu21 @DickyChant or am i somehow blind
> 
> 
> 
> Yeah if you care about indenting in a shell script? 

Ah i thought if fi closed the later coming else early on. Thought it would crash I was blind. I will just later fix it with hotfix commit for readability when i have my laptop

DickyChant:
> > > This PR has issues with indenting if and fi, no? @celia-lo @menglu21 @DickyChant or am i somehow blind
> > 
> > 
> > Yeah if you care about indenting in a shell script?
> 
> Ah i thought if fi closed the later coming else early on. Thought it would crash I was blind. I will just later fix it with hotfix commit for readability when i have my laptop

Yeah it was a pain for my `vim`...

PR #3750, opened 2024-08-20
Use the new version for the McM REST client

ggonzr:
Related to: [#23](https://github.com/cms-PdmV/mcm_scripts/issues/23), [#1138](https://github.com/cms-PdmV/cmsPdmV/pull/1138)

Use the new version for the McM REST client

1. Remove all the instructions that load this module from AFS. Defer to the user the responsibility of preparing the execution environment to include this package.
2. Raise a detailed exception explaining how to install the new version if the module is not available.
3. Update instructions that used `mangled` methods available in the McM class to avoid `DeprecationWarning` messages.
4. Delete some sections that look like dead code

### DIFF
--- a/bin/utils/request_fragment_check.py
+++ b/bin/utils/request_fragment_check.py
@@ -11,9 +11,6 @@ import glob
 import json
 import ast
 from datetime import datetime
-#
-#sys.path.append('/afs/cern.ch/cms/PPD/PdmV/tools/McM-QA/')
-#from rest import McM
 from json import dumps
 
 parser = argparse.ArgumentParser(
@@ -56,14 +53,23 @@ if args.develop is False:
 	
 # Use no-id as identification mode in order not to use a SSO cookie
 if args.local is False:
-    sys.path.append('/afs/cern.ch/cms/PPD/PdmV/tools/McM-QA/')
-    from rest import McM
+    try:
+        from rest import McM
+    except ModuleNotFoundError as e:
+        import_msg = (
+            "Unable to import the McM module. "
+            "Install this module in your execution environment "
+            "by following the instructions available at: "
+            "https://github.com/cms-PdmV/mcm_scripts"
+        )
+        raise ModuleNotFoundError(import_msg) from e
+
     mcm = McM(id=None, dev=args.dev, debug=args.debug)
     mcm_link = mcm.server
 
 def get_request(prepid):
     if args.local is False:
-        result = mcm._McM__get('public/restapi/requests/get/%s' % (prepid))
+        result = mcm._get('public/restapi/requests/get/%s' % (prepid))
         if args.download_json is True:
             with open("request_"+prepid+".json",'w') as f:
                 json.dump(result,f)
@@ -78,17 +84,8 @@ def get_request(prepid):
     result = result.get('results', {})
     return result
 
-#def get_request(prepid):
-#    result = mcm._McM__get('public/restapi/requests/get/%s' % (prepid))
-#    if not result:
-#        return {}
-#
-#    result = result.get('results', {})
-#    return result
-
-
 def get_range_of_requests(query):
-    result = mcm._McM__put('public/restapi/requests/listwithfile', data={'contents': query})
+    result = mcm._put('public/restapi/requests/listwithfile', data={'contents': query})
     if not result:
         return {}
 
@@ -97,7 +94,7 @@ def get_range_of_requests(query):
 
 
 def get_ticket(prepid):
-    result = mcm._McM__get('public/restapi/mccms/get/%s' % (prepid))
+    result = mcm._get('public/restapi/mccms/get/%s' % (prepid))
     if not result:
         return {}
 
@@ -105,7 +102,7 @@ def get_ticket(prepid):
     return result
 
 def get_requests_from_datasetname(dn):
-    raw_result = mcm._McM__get(
+    raw_result = mcm._get(
         "public/restapi/requests/from_dataset_name/%s" % (dn)
     )
     if not raw_result:
--- a/bin/utils/request_fragment_check.py
+++ b/bin/utils/request_fragment_check.py
@@ -1,5 +1,6 @@
 #!/bin/env python3
 import os
+import warnings
 import sys
 import re
 import argparse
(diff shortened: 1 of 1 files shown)


ggonzr:
Hi @efeyazgan. These changes are related to using the new version for the McM REST client and avoiding importing modules from AFS. I saw there are some CI jobs to check changes for the script. However, they do not check the contribution provided in the PR (the commits differ - PR: `a944f5b41fbf1bd9631855c68bbed7ba195ded30` vs used by the CI: `38b30bdba0758e27e99152958873b073312a0b17`). Could you test this PR and let me know if you agree to merge this contribution?

Thanks,
Geovanny

efeyazgan:
Hi @ggonzr 
When I test this locally, this is what I get:
```
./run_req.sh EXO-Run3Summer22wmLHEGS-01444
---> 1 request will be checked:
 
Current date and time: 2024-08-26 16:30:28
Prepid(s):
EXO-Run3Summer22wmLHEGS-01444
Traceback (most recent call last):
  File "/afs/cern.ch/work/e/efe/GENprod_gio/genproductions/bin/utils/request_fragment_check.py", line 497, in <module>
    res = get_request(prepid[num])
  File "/afs/cern.ch/work/e/efe/GENprod_gio/genproductions/bin/utils/request_fragment_check.py", line 72, in get_request
    result = mcm._get('public/restapi/requests/get/%s' % (prepid))
NameError: name 'mcm' is not defined
[efe@lxplus919 utils]$ vi run_req.sh 
[efe@lxplus919 utils]$ ./run_req.sh EXO-Run3Summer22wmLHEGS-01444
---> 1 request will be checked:
 
Traceback (most recent call last):
  File "/afs/cern.ch/work/e/efe/GENprod_gio/genproductions/bin/utils/request_fragment_check.py", line 57, in <module>
    from rest import McM
ModuleNotFoundError: No module named 'rest'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/afs/cern.ch/work/e/efe/GENprod_gio/genproductions/bin/utils/request_fragment_check.py", line 65, in <module>
    raise ModuleNotFoundError(import_msg) from e
ModuleNotFoundError: Unable to import the McM module. Install this module in your execution environment by following the instructions available at: https://github.com/cms-PdmV/mcm_scripts
```
This happens because of https://github.com/cms-sw/genproductions/blob/a944f5b41fbf1bd9631855c68bbed7ba195ded30/bin/utils/request_fragment_check.py#L55 when running locally. 

Is this new method to be used in the same way in lxplus (locally) and elsewhere? Then, I can drop that if. 

However, even when I drop that statement I get the same error. 

Before running this command, I do
```
python3.9 -m venv venv && source ./venv/bin/activate
```
Am I missing anything?

ggonzr:
Yes, you need to install the package by executing this [pip](https://github.com/cms-PdmV/mcm_scripts?tab=readme-ov-file#build-package) instruction. By the way, is this script meant to be used only by the McM application, or do other users/applications use it?

efeyazgan:
OK, I'll try and let you know. 

This is being used both by McM and many users locally (in lxplus).

efeyazgan:
OK, after installing the package, it worked. I think it is OK to merge.

bbilin:
Dear all,

One point before we merge, this script is run on condor (unless you changed this recently), so installation should be also implemented in the script itself, no?

Thanks
B.

bbilin:
I think it is the get test that is executed on condor. (IIRC, current PdmV and Geovanny can correct me if wrong). So indeed it can be added on the line before the script is run. 

Thanks

ggonzr:
Indeed, this script is embedded to run McM validation jobs in HTCondor and prepare other kinds of scripts that the application provides. But as described before, [#1138](https://github.com/cms-PdmV/cmsPdmV/pull/1138) takes care of this.

The only remaining thing would be to send an announcement to other users about this change, which is a breaking change in case they run the script directly in their workflows as they have to also prepare an environment that installs the package. Do you know if this situation happens often and who could be affected by this?  

Perhaps, instead of raising an error, we could try to temporarily import the package available in AFS and raise a `DeprecationWarning` alerting on this change giving users a window of 3 weeks to 1 month to update their environments before we cut the support on this old version. Do you agree with this approach?

bbilin:
Hi Geovanny

Thankw for clarification, then indeed i would add a warning and a deadline, then we can switch to error.

Cheers

B

Envoyé de mon iPhone

Le 30 août 2024 à 09:57, Geovanny González-Rodríguez ***@***.***> a écrit :

﻿

Indeed, this script is embedded to run McM validation jobs in HTCondor and prepare other kinds of scripts that the application provides. But as described before, #1138<https://github.com/cms-PdmV/cmsPdmV/pull/1138> takes care of this.

The only remaining thing would be to send an announcement to other users about this change, which is a breaking change in case they run the script directly in their workflows as they have to also prepare an environment that installs the package. Do you know if this situation happens often and who could be affected by this?

Perhaps, instead of raising an error, we could try to temporarily import the package available in AFS and raise a DeprecationWarning alerting on this change giving users a window of 3 weeks to 1 month to update their environments before we cut the support on this old version. Do you agree with this approach?

efeyazgan:
That seems OK (for the ones who want to run locally, I summarized the commands here: https://github.com/cms-sw/genproductions/blob/master/bin/utils/run_req_new.sh)

ggonzr:
Hi @efeyazgan,
Please could you test this again, the last commit loads the old version from AFS if the user has not provided a proper `venv` with the latest version of the McM package.

ggonzr:
I just rechecked this with the integration for McM described in [#1138](https://github.com/cms-PdmV/cmsPdmV/pull/1138) and it works by providing a `venv` or loading the module from AFS directly. Therefore, this PR is ready and you can merge it. Please do not forget to send an announcement to your stakeholders about this change (CMS Talk perhaps). Next month I will create a PR to remove this backward compatibility so that we do not rely on old versions

efeyazgan:
Sorry. I had missed this. I tested it and it works OK. Thanks a lot!

efeyazgan:
@bbilin @lviliani please take a look. Thanks.

lviliani:
Thanks! Looks good to me, I only have one question to make sure I understand.

I think MC contacts usually download the test script from McM (using the "Get test command") to run some quick validation locally on lxplus, before running the McM validation.

Can you please confirm that this will still work, provided you merge also https://github.com/cms-PdmV/cmsPdmV/pull/1138?

ggonzr:
Yes, it will work even if https://github.com/cms-PdmV/cmsPdmV/pull/1138 is not merged immediately as the script will load the old version without issues if it is executed in Lxplus nodes. So it is safe to merge this

PR #3753, opened 2024-08-26
Add DisplacedTop Cards

PaulVaucelle:
Providing the cards for the different benchmark of our signal process as requested by the MC&I contacts of the EXO group. (Analysis reference: Displaced top quark, EXO->LLP sub-group, no CADI-line)


lviliani:
Please do not commit the full set of cards for all the mass points.
Just put there template cards and a script to generate the full grid.

PaulVaucelle:
Removed the cards and added a script to generate the cards

lviliani:
Thanks! But it seems you are adding also some other stuff not related to your cards. 
Can you please clean it? If it's easier we can close this PR and you open a new one.

PaulVaucelle:
My bad.. Removed the files mistakenly added.
Sorry for the inconvenience

lviliani:
Thank you and sorry for the delayed reply.
Regarding the param_card, if you are using the default one and just changing a few parameters, usually we suggest to use a customize_card instead, where you include only the parameters that will be changed.
Can you do that, or is there a reason why you need to commit the full run card?

Regarding the PDF choice (306000), do you need that particular PDF set? You can also get the default recommendation by using these lines in the run card:
```
```

PaulVaucelle:
For the PDF, we can definitely use the default recommendation.
For the param card,  we can set everything in the customize card (masses + SELMIX block)

lviliani:
Thank you! I'll merge the PR.

PR #3756, opened 2024-09-04
Pair VLQ cards for Run 3, standard and exotic decays

jmhogan:
Adding 4 types of pair vector-like quark cards for Run 3:

* TT and BB with 3rd generation SM decays (copy of Run 2 UL with updated run_card). A param_card is included to set all the widths to 10 GeV and program the desired branching fractions for the VLQs.
* TT and BB with 3-body q+2tau decays via an offshell leptoquark Xi (new model for CMS). Again widths of the VLQs (and Xi's) are 10 GeV -- usually it's the default, in the case of B quarks we adjust the default from 9 to 10 in customizecards.dat.


bbilin:
Thanks, the cards look fine, one question though, why do you commit the param_cards? Do you change anything manually?

jmhogan:
@bbilin Poke? Is it possible to merge this?

jmhogan:
@bbilin Poking again! I also sent an email. We are anxious to get these samples submitted!

PR #3761, opened 2024-09-16
NLO ttH and tH cards for CP studies for Run-2

FloMau:
Cards for NLO samples as used for Run-3 ttH / tH CP analysis for Run-2. 

The only change w.r.t. the Run-3 cards is the beam energy.


lviliani:
Except for the comment above, the rest looks good to me.

PR #3771, opened 2024-10-03
 Madgraph Cards for Run2 UL ggF H->ToWW->LnuQQ Off Shell analysis

sv3048:
Hi Experts !

I am requesting to review the cards for three samples ( signal, background and Signal Background Interference) for off shell semileptonic analysis ( H-> WW-> lnuqq) .
The background samples here are already used in Run2 legacy analysis and I have used same set of cards from the sample for producing gridpack. Kindly help us to optimize other run cards ( signal and SBI ) .

Thanks,
Sadhana

[1] https://github.com/cms-sw/genproductions/tree/b9fc8d4440f1bcb14a03be2e20f8fa814d8c4f19/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/ggWW


bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/GluGluHToWW_OffShellSemilep_4f_LO/GluGluHToWWSemilep_off_shell_B_4f_LO/GluGluHToWWSemilep_off_shell_B_4f_LO_proc_card.dat
lviliani, line 1: Regarding the syntax in this card, I think you are generating also quark-induced WW production and not only the gluon-induced. Is this what you need?
If the target is off-shell gg->H->WW, the background that interferes with the signal is only gg->WW.

bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/GluGluHToWW_OffShellSemilep_4f_LO/GluGluHToWWSemilep_off_shell_SBI_4f_LO/GluGluHToWWSemilep_off_shell_SBI_4f_LO_proc_card.dat
lviliani, line 1: Same comment here.

lviliani:
In general I would suggest to also look at the diagrams produced by MG in the 3 cases to make sure the syntax is correct and you are producing what you want.

Can you also clarify on the need of adding the cuts files? Is it needed only for gg->WW or also for signal?

Let me also include our Madgraph contacts who can help validating these cards @sihyunjeon @DickyChant.

DickyChant:
> In general I would suggest to also look at the diagrams produced by MG in the 3 cases to make sure the syntax is correct and you are producing what you want.
> 
> Can you also clarify on the need of adding the cuts files? Is it needed only for gg->WW or also for signal?
> 
> Let me also include our Madgraph contacts who can help validating these cards @sihyunjeon @DickyChant.

We have been helping out for this request for O(1y) scale, I think @bbilin @menglu21 @shimashimarin have more words to say

PR #3775, opened 2024-10-08
add HZ and HW at 120 and 130

Paul-Krueper:


yihui-lai:
Hi @bbilin @lviliani , could you please have a look at these cards? These are duplicates of the SM cards, with only the mass changes to 120 and 130; the min_h_mass and max_h_mass are changed accordingly as well.

lviliani:
Is it ok to change the Higgs mass and not the width?

CERN-Finance-Club-Admin:
Hi I can reproduce the gridpacks and cards if the Higgs width needs to be changed. 

If so please let me know, along with the values for the width I should use at 120 and 130
Thanks

Paul-Krueper:
Hi I can reproduce the gridpacks and cards if the Higgs width needs to be changed.

If so please let me know, along with the values for the width I should use at 120 and 130
Thanks

lviliani:
If the width needs to be changed or not depends on how much your analysis is sensitive to this parameter.
In general, the Higgs width depends on the mass value you assume, while in these cards I think you are using the value corresponding to 125 GeV also for 120 and 130.
You can probably find the Higgs width values as a function of the mass in the LHC Higgs WG pages.

Paul-Krueper:
I just spoke with my analysis team and our analysis wont be sensitive enough to measure width effects, so it is fine to keep the width constant across 120-125-130. Let me know should anything else in the cards be unclear

lviliani:
Ok. 
Nevertheless. since this samples could be used also by other analyses and it's a very simple change, I would prefer to make things right.
Please update the width values, according to what was also used in UL requests, have a look here for example:
https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/production/2017/13TeV/Higgs/WplusHJ_HanythingJ_NNPDF31_13TeV/makecards.py

Paul-Krueper:
okay, I will change the cards accordingly. To be clear, I will have to rerun all corresponding gridpacks too right ?

yihui-lai:
> okay, I will change the cards accordingly. To be clear, I will have to reproduce all corresponding gridpacks too right ?

Yes

Paul-Krueper:
Hi all, the samples you pointed to were used for high mass/off shell analyses that are specifically sensitive to the width - the other samples we are using for our analysis,  eg GGH in https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/Higgs
use the same width across different masses. 

I have also found other Run 3 examples where the width was not changed with different mass eg for ggH again
https://github.com/cms-sw/genproductions/tree/master/bin/Powheg/production/Run3/13p6TeV/Higgs/gg_H_quark-mass-effects_NNPDF31_13p6TeV

Since it is a lot of gridpacks we'd have to reproduce and also to keep things consistent, we'd prefer to keep them as they are. Let me know what you think. Thanks

CERN-Finance-Club-Admin:
Thanks for merging.
To be clear, the gridpacks are consistent with these cards that are now merged.
Do let me know if you need anything else.

Paul-Krueper:
Thanks for merging.
To be clear, the gridpacks are consistent with these cards that are now merged.
Do let me know if you need anything else.

PR #3781, opened 2024-10-15
Pythia fragment

MohitS704:
Updating the PYTHIA filter used for gluon fusion off-shell Higgs using MCFM+JHUGen to use the MCMultiParticleMassFilter, which was introduced here: https://github.com/cms-sw/cmssw/pull/45344.

Since MCFM+JHUGen gridpacks are **separated by final state**, this formulation of the PYTHIA filters separated by the 4-lepton final-state works for each given gridpack. You can see a presentation on this at an HZZ meeting here: https://indico.cern.ch/event/1433234/#38-mc-simulation-gluon-fusion.

There is no filter planned for tau decays - so electrons and muons will be the only ones with a filter applied. For tau decays, simply use the fragment without a filter (`GluGluToZZTo4L_TuneCP5_13p6TeV-MCFM-jhugen7.5.2_nofilter.py`).


bbilin:
Hi,

Thanks, we went through the presentation to understand the details, but unfortunately failed to do so.

Why do you want to turn off MPI? If there is a specific physics need that requires such a manipulation, you are invited to present in a GEN meeting such that we can understand the needs.

Thanks a lot,

B&L

gritsan:
We did make a presentation at the GEN meeting back in October 2023, and Steve Mrenna suggested this particular configuration to have MCFM+Pythia and Pythia+Pythia in good agreement. We are not doing anything different from what Steve recommended. We assumed that this has been settled since almost a year ago, and now we proceed with the actual simulation of Run3 configuration, after implementing an additional filter.

smrenna:
Everything has to be understood in context.  At the time, we were trying to validate the transverse momentum of the hard process, which is somewhat decoupled from the MPI.   Because MPIs slow down the code considerably, I suggested we turn it off _for the sake of validation of the hard process, only_.

smrenna:
This is from the email of Sep 7, 2023:  "If you set the runtime command PartonLevel:FSR = off, then you will not have to deal with the complication of any FSR.    You would also turn off MPI in the same fashion."

But this is not for real physics.

MohitS704:
I see - then this is a misunderstanding on our end that I will fix. Thank you for weighing in @smrenna!

MohitS704:
Pinging @bbilin to look over this again, as now MPI is not turned off.

lviliani:
What about turning off `BeamRemnants:primordialKT` and `SpaceShower:MEcorrections`? 
Is this something that you need or did you turn them off just for initial validation purposes?

Also, ` 'SpaceShower:alphaSvalue = 0.118'` and `'SpaceShower:alphaSorder = 2'` are redundant settings, because they are already included in the CP5 tune parameters ([here](https://github.com/cms-sw/cmssw/blob/0e656bb0a2dd6524dd052e6584547899d31c8bac/Configuration/Generator/python/MCTunesRun3ECM13p6TeV/PythiaCP5Settings_cfi.py#L14-L15)). 
So I think you can safely drop them,

smrenna:
How can I actually see the fragment?

genfragments/ThirteenPointSixTeV/Higgs/GluGluToZZTo4L_TuneCP5_13p6TeV-MCFM-jhugen7.5.2_4e.py
```diff
@@ -0,0 +1,58 @@
+import FWCore.ParameterSet.Config as cms
+
+externalLHEProducer = cms.EDProducer("ExternalLHEProducer",
+    args = cms.vstring('Name'),
```
lviliani, line 4: Here the string should be the path to the gridpack in cvmfs

genfragments/ThirteenPointSixTeV/Higgs/GluGluToZZTo4L_TuneCP5_13p6TeV-MCFM-jhugen7.5.2_4mu.py
```diff
@@ -0,0 +1,57 @@
+import FWCore.ParameterSet.Config as cms
+
+externalLHEProducer = cms.EDProducer("ExternalLHEProducer",
+    args = cms.vstring('Name'),
```
lviliani, line 4: Same here

smrenna (commented):
Turning off primordial kT was for testing purposes only.    We did a simplified study to see how to make the Pythia shower give a result that was comparable to Powheg (if I remember correctly), so we turned off certain features that added noise.   This should not be turned off for production.

MohitS704:
> Turning off primordial kT was for testing purposes only. We did a simplified study to see how to make the Pythia shower give a result that was comparable to Powheg (if I remember correctly), so we turned off certain features that added noise. This should not be turned off for production.

Just to make sure - is this also true for SpaceShower:MEcorrections?

smrenna:
Thank you for the question.  I think the MEcorrections needed to be off so that we got the correct behavior.   I will double check our email exchanges.

MohitS704:
I think this was merged in error - I haven't edited the fragment to remove the settings that were marked as "for testing" like primordialkT - let me do that now and make another PR to correct this.

My apologies for the slowness that caused this confusion

PR #3789, opened 2024-10-28
Runcmsgridpatch req

efeyazgan:


lviliani:
The test is failing because of an error in copying the patched gridpacks to eos. Could it be that eos is not mounted within the githib CI?
@efeyazgan confirmed that local tests are fine and that similar patches worked in the past.
Could it be related to some of the recent updates from @ggonzr? 
@DickyChant @sihyunjeon

DickyChant:
> The test is failing because of an error in copying the patched gridpacks to eos. Could it be that eos is not mounted within the github CI? @efeyazgan confirmed that local tests are fine and that similar patches worked in the past. Could it be related to some of the recent updates by @ggonzr? @DickyChant @sihyunjeon

To copy a gridpack to eos, we need certain permissions.

The CI works as a runner on a self-hosted openstack VM, the linux user is "runner" and I definitely do not think it can copy anything to the GEN eos dir.

efeyazgan:
Thanks for the clarification. Yes, until now we had no problem accessing eos in McM jobs. So, we can bypass this test (but let me do a few final checks firtst). 

In fact, I think I can add something to check if the user is "runner", then it can bypass checks that includes access to eos. 

@lviliani @bbilin in any case, after all gridpacks are patched, the script will bypass this one as well.

mseidel42:
Hi all, I can confirm that there is no EOS access from the runner, and it is cumbersome to set up with a service account.
If there is a Jenkins-based solution by PdmV already then it might be good to switch to that at some point. 

> In fact, I think I can add something to check if the user is "runner", then it can bypass checks that includes access to eos.

That sounds very fragile! Better to add another option like this and call the CI with `--apply_run3_patch=false`
```
parser.add_argument('--apply_run3_patch', help="add description what this actually does", action='store_true')
```

Cvico:
Hi, 

As mentioned somewhere else in the thread, we have a set of requests that we want to inject, and they would need this patch applied. Is there anyway TOP can help to validate this, and also benefit from the patch?

Cvico:
Requests are listed in [these tickets](https://gitlab.cern.ch/cms-gen/mccm/-/issues/359). So:
      - 2016APV: [TOP-2024Oct30-00001](https://cms-pdmv-prod.web.cern.ch/mcm/mccms?prepid=TOP-2024Oct30-00001)
      - 2016: [TOP-2024Oct30-00002](https://cms-pdmv-prod.web.cern.ch/mcm/mccms?prepid=TOP-2024Oct30-00002)
      - 2017: [TOP-2024Oct30-00003](https://cms-pdmv-prod.web.cern.ch/mcm/mccms?prepid=TOP-2024Oct30-00003)
      - 2018: [TOP-2024Oct30-00004](https://cms-pdmv-prod.web.cern.ch/mcm/mccms?prepid=TOP-2024Oct30-00004) 
      
There are quite a lot of requests in this case.

PR #3791, opened 2024-11-04
Charged higgs taunu run3

slehti:
Datacards for H+ -> taunu for Run3.


bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/ChargedHiggs_taunu/ChargedHiggs_taunu_heavy_NLO/ChargedHiggs_taunu_heavy_NLO_proc_card.dat
```diff
@@ -0,0 +1,13 @@
+import model 2HDMtII_NLO
+define p = g u c d s u~ c~ d~ s~
+define j = g u c d s u~ c~ d~ s~
+define l+ = e+ mu+
+define l- = e- mu-
+define vl = ve vm vt
+define vl~ = ve~ vm~ vt~
```
lviliani, line 7: Can you please remove these `define` lines? You are not using these definitions to generate the process.

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/ChargedHiggs_taunu/ChargedHiggs_taunu_heavy_NLO/ChargedHiggs_taunu_heavy_NLO_customizecards.dat
lviliani, line 1: Is it fine to change only the mass without touching the width?

slehti:
These datacards were in use in Run2 (only modification for the first version was the CM energy). Removed now the unused definitions. About the width, I think we got example datacards originally from H+ theory people, so if the width is now required, we need to consult the theory experts about it.

PR #3793, opened 2024-11-04
Create JHUGen_2l2v.input

YujiLee301:
Dear GEN experts,

     I hope you are doing well. 
     We are HZZ MC contacts. Currently, HZZ people are asking for HZZ 2l2v samples for Run3 analyzes. For the samples generation, we will use powheg for H production and JHUGen for H decay. The new commit describes H ZZ 2l2nu decay mode. Thanks for your review. 

Best,
Yuji & Lazar [HZZ MC Contacts]


lviliani:
Hello, can you explain what is the `WidthSchemeIn=3` setting?
Also, do you need to also commit the `PMZZdistribution.out` file?

bbilin:
any answers?

YujiLee301:
Hi @lviliani ,

Sorry for late reply, here's my explanation.
PMZZdistribution.out can be used to reweight POWHEG samples to correct for P_decay(m4l) when we use JHUGen on H decay, which helps a more precise generation on low mass Z. In this case, WidthSchemeIn=3 have to be added in the inputcard when we want to use PMZZdistribution.out.
And this is the reference: https://github.com/cms-sw/genproductions/tree/master/bin/JHUGen/Pdecay

Cheers,
Yuji & Lazar [HZZ MC Contacts]

PR #3795, opened 2024-11-05
Basic POWHEG debugging changes for EL8

covarell:
Could finally do it, after approval of #3716 .

Fix to runGetSource_template.sh is very important and caused buggy WH and ZH signal samples to be produced in Run3. Other two are just for expert debugging.

@agrohsje @jshin96 .


covarell:
This actually grew to a much bigger scope, i.e. fixing many POWHEG-RES broken processes

PR #3797, opened 2024-11-06
Fix for powhegUL branch to run on lxplus (Redhat9)

jaebak:
Made a fix for the powhegUL branch to run on lxplus (Redhat9).

Used the approach that was used in the existing scripts for slc6. The approach was extended to slc7.

The fix was discussed in cmstalk in the below link.

https://cms-talk.web.cern.ch/t/issue-with-large-generator-weights-for-run-3-powheg/50579/6

### DIFF
--- a/bin/Powheg/Templates/createTarBall_template.sh
+++ b/bin/Powheg/Templates/createTarBall_template.sh
@@ -19,10 +19,6 @@ FULLGRIDRM=`ls $${WORKDIR}/$${folderName} | grep fullgrid-rm | head -n 1`
 FULLGRIDBTL=`ls $${WORKDIR}/$${folderName} | grep fullgrid-btl | head -n 1`
 UBOUND=`ls $${WORKDIR}/$${folderName} | grep ubound | head -n 1`
 PWGSTAT=`ls $${WORKDIR}/$${folderName} | grep st3-stat | head -n 1`
-FULLGRIDRM=$${FULLGRIDRM%?}
-FULLGRIDBTL=$${FULLGRIDBTL%?}
-UBOUND=$${UBOUND%?} 
-PWGSTAT=$${PWGSTAT%?} 
 
 if [ $${#FULLGRIDRM} -gt 0 -a $${#FULLGRIDBTL} -gt 0 ]; then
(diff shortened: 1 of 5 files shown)


covarell:
https://github.com/cms-sw/genproductions/pull/3797/files#diff-a946031d265db406992c699bd5f0960280050183f2643c5e902085667686c173L22-L25

Why are these lines removed?

covarell:
https://github.com/cms-sw/genproductions/pull/3797/files#diff-306a635579daba4fb10220a64e87ff0b29416fe9cfa2acf8bb8abd5391f8b7cdR48

Here, can't you just import subprocess and use subprocess.getstatusoutput?

covarell:
Thanks anyway for this important update

jaebak:
> https://github.com/cms-sw/genproductions/pull/3797/files#diff-a946031d265db406992c699bd5f0960280050183f2643c5e902085667686c173L22-L25
> 
> Why are these lines removed?

The %? in ${UBOUND%?} means that a single character would be removed from $UBOUND,
where when $UBOUND is "pwgubound-0001.dat", 
${UBOUND%?} would be "pwgubound-0001.da"

Because this was strange, I removed those lines.

Would there be a case where removing a single character would work?

jaebak:
> https://github.com/cms-sw/genproductions/pull/3797/files#diff-306a635579daba4fb10220a64e87ff0b29416fe9cfa2acf8bb8abd5391f8b7cdR48
> 
> Here, can't you just import subprocess and use subprocess.getstatusoutput?

Yes. I think we can. I changed it to use subprocess.getstatusoutput

jshin96:
I actually had the same problem of missing letter at the end that made the code to not copy this “UBOUND” dat file. I asked you in February if this needs to be fixed or not, and you said that it is OK for it having a missing letter.

covarell:
I found the conversation we had about this. I said:
"
OK for the missing final character in $UBOUND (for this, please make a PR on master, for the tutorial I don't think we need it).
"
So I meant it was OK to fix it. :-) Never mind, I wasn't clear. since I have an open PR on the master branch I will fix it.

For me it is a +1 on this PR.

PR #3798, opened 2024-11-07
Update 0007-fix-cluster-local-path.patch

tyy99phy:


sihyunjeon:
Hi,
Though, is this the reason for the patch failures we saw in your logs?
I thought those lines were irrelevant

@joonblee

joonblee:
No. It cannot affect to any physical system, but we can accept this update so that others can follow the code easily (?)

lviliani:
Is this ready to be merged?

bbilin:
kind ping @joonblee

joonblee:
This does not affect to the physical system since it only changes the commented-out lines (Thanks to Tianyi he double-checked this). I thus think accepting this request fully depends on the code style. I guess Tianyi and you guys will keep maintaining it so I'll follow your preference :)

PR #3799, opened 2024-11-08
Polarized VVjj VBS cards

ldellape:
Dear GEN Expert,
We have added some cards for semileptonic polarized VBS processes (ssWW, ZZ, WZ) in 
cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/VBS/VVjj_semilep_Polarized
Please do not consider the changes to source_condor.sh
Could you please check the syntax of our cards?
We have already tested the ssWW cards by generating both the tar and LHE outputs.


bin/MadGraph5_aMCatNLO/Utilities/source_condor.sh
lviliani, line 1: Can you please remove the changes in this file?

bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/VBS/VVjj_semilep_Polarized/WZ/VBS_WZ_PolarizLL_semilep/VBS_WZ_PolarizLL_WLtoHAD_ZLtoLEP/VBS_WZ_PolarizLL_WLtoHAD_ZLtoLEP_run_card.dat
```diff
@@ -0,0 +1,263 @@
...
+#
+#*********************************************************************
+# BW cutoff (M+/-bwcutoff*Gamma) ! Define on/off-shell for "$" and decay
+#*********************************************************************
+  15.0	= bwcutoff ! (M+/-bwcutoff*Gamma)
+ #*********************************************************************
+ # Apply pt/E/eta/dr/mij/kt_durham cuts on decay products or not
+ # (note that etmiss/ptll/ptheavy/ht/sorted cuts always apply)
+ #*********************************************************************
+  True	= cut_decays ! Cut decay products
```
lviliani, line 97: Given that you are setting this to True, all the cuts defined below should apply to all jets, even to the jets from W/Z hadronic decays.
I assume you want to apply those cuts (like `drjj` or `mmjj`) only to the VBS jets, so in this case you should set `False = cut_decays`.
You can double check this by producing some LHE events and checking the distribution of the W/Z mass from hadronic decays.

lviliani:
Please check the comments inline, they apply to all the cards.

bbilin:
any news?

ldellape:
Thank you for your comments. We have set the cut_decays flag to False for all the run cards. We checked this by producing mass plots of the W using the LHE (with 10000 events), and we now correctly observe the peak around the mass value for hadronically decaying W bosons. This peak was indeed missing in the previous version due to the cut on m_jj as you pointed out.

PR #3801, opened 2024-11-12
Add Run3 PbPb single top t-channel POWHEG cards

stahlleiton:
This PR adds two new POWHEG cards and PYTHIA8 genfragments associated to the single top and the antitop t-channel production using POWHEG (ST_tch mode)l for Run3 PbPb MC production.

@DickyChant @bbilin


stahlleiton:
@DickyChant @bbilin : are there further comments on this PR?

PR #3807, opened 2024-11-23
Add Run3 pp 5.36TeV Powheg input cards

stahlleiton:
This PR adds the Run3 pp reference at 5.36 TeV Powheg input cards for HIN PAG.
It includes DY, W and top production cards. They are similar to the input cards used for PbPb 5.36 TeV except that the PDF is changed to NNPDF31 (325300) as used in standard pp production.

@DickyChant @bbilin


stahlleiton:
@DickyChant @bbilin : are there further comments on this PR?

PR #3808, opened 2024-11-24
Add Run3 pp 5.36TeV MadGraph input cards

stahlleiton:
This PR adds the Run3 pp reference at 5.36 TeV MadGraph input cards for HIN PAG.
It includes DY, W and top production cards. They are similar to the input cards used for PbPb 5.36 TeV except that the PDF is changed to standard pp production (DEFAULT_PDF_SETS) .

@DickyChant @bbilin @sihyunjeon


stahlleiton:
@DickyChant @bbilin @sihyunjeon : are there further comments on this PR?

PR #3311, opened 2022-12-05
adding ttHH HEFT cards at different values of kl,kt and c2

gekobs:
I am opening this PR to add the cards needed to produce ttHH events, varying the dim-6 HEFT couplings.
The generation is based on a UFO model provided by a group of theorist and it is not centrally available atm. Could you please advise on  where should I upload the UFO?
Thanks!

PR #3809, opened 2024-11-25
TTHH HEFT Cards for RUN3  + additional kl=1 point in RUN2-UL Card  

ats2008:
* ports the ttHH HEFT cards for RUN3 ( original PR https://github.com/cms-sw/genproductions/pull/3311 )
* Adds additional kl=1 point to the card customization script

@acarvalh


acarvalh:
Hello! These are exact copies of run 2 with energy and pdf updates

PR #3815, opened 2024-12-05
Generation of TTZZ sample without MadSpin card

G-Vian:
Dear GEN team,

My team wants a ttZZ sample with 13.6 TeV center of mass energy. The other TTZZ card, available at  [[1](https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/TTZZ_5f_LO)], contains a MadSpin Card that decays the ZZ, that's why I created these cards for a TTZZ process without Z decaying.
Thank you very much.


bbilin:
Thanks. I think the tops need to be decayed in the Madgraph. (or madspin)

G-Vian:
Dear Bugra,

I didn't request the decay of the top, as my team wants it inclusive.

lviliani:
Dear @G-Vian, in this case the decays will be handled by pythia8, which does not take into account spin correlations effects in the decays. This will lead to wrong distributions for spin sensitive observables.
If you want inclusive decays, the best way is still to use Madgraph or Madspin to handle them.

G-Vian:
Dear Lorenzo, 
Thanks for your prompt answer.
I modified the proc card, adding the decay of the tops.

G-Vian:
Dear Lorenzo, 
We would like to keep the W undecayed because we want to explore different channels from these decays. 
Additionally, the previous TTZZ card had a MadSpin Card that decays the ZZ, which was undesirable.

lviliani:
Sorry, I'm not following.
In the previous TTZZ card you are decaying top and Z to all, which means they are inclusive in the decays and your samples will contain all the decay channels.
Now, if you don't decay Z and top in madgraph/madspin, it will be decayed later on by Pythia8 to all possible channels, so you will get an inclusive sample anyway.

G-Vian:
Yes, thats correct.

lviliani:
Ok thanks, then I don't understand why you need these new TTZZ cards.

G-Vian:
It is because we dont want the MadSpin decay of the Zs

sihyunjeon:
But then why decay top in madgraph?

Setup is not superbly convincing

Either you drop all the decays in madgraph/madspin to enhance computing performance (pythia is faster) OR decay both t and z

G-Vian:
I have decayed the Tops because was informed in the previous comment that leaving it to Pythia decay would lead to wrong spin correlations.

sihyunjeon:
Same story goes for z, it loses spin correlation when pythia handles it

G-Vian:
Dear GEN conveners,

I apologize for the delay in responding, but I have discussed this with my analysis team, and they have concluded that we will perform all decays in Pythia. We will do this in the same way as it was done in our analysis of Run 2.

Thank you very much, and please forgive the delay.

G-Vian:
Dear GEN Conveners,

I have also added cards for the TTZH process for Run 3, as the cards from this link [[1]](https://github.com/cms-sw/genproductions/tree/master/bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/TTZH_5f_LO) for Run 3 had a BW Cutoff value of 15000.

lviliani:
Hi @G-Vian, the recommendation from GEN is to use Madgraph or Madspin for the decays, and not Pythia8 because the spin correlations are lost in this case. As we wrote previously, using Pythia8 will lead to wrong kinematic distributions for variables sensitive to spin correlations.

G-Vian:
Dear Lorenzo, we are aware of this problem, but we still want to follow this approach.

lviliani:
Hi @G-Vian, before merging this PR, we would like to ask you or your team to present at one of the upcoming GEN meetings (Mondays at 2pm) to discuss whether this approach is sensible in your analysis.
Please get in touch with us via email or mattermost to organize the presentation.
Thanks!

G-Vian:
Dear GEN conveners,

Following your message, we decided to perform the decays of the Z boson and the top quark in MadGraph (or in a MadSpin card) to preserve spin correlations. Initially, we planned to handle the decays in Pythia, during the event generation, following what was done in the generation of these processes for Run 2.

With this in mind, we are thus considering modifying the other cards we submitted for Run 3 (i.e. for TTZH, TTHH and TT4b), to be consistent with the prescription of performing the decays through MadGraph (or MadSpin). The tops (inclusive) and Zs (Z > bb~) will be decayed following this new prescription, except for the Higgs, which will be decayed in Pythia (H > bb~).

Additionally, would you recommend the decays be done on the process card or in a MadSpin Card?

G. Vian, A. Savoy-Navarro, W. Wei

G-Vian:
Dear GEN conveners,  

In the spirit of the changes we discussed, which include adding MadSpin cards to perform the decays, I have modified the TTZZ, TTZH, TTHH, and TTHH folders to include MadSpin cards executing the decays. Additionally, I updated the names to explicitly indicate the decays of each boson in the process name.

Best Regards, 
G. Vian.

lviliani:
Thanks a lot @G-Vian!
The new cards look good to me, @sihyunjeon can you please double check before merging this PR?

G-Vian:
Dear GEN conveners,
I have modified also the  TT4b folder to include MadSpin cards executing the decays.

PR #3819, opened 2024-12-13
Add cards for producing Run2 UL VBF HH in Run3 sheme

tolange:


lviliani:
Hello, can you please commit only a set of template cards plus a python or bash script that will produce all the cards that you need changing the relevant parameters?

lviliani:
There are no recent updates on this PR, so I will close it.
In case it is still relevant, please migrate it to the new gitlab repository:

PR #3820, opened 2024-12-17
Check_dag_success.py should run in python3 in master

adewit:
The master branch is, according to https://cms-gen.gitbook.io/cms-generator-central-place/how-to-produce-gridpacks/powheg-box, to be used with CMSSW_13, in which python 2 is not available anymore, so updating check_dag_success to use python3


lviliani:
I think this is good to go.
@covarell @agrohsje @jshin96  can you confirm?

PR #3825, opened 2024-12-19
adding VBF off-shell semileptonic samples [ Run2 UL] 

sv3048:
Hi !

I am adding the sample cards from VBF production mode for Semileptonic Off-Shell Analysis ( H->WW->2l2nu). Here both the W's are in on-shell mode.

Kindly take care of the PR and kindly let me know if there is anything needed.

Thanks,
Sadhana


bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/VBFHToWWSemilep_off_shell_4f_LO/VBFHToWWSemilep_off_shell_S_4f_LO/VBFHToWWSemilep_off_shell_S_4f_LO_proc_card.dat
```diff
@@ -0,0 +1,6 @@
+import model loop_sm
+define l- = e- mu- ta-
+define l+ = e+ mu+ ta+
+generate p p > h > j j w+ w- $h, w+ > l+ vl, w- > l- vl~
```
lviliani, line 4: I just noticed that here you are defining fully leptonic decays instead of semi-leptonic.

DickyChant:
The issue of `etaj` is that, if you target a vbf topology involving forward jet, then for typical CMS setup, a forward ak4 jet means a margin around 4.7 (the HF acceptance) + 0.4 (ak4 cone) = 5.1, you are really at the boundary.

I don't know why you bring MadSTR in the game since you are really sending LO cards instead of NLO, the key information is indeed the ttbar as Olivier pointed out, for LO, that could be done with a single $. 

While it is true that you have the freedom of doing 4f or 5f (and the handling of top is indeed tricky), in your slides you are trying to compare your generation with powheg, do you know if that's 4f or 5f to compare with?

lviliani:
What Sitian is saying is that you are removing GEN jets with eta>5, which might be too aggressive for a VBF process.
It would be interesting to check how much phase space you are cutting out, comparing the RECO jet eta distributions (and detajj) between the current sample, and a new one with etaj set to a larger value (e.g. 6.5).

PR #3826, opened 2024-12-19
add 13p6TeV DiPhotonJetsBox1BJet (2BJet) card for Sherpa

cao-yu-x:
In this request, we add the card for 13p6TeV Sherpa process DiPhotonJetsBox1BJet_MGG-80toInf and DiPhotonJetsBox2BJets_MGG-80toInf in which we simply change the beam energy based on the Run2 13TeV card.


shimashimarin:
Hi @cao-yu-x thanks for adding the configurations. In our private discussion, you have already done gen-level comparison with 13TeV. Which looks reasonable. @lviliani and @bbilin, can you merge this at your convenience? Thanks.

PR #3828, opened 2025-01-05
Fix check_dataset_names script for bcvegpy2-pythia8-evtgen generators

Kiarendil:
Dear colleagues, 

I've found that the current version of check_dataset_names.py script does not allows names which uses the following generators: "bcvegpy2-pythia8-evtgen". 
For e.g. for "BcToJpsiPi_TuneCP5_13p6TeV_bcvegpy2-pythia8-evtgen" I got "--Invalid ME-PS format" error. 

I suggest a minor fix to the script to allow it supports the names with these generators. The local check with the example above solves the problem. 

Best,
Kirill


lviliani:
Thanks Kirill for spotting this and proposing a solution!

PR #3184, opened 2022-05-25
Wprime to e+nu cards

valesark:
Wprime to e+nu process cards for different masses and couplings.

PR #3832, opened 2025-01-10
Add 2024 WprimetoENu .dat files to Run-3 production directory

miozdemi:
Hi, in this request I add the Wprime to e+nu .dat files for different masses and couplings.
Here only for M-1000 and kR-0p01, but the customize_card.py will generate the directories and files for all the other samples.
This is a copy for 2024 gridpack generation, that was approved here https://github.com/cms-sw/genproductions/pull/3165 and here https://github.com/cms-sw/genproductions/pull/3184 for 2022 and 2023 (both had the same gridpacks)


lviliani:
Hi @miozdemi, can you please remove `gridpack_submitter.py`?

miozdemi:
Hi @lviliani, sure! I remove it and pushed it to git again.

lviliani:
Thanks, can you also please explain why you need this in the run card?
`1e-12 = small_width_treatment`

According to [this](https://answers.launchpad.net/mg5amcnlo/+faq/3053), with this setting you are deactivating the default treatment.

miozdemi:
Hi, yes we are deactivating the default treatment by that, because in our case a W' with a low coupling (kR around 0.001) will have a much narrower width than the threshold allows. Therefore to accurately model the shape of the distribution for our limit setting we are deactivating it with that.
One gets for example (using the [WprimeWidth.py](https://github.com/miozdemi/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/13p6TeV/WprimeToENu_LO/WprimeWidth.py)):
WprimeWidth(mass=6000,coupling=0.001) / 6000 of around 3.39e-08 which exceeds the default threshold of 1e-6

lviliani:
Alright, thanks a lot for explaining.
I will merge this PR straight away.

PR #3839, opened 2025-01-23
Inputs for MiNNLO Z to ee/mumu/tautau in Z mass bins

ikalaitz:
Added inputs for MiNNLO DY ee/mumu/tautau in bins of Z mass [10, 50, 100, 200, 400, 600, 800, 1000, 1500, 2000, 4000, 6000, Infinity]


lviliani:
Hi @ikalaitz, can you please commit only a template card and a script (sh or python) to produce all the cards that you need?

ikalaitz:
Hi @lviliani, thank you for the comment, I added a script to generate all cards from the existing "ZJToMuMu-13p6TeV-suggested-nnpdf31-ncalls-doublefsr-q139-powheg.input" card inclusive in mZ.

PR #3841, opened 2025-01-28
Update gridpack_generation.sh

michael-pitt:
minor typo that prevents running this branch (3.6->3.5)

### DIFF
--- a/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
+++ b/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
@@ -137,7 +137,7 @@ make_gridpack () {
     
     MG_EXT=".tar.gz"
     MG=MG5_aMC_v3.5.6$MG_EXT
-    MGSOURCE=https://launchpad.net/mg5amcnlo/3.0/3.6.x/+download/$MG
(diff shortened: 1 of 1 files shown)


lviliani:
Thanks @michael-pitt for spotting this.
This is a bit puzzling though, as this branch has been used for the validation of MG 356...
@joonblee @tyyang99 am I missing something? Can you please have a look?

michael-pitt:
Looks like in https://github.com/cms-sw/genproductions/commit/c5fd9ff4467f523049f95ddbc2f2271629842a61 when the version was updated from 3.5.5 to 3.5.6 this additional change was introduced

joonblee:
Michael is right. After [c5fd9ff](https://github.com/cms-sw/genproductions/commit/c5fd9ff4467f523049f95ddbc2f2271629842a61), we are testing not only 35X but also 36X so we need to privately change those lines several times. I had tried to share that update only to genval members letting them know how to change the mg version, but I guess I mis-pushed it to the official branch. Thanks a lot Michael for checking it out.

lviliani:
Alright thanks. I will merge this PR then.

PR #3845, opened 2025-02-05
Updates for condor on CMS Connect in singularity

bryates:
This PR fixes condor on CMS Connect. The `cmssw-cc7-condor-python27` singularity container is required to run, and must be activated from the top `genproductions` directory.

### DIFF
--- a/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
+++ b/bin/MadGraph5_aMCatNLO/gridpack_generation.sh
@@ -695,8 +695,6 @@ else
         cmssw_version=CMSSW_10_2_24_patch1 
     elif [[ $SYSTEM_RELEASE == *"release 7"* ]]; then 
         cmssw_version=CMSSW_10_6_19 
-    elif [[ $SYSTEM_RELEASE == *"release 9"* ]]; then
-	cmssw_version=CMSSW_13_2_9
     else 
         echo "No default CMSSW for current OS!"
         if [ "${BASH_SOURCE[0]}" != "${0}" ]; then return 1; else exit 1; fi        


bin/MadGraph5_aMCatNLO/gridpack_generation.sh
```diff
@@ -176,6 +176,9 @@ make_gridpack () {
       wget --no-check-certificate ${MGSOURCE}
       tar xzf ${MG}
       rm "$MG"
+
+      # agrohsje 
+      cp -rp ${PRODHOME}/addons/models/* ${MGBASEDIRORIG}/models/.
```
sihyunjeon, line 181: this is not needed, no? @agrohsje

bin/MadGraph5_aMCatNLO/gridpack_generation.sh
```diff
@@ -581,6 +585,10 @@ make_gridpack () {
       echo "cleaning temporary gridpack"
       rm $WORKDIR/pilotrun_gridpack.tar.gz
 
+      # awightma start: Force the re-weight step to only use 1 core
+      echo "nb_core = 1" >> $WORKDIR/process/madevent/Cards/me5_configuration.txt
```
sihyunjeon, line 589: this does not affect phase space integration?

bin/MadGraph5_aMCatNLO/gridpack_generation.sh
```diff
@@ -595,6 +603,7 @@ make_gridpack () {
         $WORKDIR/$MGBASEDIRORIG/MadSpin/madspin madspinrun.dat 
         rm madspinrun.dat
         rm -rf tmp*
+        cp $CARDSDIR/${name}_madspin_card.dat $WORKDIR/process/madspin_card.dat
```
sihyunjeon, line 606: Is this to keep the card once more in madspin directory although it's not 100% needed? or?

bin/MadGraph5_aMCatNLO/gridpack_generation.sh
```diff
@@ -728,6 +739,7 @@ fi
 
 #catch unset variables
 set -u
+unset PERL5LIB
```
sihyunjeon, line 742: This is to do what?

bin/MadGraph5_aMCatNLO/gridpack_generation.sh
bryates: I suppose it is not needed in general, though we use it when we make our EFT gridpacks.

bin/MadGraph5_aMCatNLO/gridpack_generation.sh
bryates: No, this is the step after when EFT reweighting is done. I can remove this too since it's not for general use.

bin/MadGraph5_aMCatNLO/gridpack_generation.sh
bryates: When running in a cc7 singularity there's a conflict with the version of Perl on CMS Connect. This makes it default to whatever the desired CMSSW version uses.

bin/MadGraph5_aMCatNLO/gridpack_generation.sh
sihyunjeon: we already have a way to deal with model files (although not significantly different in terms of logic) through `_extramodels.dat`
this will also just complain if you don't have that directory, should remove this

bin/MadGraph5_aMCatNLO/gridpack_generation.sh
bryates: Can probably remove

bin/MadGraph5_aMCatNLO/gridpack_generation.sh
bryates: Removed

bin/MadGraph5_aMCatNLO/submit_cmsconnect_gridpack_generation.sh
```diff
@@ -222,8 +227,9 @@ input_files="input_${card_name}.tar.gz"
 patches_directory="./patches"
 utilities_dir="./Utilities"
 plugin_directory="./PLUGIN"
+addons_dir="./addons"   # awightma
 if [ -e "$input_files" ]; then rm "$input_files"; fi
-tar -zchf "$input_files" "$card_dir" "$patches_directory" "$utilities_dir" "${plugin_directory}"
+tar -zchf "$input_files" "$card_dir" "$patches_directory" "$utilities_dir" "${plugin_directory}" "${addons_dir}"
```
sihyunjeon, line 232: this is not safe since not all people are going to be using a model that lives under addons directory. why not just steer it with `extramodels.dat` ? 

e.g.
https://github.com/cms-sw/genproductions/blob/master/bin/MadGraph5_aMCatNLO/cards/production/13TeV/SingleLepton_TypeIHeavyN/DYtoLNuJJ/DYtoLNuJJ_TypeIHeavyN-El_MN500_TuneCP5_13TeV-amcatnlo-pythia8/DYtoLNuJJ_TypeIHeavyN-El_MN500_TuneCP5_13TeV-amcatnlo-pythia8_extramodels.dat

this will be taken from [link](https://cms-project-generators.web.cern.ch/cms-project-generators/)

bryates: Thanks for explaining, I've removed this. Our current model is not in https://cms-project-generators.web.cern.ch/cms-project-generators/. Can we add it or can only people in the gen group make changes?

bin/MadGraph5_aMCatNLO/gridpack_generation.sh
sihyunjeon: This is due to a bug in madgraph that doesn't work when nb_core is not 1? If so, it might be worth keeping it

bin/MadGraph5_aMCatNLO/submit_cmsconnect_gridpack_generation.sh
sihyunjeon: Should ask GEN conveners

bin/MadGraph5_aMCatNLO/gridpack_generation.sh
```diff
@@ -694,6 +703,8 @@ else
         cmssw_version=CMSSW_10_2_24_patch1 
     elif [[ $SYSTEM_RELEASE == *"release 7"* ]]; then 
         cmssw_version=CMSSW_10_6_19 
+    elif [[ $SYSTEM_RELEASE == *"release 9"* ]]; then
+	cmssw_version=CMSSW_13_2_9
```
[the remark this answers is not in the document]
sihyunjeon: I think we should officially disable el9 instead of making it work like this (el9 gridpacks are not going to be usable for official production) , talked to sitian

bin/MadGraph5_aMCatNLO/gridpack_generation.sh
Andrew42: this might no longer be the case, but when the reweighting step gets run using a multicore setup it would crash or otherwise incorrectly combine the jobs produced by each core.

From what I remember it seemed like some sort of race condition where the code would crash if the first/initial job didn't finish first and would complete when the first job finished first, but in this case I was skeptical of the output since it seemed like the subsequent jobs would overwrite the output in a way that did not seem correct (e.g. the output file size seemed to bounce around as the jobs completed with no indication that there was some sort of post-job merging going on).

The other reason to limit the number of cores explicitly is b/c otherwise madgraph will default to try and use as many cores as possible on the host machine and these gridpacks were produced using CMSConnect, which meant that the reweighting step would end up saturating every possible core on the _submission_ node for CMSConnect, which would be very bad.

sihyunjeon: Ah... yes madgraph had this bug for the initial phase space integration long time ago (eating up all cores) but looks like for reweighting this behavior is still there in 2.6.5 (not sure about 2.9.x versions).

@bbilin @lviliani maybe consider this as well in master branch? or i am raising this up again which @DickyChant and I were hoping for : it's a lot easier just to keep master as the default for all campaigns instead of spending time to fix sth on both branches and deprecate old ones if the analyses are just new

lviliani:
Hello, as we discussed this PR should go in the master branch as well.
We can merge it also in `mg265UL` given that it's ready and then we don't touch this branch anymore.
@sihyunjeon is this PR ready to be merged?

PR #3848, opened 2025-02-18
New HV from Carlo Oleari

covarell:
All HV, HVJ and HVJ_ew processes again operational in CMSSW13 (from Carlo Oleari).

Very important for all Run3 VH NLO and MiNLO productions. 

@agrohsje @jshin96 @yihui-lai @mlizzo


lviliani:
Thank you @covarell !
I see you have also edited the `b_bbar_4l` part in `helpers.py`.
Is it also related to the update from Powheg authors?

covarell:
Yes, since this is using OpenLoops2 now, the old patch does not apply anymore (same for all VH processes)

PR #3849, opened 2025-02-19
Diboson wv SMEFT cards

GiacomoBoldrini:
Datacards for SMEFT interpretation of diboson WV with Run-2 data, split in WW and WZ production


anmehta:
Dear @bbilin, @lviliani,

We pushed the generator cards for our SMEFT signal samples (SMP-25-001). We are targeting dim6 operators using WV final state with fullrun2 dataset in this analysis and doing a private signal sample generation. Please let us know about your comments and suggestions.

Best,
Giacomo and Ankita

lviliani:
Hi @GiacomoBoldrini @anmehta, sorry for the delay.

I think there's an inconsistency between proc card and run card.
As far as I understand you want LO processes with MLM merging of 0 and 1 extra jet, can you confirm?

In this case you should change the run card and se ickkw=1. Also in this case xqcut should not be 0.

You can have a look at other MLM LO samples as a reference, e.g. this one (just a random example):
https://cms-pdmv-prod.web.cern.ch/mcm/requests?prepid=GEN-Run3Summer23wmLHEGS-00581&page=0&shown=140737488355327

GiacomoBoldrini:
Hi @lviliani thanks for checking, indeed your point is correct. I have fixed the run cards

PR #3850, opened 2025-02-20
DarkHiggs ZZ Run2UL cards

vhegde91:
Cards for DarkHiggs to ZZ. Similar to https://github.com/cms-sw/genproductions/tree/mg265UL/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/DarkHiggs_WW except for ZZ decays of dark higgs.


lviliani:
Hi @vhegde91, could you please remove the model tar file? That should go to our website and we can copy it for you.
Thanks.

vhegde91:
Hi @lviliani , thanks for the follow up. Yes, model tar file has been removed and it can be copied from `/afs/cern.ch/work/v/vhegde/public/Physics/darkHiggsSampleGen/ZpHiggs_UFO.tar.gz`. We would like to keep these run cards (DarkHiggs_ZZ) similar to DarkHiggs_WW. Do I need to copy these lines from DarkHiggs_WW? https://github.com/cms-sw/genproductions/blob/mg265UL/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/DarkHiggs_WW/DarkHiggs_WW_run_card.dat#L280-L291

vhegde91:
Hi @lviliani, are there any further comments on this?

lviliani:
Hi, your model has been copied to:
Regarding the additional lines, I think they are not needed.

PR #3851, opened 2025-02-25
Adding RPV single stop Run 2 MadGraph cards

devinmahon:


bin/MadGraph5_aMCatNLO/cards/production/13TeV/singleStop/coupling312_minus/singleStop_Run2_312_MSTOP_MCHI_minus_proc_card.dat
```diff
@@ -0,0 +1,40 @@
...
+set default_unset_couplings 99
+set group_subprocesses Auto
+set ignore_six_quark_processes False
+set loop_optimized_output True
+set low_mem_multicore_nlo_generation False
+set loop_color_flows False
+set gauge unitary
+set complex_mass_scheme False
+set max_npoint_for_channel 0
+import model ../../../RPVMSSM_UFO --modelname
```
lviliani, line 32: I think this will not work when you create the gridpack.
It should probably be just:
`import model RPVMSSM_UFO`
And the model should be uploaded to the GEN webpage (we can do it if it's not there already)

bin/MadGraph5_aMCatNLO/cards/production/13TeV/singleStop/coupling312_plus/singleStop_Run2_312_MSTOP_MCHI_plus_madspin_card.dat
lviliani, line 1: Is there a reason to use MadSpin instead of Madgraph directly for the decays?

bin/MadGraph5_aMCatNLO/cards/production/13TeV/singleStop/coupling312_plus/singleStop_Run2_312_MSTOP_MCHI_plus_param_card.dat
lviliani, line 1: The full param card can be dropped and you can add a customize card changing only the parameters that you need.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/singleStop/coupling313_plus/singleStop_Run2_313_MSTOP_MCHI_plus_run_card.dat
```diff
@@ -0,0 +1,282 @@
...
+#***********************************************************************
+ -1.0  =  ktdurham        
+ 0.4   =  dparameter
+ -1.0  =  ptlund
+ 1, 2, 3, 4, 5, 6, 21, 1000001, 1000002, 1000003, 1000004, 1000005, 1000006, 1000021, 2000001, 2000002, 2000003, 2000004, 2000005, 2000006  =  pdgs_for_merging_cut ! PDGs for two cuts above   
+#*********************************************************************
+# maximal pdg code for quark to be considered as a light jet         *
+# (otherwise b cuts are applied)                                     *
+#*********************************************************************
+ 4 = maxjetflavor    ! Maximum jet pdg code
```
lviliani, line 271: Here it should be 5 if you are using 5 flavor scheme.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/singleStop/coupling313_minus/singleStop_Run2_313_MSTOP_MCHI_minus_run_card.dat
```diff
@@ -0,0 +1,282 @@
...
+#***********************************************************************
+ -1.0  =  ktdurham        
+ 0.4   =  dparameter
+ -1.0  =  ptlund
+ 1, 2, 3, 4, 5, 6, 21, 1000001, 1000002, 1000003, 1000004, 1000005, 1000006, 1000021, 2000001, 2000002, 2000003, 2000004, 2000005, 2000006  =  pdgs_for_merging_cut ! PDGs for two cuts above   
+#*********************************************************************
+# maximal pdg code for quark to be considered as a light jet         *
+# (otherwise b cuts are applied)                                     *
+#*********************************************************************
+ 4 = maxjetflavor    ! Maximum jet pdg code
```
lviliani, line 271: Here it should be 5 if you are using 5 flavor scheme.

lviliani (commented):
Some of the inline comments apply to all cards.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/singleStop/coupling312_minus/singleStop_Run2_312_MSTOP_MCHI_minus_proc_card.dat
devinmahon: I did manage to successfully create the gridpacks this way by tarring up the model and placing it in the right location. I don't think it worked when I ran it just as `import model`, which is why I did it this way, so maybe it does need to be added.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/singleStop/coupling312_plus/singleStop_Run2_312_MSTOP_MCHI_plus_madspin_card.dat
devinmahon: In consultation with the theorist who was the author of the `RPVMSSM_UFO` model, I was advised to not do, e.g. `generate p p > su1 > b b~ d~ s~` and to stick to the 2-to-2 process followed by a decay of the chargino in MadSpin with `spinmode = none`.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/singleStop/coupling312_plus/singleStop_Run2_312_MSTOP_MCHI_plus_param_card.dat
devinmahon: Do you have an example? I haven't seen this before.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/singleStop/coupling313_plus/singleStop_Run2_313_MSTOP_MCHI_plus_run_card.dat
devinmahon: OK, I'm not very familiar with this, but I was reading that at leading order without matching, it doesn't matter which is used. Feel free to correct me.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/singleStop/coupling312_minus/singleStop_Run2_312_MSTOP_MCHI_minus_proc_card.dat
lviliani: The model should be uploaded to the GEN webpage as I said. Then the syntax I proposed should work.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/singleStop/coupling312_plus/singleStop_Run2_312_MSTOP_MCHI_plus_param_card.dat
lviliani: Here is just a random example:
https://github.com/cms-sw/genproductions/blob/35da69a5f0be4c9fabc9dc44d94be5be3862741f/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/TprimeTprime_ttHH_LO/TprimeTprime_ttHH_customize_card.dat#L4

You can find several similar cases in this repository.

bin/MadGraph5_aMCatNLO/cards/production/13TeV/singleStop/coupling313_plus/singleStop_Run2_313_MSTOP_MCHI_plus_run_card.dat
lviliani: I think you are right, but now I'm confused.
Why are you adding b's in the proton definition for this process?

devinmahon: The stop production with the RPV UDD 313 coupling requires a b in the proton. I'm not really sure what this `maxjetflavor` line is doing. I redefine the proton to contain b's in the `proc_card`. Do I need to change this as well for anything to be produced, or is this `maxjetflavor` related to something else?

lviliani: If you need 5 flavor scheme for this process, then it's fine to redefine the proton to include b quarks. Usually in this case we also include the b quarks in the jet definition and we set `maxjetflavor` to 5 (which means also that all the jet cuts in the run card will be applied to b quarks as well).

bin/MadGraph5_aMCatNLO/cards/production/13TeV/singleStop/coupling312_plus/singleStop_Run2_312_MSTOP_MCHI_plus_param_card.dat
devinmahon: Thanks. Can I ask if it's necessary to use this customize card? I don't remember what all I had to change from the default in the param card, and I've already setup the infrastructure to generate all the gridpacks this way and would prefer not to mess with things and have to re-validate the workflow.

devinmahon:
I have implemented the needed suggestions above as well as some other bug fixes I noticed.  I've done the same for the Run 3 cards in #3852. Regarding the flavor scheme for the 313 coupling, I have changed `maxjetflavor` to 5, set the b mass to 0, and added b's to the light jet definition. Could you please have another look and see if everything looks good now?

devinmahon:
@lviliani, just wanted to check that you saw my update above. If you could take another look when you get the chance, that would be great.

PR #3859, opened 2025-02-28
Add run and proc cards for V+Gamma, EWKV leptonic, EWKV hadronic

weissben2000:
Add run and proc cards for V+Gamma, EWKV leptonic, EWKV hadronic. Without associated gridpacks this time!


sihyunjeon:
There are more or less equivalent samples that i produced in run3. Any particular reason to ask for another set?

weissben2000:
> @sihyunjeon There are more or less equivalent samples that i produced in run3. Any particular reason to ask for another set?

Thanks for letting me know! The cards added here are for runs at 13.6 TeV but are close copies of cards at 13 TeV. If you produced some 13.6 TeV cards for these processes can you send a link to them so we can compare?

sihyunjeon:
https://github.com/cms-PdmV/GridpackFiles/tree/master/Cards/MadGraph5_aMCatNLO/VBF

sihyunjeon:
https://github.com/cms-PdmV/GridpackFiles/blob/master/Cards/MadGraph5_aMCatNLO/Z/ZGto2QG-1Jets_Bin-PTG-100to200_amcatnloFXFX-pythia8/

sihyunjeon:
https://github.com/cms-PdmV/GridpackFiles/tree/master/Cards/MadGraph5_aMCatNLO/W/WGto2QG-1Jets_Bin-PTG-100to200_amcatnloFXFX-pythia8

weissben2000:
> https://github.com/cms-PdmV/GridpackFiles/tree/master/Cards/MadGraph5_aMCatNLO/VBF

Thank you! It looks like the V+gamma cards in this PR are definitely redundant. 

What is the reason for choosing the $$ syntax in your VBF W cards such as: https://github.com/cms-PdmV/GridpackFiles/blob/master/Cards/MadGraph5_aMCatNLO/VBF/VBFWto2Q_madgraph-pythia8/VBFWto2Q_madgraph-pythia8_proc_card.dat ?
We believe we should use a singe $ instead but perhaps the difference would be minimal.

lviliani:
The `$$` syntax will remove all diagrams with s-channel propagators, while the single `$` will only remove the on-shell contribution but not the diagram IIRC.
But I will let @sihyunjeon comment further.

PS: please close this PR if it's not relevant anymore

sihyunjeon:
sorry i forgot to reply here. lorenzo is correct.
and in terms of gauge invariance safety, as long as `$$` works, there is no reason to use `$` as this is more dangerous

jennetd:
Thanks all for your help on this. We are happy with the syntax in the existing samples. I have one last question about interference with the QCD samples. Do we need to generate this separately? Has the size of this contribution been studied?

nsmith-:
Wait, I was under the impression $$ is more dangerous because it rejects off-shell contributions as well (M beyond bwcutoff) that then are never considered
See e.g. slides 27-31 in https://cp3.irmp.ucl.ac.be/projects/madgraph/raw-attachment/wiki/Madrid2015_2%20MC/Tutorial.pdf

sihyunjeon:
ah nick, sorry my memories have swapped. let me try to clarify (altho i still am very vague...), my statement on gauge invariance was just totally wrong above, $$ is more dangerous in that sense yes.

the problem for VBF W in 5FS is with top showing up
"pp > wjj" includes "pp > wbb~" which can be from "pp > tb~ > wbb~" which actually is a single top process and we don't want this for sure.

now the next question is, between $ or $$ what should we use.
$$ : remove "all top related contributions both on and offshell", thinking these are considered "unwanted" for "VBF samples".
$ : remove "only onshell top related contributions".
and the decision was on former. i did check the differences between them in LHE level the past and it was negligible (since it's offshell top that we need to see)

sihyunjeon:
> Thanks all for your help on this. We are happy with the syntax in the existing samples. I have one last question about interference with the QCD samples. Do we need to generate this separately? Has the size of this contribution been studied?

From our side (common bkg) we haven't considered other contributions (i guess you are talking about cases when the quarks radiates two gluons and those two meet to make W/Z?), we can add them easily (just changing QED=X QCD=Y numbers in the proc card) as long as anyone can volunteer to see if what we made is done as intended

jennetd:
> > Thanks all for your help on this. We are happy with the syntax in the existing samples. I have one last question about interference with the QCD samples. Do we need to generate this separately? Has the size of this contribution been studied?
> 
> From our side (common bkg) we haven't considered other contributions (i guess you are talking about cases when the quarks radiates two gluons and those two meet to make W/Z?), we can add them easily (just changing QED=X QCD=Y numbers in the proc card) as long as anyone can volunteer to see if what we made is done as intended

Ok. I started generating some gridpacks with QED=X, QCD=Y etc to get an idea of the interference. Good to confirm that this is the right approach. If we find that it is large and needs to be simulated then I will open a new PR

PR #3853, opened 2025-02-25
cards for very-low-mass DYToEE (madgraph, LO)

missirol:
Adding cards for a very-low-mass DY to dielectron sample (~~up to 1~~ no additional parton in the ME, madgraph-MLM, LO).

To be used for trigger studies for 2025 (`Run3Winter25` MC campaign).

The cards are based on the ones used for
https://cms-pdmv-prod.web.cern.ch/mcm/public/restapi/requests/get_fragment/GEN-Run3Winter25wmLHEGS-00003/0

FYI: @lathomas

PR #3860, opened 2025-03-02
replace `DYToEE_MEE-0p1To4_LO_5f` with `DY1jToEE_MEE-0p1To4_PtE7_PtJ15_LO_5f`

missirol:
The cards introduced in #3853 are incorrect, because of an inconsistency between
 - the process, i.e. LO DY without extra partons in the ME,
 - the GEN-level cuts on the inv-mass of the dielectron system, i.e. `0.1 < Mee [GeV] < 4`, and
 - the GEN-level cuts on the electrons in the ME final state, i.e. `pT_e > 7 GeV`),

which led to a cross section of "zero", and a runtime error when trying to generate the gridpack.

This PR replaces those cards with new ones where
 - the process is changed to `p p > e+ e- j` (1extra parton in the ME)
 - a GEN-level cut on the parton pT is applied (`ptj > 15`), to be consistent with the GEN-level cuts applied to the electrons in the ME final state.

This makes the sample exclusive (does not cover the full phase space), but this is considered okay for the purpose of the studies for which this sample is being produced (trigger-efficiency studies for low-mass dielectron pairs, as opposed to a physics measurement).

For clarity, the main GEN-level cuts are mentioned in the name of the files.

With these cards, the gridpack-creation step works without issues.

FYI: @lathomas


missirol:
Sorry to push, but we would like to proceed quickly with the production of a small MC sample based on these cards. Could you please review the cards today ?

Thanks !

lviliani:
Hi @missirol, as we discussed already, we will approve these cards to proceed with the trigger studies that you need to perform quickly.

But let me stress that we do not guarantee that this sample will be sound from a physics perspective, given the extremely low mass phase space and the custom cuts.
The validation of this sample would normally require more time and can't be done in such a short timescale.

PR #3861, opened 2025-03-04
Adding HJ_MiNNLO cards

adewit:
This adds cards for HJ_MiNNLO, with the settings described in the [GEN meeting on 24/02](https://indico.cern.ch/event/1516779/contributions/6382701/attachments/3019221/5326259/ggH_MiNNLO_AdeWit_2402.pdf)

Note: I'm not sure the pdf sets used are correct, so comments on that welcome.

In the gridpacks, I'd propose to include the scale weights and the pdf weights for the main set of pdfs - I have been removing weights manually (editing the pwg-rwl file and recreating the tarball), but if there is another method that is recommended for this I'd be happy to hear it.


lviliani:
Hi Adinda, I think the default PDF set should be 325300 (NNPDF31_nnlo_as_0118_mc_hessian_pdfas).

lviliani:
And regarding the weights to include you can check this script for editing the pwg-rwl file:
https://github.com/cms-sw/genproductions/blob/master/bin/Powheg/make_rwl.py

adewit:
Thanks @lviliani and apologies for the delay. I've updated the PDF set in the cards.

PR #3863, opened 2025-03-05
Cards for generating TTHH with ttbar dileptonic and semileptonic decay

tiyang-hanyang:
TTHH generation with ttbar decay modes in di-leptonic (excluding tau) and semi-leptonic channel.
Using the genfragments/ThirteenPointSixTeV/Higgs/HH/HHto2B2Tau_dipoleRecoilOn-TuneCP5_13p6TeV_madgraph-pythia8.py and genfragments/ThirteenPointSixTeV/Higgs/HH/HHto2B2V_dipoleRecoilOn-TuneCP5_13p6TeV_madgraph-pythia8.py for production.


tiyang-hanyang:
> We merge, but a question on tau final state, how will you handle those cases? It will still be a background or even signal.

Sorry that I missed the notification of the reply. Yes, this version indeed does not include the tau final state. Meanwhile I also created the cards that add the tau lepton final state and I tested the generation of both. 
I will add one additional commit to merge the tau lepton final state.
Many thanks!

PR #3867, opened 2025-03-19
sherpa ttbar lnqq cards

javi-llorente:


bin/Sherpa/cards/production/13TeV/ttbar_semilepton/Run.dat_ttbar_SemileptonM_13TeV
```diff
@@ -0,0 +1,61 @@
...
+  INTEGRATION_ERROR=0.05;
+
+  % collider setup
+  BEAM_1 2212; BEAM_ENERGY_1 6500.;
+  BEAM_2 2212; BEAM_ENERGY_2 6500.;
+  PDF_LIBRARY     LHAPDFSherpa;
+  PDF_SET         NNPDF31_nnlo_as_0118;
+  PDF_SET_VERSION 0;
+  PDF_GRID_PATH   PDFsets;
+  PDF_VARIATIONS=NNPDF31_nnlo_as_0118[all]
```
sihyunjeon, line 35: NNPDF30 was the default PDF set IIRC @shimashimarin

bin/Sherpa/cards/production/13TeV/ttbar_semilepton/Run.dat_ttbar_SemileptonM_13TeV
```diff
@@ -0,0 +1,61 @@
...
+
+  % decays
+  HARD_DECAYS On;
+  SOFT_SPIN_CORRELATIONS 1;
```
sihyunjeon, line 39: what is the benefit of SOFT_SPIN_CORRELATIONS over HARD_SPIN_CORRELATIONS? it just ignores the tau effect with SOFT_SPIN_CORRELATIONS

bin/Sherpa/cards/production/13TeV/ttbar_semilepton/Run.dat_ttbar_SemileptonM_13TeV
```diff
@@ -0,0 +1,61 @@
...
+  HDH_STATUS[24,2,-1]=2
+  HDH_STATUS[24,4,-3]=2
```
sihyunjeon, line 41: CKM mixing is not allowed in Sherpa?

bin/Sherpa/cards/production/13TeV/ttbar_semilepton/Run.dat_ttbar_SemileptonM_13TeV
```diff
@@ -0,0 +1,61 @@
...
+
+(processes){
+  Process : 93 93 ->  6 -6 93{NJET};
+  Order (*,0); CKKW sqr(QCUT/E_CMS);
+  NLO_QCD_Mode MC@NLO {LJET};
+  ME_Generator Amegic {LJET};
+  RS_ME_Generator Comix {LJET};
+  Loop_Generator LOOPGEN {LJET};
+  Associated_Contributions EW|LO1|LO2|LO3 {LJET};
+  Enhance_Function VAR{pow(max(sqrt(H_T2)-PPerp(p[2])-PPerp(p[3]),(PPerp(p[2])+PPerp(p[3]))/2)/30.0,2)} {3,4,5,6}
```
sihyunjeon, line 59: What does this variable mean qualitatively?

bin/Sherpa/cards/production/13TeV/ttbar_semilepton/Run.dat_ttbar_SemileptonM_13TeV
```diff
@@ -0,0 +1,61 @@
...
+  METS_BBAR_MODE 5;
+  NLO_CSS_PSMODE=1;
+  NLO_SUBTRACTION_SCHEME=2;
+
+  % tags for process setup
+  NJET:=4; LJET:=2,3; QCUT:=30.;
+
+  % me generator settings
+  ME_SIGNAL_GENERATOR Comix Amegic LOOPGEN;
+  EVENT_GENERATION_MODE Weighted;
```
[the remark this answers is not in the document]
javi-llorente: Hi, I will implement these comments and see how things work. Thanks.

bin/Sherpa/cards/production/13TeV/ttbar_semilepton/Run.dat_ttbar_SemileptonM_13TeV
shimashimarin: Yes. Sherpa tuned some parameters based on NNPDF30. So please replacing by the following:

```
  PDF_SET NNPDF30_nnlo_as_0118;
  PDF_VARIATIONS NNPDF30_nnlo_as_0118[all];
```

bin/Sherpa/cards/production/13TeV/ttbar_semilepton/Run.dat_ttbar_SemileptonM_13TeV
javi-llorente: This is the maximum value of the scalar sum of pT of the radiation (the recoil against the ttbar system) and the scalar sum of pT of the ttbar system. What this command does is to enhance the region of the phase space with hard radiation, so analysis looking at tt+jets configurations get enough statistics.
This is the same approach as used by ATLAS in 2112.09588 for V+jets (see Section 4.2)

sihyunjeon: but then doesn't that mean we loose statistics for soft radiation phase space ?
not sure what the use case for this will be but since we are even yet missing nominal ttbar, we should first have nominal ttbar and then move on to biased samples?
Or at least see if the bias module works fine? This is not always 100% guaranteed to work (sample with bias module doesn't really follow nominal sample distributions) from previous examples I saw

javi-llorente: The cross section of the LO ttbar configuration (no additional radiation) is still much larger, so there are plenty of statistics for it. This configuration has been validated here:
https://indico.cern.ch/event/1523797/contributions/6411006/attachments/3028502/5346126/sherpa.pdf

J.

javi-llorente: Hi @sihyunjeon, @shimashimarin,
I opened a new PR here with these comments implemented: https://github.com/cms-sw/genproductions/pull/3870
Let's keep the discussion there, if any.

Cheers,
Javier.

sihyunjeon: > The cross section of the LO ttbar configuration (no additional radiation) is still much larger, so there are plenty of statistics for it. This configuration has been validated here:
> 
> https://indico.cern.ch/event/1523797/contributions/6411006/attachments/3028502/5346126/sherpa.pdf
> 
> 
> 
> J.
> 
> 

Ehhh the whole point of using bias module is that smaller the cross section is, you get compensated with larger statistics. So larger the cross section is, you get disadvantage in statistics so what you wrote is contradictory (otherwise I am missing something in between)

Who are the customers for these samples (== are they in desperate need for hard radiation / boosted tt system phase space)? At least did you compare sherpa with bias module and without bias module gives you expected compatible results? Youve compared it against other generators which is meaningful but still does not give clear picture on whether it functions properly or not

javi-llorente: This is not only compared to other generators, but also to real data from ATLAS and CMS, which Sherpa describes better than either Powheg or MG. What we want to describe is the data, not the unbiased sample, and we proved that this sample does it better than the current ones. Moreover, if you look at slides, you'll see that we don't have any statistics problems in any of the regions, either inclusively or with additional jets.

The main customer for this sample, for the moment, is the ttbar+jet analysis that I'm carrying over together with my student, you can have a look at this talk:
https://indico.cern.ch/event/1430165/contributions/6016304/attachments/2889519/5065133/topR32.pdf
So yes, we need additional radiation. Lots of it.
 
Please, make it happen. Thanks.

javi-llorente: Maybe MiNNLO has issues if you can find the reference but, fortunately, I'm using Sherpa. For Sherpa, the enhancing factors have been thoroughly studied by ATLAS in the reference I posted above already two weeks ago (Fig. 10 and 11 and, in general, Section 4.2).
If you look at the plots attached below, you can see that there is no difference on the cross section between the enhanced and the usual sample, just as in the ATLAS paper.

We need to generate this sample as soon as possible, as we expect major improvements in our analysis from it.

d10-x01-y01
d279-x01-y01

lviliani: Thanks a lot for sharing these plots!

Could you perhaps compare the two plots (biased and unbiased, even without data) showing the statistical uncertainties? If the bias procedure is working as expected, you should have smaller statistical uncertainties compared to the unbiased in the tail of the distributions where the cross section is smaller.

From my side, this is not a showstopper to produce the samples, but I would really like to see the biasing in action. And I think it would also be a nice comparison to showcase this approach.

javi-llorente: Hi Lorenzo,
This is what is shown in the plots, but the stat. uncertainties are small enough so they're just seen in the ratio panels.
The plot on the top pT is for the inclusive sample, so the red line has a slightly larger error.
The plot on deltaPhi is for Njet = 6 (i.e. one additional jet, since we're looking at the semileptonic channel), so in this case it's the blue line having a larger uncertainty.

Here are some more plots for Njet=0, Njet=1, Njet=2 and Njet >= 3, where you can see the effect for increasing jet multiplicity.

d104-x01-y01
d105-x01-y01
d106-x01-y01
d107-x01-y01

javi-llorente: Hi @lviliani,
A few plots with the relative statistical uncertainties, so it is more clear.

Cheers,
J.

jet0
jet1
jet2
jet3

PR #3871, opened 2025-03-24
Add VBF and modify ttH

Paul-Krueper:
Added VBF H->GG in PowHEG at 120,125,130

existing ttH in PowHEG is modified to include the varying decay widths at 120,125,130

### DIFF
--- a/bin/Powheg/production/Run3/13p6TeV/Higgs/VBFHToGG_TuneCP5_13p6TeV-powheg-jhugen-pythia/vbfnlo.input
+++ b/bin/Powheg/production/Run3/13p6TeV/Higgs/VBFHToGG_TuneCP5_13p6TeV-powheg-jhugen-pythia/vbfnlo.input
@@ -1,0 +1,18 @@
+!ccccccc THIS IS THE INPUT FILE RELATED TO VBFNLO-INPUT ccccccccccccccccc
+!c
+!c      adapted from vbfnlo.dat 
+!c
+
+!   Physics parameters
+!------------------------ 
+HMASS        125.0d0         ! Higgs mass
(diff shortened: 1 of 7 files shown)


bin/Powheg/production/Run3/13p6TeV/Higgs/VBFHToGG_TuneCP5_13p6TeV-powheg-jhugen-pythia/vbfnlo.input
lviliani, line 1: Do you need this file?

Paul-Krueper: "VBF_H" in Powheg is NLO, and the genproduction powheg twiki mentioned that this file needs to be added to the directory for the process "VBF_HJJJ", which is also NLO. 
So since both are NLO, I just added the nlo input just in case it is needed. 
Let me know if this is incorrect. Thanks

lviliani: I'm not sure...
@covarell do you know if we need the `vbfnlo.input` for this process?

lviliani: Ok thanks. Then I suggest removing it to avoid misunderstandings.

PR #3872, opened 2025-03-24
14 TeV VH & ttH Powheg cards for phase 2 L1T studies

portalesHEP:
As said in the title, these cards are meant to be used to generate samples at 14 TeV for phase 2 studies, and are needed by the L1T DPG. 

They are exact copies of the equivalent Run 3 cards already used for production, with only the beam energies updated.


lviliani:
Hi @portalesHEP, given that these cards are for Phase 2, I would not put them inside the 13p6TeV directory.

Can you please move them to a new directory, which for example could be `Powheg/production/Phase2/14TeV/`?

PR #3873, opened 2025-03-25
Adding offshell qqH input cards

MohitS704:
Adds the Madgraph quark-antiquark Higgs signal input cards and diagram filter to the GenProductions repo so that it can be produced with Monte Carlo.


lviliani:
I haven't gone through all your diagram filtering logic yet, but I have a question before doing that.

Can't you achieve the same results using the standard Madgraph syntax?

For example, your BKG could be (if I understand correctly what you want):
`generate q q~ > l+ l- l+ l- / h  NP<=1 SMHLOOP=0 NPprop=0`
Here `/` is removing all the diagrams containing `h`.

Then SIG would be:
`generate q q~ > h > l+ l- l+ l- NP<=1 SMHLOOP=0 NPprop=0`

And BSI:
`generate q q~ > l+ l- l+ l-  NP<=1 SMHLOOP=0 NPprop=0`

Or would the syntax above generate something different wrt your needs?

MohitS704:
> I haven't gone through all your diagram filtering logic yet, but I have a question before doing that.
> 
> Can't you achieve the same results using the standard Madgraph syntax?
> 
> For example, your BKG could be (if I understand correctly what you want): `generate q q~ > l+ l- l+ l- / h NP<=1 SMHLOOP=0 NPprop=0` Here `/` is removing all the diagrams containing `h`.
> 
> Then SIG would be: `generate q q~ > h > l+ l- l+ l- NP<=1 SMHLOOP=0 NPprop=0`
> 
> And BSI: `generate q q~ > l+ l- l+ l- NP<=1 SMHLOOP=0 NPprop=0`
> 
> Or would the syntax above generate something different wrt your needs?

We filter diagrams to ensure only HZZ is simulated as well - the filter also removes ZH-like diagrams where there's Z->ll and H->ll decay. 

We were also been reluctant to use the `>` syntax in off-shell, as (in our experience) it forces Madgraph to make resonant diagrams, ignoring the t/u channels. There are only s-channel diagrams in the signal we are modelling here, but there the filter is relevant once again.

In hindsight, I should probably remove the background and BSI cards from this PR, as we are only requesting signal simulation.

MohitS704:
@lviliani Is this okay?

lviliani:
Hi @MohitS704, sorry for the delay.
Thanks for explaining, but I'm still a bit confused.
For BKG and your filter should be equivalent to the syntax I suggested, right?
For the other two cases, did you look at the MG diagrams to check that the filter is working as expected?

MohitS704:
Hello @lviliani , to introduce this, the gridpacks actually already exist and were submitted to and tested by the HZZ Monte Carlo contacts. All that remains is this PR. 

They were created independently of the gridpack creation script - but they do use the `runcmsgrid_LO.sh` script that exists for Madgraph in this repo. Basically, I generated the gridpacks myself in my own area to leverage my diagram filters and other personal edits (we also made our own factorization scale scheme of sqrt(s)/4, since doing it with the normal setup and a division prevents Madgraph from calculating PDF uncertainties).

> Hi @MohitS704, sorry for the delay. Thanks for explaining, but I'm still a bit confused. For BKG and your filter should be equivalent to the syntax I suggested, right? For the other two cases, did you look at the MG diagrams to check that the filter is working as expected?

Yes they would be equivalent - I have looked at the MG diagrams to check that the filter is working. 


> I also have some minor comments:
> 
>     * please put the cards in `MadGraph5_aMCatNLO/cards/production/13p6TeV` (if they are for Run3)

The Cards are for both Run2 and Run3 - I can also copy them over there as well

>     * the name of the proc cards should be `PROCESS_proc_card.dat`, where `PROCESS` is what you specify in the proc card output (e.g. `output PROCESS`). Otherwise I think our gridpack production script will not work.

I did not use the gridpack production script to create these - should I upload something else?

>     * please add also the run card.
 Will do
>     * what is the meaning of `--prefix=int` in the proc card output?
That changes the naming scheme of the subprocess functions to add a prefix of `_0` or `_1` to the functions. It's purely a cosmetic thing to make things easier to track internally - as now all the functions are not named the same thing. Makes for easier debugging when stepping through code in something like `gdb`.

lviliani:
Hi @MohitS704, the issue here is approving something that is not using the official CMS production tools.

Given that we don't know how the gridpacks are generated, it is difficult for us to judge if they are ok or not, not only physics-wise, but especially computing-wise for central production.

Is it possible to switch to the recommended tools for producing the gridpacks? If specific changes are needed for this process, those should be discussed and, if necessary, integrated in the CMS scripts.

mlizzo:
Hi @MohitS704 , HIG MC contact's speaking (adding @yihui-lai as well). We are in touch with Lazar for this request, and we have already shared with him our worries, which are in line with what @lviliani mentioned above. We weren't aware of the special needs you wanted to add for simulating this sample. This sort of things must be discussed upfront with us and GEN, so that we can try to accomodate your request. It's a pity that we had to find it out after asking you explictly, there wasn't any mention to this at the beginning of your PR. Just to be clear, we were neither informed by Lazar regarding these modifications to the code.

With that said, it's up to GEN to decide whether approving or not your request, but you cannot expect them to do this blindly. If you have to tweak the official tool, then please open a PR to genproductions. By the way, you can discuss whatever topic during HZZ meetings, but you need eventually to include responsible people for the final decision. 

To this extent, I see that you have already [some slides](https://indico.cern.ch/event/1495523/#32-update-on-the-study-of-the) documenting your studies. Since we have recently set up a HIG MC working meeting, maybe you could present and discuss them with us as well, also to prove that your procedure "works just fine" and as expected in terms of physics content - and I guess GEN will be interested in that too.

MohitS704:
That makes sense to me. To be entirely honest, I was not aware that the script was sort of "mandated" - this is I think the first time our group has requested MC with Madgraph.

I'll prep some slides detailing the technical procedure.

mlizzo:
Hi @MohitS704 , unfortunately our next HIG MC working meeting (Tuesday, April the 15th at 4pm CERN time) would overlap with the Physics Days on modelling and MC generators, therefore I don't think GEN will be able to attend it. @lviliani would it be possible to find a slot during next Monday's GEN meeting, if scheduled? Otherwise we can organize differently

MohitS704:
In the meantime, @mlizzo @lviliani, I've crafted some slides.

These will be the same slides I would present at a later date, but seeing as how that is up in the air, I thought it would be better if I shared them here as well.

The slides are linked [here](https://msrivast.web.cern.ch/.totally_not_a_secret_folder/misc_stuff/Gridpack_Creation_4_22_2025.pdf)

lviliani:
Dear @MohitS704, thanks for sharing the slides.

I think we can skip the presentation in a GEN meeting (unless you want to have it) and approve this PR.

For future requests, please get in touch with GEN and HIG MC contacts well in advance, so that we can discuss your needs and if they can be achieved with the central CMS tools. 

If not, we can always try to update the central tools to include the needed features, which can be useful for the whole collaboration, instead of implementing them in a private code.

MohitS704:
Hello @lviliani,

Thank you! And yes, in the future - we will definitely get into contact about any private gridpack creation styles, apologies about that.

PR #3874, opened 2025-03-27
fixed WV semilep proc cards, no madspin decay with MG

GiacomoBoldrini:
Removed madspin, decay in MG. Everything else is identical to previous version of the cards except for the process output name


lviliani:
Shouldn't `$$ t t~ h` be just after the first process?
`generate p p > ell+ vl w-  $$ t t~ h QCD=1 SMHLOOP=0 NP=1, w- > j j `

Also, I was wondering if you can have EFT in decays adding `NP=1` also after `w- > j j`.

Looking at these [slides](https://indico.cern.ch/event/928785/contributions/3905130/attachments/2058247/3452114/20_06_16_EFT_CMS.pdf#page=8) it seems possible (with `NP<=1`).

@sihyunjeon @Cvico could you please take a look as well? Thank you!

Cvico:
> Shouldn't `$$ t t~ h` be just after the first process? `generate p p > ell+ vl w- $$ t t~ h QCD=1 SMHLOOP=0 NP=1, w- > j j `
> 
> Also, I was wondering if you can have EFT in decays adding `NP=1` also after `w- > j j`.
> 
> Looking at these [slides](https://indico.cern.ch/event/928785/contributions/3905130/attachments/2058247/3452114/20_06_16_EFT_CMS.pdf#page=8) it seems possible (with `NP<=1`).
> 
> @sihyunjeon @Cvico could you please take a look as well? Thank you!

Hi! So NP=1 and NP<=1 I think will do essentially the same, as far as I know, only QCD orders are fixed in madgraph (and I think that is only when doing NLO). Another example is e.g. [QCD=1] QED=2, this will do QCD=1 exactly (unless something has changed recently), and QED<=2.

In any case, I would tend to be explicit in which couplings can be activated at each step (and look at the diagrams running standalone madgraph). As of now, I'd say EFT is activated in both production and decay. Maybe @sihyunjeon can confirm. I would check the diagrams just in case though.

lviliani:
For SMEFTsim I think you don't need the extra `NP<=1` at the end, the syntax should be:
`generate p p > ell+ vl w- $$ t t~ h QCD=1 SMHLOOP=0 NP<=1, w- > j j NP<=1 @0`

Then for EWdim6NLO I'm not sure, but don't you need to add NP as well? But I don't know the details of this model.

GiacomoBoldrini:
From the slides you mentioned I see:

Generate p p > h j NP<=1, h > b b~ NP<=1 @0 NP<=1

anmehta:
if we are happy about cards for the smeft samples, can we already ask Giovanni to prepare McM requests? We want to ask for central production as soon as we can to progress with the review. 
thanks!

Cvico:
> `generate p p > ell+ vl w- $$ t t~ h QCD=1 SMHLOOP=0 NP<=1, w- > j j NP<=1 @0`

Hi, for me it looks fine. The last NP<=1 is probably there for redundancy and preventing madgraph for adding EFT effects in propagators after the decay

anmehta:
Hi @lviliani, @Cvico,

thanks for your comments and suggestions. I guess if there are no more comments, we can ask our MC contacts to prepare the request? Are there any objections to this?

thanks and Best,
Giacomo, Ankita

GiacomoBoldrini:
> Ah sorry, you are right about the slides. Still looks a bit weird to me, but ok.
> 
> For EWdim6NLO that's fine. Btw could you please put these cards in separate directories wrt SMEFTsim just for the sake of clarity?
> 
> @Cvico @sihyunjeon any objections regarding the current cards? If not I guess we can merge this PR.

Hi @lviliani I distributed the cards in two separate directories with obvious names.

lviliani:
Thank you! I can you can also remove the `WZ_semileptonic_01J` directory?

GiacomoBoldrini:
Done!

PR #3876, opened 2025-04-01
Corrected the collider energy for ZH cards with anomalous couplings.

pianonick411:
Hello, 

The current Run3 JHUGen cards for ZH do not have the ColliderEnergy parameter correctly specified. The cards in this PR correctly specify the collider energy to be 13.6 TeV.


lviliani:
Thanks @pianonick411, I will merge this PR straight away.

Just a question, did we produce any 13.6 TeV sample that used these cards by mistake? If so, they should be invalidated.

pianonick411:
Hi @lviliani, I don't believe any gridpacks were produced with these cards.

PR #3877, opened 2025-04-01
Adding offshell EW input cards

lk11235:
Adding the JHUGen input cards for new Run3 EW off-shell Higgs samples so that we can proceed with our Monte Carlo requests. Samples have been documented on the twiki: https://twiki.cern.ch/twiki/bin/view/CMS/Run3MCProductionforHiggsProperties#JHUGen_MCFM_electroweak_producti
Also including updated Run2 cards which were remade with more inclusive gen cuts.


mlizzo:
Hi @lviliani @bbilin could you please take a look at this pending PR when you have some time? Thank you for the cooperation

lk11235:
Hi @lviliani thank you for your review and pointing this out. These samples do not generate many (or any, really) events at those high energies so we believe that `m4l_max=13000` is acceptable for now. But we will make this change to the mass range with our next set of gridpacks.

PR #3864, opened 2025-03-05
Add ADD fragment for dijets

ahinzmann:
This is the Pythia8 fragment for ADD extra dimensions in dijet angular distributions used in EXO-24-011.

PR #3881, opened 2025-04-15
Add ALP and EFT cards for dijets

ahinzmann:
These are the Madgraph cards for virtual ALP exchange and EFT triple-gluon coupling in dijet angular distributions used in EXO-24-011.

(it is build on top of #3864, maybe consider this after the other)


lviliani:
Once we have resolved the comments above, can you please migrate this PR to the new gitlab repository?
Thanks!

PR #3884, opened 2025-04-23
Same-sign WWGamma cards

Ruobing2023:
These are the process card and run card to generate the same-sign WWGamma signal sample.


Ruobing2023:
These cards are for the SMP part generation.

PR #3887, opened 2025-04-24
add CRMC generator

michael-pitt:
This PR adds gridpack generator for CosmicRayMonteCarlo (CRMC) capable of generating hadron collisions using four different event generators: EPOS LHC-R, Sibyll_2.3e, DPMJet-III_2019.1, QGSJETIII

The current version is using [crmc-v2.2.1](https://gitlab.iap.kit.edu/AirShowerPhysics/crmc/-/releases/v2.2.1), which has been available since 29-Apr 2025 and suited for pO/OO collisions at the LHC.

NOTE: 
* Sibyll and QGSJET are asymmetric generators (pO will be different from Op) as they designed for air showers and don't treat target remnants which as momentum 0 in lab frame. 
* Sibyll is limited to pA with A<19
* QGSJET were never designed for HI collisions (no final state interactions)

Usage:
```
./gridpack_generation.sh GEN_COL el9_amd64_gcc12 CMSSW_15_0_0_pre2
```
where 
`GEN=[eposlhcr, sibyll, dpmjetIII.19, qgsjetIII]` is the generator type
`COL=[pO, Op, OO, NeNe]` is the collisions type

The beam energy is hardcoded:
pO collisions E_beam = 6.8 ZTeV
OO/NeNe collisions E_beam = 5.36 ZTeV

### DIFF
--- a/bin/CRMC/genfragments/gen_fragment.py
+++ b/bin/CRMC/genfragments/gen_fragment.py
@@ -1,0 +1,25 @@
+import FWCore.ParameterSet.Config as cms
+
+externalLHEProducer = cms.EDProducer("ExternalLHEProducer",
+    args = cms.vstring(PATH_TO_TARBALL),
+    nEvents = cms.untracked.uint32(1),
+    numberOfParameters = cms.uint32(1),
+    outputFile = cms.string('cmsgrid_final.lhe'),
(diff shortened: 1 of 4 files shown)


michael-pitt:
The gridpack code is updated for the latest (final) version of CRMC for oxygen run at the LHC

PR #3888, opened 2025-05-05
UL cards for missing samples: vector-like quark B -> tW, t-associated production

jmhogan:
Adding cards for B2G-24-013 vector-like B quark paper. 
Analog of existing cards [BprimeBToTW_LH](https://github.com/cms-sw/genproductions/tree/mg265UL/bin/MadGraph5_aMCatNLO/cards/production/2017/13TeV/BprimeBToTW_LH), but with t-associated production. Most cards are identical, with names changed to BprimeTToTW. The proc card shows the change from a qbtW final state to a qttW final state.


jmhogan:
@lviliani How do I indicate Run2 UL, just by using the "13TeV" folder in the master branch?

lviliani:
Yes, looks correct. Thanks!

PR #3896, opened 2025-06-18
TTHH LO generation with dileptonic and semileptonic (including tau) final state

tiyang-hanyang:
The cards including the tau leptons in the final state are uploaded following the previous merge request https://github.com/cms-sw/genproductions/pull/3863.
The cards guides the decay of the top pairs in dileptons (e mu tau) and semileptons respectively. The Higgs boson decay is handled by pythia with decay restrictions in corresponding fragments as before. 
In usage of the corresponding gridpacks, the fragments like HHto2B2V_dipoleRecoilOn-TuneCP5_13p6TeV_madgraph-pythia8.py fragment without restrictions on the tau leptons will decay the tau leptons to any final states. Thus the leptonic decay of top will include hadronic tau final state, but not only e/mu. In most of the scenario, the DL and SL generated samples is designed to be used together. 
The gridpack production and the crab generation using the gridpack and official fragment has been tested.


tiyang-hanyang:
Yes, many thanks for reminding me this. These two are almost the same. Let's migrate to the gitlab.

PR #3901, opened 2025-09-03
Run 3 PolZZ to 4l information

isehle:
Adding MadGraph run and proc cards for PolZZ to four lepton (Run 3), as well as the corresponding fragments.


lviliani:
Can you please move this PR to the new gitlab repository:

This one is only kept as an archive.

